PR #141, opened 2012-10-15
Support HTML and XML escape entities

ntrrgc:
Otherwise, code with them break.

### DIFF
--- a/src/languages/xml.js
+++ b/src/languages/xml.js
@@ -55,6 +55,11 @@ function(hljs) {
         relevance: 10
       },
       {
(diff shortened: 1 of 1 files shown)


jdagerot:
This is actually a problem, I'm building a site with some localized strings, for example words like "får, häst, ödla" The will be f&aring;r, h&auml;st, &ouml;dal when entetizied. We are doing this out of recommendation and that we've found that the character encoding is somewhat flaky within out rather huge development environment.

I would love to be able to not have highlight.js deentetize these entities in some way, could anyone maybe help where I should start?

PR #250, opened 2013-07-24
Add Jade language support

h2non:
Still needs support for indented multi line comments and strings


mrmlnc:
Maybe take a look at implementation in [Prism](https://github.com/PrismJS/prism/blob/gh-pages/components/prism-jade.js)? Almost like in WebStorm.

Test: [link](http://prismjs.com/test.html)

patarapolw:
The downside is albeit PrismJS supports Pug, it does not support `*.vue`. (More specifically, `<template lang="pug">`).

I have to stick with `highlight.js` for now.

PR #467, opened 2014-05-01
Adding FORTRAN 90 language file

demitri:
First pass at a FORTRAN 90 language file. The .eq./.and./etc. piece can probably be simplified!


Sannis:
Fortran definition merged in #727.

PR #562, opened 2014-08-30
Add hljs.PHRASAL_WORDS_MODE to all comment modes

Sannis:
I'm not think that this is so needed, but...

@isagalaev, what do you think about this?


Sannis:
If `hljs.COMMENT()` will be accepted, I'll make same thing for numbers - most of that should have zero relevance to avoid wrong detection.

src/highlight.js
```diff
@@ -669,21 +669,23 @@ https://highlightjs.org/
...
+      className: 'comment',
+      begin: begin, end: end
+    };
+    if (inherits) {
+      mode = hljs.inherit(mode, inherits);
+    }
+    if (!mode.contains) {
+      mode.contains = [];
+    }
+    mode.contains.push(hljs.PHRASAL_WORDS_MODE);
```
isagalaev: These four lines could be spelled shorter: `mode.contains = (mode.contains || []).push(hljs.PHRASAL_WORDS_MODE);`

isagalaev:
Merged it regardless of code in hljs.COMMENT, as it is ultimately a good change anyway. We could refactor stuff right in master.

isagalaev:
Thanks a lot!

Sannis:
Nice, will do TODO and doctag highlighting as next step.

Issue #527, opened 2014-07-31
Add Julia language support


marcusps:
[Julia](http://julialang.org/) is becoming quite popular, and projects like GitLab now rely on `highlight.js` for syntax highlighting, so supporting Julia seems like a good idea.

There is CodeMirror [support](https://github.com/JuliaLang/IJulia.jl/blob/master/deps/julia.js) in the in the IJulia project that may be useful to build such a feature in `highlight.js`.

PR #596, opened 2014-10-06
Support new language: Julia

bicycle1885:
As mentioned in #527 by @marcusps, it would be nice to support the Julia language. Here it is!

I've checked the "Language contributor checklist" and confirmed that this passed the unit test.
But before completing the job, let me ask some questions.

1. To avoid misclassification with Matlab, I slightly modified the relevance of comment style of it. Is this permissible?
2. I want to escape expressions surrounded by '$(' and ')' in a string literal like Ruby or CoffeeScript, but the way of Ruby and CoffeeScript is not satisfactory to me because the escaped expressions are highlighted awkwardly. Is there any way to completely escape these expressions as if they are in normal code?

When there is any inappropriate manner in this pull request, please let me know.

Thanks.

### DIFF
--- a/src/languages/matlab.js
+++ b/src/languages/matlab.js
@@ -92,7 +92,8 @@ function(hljs) {
       },
       {
         className: 'comment',
-        begin: '\\%', end: '$'
+        begin: '\\%', end: '$',
+        relevance: 10
       }
     ].concat(COMMON_CONTAINS)
   };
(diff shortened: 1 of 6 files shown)


src/languages/matlab.js
```diff
@@ -92,7 +92,8 @@ function(hljs) {
       },
       {
         className: 'comment',
-        begin: '\\%', end: '$'
+        begin: '\\%', end: '$',
+        relevance: 10
```
Sannis: There is no special usage of `%` in Julia so I'll revert this.

Sannis:
Finally merged, thanks!

bicycle1885:
oh, I completely forgot the pull request, sorry.
Anyway, thank you for merging!

PR #599, opened 2014-10-07
AspectJ syntax

ozlerhakan:
Syntax highlighting for the AspectJ language.

Just note that: what you are doing in Java you can also do in any aspect module, so we cannot expect a low matching value between Java and AspectJ.

### DIFF
--- a/src/styles/monokai_sublime.css
+++ b/src/styles/monokai_sublime.css
@@ -93,7 +93,8 @@ Monokai Sublime style. Derived from Monokai by noformnocontent http://nn.mit-lic
 
 .hljs-variable,
 .hljs-params,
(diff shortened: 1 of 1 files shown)


jolkdarr:
Hello.
A minor issue with `final` keyword: not highlighted for pointcut parameters.
Example:
```aspectj
privileged final aspect Message_validation {
    pointcut message_receipt(final Converter converter, final Message m) :
        this(converter) && execution(void receive(*)) && args(m);

    before(final Converter c, final Message m) : message_receipt(converter, m) {
        if (c.provider != null)
            m.get_origin().set_identifier(c.provider.get_identifier(c.get_network(), m.get_originator()));
    }
}
```
Thanks for the contribution.

PR #640, opened 2014-11-13
Add Prolog language support

rla:
Attempted to add Prolog highlighter. Uses small set of tokens: atom, string, number, variable and comment. The highlighter code was partially inspired from Erlang (which is historically based on Prolog) and Lisp (recursive rules).

Detection relevance tuning:

 * Operator `:-` boosted to 10. Helps to distinguish from Erlang.
 * Dot at the end of line. Gives similar boost as in the Erlang's grammar.
 * Atoms demoted to 0. Helps to distinguish from languages that can have many identifiers
   occur sequentially like in Applescript.
 * Variables demoted to 0. Similar to atoms but for languages that tend to use identifiers starting
   with a capital letter. Variables cannot occur sequentially in Prolog, neither can they occur before
   opening parens `(` but I found no ways to express such constraints.

Prolog allows definition of custom operators that cannot be syntactically distinguished from atoms. This makes situation similar to C/C++ where enough preprocessor magic can make a program in language X a valid C/C++ program.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -150,3 +150,4 @@ Contributors:
 - mucaho <mkucko@gmail.com>
 - Dennis Titze <dennis.titze@gmail.com>
 - Jon Evans <jon@craftyjon.com>
+- Raivo Laanemets <raivo@infdot.com>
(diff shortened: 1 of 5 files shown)


isagalaev:
I didn't get around to resolving it for 8.4, will try for 8.5.

isagalaev:
Finally got to merging it. In a nutshell:

- Played with relevances to get Mercury and Prolog detected properly. I'm sure there will be mis-detections on real-world code, we'll deal with those as they happen.
- Chose a very neutral name 'name' for variables. The important thing is that it wasn't used before, so there won't be any backwards incompatible problems.

PR #653, opened 2014-11-23
Fixed PHP ::class syntax

ksimka:
PHP 5.5 has new syntax for resolving class names: \Some\ClassName::class. It was broken in hljs by 'class' mode. Here is nasty fix. Will appreciate if you will help to improve it.


isagalaev:
Fixed.

PR #683, opened 2014-12-23
Improvements to C# language

citizenmatt:
* Namespaces, classes, enums, interfaces and structs now properly use
  "title" class
* Qualified identifiers supported in namespaces, method/property return
  types, inherited types, etc.
* Multiple and nested generic arguments supported by type, method and
  property declarations
* Generic constraints on type declarations
* Better keywords in function parameters
* XML doc tags treated as XML language
* True, false and null treated as literals instead of keywords
* Pragma disable/restore
* Type declarations (class, interface, struct, enum) get "class" class
* Type declarations wrap inheritance declarations in "inheritance" class
* Each inherited type is wrapped in "parent" class
* Properties wrapped in "property" class
* Attributes wrapped in "attribute" class

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -153,3 +153,4 @@ Contributors:
 - Brian Quistorff <bquistorff@gmail.com>
 - Jonathan Suever <suever@gmail.com>
 - Alexis Hénaut <alexis@henaut.net>
+- Matt Ellis <m.t.ellis@gmail.com>
--- a/docs/css-classes-reference.rst
+++ b/docs/css-classes-reference.rst
@@ -643,12 +643,20 @@ C# ("cs", "csharp")
 
 * ``keyword``:          keyword
 * ``number``:           number
+* ``literal``:          literals ``true``, ``false`` and ``null``
 * ``string``:           string
 * ``comment``:          comment
-* ``xmlDocTag``:        xmldoc tag ("///", "<!--", "-->", "<..>")
-* ``class``:            class header from "class" till "{"
+* ``preprocessor``:     pre-processor directives, e.g. #if, #region, etc.
+* ``xmlDocTag``:        xmldoc tag ("///", "<!--", "-->", "<..>"). XML tags are marked up with the "xml" sub-language
+* ``class``:            class header from "class" till "{". Also applies to "interface", "enum" and "struct"
 * ``function``:         method header
-* ``title``:            title of namespace or class
+* ``property``:         property header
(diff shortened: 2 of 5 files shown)


isagalaev:
Trying to merge this one but with no luck so far: it hangs the test suite. I *suspect* that trouble comes from all the regexes that try to be nested, like FAKE_GENERIC_IDENT_RE and GENERIC_IDENT_RE. Another my concern is probably QUALIFIED_IDENT_RE which tries to swallow the whole string of identifiers joined by dots. Now, they might not have anything to do with the hanging itself but I think we don't want to parse them like this at all. So, could you explain why these complex regexes are needed for highlighting purposes?

JoshClose:
More C# support would be great. I just implemented it for some documentation and a lot of it isn't highlighted.

citizenmatt:
@isagalaev the reason for the nasty regexes was to try and handle matching brackets, when a fully qualified name might have one or more, or even nested, generic class or method typenames. E.g.: `Namespace1.Namespace2.Type1<string, int, Namespace3.Type2<Namespace4.Type3.NestedType>>` which is pretty horrible.

If we can do this in a nicer way, I'm all for that!

isagalaev:
@citizenmatt the usual way to handle those in highlight.js is to define a mode for the `<...>` part that would take care of nesting:

```
{
    begin: /</, end: />/,
    contains: ['self']
```

But my point was that we actually don't want to highlight identifiers at all except those in class and function *definitions*. And in those cases I'd say it's better to leave out the generic part anyway.

isagalaev:
On a related note, this got me thinking about how we currently look for the function declarations:

```
var GENERIC_IDENT_RE = hljs.IDENT_RE + '(<' + hljs.IDENT_RE + '>)?';
{
  className: 'function',
  begin: '(' + GENERIC_IDENT_RE + '\\s+)+' + hljs.IDENT_RE + '\\s*\\(',
  ...
```

This doesn't take care of nested generic specifiers at all and I doubt we can do better with out current parsing (we have to rely on this hack of first swallowing the entire function definition with a single regex). In which case we have two options:

- Accept that we don't recognize function titles whose definitions contain nested generics, leading to inconsistencies.
- Drop the feature entirely.

I'm leaning to the first choice as I hope nested generics are rare and we only loose a part of presentation. What do you think?

isagalaev:
Closing this without fixing, as there's quite a lot to do to make it mergeable, and it might even be more practical to start the work anew. Thanks for the effort!

PR #689, opened 2014-12-29
Comprehensive objective-c built-ins list: constants, classes, protocols, etc.

ahfarmer:
The existing list of built-ins for Objective-C was very small - only 84 built-ins. Some notable omissions were `NSTextView` and `NSMakeRect`.

This new list contains 5732 built-ins. I tried to put together as complete of a list as I could, in part by using `objc_getClassList` to get a list of class names on both iOS and OS X.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -153,3 +153,4 @@ Contributors:
 - Brian Quistorff <bquistorff@gmail.com>
 - Jonathan Suever <suever@gmail.com>
 - Alexis Hénaut <alexis@henaut.net>
+- Andrew Farmer <ahfarmer@gmail.com>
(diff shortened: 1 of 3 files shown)


Sannis:
This is not documented, but it looks like any regex can be used in keywords definition, so you can replace many of new and old keywords with just `NS\w+ AV\w+`.

ahfarmer:
Hmmm @Sannis I tried using regexes and it didn't work so I took a look at the highlight.js code. It looks like keywords can only be exact strings. They are compiled into the mode.keywords hash.

Would it be acceptable to use a mode to find class names in Objective-C? I have created a pull request with this approach:
https://github.com/isagalaev/highlight.js/pull/698

Sannis:
Yeah, that's my fail, we don't support regexps in keywords definitions :-(

ahfarmer:
@Sannis would you mind taking a look at my other pull request? It accomplishes the same goal, but highlights the iOS/OS X class names with a 'mode' instead of a 'keyword': https://github.com/isagalaev/highlight.js/pull/698

isagalaev:
Listing a few thousand identifier is certainly not what we can have in a :common distribution, nor it is advisable anyway as such lists tend to become a maintenance nightmare. The approach in #698 feels much more sane so I'm closing this one in favor of that.

PR #698, opened 2015-01-05
Objective-C regex for iOS/OSX API class names & custom class names

ahfarmer:
Before this change, the class list is a very small subset of all the available classes in the iOS & OS X API.

I have replaced those hard-coded keywords with 2 modes: One for API class names (words like UIView & NSTextView), and one for custom class names (like AFCustomClass).


isagalaev:
I've merged this one with one change: dropping the `CUSTOM_CLASS` mode. Since it was swallowing all the identifiers starting with two capital letters we lost highlighting of words like `BOOL` and `IBAction` that wasn't being picked up by `API_CLASS` built-ins. As far as I could figure out the custom classes were intended to highlight any user-defined types which we simply don't highlight throughout the library as a design choice.

PR #703, opened 2015-01-09
pf.conf(5) support

oldlaptop:
This is experimental support for the pf.conf(5) firewall rules file format as of OpenBSD 5.6. I'm not entirely sure how updates to all the custom styles are meant to be managed; for now only the "default" style is changed.


src/languages/pf.js
```diff
@@ -0,0 +1,58 @@
...
+    className: 'variable',
+    begin: /</, end: />/
+  };
+  var QUOTE_STRING = {
+    className: 'string',
+    begin: /"/, end: /"/
+  };
+
+  return {
+    aliases: ['pf', 'pf.conf'],
```
Sannis: First 'pf' is redundant here.

src/languages/pf.js
```diff
@@ -0,0 +1,58 @@
+/*
+Language: pf
+Category: config
+Author: Peter Piwowarski <oldlaptop654@aol.com>
+Description: The pf.conf(5) format as of OpenBSD 5.6
+*/
+
+function(hljs) {
+  var MACRO = {
+    className: 'variable',
```
Sannis: Very confused. `var MACRO` contains mode definition with `variable` class, but you specify `macro` in CSS classes reference.

src/languages/pf.js
```diff
@@ -0,0 +1,58 @@
...
+  var QUOTE_STRING = {
+    className: 'string',
+    begin: /"/, end: /"/
+  };
+
+  return {
+    aliases: ['pf', 'pf.conf'],
+    lexemes: /[a-z0-9_<>-]+/,
+    keywords: {
+      action: /* block match pass are "actions" in pf.conf(5), the rest are
```
Sannis: Last time we try to minimize different classes count across languages: https://github.com/isagalaev/highlight.js/issues/348#issuecomment-58228366. I suggest to merge `action` and `parameter` keywords into `built_in`.

Sannis:
Thanks!

PR #706, opened 2015-01-11
Autodetect XML-like languages script tag content language

### LINKED ISSUE #695: Wrong highlighting in <script> bodies where content is not javascript
It seems to me that in some cases, the language of a `script` tag is incorrectly assumed to be javascript although the type attribute explicitly defines it as something else.

The case at hand is:

    ``` html
    <!DOCTYPE html>
    <html>
    <body>
    <script type="text/x-handlebars">
      <!-- Handlebars body highlighted as javascript -->
    </script>
    </body>
    </html>
    ```

The html output generated from the `script` tag above has `class="javascript"` which results in incorrect highlighting (handlebars template highlighted as javascript code).

I think the type attribute should be taken into consideration and the body of the `script` tag highlighted as such.

Sannis:
This fixes #695.

@isagalaev, what are you think about this change?

### DIFF
--- a/src/languages/xml.js
+++ b/src/languages/xml.js
@@ -81,7 +81,7 @@ function(hljs) {
         contains: [TAG_INTERNALS],
         starts: {
           end: '</script>', returnEnd: true,
-          subLanguage: 'javascript'
+          subLanguage: ''
         }
       },
       PHP,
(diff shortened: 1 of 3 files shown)


isagalaev:
The problem with this is that we introduce detection errors where there used to be none. On the other hand, non-javascript scripting becomes more common, so I think it might be a net positive. Let's merge it like this and see how it flies.

PR #709, opened 2015-01-13
Changed number definition to match VHDL decimal and based literals

senshu:
The previous language definition for VHDL used the default C number mode. Some specific features of VHDL were not covered:
* The ability to insert underscores between digits to improve readability (e.g. `12458789` can be written as `12_458_789`)
* The syntax of based literals, which is different from C (e.g. `16#5C3B#` instead of `0x5C3B`).


Sannis:
Merged with zero relevance for VHDL numbers (a8217e9299c8e1b22748bb65cb6c993cb7b13fe4).

PR #724, opened 2015-01-31
add https alias

eksperimental:
add https alias for http
in addition to PR #723


Sannis:
Thanks!

Also you can push to pull request branch and Github updates it.

PR #727, opened 2015-02-01
Add Fortran language (2nd try)

scemama:
The first try didn't build. This one should be OK


Sannis:
JFYI, you can push to pull-request branch ant new commits will appear in it.

docs/css-classes-reference.rst
```diff
@@ -605,6 +605,14 @@ AspectJ ("aspectj")
 * ``inheritance``:      keywords "extends" and "implements" inside an aspect header
 * ``title``:            aspect, (intertype) method name or pointcut name inside an aspect header
 
+Fortran ("fortran", "f77", "f90", "f95")
```
Sannis: Is this pull request only for free-form syntax?

src/languages/fortran.js
```diff
@@ -0,0 +1,85 @@
...
+      'ieee_arithmetic ieee_support_underflow_control ieee_get_underflow_mode ieee_set_underflow_mode' +
+      'is_iostat_eor move_alloc new_line selected_char_kind same_type_as extends_type_of'  +
+      'acosh asinh atanh bessel_j0 bessel_j1 bessel_jn bessel_y0 bessel_y1 bessel_yn erf erfc erfc_scaled gamma log_gamma hypot norm2' +
+      'atomic_define atomic_ref execute_command_line leadz trailz storage_size merge_bits' +
+      'bge bgt ble blt dshiftl dshiftr findloc iall iany iparity image_index lcobound ucobound maskl maskr' +
+      'num_images parity popcnt poppar shifta shiftl shiftr this_image' 
+  };
+  return {
+    case_insensitive: true,
+    aliases: ['fortran', 'f77', 'f90', 'f95'],
```
Sannis: First `fortran` can be skipped here.

src/languages/fortran.js
```diff
@@ -0,0 +1,85 @@
...
+//           '^[#{}.]|' +
+//           '.def|' +
+//           '-module|' +
+//           'cd',
+    keywords: F_KEYWORDS, 
+    contains: [
+      hljs.inherit(hljs.APOS_STRING_MODE, {className: 'string', relevance: 0}),
+      hljs.inherit(hljs.QUOTE_STRING_MODE,{className: 'string', relevance: 0}),
+      {
+        className: 'function_start',
```
Sannis: In other languages this is `function`.

Sannis:
Thanks! First Fortran definition that is quite simple and passes all tests.

docs/css-classes-reference.rst
scemama: Yes. Having both in the same plugin would probably make the syntax file much more complicated. We could probably remove the ``f77`` and keep only ``("fortran", "f90", "f95")``, and leave fixed-form syntax for another f77 definition file.

src/languages/fortran.js
scemama: OK. I didn't see that.

PR #731, opened 2015-02-04
Corrected Fortran

scemama:
* There were missing blank spaces at the end of the keyword strings.
* A few keywords were missing


scemama:
Thanks for everything! You are very reactive!

PR #737, opened 2015-02-10
Fixing regex for Fortran

TApplencourt:
In the current implementation of the Fortran Highlight a figure in a word (like var1 and va1r) is highlighted.

This merge provides a fix by changing the regex. The main idee is to check the begining of a new line or a whitespace before matching. 

You can show the usage case in this image:
Diff

Have a nice day ! :runner:


TApplencourt:
Not compiling, close this one.

Sannis:
Can you modify this regex to not include `\n` before numbers into number mode?

TApplencourt:
Yep. It's done. And you are write if more clean like this.
Sorry for the merge conflicts... I don't know why.

Working regex

src/languages/fortran.js
```diff
@@ -71,9 +71,9 @@ function(hljs) {
       },
       {
         className: 'number',
-        begin: '-?(\\d+(\\.\\d*)?|\\.\\d+)([DdEe][+-]?\\d+)?',
+        begin: '(?=\\b|\\+|\\-|\\.)(?=\\.\\d|\\d)(?:\\d+)?(?:\\.?\\d*)(?:[de][+-]?\\d+)?\\b\\.?',
```
Sannis: Is it necessary to include points around number into regex? It does not included in tests and it looks like we can skip them.

Sannis:
Anyway, merged. Thanks for your contribution.

src/languages/fortran.js
TApplencourt: If i understand correctly your question, my answer is: Yes, We need them. Because `.\d+` (like .23) can,  sadly, be a number as `\d+.` (12.).
If we only take the digit, it will be strange at first look, no? The number is `.23` not `23` .  I guess is a personal preference.

It's me to thank you for your amazing work!

PR #741, opened 2015-02-16
Prevent script tag from prematurely closing.

trusktr:
This prevents a script tag from being prematurely closed when highlight.js is injected into such script tag.

### DIFF
--- a/src/languages/xml.js
+++ b/src/languages/xml.js
@@ -80,7 +80,7 @@ function(hljs) {
         keywords: {title: 'script'},
         contains: [TAG_INTERNALS],
         starts: {
-          end: '</script>', returnEnd: true,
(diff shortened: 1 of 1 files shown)


trusktr:
I was hacking around, and I injected a browserify bundle into a script tag of an HTML template, and the browser thought the script ended at that point (the closing script tag). For example, the following makes the browser throw an error:

```js
<script>
var before = 'before'
var foo = '</script>' + 'bar'
var after = 'after'
</script>
```

which results in 

```
' + 'bar';
var after = 'after'
```

being rendered in the page.

isagalaev:
Looks like it's been reported recently, we just haven't had a chance to merge that yet: #718.
Thanks!

PR #742, opened 2015-02-16
Added Android Studio Theme

kanytu:
Android Studio theme for XML and Java added.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -157,3 +157,4 @@ Contributors:
 - Peter Piwowarski <oldlaptop654@aol.com>
 - Kenta Sato <bicycle1885@gmail.com>
 - Anthony Scemama <scemama@irsamc.ups-tlse.fr>
+- Pedro Oliveira <kanytu(at)gmail.com>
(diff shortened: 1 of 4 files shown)


Sannis:
@kanytu, why we cannot use this color scheme for other languages?

Terri5732:
Great thanks guys!!

kanytu:
@Sannis I'm not familiar with other languages so in order to avoid getting a messy css I limited only for java. Although I can spend some time fixing it if it's necessary.

Sannis:
I'll keep this PR open until you or somebody else or one of maintainers will have time to improve it.

Anyway thanks for your PR!

isagalaev:
Language-specific styles are actually an anti-pattern in highlight.js, they're supposed to be universal. So, even a suboptimal highlighting of other languages is better than none.

kanytu:
Theme has been updated and is not language-specific anymore.

Sannis:
Thanks, merged!

Sannis:
I've also add some custom styles for addition/deletion/change selectors.

PR #744, opened 2015-02-17
Add support for BigFix Relevance

briangreenery:


isagalaev:
Hey @briangreenery! What was the reason to close this pull request?

briangreenery:
This language is probably not generally useful. It's also easy to add it to highlight.js as an addon, and I created [hljs-bigfix-relevance](https://www.npmjs.com/package/hljs-bigfix-relevance) to do this. Also, the pull-request was open for a while, so I thought maybe there wasn't any interest in including it (which is  okay!)

jgstew:
If possible, it would be useful to see this in highlight.js but the add on functionality might be sufficient since bigfix relevance is so specific to a particular application.

isagalaev:
@briangreenery sorry, we, frankly, suck at reacting fast on new contributions :-( But I believe you're right and it's better to keep it as an add-on. I'm trying to figure out what to do with proliferation of highly specialized and niche languages in highlight.js that bump the overall number and add bulk but not much value for most users.

PR #753, opened 2015-02-21
SQL: Revoke does not get highlighted

iceraj:
Here is jsfiddle that demonstrates this problem:
http://jsfiddle.net/iceraj/nvn7zdj7/


isagalaev:
Thanks for catching that! Fixed.

PR #754, opened 2015-02-21
Include images in the node.js build

kemar:
Hi, images are missing in the `styles` folder of the [`npm`](https://www.npmjs.com/package/highlight.js) package.

This is an attempt to fix the problem but you may prefer a `*` to match all files.


isagalaev:
I prefer the explicit way. Thanks a lot for catching that and for the fix!

PR #762, opened 2015-02-25
XQuery

dirkk:
Adds supports for XQuery 3.1 (http://www.w3.org/TR/xquery-31/) with XQuery Update.
This also adds support for XPath, as XQuery is a superset of XPath


Sannis:
Thanks, finally merged!

isagalaev:
I was wondering… All the strings in the test snippet are single-quote, yet the language definition only includes double-quote variety. Is it intentional?

dirkk:
I guess you are refering to https://github.com/isagalaev/highlight.js/pull/762/files#diff-0807bf531adf0615be7b7c0375700491R31?

Indeed, this should include single and double quotes, as both is permitted to be used by the XQuery spec.

isagalaev:
OK, fixed those in a3245ea.

PR #763, opened 2015-02-26
Adding support for highlighting C# keywords: dynamic and 'when'

hmemcpy:
'when' is a part of the new Exception filter syntax in C# 6


Sannis:
Thanks, merged!

PR #764, opened 2015-02-26
fixed: language is added to className every run the parse

YusukeHirao:
When real time parse, language is added to className infinity.

```html
<!-- converted html -->
<code class="js lhjs javascript javascript javascript javascript javascript javascript javascript javascript javascript ... javascript">...code...</code> 
```


isagalaev:
Thanks for catching that!

PR #768, opened 2015-02-27
Use Travis container-based architecture for faster builds

danvk:
See http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

In my own projects, I've seen this dramatically improve build start time.


isagalaev:
Seems like the way forward. Thanks!

PR #770, opened 2015-03-01
Highlight JS Loader v6.2

alidbg:
Added highlightjs.loader v6.2


isagalaev:
@alidbg , your loader is a perfect satellite project but it doesn't make sense to include it as a part of highlight.js code. The reason why highlight.js doesn't do all that is because we want to let people use it in any way they want: with any markup, any scrolling mechanism, with their custom markup, etc.

PR #771, opened 2015-03-02
Add support for Gams (.gms)

b-pos465:
http://www.gams.com/

### DIFF
--- a/docs/css-classes-reference.rst
+++ b/docs/css-classes-reference.rst
@@ -5,6 +5,13 @@ This is a full list of available classes corresponding to languages'
 syntactic structures. The parentheses after language name contain identifiers
 used as class names in ``<code>`` element.
 
+
+Gams ("gms", "gams")
+--------------------
+
+* ``keyword``:          keyword
+
+
 Python ("python", "py", "gyp")
 ------------------------------
 
--- a/src/languages/gams.js
+++ b/src/languages/gams.js
@@ -1,0 +1,30 @@
+module.exports = function (hljs){
+
+    var keywords = 'abort acronym acronyms alias all and assign binary card diag display else1 eps eq equation equations file ' +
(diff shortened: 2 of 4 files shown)


docs/css-classes-reference.rst
```diff
@@ -5,6 +5,13 @@ This is a full list of available classes corresponding to languages'
 syntactic structures. The parentheses after language name contain identifiers
 used as class names in ``<code>`` element.
 
+
+Gams ("gms", "gams")
+--------------------
+
+* ``keyword``:          keyword
+
```
isagalaev: The language also defines "string", "comment" and "number", please include those in the reference too.

src/languages/gams.js
```diff
@@ -0,0 +1,30 @@
+module.exports = function (hljs){
```
isagalaev: This won't let the build system to properly wrap this file. A language file should contain a bare JavaScript function and some meta data (http://highlightjs.readthedocs.org/en/latest/language-contribution.html#provide-meta-data).

src/languages/gams.js
```diff
@@ -0,0 +1,30 @@
...
+
+    var keywords = 'abort acronym acronyms alias all and assign binary card diag display else1 eps eq equation equations file ' +
+        'files for1 free ge gt if inf integer le loop lt maximizing minimizing model models na ne negative no not option   options or ord ' +
+        'parameter parameters positive prod putpage puttl repeat sameas scalar scalars semicont semiint set1 sets smax smin solve sos1 sos2 ' +
+        'sum system table then until using variable variables while1 xor yes';
+
+    keywords += keywords.toUpperCase();
```
isagalaev: It's easier to just set `case_insensitive: true` instead of duplicating keywords.

Sannis:
@b-pos465 are you plan to update CSS reference and add some test snipper got Gams?

PR #776, opened 2015-03-06
Elixir: add 'defp' to list of function keywords.

c-rack:
See http://elixir-lang.org/getting-started/modules.html#named-functions for documentation of ```defp```, which indicates a private function.


isagalaev:
Merged, thanks!

PR #778, opened 2015-03-12
Introduce Kotlin language

cy6erGn0m:
see http://kotlinlang.org for details

### DIFF
--- a/src/languages/kotlin.js
+++ b/src/languages/kotlin.js
@@ -1,6 +1,7 @@
 /*
- Language: Kotlin
- Author: Sergey Mashkov <cy6erGn0m@gmail.com>
+Language: Kotlin
+Author: Sergey Mashkov <cy6erGn0m@gmail.com>
+Category: common
  */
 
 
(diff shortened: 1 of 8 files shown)


src/languages/kotlin.js
```diff
@@ -0,0 +1,105 @@
+/*
+ Language: Kotlin
+ Author: Sergey Mashkov <cy6erGn0m@gmail.com>
```
Sannis: This is not necessary, but you can add `Category` field here.

Sannis:
LGTM

src/languages/kotlin.js
cy6erGn0m: Are there any rules to choose valid "category"? Is it fine to use "common" here?

src/languages/kotlin.js
```diff
@@ -0,0 +1,106 @@
+/*
+Language: Kotlin
+Author: Sergey Mashkov <cy6erGn0m@gmail.com>
+Category: common
```
isagalaev: "common" is the only category that at this moment is special: it defines a small subset of very commonly used languages that goes into a pre-built script we host on CDNs. So it's best not to use it for new languages. Here it can simply be dropped unless it fits well into any of the existing categories (https://highlightjs.org/static/demo/).

cy6erGn0m: OK, changed it to misc as similar languages here (Groovy and Scala)

Sannis:
Thanks, merged!

PR #789, opened 2015-03-21
Lisp syntax fix: identifiers with digital suffix, uppercase exponent in floating literals

CodeArhat:
hljs_lisp_issues


isagalaev:
Merged, thanks!

PR #592, opened 2014-10-04
Export compileLanguage

dbkaplun:

PR #790, opened 2015-03-22
Refactor highlight()

eemeli:
For a project, we needed to have a tree-like representation of the highlighted source of a document, so I refactored the core `highlight()` function to support this; here's the result. With this PR, the parser no longer outputs HTML directly, but a tree of `HighlightNode` objects, which may then be [transformed into HTML](/eemeli/highlight.js/blob/d382d820876287ca18555947baa0e6d5c9df5420/src/highlight.js#L315), or any other form ([example](/eemeli/highlight.js/blob/d382d820876287ca18555947baa0e6d5c9df5420/src/highlight.js#L330)). This should also fix the intent of PR #592.

In addition to the different output, I've refactored the internal functions of `highlight()` to be much more functional, with only a couple of references to variables coming from the surrounding scopes. I addded a new boolean parameter `dataOnly` to both `highlight()` and `highlightAuto()`; if it's `true`, the html output is suppressed. This is useful e.g. for subLanguages, the nodes of which end up in the surrounding context's tree, and are parsed into output again later.

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -265,194 +265,189 @@ https://highlightjs.org/
 
   */
   function highlight(name, value, ignore_illegals, continuation) {
-
-    function subMode(lexeme, mode) {
-      for (var i = 0; i < mode.contains.length; i++) {
-        if (testRe(mode.contains[i].beginRe, lexeme)) {
-          return mode.contains[i];
-        }
-      }
+    var _language = getLanguage(name);
+    if (!_language) {
+      throw new Error('Unknown language: "' + name + '"');
     }
+    compileLanguage(_language);
+    var _continuations = {}; // keep continuations for sub-languages
+    var _mode_buffer = '';
 
-    function endOfMode(mode, lexeme) {
-      if (testRe(mode.endRe, lexeme)) {
-        while (mode.endsParent && mode.parent) {
-          mode = mode.parent;
-        }
-        return mode;
-      }
-      if (mode.endsWithParent) {
-        return endOfMode(mode.parent, lexeme);
-      }
-    }
+    var result = {
+      relevance: 0,
+      value: '',
+      language: name,
+      mode: null
+    };
 
-    function isIllegal(lexeme, mode) {
-      return !ignore_illegals && testRe(mode.illegalRe, lexeme);
-    }
-
-    function keywordMatch(mode, match) {
-      var match_str = language.case_insensitive ? match[0].toLowerCase() : match[0];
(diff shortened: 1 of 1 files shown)


isagalaev:
Hi! This is a great effort and the result is, undoubtedly, more readable than what we have now. However there is a very practical reason why most logic currently relies on mutable external state and why there's no flexibility in output generation: it's the code size. As long as we're interested in working in a browser, it's a pretty important optimization metrics that, unfortunately, beats modularity, functional design and even readability. So, for now I can only suggest to keep this refactoring as a fork, we won't merge it.

eemeli:
I would argue that in this case the increase in code size is insignificant. To be precise, after minifaction and gzipping, the increase in file size we're talking about is the range of 393 - 454 **bytes**, depending on the number of languages that are included in the pack. For the full set, that means an increase of about 0.36% to the total size of 108kB.

For the sake of comparison, the average language definition in highlight.js is 1486 bytes when compressed, though there is admittedly quite a bit of variation there.

isagalaev:
You know, you're probably right. There was a bug push for squeezing the core as much as possible a while ago when we tried to keep the library with a handful of common languages under the magical number of 30K (un-gzipped, as it is what people usually take into account for some reason). But we've happily passed that limitation already, so it's time to have another look at the problem.

I can't promise that it will happen very fast both because we're in the process of wrapping the next release soon and because the patch is non-trivial and might provoke some philosophical discussion :-) But it's interesting, so thanks anyway!

Sannis:
BTW, @isagalaev, when you plan to release 8.5?

isagalaev:
@Sannis wanted to get it in shape this week but for various reasons I blew all my time slots for highlight.js :-( Looking at the issue list (https://github.com/isagalaev/highlight.js/milestones/8.5) I think we should postpone the big C# refactoring for later as it turned out way harder than I thought, and the other two should take about one day. So, "next week" is an ongoing estimate :-)

eemeli:
I agree that this pull request is somewhat non-trivial and deserves more thorough review. Furthermore, using this approach I'm pretty sure the `nodeStream/mergeStreams` implementation of `highlightBlock` could be replaced with something cleaner, where the `HighlightNode` tree is used instead of `nodeStream(resultNode)`. Haven't had time yet to plan or implement that, though.

isagalaev:
I decided to reject the pull request after all. The main reason being that this rewrite makes the core completely unfamiliar to me, as a principal maintainer. While I agree with the general idea of diverging tree building from rendering and making code more functional, I just can't take it as is entirely. We'll be going into that direction but one step at a time.

Sorry for taking so long to make this decision.

Issue #736, opened 2015-02-09
[NFR] Added support for Zephir Language


ovr:
Hi ;)
First thanks contributors for this library :smiley: 

https://github.com/phalcon/zephir

This will be similar for PHP
but without `<?php` open tags, `$` for variables

Can anyone help with it?

ovr:
@Sannis Any news? It's rly important. For upcomming conference
Thanks

ovr:
@Sannis ping

@isagalaev Can you help?

isagalaev:
Can't speak for @Sannis what he meant by assigning this to himself but generally we don't take requests on implementing languages, instead we accept implementations from contributors (http://highlightjs.readthedocs.org/en/latest/language-requests.html). In any case, we're all volunteers here and don't adhere to any time schedule.

Sannis:
PHP detection depends on PHP-specific things like $-started variables. I've stat working on it but currently I have many test broken :-(

ovr:
Парни вопрос жизни или смерти
Даже 1С есть, прошу молю :smiley_cat:

Sannis:
Fixed in #793.

PR #793, opened 2015-03-29
WIP: Zephir language highlighting

Sannis:
Pull-request for #736.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,16 @@
 ## Master [Unpublished]
 
+New languages:
+
+- *Zephir* by [Oleg Efimov][]
+
 Notable fixes and improvements to existing languages:
 
 - Fix encoding of images when copied over in certain builds
 - Fix incorrect highlighting of the word "bug" in comments
 - Treat decorators different from matrix multiplication in Python
+
+[Oleg Efimov]: https://github.com/Sannis
 
 ## Version 8.6
 
--- a/src/languages/zephir.js
+++ b/src/languages/zephir.js
@@ -1,0 +1,111 @@
+/*
+ Language: Zephir
+ Author: Oleg Efimov <efimovov@gmail.com>
+ */
+
+function(hljs) {
+  var STRING = {
+    className: 'string',
+    contains: [hljs.BACKSLASH_ESCAPE],
+    variants: [
+      {
+        begin: 'b"', end: '"'
+      },
+      {
+        begin: 'b\'', end: '\''
+      },
+      hljs.inherit(hljs.APOS_STRING_MODE, {illegal: null}),
+      hljs.inherit(hljs.QUOTE_STRING_MODE, {illegal: null})
+    ]
+  };
+  var NUMBER = {variants: [hljs.BINARY_NUMBER_MODE, hljs.C_NUMBER_MODE]};
+  return {
+    aliases: ['zep'],
+    case_insensitive: true,
+    keywords:
+    'and include_once list abstract global private echo interface as static endswitch ' +
+    'array null if endwhile or const for endforeach self var let while isset public ' +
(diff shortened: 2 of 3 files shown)


Sannis:
@ovr can you look on this PR to check I don't miss any significant modes or keywords? I used current PHP language definition and https://github.com/phalcon/zephir/blob/master/parser/scanner.re as references.

Also it whould be good if you can suggest some Zephir snippets to be added for tests.

ovr:
@Sannis 

It can helps
https://github.com/phalcon/zephir-lexer
https://github.com/phalcon/zephir-sublime

I will write good\big code snippet ASAP (1, 2 days)

Thank you for your work ;)

Sannis:
Thanks, I'll look on it!

ovr:
Start working on this

```zep

function testBefore(<Test> a, var b = 5, int c = 10) {
	a->method1();

	return b + c;
}

namespace Test;

function testAfter(<Test> a, var b = 5, int c = 10) {
	a->method5();

	return b + c;
}

abstract class CustomClass() {

}

interface TestInterface() {

}

class Test extends CustomClass implements TestInterface
{
	const C1 = null;

	const C2 = false;

	const C3 = true;

	const C4 = 10;

	const C5 = 10.25;

	const C6 = "test";

	//magick consts http://php.net/manual/ru/language.constants.predefined.php
	const className = __CLASS__;

	protected testVar {get, set} = 10;

	protected myArray = [
		"key": [
			"test", "test"
		]
	];

	public function method1(const string phpVersion = PHP_VERSION, var simple = self::C4) -> boolean
	{
		int b, c, sum;

		var a = [1, 2, 3, 4, 4, 5, 6, 7];
		array a = [1, 2, 3, 4, 4, 5, 6, 7];

		char h = 'A';
		string myStr = "example string";

		for ch in str {
			//built-in method toInt
			sum += ch->toInt();
		}

		for b in a {

		}

		for b, c in a {

		}

		for _, c in a {

		}

		for _ in a {

		}

		if (a == 5) {

		} elseif (a == 4) {

		} else {

		}

		switch (a) {
			case 1:
				return true;
				break;
			default:
				return false;
				break;
		}

		return true;
	}

	// See fn is allowed like shortcut
	public fn method2() -> <Test>
	{
		call_user_func(function() { echo "hello"; });


		[1, 2, 3, 4, 5]->walk(
			function(int! x) {
				return x * x;
			}
		);

		[1, 2, 3, 4, 5]->walk(
            function(_, int key) { echo key; }
        );

        array input = [1, 2, 3, 4, 5];

        input->walk(
			function(_, int key) { echo key; }
		);


		input->map(x => x * x);

		return this;
	}

	public function method3() -> int
	{
		var a, image;

		let a = new CustomClass();

		let a = 2 + 1 * 2;

		let a = this->testVar;

		let a = this->testVar + 5;


		let image = new Image();

		// named
		i->chop(width: 100); // width=100, height=400, x=0, y=0
        i->chop(height: 200); // width=600, height=200, x=0, y=0
        i->chop(height: 200, width: 100); // width=100, height=200, x=0, y=0
        i->chop(x: 20, y: 30); // width=600, height=400, x=20, y=30

		return 1;
	}

    public function ord(string! c)
    {
        uchar ch;

        let ch = c[0];

        if ch <= 0x7F {
            return ch;
        }

        if ch < 0xC2 {
            return false;
        }

        if ch <= 0xDF {
            return (ch & 0x1F) << 6  | (c[1] & 0x3F);
        }

        if ch <= 0xEF {
            return (ch & 0x0F) << 12 | (c[1] & 0x3F) << 6  |
                    (c[2] & 0x3F);
        }

        if ch <= 0xF4 {
            return (ch & 0x0F) << 18 | (c[1] & 0x3F) << 12 |
                    (c[2] & 0x3F) << 6 | (c[3] & 0x3F);
        }

        return false;
    }
}
```

PR #795, opened 2015-03-30
New language: C/AL

kfuglsang:
I added support for the C/AL language used by Microsoft Dynamics NAV C/AL.


forki:
Very very cool. +1

Sannis:
There is a trivial merge conflict with current master, can you merge it into your branch?

On first look all is ok.

kfuglsang:
Merged with latest upstream

forki:
@kfuglsang  is there a way to use this in fsreveal?

kfuglsang:
@forki Yeah once this PR is resolved.

If the syntax is not autodetected as C/AL you could use `[lang=cal]` to force it.

forki:
even if it get's merged here, there is still reveal.js which needs to be updated, right?

I wonder if we can hack it into fsreveal as a prerelease.

kfuglsang:
Thats true. I could fork reveal.js afterwards to update highlight.js in that. Then we could change fsreveal to use the fork.

kfuglsang:
@Sannis Should be ready to merge.

isagalaev:
BTW, I wouldn't call using this fork a "hack", and it won't be made any more "true" when we merge it with our master branch. So feel free not to wait for us!

PR #798, opened 2015-03-31
Add a reference to highlight in jBatch to README

BorisSchaeling:
The reference to the highlight command in jBatch should make it even easier to get started. Anyone using jBatch already can just invoke the highlight command and is done. And JavaScript beginners who have trouble customizing the library may find it easier to use a command with options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,19 @@ $('div.code').each(function(i, block) {
 For other options refer to the documentation for [`configure`][3].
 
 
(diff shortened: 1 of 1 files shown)


isagalaev:
I don't really see the need to document a system using highlight.js inside of highlight.js own docs. There's **a lot** of those and it's a slippery slope: as soon as we document one, we'll have to forever document all the others.

PR #805, opened 2015-04-05
remove the first new line character after code tag

towry:
The first new line character is useless and will cause an extra space on the top of the code block.

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -487,6 +487,7 @@ https://highlightjs.org/
       if (!getLanguage(name)) {
         return;
       }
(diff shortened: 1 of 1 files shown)


isagalaev:
This would break an informal rule of thumb that we don't normally change the way the code is laid out in HTML without highlighting. Also, I can imagine a use-case where this line-break might be put there intentionally: for example as a demonstration of a parser bug triggered by it.

isagalaev:
(But thanks for the contribution anyway! It's good to have different ideas coming.)

PR #806, opened 2015-04-05
Removes "true", "false", and "nil" from "built_in"

mariusschulz:
The literals `true`, `false`, and `nil` are listed underneath `literal` and therefore shouldn't be part of `built_in` at the same time.


isagalaev:
Could you also update the .expect.txt for the markup test that's failing?

mariusschulz:
Of course, here's the updated markup file. Sorry about that, forgot to run `node tools/build -t node` before running `npm test`.

Sannis:
Thanks, merged!

PR #807, opened 2015-04-05
Params without parentheses

mariusschulz:
Places parentheses outside of `<span class="hljs-params">` tags for JavaScript, TypeScript, and C#.


isagalaev:
Merged, thanks!
I've reverted the change that suppressed generation of an empty span, for consistency.

PR #822, opened 2015-04-09
Openscad

pzl:
added the language used for the 3D modeling program OpenSCAD

let me know if there's a better/more preferred way to do anything in the language definition


Sannis:
Thanks for your contribution!

PR #824, opened 2015-04-11
ARM Assembler language

pzl:
Adding support for ARM assembly (+ thumb, thumb2, and extensions).

---

tried something new here with the keywords. There's a helper function to generate the multiple combinations of instructions off of each base. Here's what's happening:

For a given instruction, say, `ADD`, there are 17 variants on this instruction; two-character condition codes added to the end of the instruction. So you get `ADDEQ`, `ADDNE`, `ADDCS`, `ADDCC`, `ADDMI` and so forth. Some instructions also have an additional postfix flag, a single character. This flag can be in addition to, or after all the conditional flags. Using our `ADD` example again, with the `S` flag, we have another batch of variants: `ADDS`, `ADDEQS`, `ADDNES`, `ADDCSS`, `ADDCCS`, etc. 

Nearly all instructions have these 17 condition variants, and maybe half the instructions also have an additional postfix letter. To cut down on a massive typed out list, maybe 25 times the size of the current list (I counted about 5,500 instructions), a helper function generates all the condition codes for each instruction.

I didn't see any appreciable performance differences with having the keyword list populated like this instead of a static string from startup, but if this is stylistically a problem, it wouldn't be too difficult to place the generated list in there as a string. It just makes maintenance of instructions a bit more difficult.

### DIFF
--- a/test/detect/armasm/default.txt
+++ b/test/detect/armasm/default.txt
@@ -1,0 +1,36 @@
+.text
+
+.global connect
+connect:
+    mov     r3, #2              ; s->sin_family = AF_INET
+    strh    r3, [sp]
+    ldr     r3, =server_port    ; s->sin_port = server_port
+    ldr     r3, [r3]
+    strh    r3, [sp, #2]
+    ldr     r3, =server_addr    ; s->sin_addr = server_addr
+    ldr     r3, [r3]
+    str     r3, [sp, #4]
+    mov     r3, #0              ; bzero(&s->sin_zero)
+    str     r3, [sp, #8]
+    str     r3, [sp, #12]
+    mov     r1, sp      ; const struct sockaddr *addr = sp
+
+    ldr     r7, =connect_call
+    ldr     r7, [r7]
+    swi     #0
+
+    add     sp, sp, #16
+    pop     {r0}        ; pop sockfd
+
+    pop     {r7}
+    pop     {fp, ip, lr}
+    mov     sp, ip
+    bx      lr
+
+.data
+socket_call:   .long 281
+connect_call:  .long 283
+
+/* all addresses are network byte-order (big-endian) */
+server_addr:            .long 0x0100007f ; localhost
+server_port:            .hword 0x0b1a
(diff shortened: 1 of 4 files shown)


test/detect/armasm/default.txt
```diff
@@ -0,0 +1,36 @@
...
+    pop     {r7}
+    pop     {fp, ip, lr}
+    mov     sp, ip
+    bx      lr
+
+.data
+socket_call:   .long 281
+connect_call:  .long 283
+
+/* all addreses are network byte-order (big-endian) */
```
lucaswerkmeister: typo? addres<strong>s</strong>es

Sannis:
@isagalaev, what are you think about this diff?

isagalaev:
It seems that it can be done in a less complicated manned by simply using an overarching regex without generating a space-separated line first. Like this:

```js
{
  className: 'keyword',
  begin: '(adc|([uqs]?add(8|16))| ... )(eq|ne|cs ... )?s?'
}
```

This can be split up into several lines, if needed.

Granted, this regex would cover some nonexistent  combinations but this is not a problem for a syntax highlighter, as it doesn't validate the input, simply expecting it to be already correct.

pzl:
sounds reasonable, I can rework this into the regex method in the next few days. And fix that typo noted in the code example

Sannis:
@pzl, are you plan to rework this into regex in next week?

pzl:
Yes, that wouldn't be a problem

Sannis:
I'm just thinking about adding this to 8.6 release or not.

isagalaev:
@Sannis you can add it to the milestone tentatively (meaning "we want that on principle"), and then you can simply remove it right before the release if you decide you don't want to hold the whole release for longer because of this issue. In other words, including an issue in a milestone is intention, not commitment.

pzl:
updated with regex strategy for keywords, correct the typo in the code sample as well

Let me know if there's anything else you need

Sannis:
LGTM

PR #825, opened 2015-04-11
Add Ceylon language

lucaswerkmeister:
Adds support for the Ceylon programming language ([official website](http://ceylon-lang.org/), [quick introduction](http://ceylon-lang.org/documentation/current/introduction/)).

The default test is a combination of code from the [quick introduction](http://ceylon-lang.org/documentation/1.1/introduction/#a_familiar_readable_syntax) and the [tour](http://ceylon-lang.org/documentation/1.1/tour/basics/#operators_for_handling_null_values). In both cases, the code was authored by @gavinking (ceylon/ceylon-lang.org@31c65557d6a36b893b376ff50cc0dba8f9e1ab79, ceylon/ceylon-lang.org@1f768d00e0e26851cb69d2027f61a34d26cbae97, ceylon/ceylon-lang.org@9cd84b10a4da5c749ece6fac77d4973341f73f00, ceylon/ceylon-lang.org@d047d9602528bd1afa8827beccea52b0b6d82738, ceylon/ceylon-lang.org@f0fe1921a431319244122faa44d900d7c7b60c3e). In general, sample code on `ceylon-lang.org` [falls under the ASL2](https://github.com/ceylon/ceylon-lang.org#license), but I’m not sure whether these short snippets are above the threshold of originality to warrant any copyright concerns; @gavinking wdyt?


gavinking:
@lucaswerkmeister I'm sure there are no copyright issues.

src/languages/ceylon.js
```diff
@@ -0,0 +1,71 @@
...
+  // 2.3. Identifiers and keywords
+  var KEYWORDS =
+    'assembly module package import|0 alias class|0 interface|0 object given value ' +
+    'assign void|0 function new|0 of extends|0 satisfies|10 abstracts in out return|0 ' +
+    'break|0 continue|0 throw|0 assert dynamic if|0 else|0 switch|0 case|0 for|0 while|0 try|0 ' +
+    'catch|0 finally|0 then let this|0 outer super|0 is exists|10 nonempty|10';
+  // 7.4.1 Declaration Modifiers
+  var DECLARATION_MODIFIERS =
+    'shared|10 abstract formal|10 default|10 actual|10 variable late native deprecated' +
+    'final sealed annotation suppressWarnings small';
```
isagalaev: It's better not fiddle with custom relevance until there's real need, and even in that case it's a last resort measure. Also, in this particular case they are quite wrong: for example, they would make snippets with very common variable names ("default", "actual", etc.) to be likely detected as Ceylon. And zero values would deprive Ceylon from having its fair relevance from normal keywords.

src/languages/ceylon.js
```diff
@@ -0,0 +1,71 @@
...
+  return {
+    keywords: {
+      keyword: KEYWORDS,
+      annotation: LANGUAGE_ANNOTATIONS
+    },
+    illegal: '\\$[^01]|#[^0-9a-fA-F]',
+    contains: [
+      hljs.C_LINE_COMMENT_MODE,
+      {
+        // block comment (can be nested, so no hljs.C_BLOCK_COMMENT_MODE)
```
isagalaev: A new (as yet undocumented) way of doing it is this:

```js
hljs.COMMENT('/\\*', '\\*/', {contains: ['self']})
```

src/languages/ceylon.js
```diff
@@ -0,0 +1,71 @@
...
+        end: '"|``',
+        relevance: 0
+      },
+      {
+        // character literal
+        className: 'string',
+        begin: "'",
+        end: "'",
+        relevance: 0
+      },
```
isagalaev: Zero relevance is wrong here as well (in both modes), just leave it out. These two modes are probably better collapsed in one:

```js
{
  className: 'string',
  variants: [
    {begin: '"', end: '"'},
    {begin: "'", end: "'"},
    {begin: '``', end: '``'}
  ]
}
```

isagalaev:
Hi, thanks for the contribution! I left some comments inline, with those addressed it looks good to merge.

P.S. Also, there's no formal rules for language categories and none is required for a language, by default it will reside under 'misc' in the demo page.

src/languages/ceylon.js
lucaswerkmeister: I thought that we should give zero relevance to things that are common across many languages, like “strings in double quotes” and “void means no return value”?

String parts can also start with quotes and end with backticks, or vice versa:
```ceylon
"Literal part of a string `` expression `` another literal part"
```
Perhaps I should make different modes for those (same `className` of course), with higher relevance? Backtick interpolation is far less common AFAIK

src/languages/ceylon.js
lucaswerkmeister: Nice, thanks!

src/languages/ceylon.js
lucaswerkmeister: If I remove all the `|10`s, the Ceylon sample is misidentified as Gradle.

lucaswerkmeister: It still works with `|3`, but with `|2` it’s misidentified as AspectJ. Would `|3` be acceptable?

lucaswerkmeister:
This seems to have stalled a bit. The remaining issue seems to be relevances, in two ways:

1. Zero relevance: I thought we were supposed to add zero relevance to elements that are common across many languages, like certain keywords (`void`, `new`) and string literals ([recommended by the language definition guide](https://highlightjs.readthedocs.org/en/latest/language-guide.html#relevance)). Is this incorrect?
2. High relevance: I gave certain language annotations a high relevance, because I think they’re the most reliable way to recognize Ceylon code, which otherwise looks a lot like many other curly-brace languages. If I remove these relevances, the sample code is misidentified as other languages (`|3` works, `|2` doesn’t). Would it be acceptable to keep these relevances?

lucaswerkmeister:
(merged `master`, should now merge cleanly again. let me know if you’d prefer a rebase instead)

quintesse:
Great work @lucaswerkmeister ! Thanks.
Could this be merged @isagalaev ? We're trying to get Gitter to correct the horrible highlighting our code has right now ;)

CHANGES.md
```diff
@@ -20,8 +21,13 @@ Notable fixes and improvements to existing languages:
 
 [Bram de Haan]: https://github.com/atelierbram
 [Kenneth Fuglsang]: https://github.com/kfuglsang
+<<<<<<< HEAD
```
Sannis: Please fix this merge artifact.

Sannis:
LGTM. Anyway we can play with relevance boosters in master after merge.

CHANGES.md
lucaswerkmeister: oops, sorry

lucaswerkmeister: I was looking for more overlooked artifacts. I don’t think there’s any more by me, but I found some others:
```
$ grep -r --binary-files=without-match '<<<< '
node_modules/gear-lib/node_modules/less/test/less/property-name-interp.less.orig:<<<<<<< HEAD
node_modules/gear-lib/node_modules/less/lib/less/tree/rule.js.orig:<<<<<<< HEAD
node_modules/gear-lib/node_modules/less/lib/less/tree/rule.js.orig:<<<<<<< HEAD
node_modules/gear-lib/node_modules/less/lib/less/parser/parser.js.orig:<<<<<<< HEAD:lib/less/parser/parser.js
```
Are those intentional?

lucaswerkmeister:
Great, thanks!

gavinking:
Excellent!

isagalaev:
JFYI, about relevance and all: I did some irreparable improvements to the language here: https://github.com/isagalaev/highlight.js/compare/dde27ca...10ea4dd. Mainly it's resetting all relevances to default values but instead introducing string substitutes as a real mode.

PR #828, opened 2015-04-13
Add 'async' to ES6/7 keywords

marcominetti:
You already added 'await' keyword, but it is actually used with 'async' one in function declaration by ES7 draft specification.


Sannis:
Thanks, merged!

marcominetti:
Thank you! ;)

Issue #821, opened 2015-04-09
Support DNS Zone files


enko:
It would be cool if highlight.js would support [dns zone files](http://en.wikipedia.org/wiki/Zone_file). I'm using highlight.js in my blog and have written an [article about dnssec](https://datenknoten.me/blog/dnssec-mit-bind-und-ovh) and especialy the dnskey example does not look so nice.

isagalaev:
We'd be glad to have it supported too: http://highlightjs.readthedocs.org/en/latest/language-requests.html

PR #834, opened 2015-04-19
Add support for DNS zone files.

enko:
As suggested in #821, I added initial support for dns zone files.


src/languages/dns.js
```diff
@@ -0,0 +1,33 @@
...
+    keywords: {
+      keyword:
+        'IN A AAAA AFSDB APL CAA CDNSKEY CDS CERT CNAME DHCID DLV DNAME DNSKEY DS HIP IPSECKEY KEY KX LOC MX NAPTR NS NSEC NSEC3 NSEC3PARAM PTR RRSIG RP SIG SOA SRV SSHFP TA TKEY TLSA TSIG TXT'
+    },
+    contains: [
+      hljs.COMMENT(';', '$'),
+	  {
+		className: 'operator', 
+		begin: '\\$(TTL|GENERATE|INCLUDE|ORIGIN)',
+		relevance: 10
```
Sannis: Is this so necessary? Maybe we can just add `keywords: 'TTL GENERATE INCLUDE ORIGIN'` to this mode?

src/languages/dns.js
```diff
@@ -0,0 +1,33 @@
...
+function(hljs) {
+  return {
+    aliases: ['bind','zone'],
+    keywords: {
+      keyword:
+        'IN A AAAA AFSDB APL CAA CDNSKEY CDS CERT CNAME DHCID DLV DNAME DNSKEY DS HIP IPSECKEY KEY KX LOC MX NAPTR NS NSEC NSEC3 NSEC3PARAM PTR RRSIG RP SIG SOA SRV SSHFP TA TKEY TLSA TSIG TXT'
+    },
+    contains: [
+      hljs.COMMENT(';', '$'),
+	  {
```
Sannis: Some bad indent, but we cant fix this after merge :-)

src/languages/dns.js
```diff
@@ -0,0 +1,33 @@
+/*
+Language: DNS Zone file
+Author: Tim Schumacher <tim@datenknoten.me>
+Category: common, config
```
Sannis: IMO, this is not so wide using "language" to be included in common set.

Sannis:
Thanks! Looks good for me except comment above.

src/languages/dns.js
```diff
@@ -0,0 +1,33 @@
...
+      {
+        className: 'number',
+        begin: '((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)(\\.(25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]?\\d)){3}))|:)))'
```
Sannis: Can you also split this regexp into few separate?

src/languages/dns.js
enko: I copied this from nginx.js, since both have more or the same featureset. What should I do?

src/languages/dns.js
enko: I took the expression from here:

http://nbviewer.ipython.org/github/rasbt/python_reference/blob/master/tutorials/useful_regex.ipynb#Checking-for-IP-addresses

I tried to comprehend the expression, but failed, but I did some testing and it worked for IPv4 & IPv6 addresses.

src/languages/dns.js
isagalaev: @enko just remove the "common" category (it's treated specially internally).

Sannis:
Thanks, merged!

PR #837, opened 2015-04-21
Add support for '.' in C# namespaces

DavidAnson:
First PR for highlight.js, please let me know anything that could be done better or that you'd like changed for whatever reason. (For example, this is a tiny change and I'm happy to omit the edit to AUTHORS.en.txt.) Without this edit, C# code that includes a dotted namespace (as in the example file) is not recognized as C#.


Sannis:
Thanks, merged!

PR #838, opened 2015-04-21
Add .gitattributes to test directory to avoid failures in CRLF environme...

### LINKED ISSUE #836: Multiple tests fail if run in an environment where core.autocrlf is set to true and the OS defaults to crlf
The manual workaround (for Windows users) is to run `git config core.autocrlf input` and re-checkout the repository files. It would be nice if the repository included a `.gitattributes` file that tagged the relevant files `eol=lf` so tests passed by default regardless of platform.

I can submit a pull request if there's interest in having this fixed.

The first test failure (there are more):

```
  1) hljs .highlightAuto should be detected as http:

      AssertionError: expected 'groovy' to be 'http'
      actual expected

      "httpgroovy"

      at D:\T\highlight.js\test\highlightAuto.js:27:21
      at Array.forEach (native)
      at Context.<anonymous> (D:\T\highlight.js\test\highlightAuto.js:20:14)
```

DavidAnson:
...nts (fixes #836)

Note: I started by enumerating the problematic files one-by-one, but decided after a few of them that the list seemed arbitrary and might confuse people. With this approach, *everything* in test is LF-only regardless of platform.

Regarding test/detect/less/default.txt, it seems to have been added with CRLF line endings (i.e., not as a text file), so I special-cased it to avoid bogus "change" detection by GIT. If you'd rather, I could submit it with this pull request as a text file so it doesn't need to be treated differently.


isagalaev:
Thanks, David! And yes, there's no reason to not keep all of them uniform, so please so convert the Less test as you suggested.

DavidAnson:
Done and rebased to latest master. Should be good to go.

isagalaev:
Merged. Thanks!

PR #844, opened 2015-05-04
Fix gherkin tables carrying on to next lines

RX14:
Also fixed string highlighting applying to the whole table, not just the columns.

Before:

After:


isagalaev:
Thanks a bunch!

RX14:
NP :wink:

PR #846, opened 2015-05-09
First pass at updating scala mode to new-styles.

non:
This commit standardizes some of the style names,
and adds other names for which no good equivalent
existed. It also updates the mode to handle string
interpolation/substitution, and removes some
unnecessary code from scala/default.txt.

Highlighting appears to be correct in the Pojoaque
style currently.

Review by @isagalaev.

### DIFF
--- a/docs/css-classes-reference.rst
+++ b/docs/css-classes-reference.rst
@@ -28,6 +28,15 @@ Stylable classes
 | variable                 | Variable in a config or a template file.          |
 |                          | Environment var expansion in a script.            |
 +--------------------------+---------------------------------------------------+
+| type                     | Type name used in a function signature, type      |
+|                          | parameter, generic template, etc.                 |
++--------------------------+---------------------------------------------------+
+| class                    | Class declaration                                 |
++--------------------------+---------------------------------------------------+
+| function                 | Function or method declaration                    |
++--------------------------+---------------------------------------------------+
+| title                    | Name defined within class or function declaration |
++--------------------------+---------------------------------------------------+
 
 
 (...)
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -1,23 +1,53 @@
 /*
 Language: Scala
+Category: functional
 Author: Jan Berkel <jan.berkel@gmail.com>
 Contributors: Erik Osheim <d_m@plastic-idolatry.com>
 */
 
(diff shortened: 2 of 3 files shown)


isagalaev:
That's a good start! I've merged your changes so far but I've got a few comments:

- I don't know what a SYMBOL is and I don't see it anywhere in the snippet. What is it?

- Currently TYPE is defined as anything starting with a capital letter, so in `def gsum[A: Ring](as: Seq[A]): A =` it highlights "A" which is probably wrong? Anyway, I'd like to get rid of highlighting plain user identifiers everywhere as they usually don't play a special role to the reader. I think it's better to define built-in language types in the keywords sections as "built_in".

- I also rewrote the whole strings/subst section using variants, take a look! As a side effect, backslash escapes are no longer considered substs (as is the case in other languages too).

non:
Thanks. To answer your questions:

* Scala has a syntax for literal symbols: `'foo` is a symbol, which is distinct from `"foo"` (a string), or `foo` (a bare word). I can add some symbols to the snippet if you'd like.

* Actually in that example `A` definitely should be a type -- it is a type parameter (a variable for types), in the same way that `Seq` is a type (a parameterized type). My instinct is that distinguishing values from types is pretty important in most Scala code, and Scala doesn't really have a concept of "built-in" types as distinct from user-defined types.

* Oooh, I like the variants. Thanks! As far as not highlighting backslash escapes, I like seeing them, but I understand that you prefer to leave them out for consistency and simplicity.

Thanks for merging the PR and for improving my code.

I am happy to submit another PR to add symbols to the snippet. Is there anything else you'd like to see? I would prefer to leave the type highlighting in, but if you have decided not to highlight static types in other languages (like C, Java, etc) then I understand the desire to remove them here.

isagalaev:
Yes, please do add symbols to the snippet. Also, I think that they warrant a special class, namely "symbol", as this thing is pretty common in one form or another throughout  programming languages. The generic "name" doesn't really cut it in this case.

I see your argument about types, so let's keep'em in for now. They aren't highlighted in C or Java, as they aren't that special there (and besides, can't be highlighted reliably, as they don't follow any syntactic convention, they're just identifiers). On the other hand, we do highlight them in Haskell, so this is probably something that should be decided on a language by language basis.

PR #849, opened 2015-05-13
Improve support for symbols in Scala.

non:
This commit does two things:

 1. Creates a syntactic class called "symbol"
 2. Adds symbols to the Scala snippet

As before, it seems like the Pojoaque mode is highlighting the Scala
acceptably.

Review by @isagalaev


isagalaev:
Looks great. I'll start reworking styles now, keeping Pojoaque in mind as a reference point.

PR #864, opened 2015-06-02
Add Toml language

GuillaumeGomez:


GuillaumeGomez:
An example of the result can be found on my [blog](http://blog.guillaume-gomez.fr/Rust/3/1) (sorry, it's written in french :p).

isagalaev:
Yes, we want tests too, here's a quick guide on making a test snippet: http://highlightjs.readthedocs.org/en/latest/building-testing.html#basic-testing

Merci beaucoup!

GuillaumeGomez:
I'll add them this evening then.

De rien, c'est un plaisir ! ;-)

src/languages/toml.js
```diff
@@ -0,0 +1,33 @@
+/*
+Language: Toml
+Author: GuillaumeGomez <guillaume1.gomez@gmail.com>
+Category: common
```
Sannis: Category should be `config` IMO. @isagalaev?

GuillaumeGomez: You're totally right. I change it.

GuillaumeGomez:
I added tests but I didn't test yet. I'll do it as soon as possible.

sourrust:
You need an example of toml in `test/detect/toml/default.txt`. Follow the [contributor check][1] in the documentation.

[1]: https://highlightjs.readthedocs.org/en/latest/language-contribution.html

GuillaumeGomez:
Hum... It seems I have an issue with nix language.

isagalaev:
I've fixed it in a quite unexpected way :-)
While doing inter-language relevance tuning triggered by Toml (these things tend to cascade) I realized that the only good way to do it is to redefine Toml in the same way as Ini is defined right now: with a rigid structure that would both give good relevance to Toml code as well as prevent Toml to recognize other code. And then I realized that Toml is a superset of Ini and there's no point in having them both in two separate files. So I merged all the interesting bits from toml.js into ini.js, updated the Ini snippet and attributed Guillaume's contribution in relevant headers and files.

GuillaumeGomez:
Wo, impressive ! Thanks a lot ! I'm totally taking a look at the result !

PR #868, opened 2015-06-14
Add Elm support

jvoigtlaender:
This adds support for the [Elm language](http://elm-lang.org/). The definitions are derived from the existing ones for Haskell, since the two languages are syntactically very similar.

@evancz (the designer of Elm), can you double check I got the keyword lists right?


evancz:
LGTM, thanks @jvoigtlaender!

Sannis:
Thanks, merged!

jvoigtlaender:
Thanks for merging it in!

PR #876, opened 2015-06-26
C++: fix highlighting of identifiers containing _t

### LINKED ISSUE #875: C++: incorrect highlighting of identifiers that contain _t
It seems [`CPP_PRIMATIVE_TYPES`](https://github.com/isagalaev/highlight.js/blob/master/src/languages/cpp.js#L9-L12) is a bit overzealous:

Imgur

If it's OK to revert the commits (aa342e01...badb33fb) this was introduced in I can submit a PR.

jcmoyer:
Before:

Imgur

After:

Imgur

Fixes #875

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -37,7 +37,6 @@ function(hljs) {
     keywords: CPP_KEYWORDS,
     illegal: '</',
     contains: [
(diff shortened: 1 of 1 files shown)


sourrust:
Thanks for the pull request but I fixed this with cc9b121.

PR #887, opened 2015-07-20
Added support for AutoIt language

J2TeaM:


isagalaev:
Thanks for the contribution! We'll look into it (and the build failure) as time allows.

isagalaev:
Merged! I fixed the detection issues that were breaking tests. All of them were caused by overly high relevance of constants, variables and comments (c706ab4c0), it's best to not define any special relevance unless strictly needed.

Also, if you don't mind, could you strip some code from the test snippet? It's also used on a demo page and looks best when it's short, a dozen lines or so. It shouldn't be a working code even, only give a taste of the language.

Thanks!

J2TeaM:
Thank you, @isagalaev! I've edited it into a shorter code and sent a pull request.

PR #894, opened 2015-07-31
JS: Remove illegal characters in func params for ES6 default params

STRML:
This is related to [this redux discussion](https://github.com/gaearon/redux/issues/345#issuecomment-126710039).

Highlighting fails when ES6 default parameters are used:

imgur

The characters defined as `illegal` in function parameters are no longer illegal. Added a test case.

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -81,8 +81,7 @@ function(hljs) {
             contains: [
               hljs.C_LINE_COMMENT_MODE,
               hljs.C_BLOCK_COMMENT_MODE
-            ],
-            illegal: /["'\(]/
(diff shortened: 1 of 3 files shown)


sourrust:
Merged in. Thanks for contributing.

isagalaev:
A note to self, mostly: it may be a good idea to still leave those in but instead define a `{ }`-swallowing mode…

knowbody:
any chance this could be released? thanks

Sannis:
We now have scheduled releases every 6 weeks, next one will be in early September.
You can build highlight.js from master if this is critical for your project.

PR #899, opened 2015-08-06
Add async/await keywords to Python language.

1st1:
async/await were added in Python in its 3.5 version (scheduled for
release in September 2015.

See https://www.python.org/dev/peps/pep-0492/ for more details.


sourrust:
All merged in. Thanks for contributing.

PR #911, opened 2015-08-18
Update to new styles

Sannis:
- [x] fortran
- [x] irpf90
- [x] golo
- [x] kotlin
- [x] nsis
- [x] step21
- [x] tp


Sannis:
@isagalaev you can merge this at any moment - I'll just create new branch for next bunch of languages.

isagalaev:
Yes, I should be able to look into this today!

isagalaev:
I've got one general comment.

I still haven't made my mind about calling built-in types by the "type" class name. This was intended for languages like Haskell and Scala where a type is an important concept and they are mostly user-defined, rather than being just built-in primitive types. So in languages like Java I just stick those with the other keywords. 

If this explanation does make sense, let's keep it this way for now. Also kotlin.js has many things called "type" (like stuff between `<..>` and after `:` in a function definitions) that should really be class-less.

Sannis:
I used to work on such tasks iteratively, so I set a goal to reduce CSS classes number.
But this makes sense, so I'll update this PR with this new concept.

Sannis:
@isagalaev what about `literal`s? IMO we can merge them into keywords list. But I have no thoughts about builtints.

isagalaev:
The difference between keywords and literals is quite clear: keywords define code structure, literals are just special values, along with numbers and strings. As long as the description is clear I prefer to keep things separate as it is always possible to just color them the same way if needed.

Sannis:
@isagalaev, I've merged built-in types into keywords for these languages.

Sannis:
Nice!

isagalaev:
Only three left now :-) And a bunch of other stuff (docs, cleanup, etc.)

PR #917, opened 2015-08-24
Hopscotch version bump to 1.0.1

idleberg:


Sannis:
Oh, this PR title was scare me :-)

isagalaev:
Uhm… Sorry, looks like I'll have to ruin it all and revert the patch. We deliberately don't include all this stuff in our styles to keep them consistent with each other (jumping fonts and text measurements on style switching look ugly) and to not impose the choice of fonts on users. Of all these changes, probably only `direction: ltr` does make sense, but it should go in every style then, not just one.

On a related note: the new-styles branch is nearing completion so now it makes sense to do any style updates against that branch instead of master.

idleberg:
Ah, I thought of the possibility, but then I saw the CodePen theme using a different font as well. Nevermind, if that's policy that's fine.

isagalaev:
> I saw the CodePen theme using a different font as well

I fully admit I wasn't very diligent in keeping them all in check :-). The new-styles is going to bring a lot of unification and I'll fix those common rules for every style too.

PR #926, opened 2015-09-07
Add highlighting for the crmsh DSL

krig:
crmsh is the configuration language used by
Linux HA clusters. It is documented here:

http://crmsh.github.io/man/#cmdhelp_configure


Sannis:
Thanks!

Issue #730, opened 2015-02-03
(elixir) interpolation, sigils and / functions


eksperimental:
### Update 11/2019: As far as I know #2257 leaves only item 3 open.

- [x] 1) **Interpolation**: It cannot deal properly with interpolation.
- [x] 2) **Sigils** : It cannot deal with these [`~S sigil`](http://elixir-lang.org/getting-started/sigils.html#interpolation-and-escaping-in-sigils). `~s` is fine though.
- [x] 3) **Function with a slash ( "Kernel./" )**: this one takes `/:` as the beggining of a regular expression (Closed via #2406)

---

We are adding hightlightjs to https://github.com/elixir-lang/ex_doc/
and when I ran it against the [test file for pygments](https://bitbucket.org/birkenfeld/pygments-main/src/tip/tests/examplefiles/example_elixir.ex).
In addition to #707, I have run into three new issues:

1) **Interpolation**: It cannot deal properly with interpolation.

Every single line bellow, has an issues, it takes it as a comment whatever comes after the #

``` elixir
~R'this + i\s "a" regex too'
~w(hello #{ ["has" <> "123", '\c\d', "\123 interpol" | []] } world)s
~W(hello #{no "123" \c\d \123 interpol} world)s
~s{Escapes terminators \{ and \}, but no {balancing} # outside of sigil here }
~S"No escapes \s\t\n and no #{interpolation}"
```

2) **Sigils** : It cannot deal with these [`~S sigil`](http://elixir-lang.org/getting-started/sigils.html#interpolation-and-escaping-in-sigils). `~s` is fine though.

``` elixir
defmodule Long.Module.Name do
  @doc ~S'''
  No #{interpolation} of any kind.
  \000 \x{ff}

  \n #{\x{ff}}
  '''
  def func(a, b \\ []), do: :ok

  @doc ~S"""
  No #{interpolation} of any kind.
  \000 \x{ff}

  \n #{\x{ff}}
  """
  def func(a, b \\ []), do: :ok
end
```

3) **Function with a slash ( "Kernel./" )**: this one takes `/:` as the beggining of a regular expression

``` elixir
import Kernel, except: [spawn: 1, +: 2, /: 2, Unless: 2]
```

You can see this live, 
- Original test doc page, with the problems: http://eksperimental.github.io/issues/highlightjs/Highlight.Elixir.html#pygments/0
[...]

PR #927, opened 2015-09-09
(elixir) Two of three issues fixed (which issues?)

Sannis:
Relates to #730.

@eksperimental, @isagalaev I suggest to merge this without full sigils support.

### DIFF
--- a/src/languages/elixir.js
+++ b/src/languages/elixir.js
@@ -50,11 +50,6 @@ function(hljs) {
     CLASS,
     FUNCTION,
     {
-      className: 'constant',
-      begin: '(\\b[A-Z_]\\w*(.)?)+',
-      relevance: 0
-    },
-    {
       className: 'symbol',
       begin: ':',
       contains: [STRING, {begin: ELIXIR_METHOD_RE}],
--- a/src/languages/elixir.js
+++ b/src/languages/elixir.js
@@ -44,33 +44,64 @@ function(hljs) {
     className: 'class',
     beginKeywords: 'defmodule defrecord', end: /\bdo\b|$|;/
   });
-  var ELIXIR_DEFAULT_CONTAINS = [
-    STRING,
-    hljs.HASH_COMMENT_MODE,
-    CLASS,
-    FUNCTION,
-    {
-      className: 'symbol',
-      begin: ':',
-      contains: [STRING, {begin: ELIXIR_METHOD_RE}],
-      relevance: 0
-    },
+
+  var ELIXIR_NUMBER = {
+    className: 'number',
+    begin: '(\\b0[0-7_]+)|(\\b0x[0-9a-fA-F_]+)|(\\b[1-9][0-9_]*(\\.[0-9_]+)?)|[0_]\\b',
+    relevance: 0
+  };
+
+  var ELIXIR_SYMBOL_MODES = [
     {
       className: 'symbol',
       begin: ELIXIR_IDENT_RE + ':',
       relevance: 0
     },
     {
-      className: 'number',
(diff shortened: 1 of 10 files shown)


eksperimental:
thank you :100: times @Sannis

Sannis:
Current state:
screen shot 2015-09-10 at 13 06 29
I nedd to with issues with comments in sigils before merging this.

eksperimental:
@Sanni sorry, but it is not clear what you say you need in your last comment.

here are some comments:

This is what the Elixir interactive shell (`iex`) interprets:

```elixir
iex(1)> ~R'this + i\s "a" regex too'
~r/this + i\s "a" regex too/

iex(2)> ~w(hello #{ ["has" <> "123", '\c\d', "\123 interpol" | []] } world)s
["hello", "has123c\d123", "interpol", "world"]

iex(3)> ~W(hello #{no "123" \c\d \123 interpol} world)s
["hello", "\#{no", "\"123\"", "\\c\\d", "\\123", "interpol}", "world"]

iex(4)> ~s{Escapes terminators \{ and \}, but no {balancing} # outside of sigil here }
"Escapes terminators { and }, but no {balancing"

iex(5)> ~S"No escapes \s\t\n and no #{interpolation}"
"No escapes \\s\\t\\n and no \#{interpolation}"
```

1. is working perfect!
2. Sigils, here is a link to [~w](http://elixir-lang.org/docs/master/elixir/Kernel.html#sigil_w/2) and [~W](http://elixir-lang.org/docs/master/elixir/Kernel.html#sigil_W/2). the w/W sigil splits everything inside the parenthesis taking a white space as a separator. These sigils can take a modifier at the end (that the "s" you see there).  The difference is that the lowercase version does interpolationation and character unescaping, and the uppercase version takes everything literal.

One thing to mention, is that elixir uses `#{ }` to interpolate variables and functions insie other strings. So not always everything that comes after `#` should be considered a comment.

so `~W(hello #{no "123" \c\d \123 interpol} world)s` will split the text inside the parenthesis by the whitespace. so "hello" "#{no" "\"123\"" ...etc. will be the elements of the list returns. this doesn't really matter for the highlighting, but just to explain it better.

You can see the example of how pygment highlights it as a reference: http://pygments.org/demo/1180634/

Example no.4: is perfect

Example no.5 : I think `#{interpolation}` should not be highlighted differently since it has no special meaning in `~S`. if the sigil would be ~s, It would be alright to highlight it different (interpolation will be a function or a variable)

Here's a modified version of the example above, using `~s` sigil.
```elixir
iex(6)> ~s"No escapes \s\tTabbed\n and no #{"interpolation"}"
"No escapes  \tTabbed\n and no interpolation"
iex(7)> IO.puts ~s"No escapes \s\tTabbed\n and no #{"interpolation"}"
No escapes      Tabbed
 and no interpolation
:ok
```
Characters are unescaped

Let me know if you need clarification on more issues.
I'm glad to help, or I can bring more people in, if you need further input that I am not able to provide.

thanks again for looking after this issue.

eksperimental:
@Sannis hi, just poping back in to see if there is any more feedback from my side to help have this PR merged in master? please let me know if you need anything,
thanks for looking after this.

eksperimental:
oh by the way, there was one more issue that was discovered by @memory: https://github.com/isagalaev/highlight.js/issues/730#issuecomment-157696731
> `&fun!/1` is interpreted as a regex

yyyc514:
@Sannis You still around?  Could you add a summary in your original post about what 2 or 3 items this is improving so I know what to be looking for when reviewing this?

yyyc514:
Ping. Anyone there?

isagalaev:
> Anyone there?

Apparently not :-) I am frankly too far removed from thinking about this again, so I'd say use your best judgment what to do with it.

yyyc514:
Closing in favor of: https://github.com/highlightjs/highlight.js/pull/2257

Issue #348, opened 2013-12-28
Unification of class names across languages


isagalaev:
## Problem

>  One of the early design principles for highlight.js was having just a few common class names in order to have universal styles that would work for any language. Unfortunately this principle wasn't strongly enforced. We now have language-specific classes, language-specific style rules and whole language-specific styles. This is a maintenance nightmare as the number of unique conditions that should be visually tested is a production of the number of unique language features and the number of styles. And since this is an impossible amount of work, we usually test only a small subset of languages and styles and rely on pure luck for the rest which is a majority. ([source][])

## Principles

- Class names should have stronger semantics. A "variable" in PHP is hardly worth highlighting at all while a "variable" in Prolog is essential for understanding the code and is as important as, say, function definitions in other languages. This means that we'll have to invent a separate name for every case that has to be treated differently in different languages.

- The set of available class names should be rich enough for all languages and adding a new one after we convert all we currently have should require a serious reason.

- The class reference becomes a flat list of available class names with explanations of their meaning.

- All styles become language-agnostic. We may even drop generating class names for languages altogether.

## Process

There is a a long lived branch where the conversion is slowly happening: https://github.com/isagalaev/highlight.js/tree/new-styles.

It should eliminate all language names from CSS files. Every current class name should also be evaluated and many will be either dropped or change their semantics. At every such change (removal of a language specific rule or redefinition of a class name) we should make sure that *all* styles look reasonable. This means that we'll definitely loose similarity with other highlighter engines emulated by various styles, but that's an inevitable implication of the whole task.

During the whole conversion process we will have to intelligently port new languages and styles into the new rules. Possibly encouraging contributors to do that :-)

[source]: http://softwaremaniacs.org/blog/2014/07/26/highlight-js-what-next/en/

## Status

- [x] 1c
- [x] accesslog
- [x] actionscript
- [x] apache
- [x] applescript
- [x] asciidoc
- [x] armasm
- [x] aspectj
- [x] autohotkey
- [x] autoit
- [x] avrasm
- [x] axapta
- [x] bash
- [x] brainfuck
- [x] cal
- [x] capnproto
- [x] ceylon
- [x] clojure
- [x] clojure-repl
- [x] cmake
- [x] coffeescript
- [x] cpp
- [x] crmsh
- [x] crystal
- [x] cs
- [x] css
- [x] d
- [x] markdown
- [x] dart
- [x] delphi
- [x] diff
- [x] django
- [x] dns
- [x] dockerfile
[...]

PR #934, opened 2015-09-15
css-classes reference sql

SibuStephen:
SQL variable name added as a class reference.


Sannis:
Thanks!

BTW, we will unify CSS classes in version 9.0, look on #348.

PR #935, opened 2015-09-15
css-classes reference sql

SibuStephen:
SQL variable name added as a class reference.(`Variable` spelling is misplaced)


Sannis:
Thanks, fixed in master.

PR #940, opened 2015-09-22
Add missing Swift keywords.

natecook1000:
This catches a few Swift keywords that hadn't made it in, using the [Lexical Structure](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/doc/uid/TP40014097-CH30-ID413) section of Apple's documentation as a reference. (The only things I noticed missing were the `public`/`internal`/`private` access modifiers.)

### DIFF
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -8,14 +8,14 @@ Category: system
 
 function(hljs) {
   var SWIFT_KEYWORDS = {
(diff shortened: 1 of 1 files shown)


isagalaev:
Looks good! Thanks a bunch!

PR #946, opened 2015-10-01
crmsh: Improved command start detection

krig:
Found and fixed a corner case if a resource attribute has the same name as a command.


Sannis:
Thanks!

PR #950, opened 2015-10-04
Add SQF language

senevoldsen90:
This commit adds support for the SQF language. This is used in the Arma video game series.


Sannis:
Thanks!

PR #952, opened 2015-10-06
Dockerfile - Add suport for LABEL

abtris:


Sannis:
@abtris looks like your test example should be changed.

abtris:
Test fixed, but I'm not sure if possible eliminate problem if key after label contains label word too.

Sannis:
Thanks!

Issue #954, opened 2015-10-08
Windows freeze with a huge code block parsing


Soullivaneuh:
I set up Highlight.js to load 500 ms after window loading:

```js
jQuery(window).load(function() {
    setTimeout(function() {
        $('pre code').each(function(i, block) {
            hljs.highlightBlock(block);
        });
    }, 500);
});
```

I can scroll on the page, but it freezes when Highlight.js start parsing on a huge code block, like this one: https://gist.github.com/Soullivaneuh/acc9c0ec725f094bcfbd

Do you have a solution to load parsing "in background" and avoid this page freezing?

Using `8.8.0` version.

Sannis:
You can mark this block as no-highlight to avoid freeze.

```html
<pre><code class="nohighlight">...</code></pre>
```

Soullivaneuh:
> You can mark this block as no-highlight to avoid freeze.

Well, that's a workaround... or not, because this will just disable highlighting.

I want this piece of code to be highlighted, but in background with no window freezing.

isagalaev:
Probably the only thing you could try to do is to write your own initialization function that would call `hljs.highlight()` repeatedly on chunks of the input string of some manageable size. This function returns a parsing state that you can [pass back as a 4th parameter][1] to pick up parsing from the previous point.

[1]: http://highlightjs.readthedocs.org/en/latest/api.html#highlight-name-value-ignore-illegals-continuation

Soullivaneuh:
@isagalaev Thanks for the tip.

But I'm not sur to undersand well the way to do.

Could you please provide a code sample based on mine?

frigus02:
Once we have web worker support (https://github.com/isagalaev/highlight.js/pull/963), you can highlight your code blocks in a worker asynchronously without freezing the page.

You will have to create a separate JavaScript file for the worker, which could look like this:

```javascript
importScripts('../libs/highlight.pack.min.js');

self.onmessage = function (event) {
    var result = hljs.highlightAuto(event.data);
    postMessage(result);
};
```

Then you can use it with this modified snippet of yours:

```javascript
jQuery(window).load(function() {
    $('pre code').each(function(i, block) {
        var code = $(block).text(),
            worker = new Worker('js/myworker.js');

        worker.onmessage = function (event) {
            var result = event.data;
            $(block).addClass('language-' + result.language).html(result.value);
            worker.terminate();
        };

        worker.postMessage(code);
    });
});
```

[...]

PR #963, opened 2015-10-23
Add support for use inside of a web worker

frigus02:
I want to use highlight.js inside of a web worker, because I need to highlight large code blocks and don't want to do this on the main thread. All it takes to support web workers is make a small change in the factory function.

WIth this change I can just use highlight.js in my worker:

```javascript
importScripts('bower_components/highlightjs/highlight.pack.min.js');

self.onmessage = function (e) {
    var result = hljs.highlight(e.data.language, e.data.code);
    return result.value;
};
```

This works great for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "jsdom": "3.1.2",
     "lodash": "^3.3.1",
     "mocha": "^2.0.1",
-    "should": "^7.0.1"
+    "should": "^7.0.1",
+    "webworker-threads": "^0.6.2"
   }
 }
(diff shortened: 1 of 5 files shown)


Sannis:
Nice, this is one of the solutions for #954.

Sannis:
If you can add some tests for that or have any ideas abouit that - you are welcome! If not - I'll merge this soon.

Sannis:
@isagalaev, should we add any usage examples like this to readme or docs?

frigus02:
I can add some tests I think. Could you point me to the correct tests file/folder. Or will this be a completly new one?

frigus02:
I added a test for the web worker support. Please let me know, if this fits.

Sannis:
@frigus02 good work, thanks!

Soullivaneuh:
Is that pushed on a new stable release?

Sannis:
Yes, this will be released in 8.10.

PR #964, opened 2015-10-24
Update GLSL syntax

tikhomirov:
Add qualifiers, built-in functions, variables and constants for all GLSL versions up to and including 4.30 and GLSL ES 1.00 – 3.00

Also update author's email address :)


Sannis:
Thanks!

PR #969, opened 2015-10-29
Add MIPS assembly as a language

Nebuleon:
This should cover a lot of the MIPS assembly syntax, including opcodes, registers and their aliases, literals and comments, up to MIPS32R2. Everything was taken from volume 2 of the MIPS architecture manuals.

Missing here are the names of MIPS32R3 and later opcodes, MIPS64 and later, MSA, MDMX, MIPS 3D and other application-specific extensions.

### DIFF
--- a/src/languages/mipsasm.js
+++ b/src/languages/mipsasm.js
@@ -1,0 +1,91 @@
+/*
+Language: MIPS Assembly
+Author: Nebuleon Fumika <nebuleon.fumika@gmail.com>
+Description: MIPS Assembly (up to MIPS32R2)
+Category: assembler
+*/
+
+function(hljs) {
+    //local labels: %?[FB]?[AT]?\d{1,2}\w+
+  return {
+    case_insensitive: true,
+    aliases: ['mips'],
+    lexemes: '\\.?' + hljs.IDENT_RE,
+    keywords: {
+      meta:
+        //GNU preprocs
+        '.2byte .4byte .align .ascii .asciz .balign .byte .code .data .else .end .endif .endm .endr .equ .err .exitm .extern .global .hword .if .ifdef .ifndef .include .irp .long .macro .rept .req .section .set .skip .space .text .word .ltorg ',
+      built_in:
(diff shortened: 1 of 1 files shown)


Sannis:
@Nebuleon suddenly there is mis-detection with already existing ASM dialect. Can you make test samples more specific?

Nebuleon:
Assembly languages have many things in common: the use of commas, the similar ways to form numbers, symbols and register names, and some opcode names (ADD, SUB, MUL, DIV, OR, XOR, AND, in particular).

I could give you some code samples that use the numeric and named registers for MIPS, demonstrates stack usage with `SW` and `LW`, and maybe one that is the output of a compiler so that more linker directives appear.

What's the best code license to put the samples under?

src/languages/mipsasm.js
```diff
@@ -0,0 +1,91 @@
...
+            'p[lu][lu]\.ps|recip\.fmt|r?sqrt\.[ds]|sdx?c1|sub\.[ds]|suxc1|' +
+            'swx?c1|' +
+            // system control instructions
+            'break|cache|d?eret|[de]i|ehb|mfc0|mtc0|pause|prefx?|rdhwr|' +
+            'rdpgpr|sdbbp|ssnop|synci?|syscall|teqi?|tgei?u?|tlb(p|r|w[ir])|' +
+            'tlti?u?|tnei?|wait|wrpgpr'+
+        ')',
+        end: '\\s'
+      },
+      hljs.COMMENT('[;#]', {relevance: 0}),
```
Sannis: Second argument missing here. If added, current test suite passed.

Nebuleon: The second argument to `hljs.COMMENT` is `{relevance: 0}`. Is that what you mean? Or am I missing a named argument inside the `{}`, and if so, what is its name?

isagalaev:
The second argument for `hljs.COMMENT()` is the ending expression. Without it comments were being ended right after they were started. Incidentally, this was what caused the mis-detection with avrasm. I've fixed this along with another mis-detection and added a MIPS snippet for testing and the demo.

PR #973, opened 2015-11-04
Update building-testing.rst: add ES6 note

ZitRos:
It is not mentioned in dev docs that ES6 is required for building the project.


ZitRos:
Didn't notice world moved forward and brought node v5 when I was still using v0.12.*

Issue #892, opened 2015-07-27
YAML Support


tedivm:
It would be fantastic if this library supported yaml.

Sannis:
Hi! This is not exactly how it works, here's how: http://highlightjs.readthedocs.org/en/latest/language-requests.html

Issue #835, opened 2015-04-20
Feature: Add styles for yaml files


simonschaufi:
Example and official site: http://yaml.org/

isagalaev:
Hi! This is not exactly how it works, here's how: http://highlightjs.readthedocs.org/en/latest/language-requests.html

Issue #119, opened 2012-08-23
Add YAML


gka:
YAML syntax highlighting would be really nice.

Awesome lib, btw! Thanks so much.

wwi-johnbuhay:
isagalev,

Here is some info, let me know if I can do anything else to help. Thank you

http://docs.ansible.com/YAMLSyntax.html
https://github.com/clifford-github/sublime-ansible

isagalaev:
@wwi-johnbuhay new languages in highlight.js don't happen this way, here's how: http://highlightjs.readthedocs.org/en/latest/language-requests.html

lukeroge:
This issue server as at least an expression of interest in case a
contributor wants to submit a language.

PR #976, opened 2015-11-07
YAML syntax added

zealot128:
Hi!
I've added a basic yaml syntax highlighting. It handles:

* YAML literals
* Keywords / hash keys
* Special blocks, like "---", references, comment,
* different Multi-line strings
* Special highlighting: YAML references, Ruby ERB tags, Ansible Jinja Tags, Ruby I18n

It ignores indenting, though

example output:

bildschirmfoto 2015-11-07 um 20 35 45

relevant issues: #892 #119 #835

### DIFF
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -3,7 +3,7 @@ Language: YAML
 Author: Stefan Wienert <stwienert@gmail.com>
 Requires: ruby.js
 Description: YAML (Yet Another Markdown Language)
-Category: common, protocols
+Category: common, config
 */
 function(hljs) {
   var LITERALS = {literal: '{ } true false yes no Yes No True False null'};
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -1,0 +1,90 @@
+/*
+Language: YAML
+Author: Stefan Wienert <stwienert@gmail.com>
+Requires: ruby.js
+Description: YAML (Yet Another Markdown Language)
+Category: common, protocols
+*/
+function(hljs) {
+  var LITERALS = {literal: '{ } true false yes no Yes No True False null'};
+
+  var keyPrefix = '^[ \\-]*';
+  var keyName =  '[a-zA-Z_][\\w\\-]*';
+  var KEY = {
+    className: 'attr',
+    variants: [
+      { begin: keyPrefix + keyName + ":"},
+      { begin: keyPrefix + '"' + keyName + '"' + ":"},
+      { begin: keyPrefix + "'" + keyName + "'" + ":"}
+    ]
+  };
+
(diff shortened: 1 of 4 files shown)


src/languages/yaml.js
```diff
@@ -0,0 +1,90 @@
...
+Description: YAML (Yet Another Markdown Language)
+Category: common, protocols
+*/
+function(hljs) {
+  var LITERALS = {literal: '{ } true false yes no Yes No True False null'};
+
+  var keyPrefix = '^[ \\-]*';
+  var keyName =  '[a-zA-Z_][\\w\\-]*';
+  var KEY = {
+    className: 'keyword',
```
isagalaev: More appropriate class name for key names is 'attr'.

src/languages/yaml.js
```diff
@@ -0,0 +1,90 @@
...
+    variants: [
+      { begin: keyPrefix + keyName + ":"},
+      { begin: keyPrefix + '"' + keyName + '"' + ":"},
+      { begin: keyPrefix + "'" + keyName + "'" + ":"}
+    ]
+  };
+
+  var TEMPLATE_VARIABLES = {
+    className: 'subst',
```
isagalaev: This one is probably better served by 'template-variable'.

isagalaev:
Thanks for the contribution! YAML is one of the most glaring omissions in highlight.js, looking forward to merging it after you've addressed two comments I left against the diff.

zealot128:
thanks!
I've changed the classNames as you suggested.

zealot128:
I have also changed the category for YAML to [common, config ] (removed "protocol"; I copied the header from json.js originally)

Sannis:
Thanks for your contribution!

@isagalaev is it ok to add YAML to common category?

isagalaev:
Ah… No, I'd rather not add anything to :common without us finding some way of reducing the size of the bundle. I removed it in 01147f1.

estan:
Just a question about this feature: The example above shows keys in bold, but for me they are in normal weight:
 
highlightjs_yml_in_gogs

This is from the UI of Gogs (https://gogs.io) which is using highlightjs 9.6.0.

Was the YAML support changed so that keys are normal weight after this feature was merged?

estan:
This can be seen in the demo page at https://highlightjs.org/static/demo/ as well. The keys are not bold as in the example.

isagalaev:
This is style-dependent. There was a big overhaul of styles recently, and currently the default style along with many others don't apply any special highlighting to keys in such languages as JSON and YML. But there are others that do.

isagalaev:
However, to avoid vastness of non-highlighted YAML I just changed it to treat strings more consistently: b144fc6

estan:
@isagalaev: Excellent, thanks.

PR #978, opened 2015-11-10
Add built-in names introduced in Julia v0.4

bicycle1885:
Hi,

This pull request adds built-in names introduced in Julia v0.4, which is recently released.
Since v0.3 is still common, I didn't remove any names that exist in v0.3.
Also, I moved/removed a few names that didn't fit in their positions.


Sannis:
Thanks!

PR #985, opened 2015-11-15
New language add: Caché Object Script

ZitRos:
Hello developers!

Thank you for providing a great library for syntax highlighting. This time I have been working on implementing the [Caché Object Script language](http://docs.intersystems.com/ens20102/csp/docbook/DocBook.UI.Page.cls?KEY=GCOS_intro) (the primary language of [InterSystems'](http://www.intersystems.com/) technologies).

If you think this language is worthy to be in highlight.js library, I will be happy to hear any comments and see Caché Object Script in the library near release!

Thank you!

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -192,3 +192,4 @@ Contributors:
 - Ladislav Prskavec <ladislav@prskavec.net>
 - Jan Kühle <jkuehle90@gmail.com>
 - Stefan Wienert <stwienert@gmail.com>
+- Nikita Savchenko <zitros.lab@gmail.com>
--- a/AUTHORS.ru.txt
+++ b/AUTHORS.ru.txt
@@ -192,3 +192,4 @@ URL:   https://highlightjs.org/
 - Ладислав Пршкавеч <ladislav@prskavec.net>
 - Ян Кюхле <jkuehle90@gmail.com>
 - Стефан Вайнерт <stwienert@gmail.com>
+- Никита Савченко <zitros.lab@gmail.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## MASTER (UNRELEASED)
 
+New languages:
+
+- *Caché Object Script* by [Nikita Savchenko](https://github.com/ZitRos)
+
 Notable fixes and improvements to existing languages:
 
 - ECMAScript 6 modules import now do not require closing semicolon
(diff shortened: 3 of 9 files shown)


isagalaev:
@ZitRos I've tweaked your definitions somewhat in c69dcd7. I've a question about the `class##` thing: it doesn't look like a keyword at all (especially including the parens and everything within it). Is it some sort of a built-in function? I think it's okay to just leave it unstyled, as this double-# makes it stand out already, but may be you could think of an appropriate class name from [the list](https://github.com/isagalaev/highlight.js/blob/master/docs/css-classes-reference.rst).

isagalaev:
The more I look, the more it seems like a 'meta' to me. We use it for all sort of processor directives, annotations, decorators, etc.

ZitRos:
@Sannis, @isagalaev, thank you very much!

@isagalaev, the `##class` keyword in Caché Object Script stands for `##class(PackageName.ClassName).MethodName(MethodArgument)` - it is just a static reference to the class.

As for class names, I have chosen only those which do a better visual highlighting and tested it on different color themes. This is because in some cases the colors (styles) have no difference between different stylable classes. In other words, the usage of `class: "keyword"` for `##class` construction is only because of better styling.

Do you thing it is a wrong way?

Anyway, thank you a lot for your hints about the code!

isagalaev:
> the ##class keyword in Caché Object Script stands for ##class(PackageName.ClassName).MethodName(MethodArgument) - it is just a static reference to the class.

From this I'd say the best way forward is to highlight `##class` as keyword and leave the class name in parentheses un-highlighted, as it can be any class, not necessarily a built-in one.

> As for class names, I have chosen only those which do a better visual highlighting and tested it on different color themes. This is because in some cases the colors (styles) have no difference between different stylable classes. In other words, the usage of class: "keyword" for ##class construction is only because of a better visual styling.
>
> Do you thing it is a wrong way?

Yes, it's definitely not how it's supposed to work. We can't expect a language author to test the language against all the available styles (and keep doing that whenever any style changes), and neither a style author to test it against all languages. Instead we've defined a limited set of class names semantically, so they can be highlighted consistently in any language and in any style.

Whenever a particular combination doesn't work we could work it out and decide what to do: fix a particular style, use another class name, or even invent a new one. In this particular case I think leaving the class name un-highlighted is a good solution, as we consistently do this in all languages (unless it is indeed a built-in object).

ZitRos:
@isagalaev, thank you for your notes and fixes that made COS appear in the new library release!

I am going to make a new pull request with all the fixes and improvements related to the COS language.

isagalaev:
@ZitRos thanks! We love active maintainers :-)

PR #989, opened 2015-11-20
Qml - new script language

jf990:
request to add new script lanaguage QML, based off JavaScript


patrickarlt:
@Sannis Did this miss the 9.1 release due to the merge conflicts? If so @jf990 can you update this so it can get included in the next release.

jf990:
the [contrib instructions step 5](http://highlightjs.readthedocs.org/en/latest/language-contribution.html) says to update those files, but doesn't seem to be a useful idea on a branch. I spend more time resolving conflicts on those files! I just removed anything I changed there no point trying to fight the system.

isagalaev:
We usually resolve merge conflicts ourselves, AUTHORS.* and CHANGES conflicts are pretty much inevitable and easy to fix. The only reason of anything missing the release is simply the lack of time that any of the core devs can devote to the project. We'll get to this pull request eventually!

patrickarlt:
Thanks @Sannis and @isagalaev!

isagalaev:
The QTCreator style here appears to predate our new [style guide][], so I fixed it according to my best understanding in d7e937a. Since the translation isn't automatic I encourage you guys to look at the result and fix what's wrong and/or missing.

[style quide]: http://highlightjs.readthedocs.org/en/latest/style-guide.html

isagalaev:
Oh, and the best way to look at it is to pull the master code, build it for the browser (`node tools/build.js`) and look at the demo in build/demo/index.html.

jf990:
OK, thanks for the heads up I will look into it.

jf990:
I did some clean up on it but I didn't see anything big that was wrong. I used another style as the model. If I missed something let me know I will fix it.
thanks!

isagalaev:
One thing I noticed is that in the light version comments, meta and built-ins are all one color (#008000). Probably not intentional?

jf990:
I agree with you and double checked it, that is the setting in the default light style in the IDE. I actually am using a few more colors than the default. I think we should go with this for now as I did a side-by-side compare with the IDE and it is pretty close match. If any issues come in from any users I can submit a new PR for it.

isagalaev:
OK. You're a better judge anyway.

PR #995, opened 2015-11-24
Add Language file for EEx

jeffkreeftmeijer:
I've added a language file for [EEx](http://elixir-lang.org/docs/v1.0/eex/) (embedded Elixir) using a tweaked ERB language file.

Please let me know if there's anything I forgot to add or if there's anything that could be improved. Thanks!

### DIFF
--- a/src/languages/eex.js
+++ b/src/languages/eex.js
@@ -1,0 +1,22 @@
+/*
+Language: EEx (Embedded Elixir)
+Requires: xml.js, elixir.js
+Author: Jeff Kreeftmeijer <jeff@kreeftmeijer.nl>
+Description: "Bridge" language defining fragments of Elixir in HTML within <% .. %>
+Category: template
+*/
+
+function(hljs) {
+  return {
+    subLanguage: 'xml',
+    contains: [
+      hljs.COMMENT('<%#', '%>'),
+      {
+        begin: '<%[%=]?', end: '%>',
+        subLanguage: 'elixir',
+        excludeBegin: true,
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer:
Alright, this obviously needs some more work to get the detection working. I'll check it out as soon as I can and will report back.

jeffkreeftmeijer:
I think I need some guidance here. EEx support works and it’s properly detected, but the build is failing because the ERB file is wrongly marked as an EEx file.

Since EEx and ERB are very alike and probably because EEx comes before ERB alphabetically, this is to be expected. However, I think it’d probably be a good idea to turn off auto-detection for EEx to prevent false positives. 

What do you think? Is it possible to turn off auto-detection for EEx?

Sannis:
@isagalaev what do you think? Maybe we should embed EEx into Elixir definition?

isagalaev:
The general set of rules for very similar languages is this:

- In the wild we don't actually care what language name is assigned to a snippet, as long as it gets highlighted in a useful way (e.g. in `if (name == "something") { ... }` we don't care if it's Java, C++ or Rust as long as `if` is a keyword and `something` is a string).

- For a test snippet it *does* make sense to tweak it so it has more distinguishing language features and gets detected properly. After all, Ruby and Elixir aren't identical.

So, in practical terms: let's identify what Ruby has that Elixir doesn't (and vice versa) and make sure those features are present in ERB and EEx snippets. The relevance difference doesn't have to be big.

PR #996, opened 2015-11-24
Topic/more scala fixes

non:
Fix some remaining issues with scala highlighting.

After the last fix to case classes, types in the
parameter list were not highlighting correctly.
There were also some remaining issues with the
delimiters being highlighted.

This PR cleans all that up, and contains small
fixes to two styles which lacked an important
class (`.hljs-type`) which Scala needs.

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -73,10 +73,22 @@ function(hljs) {
         relevance: 10
       },
       {
+        className: 'typeParams',
+        begin: /\[/,
+        end: /\]/,
+        excludeBegin: true,
+        excludeEnd: true,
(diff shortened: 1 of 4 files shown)


non:
Ugh sorry, looks like my changes messed up source detection. I'll try to figure out what happened.

src/languages/scala.js
```diff
@@ -73,10 +73,22 @@ function(hljs) {
         relevance: 10
       },
       {
+        className: 'typeParams',
```
Sannis: Looks like this should be just `type`.

isagalaev:
Merged. Thanks!

PR #999, opened 2015-11-25
Add htmlbars parser

lastobelus:
* parses `attr=variable`, using `attr` and `variable` classnames
* makes `as` the only `keyword`, with `each`, `if`, etc. styled as
* `built_in`
* uses `name` for the name of tags
* only allows dasherized names, built-ins, or any block-form to be a
* name,
 otherwise assumes variable. However, only uses `template-tag`, and all
 tags are either `template-tag` or `template-yield` (see below)
* adds `template-yield`, because when working with templates this is
 very often the one thing I would like to be able to guickly find in
 a tag soup. `template-yield` matches {{yield...}}
 tags that have a . in the first token: {{bob.mary}} will get parsed as
 `<span class="template-tag">{{<span
class="variable">bob.mary</span>}}</span>`
* `hljs.QUOTE_STRING_MODE` and `hljs.NUMBER_MODE` are recognised inside
* tags
* adds the ember 2.0 set of template helpers as built_ins

### DIFF
--- a/src/languages/htmlbars.js
+++ b/src/languages/htmlbars.js
@@ -1,0 +1,118 @@
+/*
+Language: Handlebars
+Requires: xml.js
+Author: Michael Johnston <lastobelus@gmail.com>
+Description: Matcher for HTMLBars
+Category: template
+*/
+
+function(hljs) {
+  var BUILT_INS = 'action collection component concat debugger each each-in else get hash if input link-to loc log mut outlet partial query-params render textarea unbound unless with yield view';
+  var EXPRESSION_KEYWORDS = 'as';
+  var BUILT_IN_HELPERS = {
+    beginKeywords: BUILT_INS,
+    keywords: {built_in: BUILT_INS}
+  };
+  var ATTR_ASSIGNMENT = {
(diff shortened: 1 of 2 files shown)


lastobelus:
Would welcome some advice on how to set up the relevance.

For example, to make the existing handlebars win unless `as` or `mut` etc. are present.

isagalaev:
Got to merging it, finally. I've reworked the definition quite heavily for better consistency with other template languages. One thing that still bother me is the 'built_in' keywords: this is used mostly for standard library objects in procedural languages, but in HTMLBars they can seemingly be everywhere. For example, in the snippet that we have as a test:

    <div class="entry">
      {{!-- only output this author names if an author exists --}}
      {{#if author}}
        {{#playwright-wrapper playwright=author action=(mut author) as |playwright|}}
        <h1>{{playwright.firstName}} {{playwright.lastName}}</h1>
        {{/playwright-wrapper}}
      {{/if}}
      {{yield}}
    </div>

It's "action", "mut", "yield" (but not "as"). May be it's better to just call them "keyword", not "built-in".

PR #1001, opened 2015-11-26
Added some built_in keywords.

g-tv:
Added some built_in keywords which are part of the C++ standard.
see http://en.cppreference.com/w/cpp/io/manip/endl

Added some built_in keywords which are part of the C++11 standard.
see http://en.cppreference.com/w/cpp/utility/initializer_list
see http://en.cppreference.com/w/cpp/memory/unique_ptr


Sannis:
Thanks!

PR #1003, opened 2015-11-30
Verilog improvements

raphnet:
- Improve number support
 - Verilog is case sensitive


Sannis:
@raphnet can you please add short markup test for Verilog?

raphnet:
@Sannis Done. I created two tests. First a verilog number test, since fixing number highlighting was specifically what my original commit was about, and a second test based on a copy of the language detection code which will be I think a good sanity test.

isagalaev:
Thanks for keeping an eye on the language!

PR #1022, opened 2015-12-14
New language: Smarty

arnoldstoba:
Support for the [Smarty](http://www.smarty.net/) templating engine.
Based on [this](https://github.com/isagalaev/highlight.js/issues/998) issue.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -197,3 +197,4 @@ Contributors:
 - prince <MC.prince.0203@gmail.com>
 - Brendan Rocks <rocks.brendan@gmail.com>
 - Raphaël Assénat <raph@raphnet.net>
+- Arnold Stoba <arnold.stoba@web.de>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ New languages:
 
 - *Stan* by [Brendan Rocks][]
 - *BASIC* by [Raphaël Assénat][]
+- *Smarty* by [Arnold Stoba][]
 
 [Brendan Rocks]: http://brendanrocks.com
 [Raphaël Assénat]: https://github.com/raphnet
--- a/src/languages/smarty.js
+++ b/src/languages/smarty.js
@@ -1,0 +1,102 @@
+/*
+Language: Smarty
+Requires: xml.js, javascript.js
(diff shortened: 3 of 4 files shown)


pgorod:
Why was this closed without merging? I'm also looking for Smarty support in highlight.js...

yyyc514:
The original author closed it - but we're no longer merging language contributions to the core library now.  Your more than welcome to contribute this as a 3rd party language grammar though.  You could even turn this work into one if you wanted.

https://github.com/highlightjs/highlight.js/blob/master/extra/3RD_PARTY_QUICK_START.md

pgorod:
Ah ok. I don't think I can take on this work, but thanks anyway for all the nice things highlight.js already does. 🎉

yyyc514:
Are you unable to use a custom build or add a JS include to whatever project you're using HLJS on?

pgorod:
I am unable to contribute to this project due to being waaaay overstretched. Too much learning curve, even if it's not very complicated. I am not a JS guy, I am not a grammar guy, I am not even a Smarty guy 😄 

I have another highlight.js problem which is much more relevant to me, I'm not getting PHP highlighting, only HTML (see [example](https://docs.suitecrm.com/developer/views/), scroll to the bottom), and I can't figure out which part of the stack is to blame, if it's Hugo, or Asciidoctor, or highlight.js itself... this used to work but I don't know how long it's been broken, so I can't check the exact change that caused it.

If you can give me a preliminary look and tell me if I should open an issue for it?

yyyc514:
>  I am not a JS guy, I am not a grammar guy, I am not even a Smarty guy 😄

You sound like the perfect contributor!

> and I can't figure out which part of the stack is to blame, if it's Hugo, or Asciidoctor, or highlight.js itself... 

It might help if you used a version of HLJS that actually included the PHP grammar. ;-)

pgorod:
Aaahhh you speak with wisdom.

Somebody checked the wrong checkboxes when building the pack into our Hugo Theme.

That kicked in the Great Law of the Universe that _Code That Is Not There Shall Not Run_, and our highlighting was broken. 

Thanks for putting us on the right track! Cheers

yyyc514:
Glad to help.

PR #1033, opened 2015-12-23
removing define module name as it isn't needed and breaks aliasing

adamcarr:
This breaks the ability to do the following.

```javascript
// config.js
requirejs.config({
    paths: {
        'mymodule-highlightjs': '/path/to/highlightjs'
    }
});

// file.js
define(['mymodule-highlightjs'], function(highlightjs){
    //...
});
```

Removing the named define fixes this.

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -16,7 +16,7 @@ https://highlightjs.org/
 
     // Finally register the global hljs with AMD.
     if(typeof define === 'function' && define.amd) {
-      define('hljs', [], function() {
+      define(function() {
         return self.hljs;
       });
     }


sourrust:
The explicit name was added because of #853 and this pull request will most likely break use of almond angular agian. It's probably best to use the defined module name or modify the file locally.

adamcarr:
I have forked this so I can use my solution. Such is the ugliness of module loaders today.

adamcarr:
One thing to consider here is that you are in essence preferring Almond.js over require.js. Granted, this is a specific use case of require.js and most scenarios that use require.js don't need this, so they aren't broken. That said, ideally, in your module detection code block you could detect if using almond.js and if so, include the define module name; otherwise, omit it. I am not blocked here as I have the fork I can use with bower, but that offloads the maintenance to myself, and others, if they also encounter this problem.

isagalaev:
I got around this (finally), and while it's seems to be settled, there's one thing I want to clear up. One reason for having a separate building infrastructure is to keep the actual source free of decisions like this.

I found the point where this environment detecting code was added in 3ed3be7 but I don't remember the discussion leading to that. Looking at it all afresh I think it should be possible to move the header into the build tool, restore a separate target for AMD and control adding an explicit define name with a parameter. It seems to be more consistent with how other build targets are handled.

@sourrust: what do you think about it?

sourrust:
Looking into it, it seems that [anonymous modules][1] is the way to go. And I found that require.js has a tool that will name the modules for you, it is something like this:

```
```

[1]: https://github.com/jrburke/requirejs/wiki/Updating-existing-libraries#register-as-an-anonymous-module-

isagalaev:
Okay. Seems to be the least complicated thing to do, as it keeps it out of our code.

Issue #955, opened 2015-10-12
Q: Extra mode for the `end`


tenbits:
Hi Team,
how it is possible to define the ending mode, similar to the `starts`:
```
begin: /~\[/, 
className: "subst smartquote",
starts: {
  end: /\]/,
  subLanguage: 'javascript',              
  excludeEnd: true,
  className: 'subst',
}
```
This should be the interpolation quotes within string mode and with javascript mode inside. Like this:
```mask
'Hello ~[foo + 2] bar'
```
Now the problem I have, the end quote `]` does not belong to `"subst smartquote"` mode, but to the parent _(`"string"`)_. And when I set `excludeEnd: true,` to `false`, the `]` belongs to javascript mode. How I can highlight the end matching?

Thanks in advance!

isagalaev:
From my understanding of the problem something as simple as this should work:

```js
{
  className: 'subst',
  begin: /~\[/, end: /\]/,
  subLanguage: 'javascript'
}
```

tenbits:
Thanks Ivan, but how I put the interpolation quotes _(the beginning match and the ending)_ to the seperate mode, for instance the `"smartquote"`, but not to the javascript content itself.

And how I make the content to be ungreedy, I thought actually the `starts` feature is for this, sothat the sublanguage mode is ended when we are in the root mode of that language **and** the ending is matched, like in TextMate.

So that such things would work:
```mask
"Be ~[ users[0].name ] good"
```

When we use `begin: /~\[/, end: /\]/,`, then we will stop the interpolation after `users[0]`.

isagalaev:
Oh, that's a whole different can of worms then :-(

Unfortunately `starts` won't help here, it starts a new mode unconditionally but that mode is just a regular one, it still needs `end` or `endsWithParent` to be closed. A sub-language doesn't affect parsing of the outer language at all, it's actually applied during a second pass when the whole outer mode is parsed up to the next terminator. Which means that dealing with nested brackets and a sub language is impossible as every part between `[ ... [ ... ] ... ]` will be treated separately.

To make it work we'll need to rework this logic so a sub-language syntax tree would actually be a part of the outer language, and it's going to be quite an undertaking!

tenbits:
Thank you, Ivan, I appreciate your explanation. And such sub-language tree nesting would be much desired. As for now, I'm trying to create the syntax highlighting for a template markup, which accepts not only the javascript interpolations, but also javascript and css style blocks. For instance:
```mask
header {
[...]

PR #1037, opened 2015-12-29
Add MaskJS language

tenbits:
Hallo Team,

adding here [MaskJS](https://github.com/atmajs/maskjs) syntax highlighting.

I have also implemented the `clone` method to properly deep-clone the mods. As the `hljs.inherit` method should be rethinked, it performs only the shallow copy of the object - these could lead to compile problems. For example, using `hljs.COMMENT_LINE_BLOCK` without cloning, breaks the detection test: for some reason `cmake` is recognized then as `sql`.
- [ ] **⇒** Block Syntax Pattern _(`beginClassName`, `endClassName`, nesting)_
- [ ] Helper Functions _(Tree Factories)_
- [ ] Root tree reference in submodes (_`Mask.contains`_)
- [ ] Disable `relevance` per default. _(`disableRelevance()`)_
- [ ] Sublanguage cloning. _(`clone()`)_

Cheers, Alex

### DIFF
--- a/src/languages/mask.js
+++ b/src/languages/mask.js
@@ -1,0 +1,396 @@
+/*
+  Language: MaskJS
+  Category: template
+  Requires: css.js, javascript.js, xml.js
+*/
+function(hljs) {
+  var Mask = {
+    aliases: ['mask'],
+    contains: []
+  };
+  var clone;
+  (function(){
+    var cached = [];
+    clone = function(obj){
+      cached.length = 0;
+      return cloneObject(obj);
+    };
+    
+    function cloneObject(obj) {
+      if (obj == null){
+        return null;
+      }
+      if (typeof obj !== 'object' || obj instanceof RegExp){
+        return obj;
+      }
+      var i = cached.indexOf(obj);
+      if (i !== -1)
+        return cached[i + 1];
+
+      var cloned = {};
+      cached.push(obj, cloned);
+      for(var key in obj) {
+        if (COPY[key] === 1) {
+          cloned[key] = obj[key];
+          continue;
+        }
+        if (CLONE[key] === 1){
+          var val = obj[key];
+          if (val == null) {
+            continue;
+          }
+          if (val instanceof Array) {
+            var arr = cloned[key] = [];
+            var imax = val.length;
+            var i = -1;
+            while(++i < imax) {
+              arr.push(cloneObject(val[i]));
+            }
+            continue;
+          }
+          cloned[key] = cloneObject(val);
+          continue;
+        }        
+      }
+      return cloned;
+    };
+    var COPY = {
+      begin:1,
+      end: 1,
+      beginKeywords: 1,
+      keywords: 1,
+      variants: 1,
+      subLanguage: 1,
+      className: 1,
+      relevance: 1,
+      lexemes: 1,
+      illegal: 1,
+      excludeBegin: 1,
+      excludeEnd: 1,
+      returnBegin: 1,
+      returnEnd: 1,
+      endsParent: 1,
+      endsWithParent: 1
+    };
+    var CLONE = {
+      contains: 1,
+      starts: 1,
+    };
+  }());
+
+  var jsLang = hljs.getLanguage('javascript');
+  var cssLang = hljs.getLanguage('css');
+  var cssMods = cssLang.contains;
+  var jsMods = [{begin: /\[/, end: /\]/, relevance: 0, contains: ['self']}]
+        .concat(clone(jsLang).contains);
+
+  function createBlock(def, isSub) {
+    return {
+      begin: def.begin, end: def.end,
+      className: def.className,
+      endsParent: 'endsParent' in def ? def.endsParent : false,
+      returnBegin: true,
+      returnEnd: 'returnEnd' in def ? def.returnEnd : false,      
+      relevance: def.relevance,
+      contains: [
+        {
+          begin: /\s+/,
+        },
+        {
+          begin: def.begin,
+          className: 'beginClassName' in def ? def.beginClassName : 'meta',
+        },
+        {
+          begin: def.end,
+          className: 'endClassName' in def ? def.endClassName : 'meta',
+          endsParent: true,
+          returnBegin: 'returnEnd' in def ? def.returnEnd : false,
+        },
+        {
+          end: def.end,
+          returnEnd: true,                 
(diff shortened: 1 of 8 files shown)


isagalaev:
Hi Alex,

Took me some time to look through the definition as it is very complicated. As a rule, a new language shouldn't get out of its way to define a whole new language to define languages, it makes further maintenance prohibitively expensive.

I don't have a detailed understanding of everything in your code, so instead of giving a comprehensive review let's start with *very* high level comments and get to more details iteratively:

- Instead of populating `Mask.contains` imperatively, it'd be much easier to read if it was an actual array. Is there a reason why it's this way?

- Speaking more generally, in-place declarative definition is always preferred to defining temporary variables beforehand and then programmatically modifying them later. The more declarative a language definition the better.

- Functions like `createBlock()` and `createTag()` should not be necessary as we have `variants` (which you already use in a couple of places)

- Blanket `disableRelevance()` is also something you should not need to do. Any mis-detections are usually resolved on a case-by-case basis, not by burning the land with napalm :-)

These are points which I'd like to see resolved before attempting to look deeper. Thanks for your patience!

tenbits:
Hi Ivan,

first of all, thank you for looking at this! Now going through the points:

- `Mask.contains`: I have to initialize an Array first, so that it will be accessed via reference later while creating the grammar. Some modes should have as `contains` the root `MaskJS` grammar. It is like `"self"`, which is supported by `highlight.js`, but it does not have something like `"root"`. Or with other words, at some points I want to use `hljs.getLanguage('mask')`, but it would not be possible, as the language has not been defined at that moment. Such `self-looping` is required for `Block` definitions, which will be covered later in `createBlock` explination.

- > The more declarative a language definition the better.

	Totally agree with you. Like having for instance, the `json`, as a syntax grammar would be nice. Similar like `Textmate` grammars. But unfortunatelly JavaScript regular expressions are not so powerfull like in Ruby for example. A lot of features are not supported, but I'm happy with `highlight.js` as such things like `lookbehind` can be partially simulated with `returnBegin|returnEnd: true`, so that the next mode can consume the matched portion from previous mode. But `highlight.js` has a lack of some features, or better to say, some API, thats why I had to implement local (in a language definition), some helpers and some things to achieve the desired behaviour, and also to prevent myself from repeating, which leads to better maintainance. See my farther explination to `createBlock`. 

- `createBlock`, `createTag`, `createXmlNode`

	All this functions are here due to DRY. You can consider them, as `Factories`, which create some tree of modes. For instance, there are some XML Nodes with different sublanguages, so I define generic method `createXmlNode`, wich creates the grammars mode node structure with some arguments. Such approach is even better for the maintainance.

- `createBlock`

	The function is a factory for the grammar node structure described in #955. Without the helper function, the grammar tree would be much more complicated with deep nesting, if we would write the structure directly in-place.

	Now, to the `createBlock` pattern itself. As it was described, it creates a Mode which ends with the match, but only when we are in the current mode. I will try to explain with an example.
	`'~[ foo[property] + 2 ]'`: from here the resulting mods are expected as such: `~[ foo[property] + 2 ]` is an interpolation mode, and `foo[property] + 2` is a javascript mode. So when defining the `InterpolationMode`, we must say the `highlight.js`, that the mode ends with `]` match, but only when we are in a top-level mode of a child. Additionally we want to style the beginning (`~[`) and the ending (`]`) of the `InterpolationMode`. Right now there is no way to define desired behaviour direclty in one Node. And the `createBlock` introduces this new node declaration API and encapsulates part of the Tree definition. 

	```javascript
	createBlock({
		begin: /~\[/,  end: /\]/, 
		beginClassName: 'meta', endClassName: 'bold'
		contains: CHILD_MODES, /* e.g., hljs.getLanguage('javascript').contains */
		className: 'subst hljs-emphasis'
	});
	```

	And if we extract the encapsulated pattern we get:

	```javascript
	{
		begin: /~\[/, 
		returnBegin: true,
		className: 'subst hljs-emphasis',
		contains: [
			{
				// if not handled in a CHILD_NODES, skip whitespace and stay in the parent mode			
				begin: /\s+/, 
			}, {
				begin: /~\[/,
				className: 'meta',
			}, {
				begin: /\]/,
				className: 'bold',
				// when this mode was matched, that we are in a top level mode
				endsParent: true
			}, {
				// Immeditaly begin the CHILD_MODES, and end them later when `\]` is matched.
				// We stay at the parent mode afterwards, and next matched child mode is the previous one, 
				// where we highlight it and end parent.
				end: /\]/,
				returnEnd: true,
				contains: CHILD_MODES
			}
		]
	}
	```

	The resulted highlighting: 
image


	Please share if such highlighting could be achieved in a another way. So that I can simplify the block pattern and use the correcter way.


- `disableRelevance`

	In `Mask` grammar should be almost everything be set to `relevance:0` per default, otherwise almost every code would be recognized as `Mask`. 


So, if you mind, based on your comment, I will add the checkmarks with todos we should to discuss to the pr description, so we can go through, one by one. 

I suggest to start with the `Block Pattern`. What do you think? Should we leave as is, do you want to add the pattern directly to `highlight.js`, as I think this api could be useful for others (`beginClassName`, `endClassName`, nesting, etc.). Or are there other ways to do this?

Thank you for your time!

isagalaev:
Hello again!

I've been quiet for a while because I tried to answer the entire thing and it just took too long every time, going deep into software engineering principles, etc. Since it's not working, let me try and get to it from the other side. Here's some things that absolutely need changing before we can tackle the bigger stuff:

- Get rid of the anonymous self-calling modules (`(function(){..})();`), they serve no purpose here and only add clutter.

- Instead of calling `Mask.contains.push` refactor it into this pattern that we have in other languages:

    ```js
var SOME_MODE = { .. };
var DEFAULT_CONTAINS = [
  SOME_MODE,
  ...
];
SOME_MODE.contains = DEFAULT_CONTAINS;
    ```

    It's not inherently "better", it's just more recognizable by core maintainers.

- Get rid of multi-word class names, like "subst hljs-emphasis". I understand the intention, but it's exactly what we don't want to have. We want highlighting to be consistent across languages.

- Drop specific relevances everywhere and the whole `disableRelevance` with them. Don't worry by broken tests for now, let's first get the language right and then it will either fix itself (I have a pretty good idea why it happens) or we'll fix it by more precise means.

isagalaev:
Closing this, is the approach taken is too complicated for future maintenance.

PR #1039, opened 2016-01-03
Request to merge fork for Maxima highlighting

robert-dodier:
Hi, I've created a highlighting mode for Maxima, a computer algebra system. See: http://maxima.sourceforge.net

It seems to work pretty much as expected, although I'm not yet able to detect integers with trailing period (e.g. 1234. is an integer) and floats with preceding period (e.g. .1234 is a float). Also, it doesn't detect nested comments, which are allowed in Maxima. I put in the obvious contains:[self] in the comment mode, but that seems to cause trouble -- I get a "unresponsive script" or some error message like that in the browser (Firefox). Anyway, at this point I think I'm ready to unleash it upon the world.

Let me know if there is anything else I need to do in order to have this PR accepted.

best,

Robert Dodier


src/languages/maxima.js
```diff
@@ -0,0 +1,415 @@
...
+      },
+      hljs.QUOTE_STRING_MODE,
+      {
+        className: 'number',
+        variants: [
+          {
+            // float number w/ exponent
+            // hmm, I wonder if we ought to include other exponent markers?
+            begin: '\\b(\\d+|\\d+\\.|\\.\\d+|\\d+\\.\\d+)[Ee][-+]?\\d+\\b',
+            relevance: 1
```
Sannis: We usually try to avoid relevances different from 0 and 10 because so precise tuning causes problems with next merged languages. I'll try to play with this.

robert-dodier: Thanks for taking the time to look at it. I agree that fine-tuning the
relevance values introduces the possibility of conflict between
languages, but I found that was the only way to get it to pass the
tests. In particular, in the autodetection tests, Maxima was confused
with other languages (GAMS, COS, or Matlab) when the relevance values
were even slightly different.

I wonder if it is possible to get calculated total relevance scores
from a failed autodetection test.


best, Robert.

isagalaev: Robert, we can fix the detection later. Sometimes it's the other language that has to be changed. Let's first make this language correct :-).
I wonder, was the relevance tuning the only reason to split up the number definition? Could you simply use `hljs.NUMBER_MODE`?

src/languages/maxima.js
```diff
@@ -0,0 +1,415 @@
+/*
+Language: Maxima
+Author: Robert Dodier <robert.dodier@gmail.com>
+Category: scientific
+*/
+
+function(hljs) {
+  var KEYWORDS = 'if then else elseif for thru do while unless step in and or not';
+  var LITERALS = 'true false unknown inf minf|2 ind und %e|2 %i|2 %pi|2 %phi|3 %gamma|3';
```
isagalaev: Please remove these keyword relevances too.

src/languages/maxima.js
```diff
@@ -0,0 +1,415 @@
...
+      + ' yv_grid y_voxel yx_ratio zaxis zaxis_color zaxis_type zaxis_width zeroa zerob'
+      + ' zerobern zeta%pi zlabel zlabel_rotate zlength zmin zn_primroot_limit zn_primroot_pretest';
+
+  return {
+    lexemes: '[A-Za-z_%][0-9A-Za-z_%]*',
+    keywords: {
+      keyword: KEYWORDS,
+      literal: LITERALS,
+      built_in: BUILTIN_FUNCTIONS,
+      symbol: BUILTIN_VARIABLES
```
isagalaev: With the exception of `_`, `__`, `%` and `%%` all of these BUILTIN_VARIABLES should be just built-ins, not symbols. For those four, let's define a separate mode like `{className: 'symbol', begin: /(_|__|%|%%)/}`.

src/languages/maxima.js
```diff
@@ -0,0 +1,415 @@
...
+    },
+    contains: [
+      {
+        // C-style comments, with nonstandard color.
```
isagalaev: What are these semantically? `meta` is for meaningful things like compiler directives, `comment` is just a comment.

src/languages/maxima.js
```diff
@@ -0,0 +1,415 @@
...
+        // Dunno if this recognizes nested comments.
```
isagalaev: Use `contains: ['self']` for that.

src/languages/maxima.js
```diff
@@ -0,0 +1,415 @@
...
+        // I wonder, should we try to recognize quoted strings in comments?
```
isagalaev: *If* it's a proper meta and has internal structure, then yes. We have "meta-string" and "meta-keyword" classes for that.

test/detect/maxima/default.txt
```diff
@@ -0,0 +1,79 @@
...
+foo(%pi, %e, bessel_j(1, u));
+is (% > 0);
+(sin(1) - 1, 2*%%, 1 - %%);
+print ("most recent output is", %);
+
+blurf[foo](a) := a^foo - foo^a $
+blurf[2];
+is (blurf[2] = lambda([a], a^2 - 2^a));
+
+blurf[%pi](%e) - matrix([%phi])^^2;
```
isagalaev: The demo snippet is too long. It should present a language with most interesting things in it but it shouldn't be neither a comprehensive test of features (that's markup tests), nor a big chunk of actual working code with many things repeating over and over. Most demo snippets in highlight.js do come from real code but get ruthlessly sequestrated :-)

test/markup/maxima/example.txt
```diff
@@ -0,0 +1,79 @@
...
+foo(%pi, %e, bessel_j(1, u));
+is (% > 0);
+(sin(1) - 1, 2*%%, 1 - %%);
+print ("most recent output is", %);
+
+blurf[foo](a) := a^foo - foo^a $
+blurf[2];
+is (blurf[2] = lambda([a], a^2 - 2^a));
+
+blurf[%pi](%e) - matrix([%phi])^^2;
```
isagalaev: This example markup test seems to duplicate the detection/demo snippet. It's also unnecessary.

isagalaev: Uhm, let me rephrase that. When the demo snippet gets shortened, it does make sense to keep various markup as markup tests but they should test separate concepts (like the one for numbers below), not be a blanket test of the whole language. It tends to break on every maintenance change and hence stops being a good test.

src/languages/maxima.js
robert-dodier: Hi, thanks for your comments. I'll try to address your comments one by one. You wrote: "Could you simply use hljs.NUMBER_MODE?" -- Well, Maxima numbers aren't exactly the same as in other languages, so it seemed easier to me to just work out the regexes for Maxima numbers, even though there is some overlap with other languages. In particular, Maxima has so-called bigfloat numbers (variable precision floats) which have exponent marker "b", which other languages don't have, I think.

src/languages/maxima.js
robert-dodier: OK, removed in commit fd535d2. Passes tests.

src/languages/maxima.js
robert-dodier: Well, I want built-in variables to have a different color than built-in functions. If I do `built_in: BUILTIN_FUNCTIONS + ' ' + BUILTIN_VARIABLES`, as I think you are suggesting, they will be the same. Is there a better way for me to make built-in variables different from built-in functions?

src/languages/maxima.js
robert-dodier: About comments, I chose 'meta' because I like comments to be blue. Is there a better way to achieve that?

src/languages/maxima.js
robert-dodier: OK, handle nested comments via `self` (commit b19de89).

src/languages/maxima.js
robert-dodier: I've dropped the idea of detecting quoted strings in comments (commit c0bfbcd).

test/detect/maxima/default.txt
robert-dodier: OK, I've made the demo program shorter (commit 57fcb90).

test/markup/maxima/example.txt
robert-dodier: OK, I've made the markup test shorter (commit 25d5a7a).

robert-dodier:
I've changed any relevance values to 0 or 10 (or removed them. The current version (commit a72e3ad) passes all tests. Ivan, perhaps you can take another look at this PR? Thanks for your help.

src/languages/maxima.js
isagalaev: I'm afraid that'd go against some of the design decisions in highlight.js.

One of the goals is for highlighting to be consistent across languages, so we have a limited set of stylable features (class names) that style authors rely on when producing styles. Re-using a class name for another purpose may cause undesired effects and adding a new class name means updating all of the existing styles and complicating creation of new ones. I would agree though, that in this particular case styling built-in vars as "symbol" probably won't cause any artistic shock anywhere :-) But this does explain why we don't simply treat each language individually.

However! There's another design choice of keeping highlighting minimalistic (and more useful). This is subjective, of course, and in this case I simply follow a precedent set in 9739df60a: upper-case gives enough distinction to those vars.

src/languages/maxima.js
isagalaev: You can't target a particular color as it is different in every theme anyway (and may be changed in the future). The right way to make comments blue is to make a separate style where all comment would be blue, not re-purpose "meta" for one language. And by "separate style" I don't mean creating an entire new theme (though you're welcome to do that too!), but rather adding a local style rule `.hljs-comment { color: your-favorite-blue; }`

isagalaev:
I went ahead and merged this with the modifications to built-ins and comments I was talking about, so we could have a base working implementation in master. We can continue discussing some points further and make changes later if necessary.

Thanks for your contribution, Robert!

robert-dodier:
Hi Ivan, I took a look at the merged version in highlight.js and it looks pretty good to me. I'm a little puzzled that symbolic constants, built-in variables, and built-in functions all have to be the same color. Other than that it seems fine. Let me know if there's anything further you want me to do. Best, Robert.

isagalaev:
Well, symbols actually do differ from all the built-ins. And all the built-ins under one color is just how highlight.js has it right now.

PR #1049, opened 2016-01-09
Add support for base64-encoded urls in CSS

bcherny:


src/languages/css.js
```diff
@@ -15,7 +15,17 @@ function(hljs) {
           endsWithParent: true, excludeEnd: true,
           contains: [
             {
-              begin: /[\w-]+\s*\(/, returnBegin: true,
+              begin: /[\d\w-]*\(/, end: /\)/, className: 'function',
+              contains: [
+                {
+                  className: 'string',
+                  begin: /[\s\w-\/=:;,"'\.]+/,
```
bcherny: more permissive than normal strings, and `"`/`'` are optional

src/languages/css.js
```diff
@@ -15,7 +15,17 @@ function(hljs) {
...
+                  relevance: -1
```
bcherny: avoid a false positive for less lang detection

isagalaev:
Finally got to merging it. Thanks!

I've got a bit carried away with post-merge fixes and summarized them in c064f1fab1bb. I can clarify any point if needed, feel free to ask!

PR #1054, opened 2016-01-17
Additional Dart keywords.

dikmax:
Support for new Dart keywords, that were introduced into language not so long ago.


Sannis:
Thanks!

PR #1060, opened 2016-01-25
Add 'static' keyword to JavaScript language

anthonydugois:


isagalaev:
I don't see this keyword in [the language spec][1]. How come?

[1]: http://www.ecma-international.org/ecma-262/6.0/#sec-keywords

anthonydugois:
It is in [Class Definitions](http://www.ecma-international.org/ecma-262/6.0/#sec-class-definitions).

isagalaev:
Merged. Thanks!

PR #1068, opened 2016-01-31
style: Add gruvbox style (light and dark)

Qeole:
Based on Gruvbox style by Pavel Pertsev (@morhetz), available at https://github.com/morhetz/gruvbox.

### DIFF
--- a/src/styles/gruvbox_dark.css
+++ b/src/styles/gruvbox_dark.css
@@ -1,0 +1,107 @@
+/*
+Original style from Pavel Pertsev (morhetz/gruvbox)
+*/
+
+.hljs {
+  display: block;
+  overflow-x: auto;
+  padding: 0.5em;
+  background: #282828;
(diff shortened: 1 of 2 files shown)


isagalaev:
The CSS in these themes seems to be based on some old pre-new-styles file. We now have a new [style guide][] for those, so could please bring the themes in accordance to it. It can be done either by fixing the existing code (removing language-specific rules, removing non-existent classes, adding missing important classes) or it may be even easier to just make it from scratch, going through the style guide.

[style guide]: http://highlightjs.readthedocs.org/en/latest/style-guide.html

Qeole:
You are right, I used the _Solarized Dark_ file as a basis for creating this one. I will do as you suggest and start again after reading the guide, thanks for the pointer!

isagalaev:
It's quite okay to use another file as a basis, it just has to be up to date :-)

PR #1090, opened 2016-02-23
Update HTML attributes syntax

AntonKhorev:
Couple of fixes for XML with test cases:

* unquoted attributes can contain slash
* can have spaces around `=` in attribute definition

See [attribute syntax in HTML spec](https://www.w3.org/TR/html5/syntax.html#attributes-0)


isagalaev:
Thanks! I love comprehensive markup tests that make sense :-)

Issue #1086, opened 2016-02-19
Provide access to a parse tree as an alternative to HTML output


derhuerst:
I've just had a quick look at hightlight.js and didn't any discussion on this. So forgive me if this is not the right place or has already been discussed.

**I'd like to propose to make highlight.js more generally usable, or at least pull out the core highlighting logic to make it independent from the output format.** Just like the famous [Pygments has formatters](http://pygments.org/docs/formatters/), the HTML generation would be done by a separate module/package.

highlight.js seems to be the de factor standard from syntax highlighting, even among porject not related to HTML/browsers. See [the slap editor](https://github.com/slap-editor/editor-widget/blob/master/lib/highlight/server.js) and [hicat](https://github.com/derhuerst/hicat/blob/master/index.js) for examples: Both use it and later on manually walk/manipulate the HTML generated by highlight.js.

In the world of lightweight do-one-thing NPM packages, it would make sense to split highlight.js into two parts, **one generating an treeish object representation** (possibly streaming), and **the second generating HTML from it**.

isagalaev:
This idea is hanging in the air for a while. It's never been a priority as it lacked real use-cases, but now as you mentioned those two it makes very real sense. Another problem is that in one or two attempts at doing this it was done as a part of bigger (and controversial) refactorings of the whole core library, so it haven't got merged. I will probably get to it myself eventually, but if you'd fancy giving it a try, I could give some guidance.

derhuerst:
> I will probably get to it myself eventually, but if you'd fancy giving it a try, I could give some guidance.

I'm having difficulties in understanding the inner workings of highlight.js, both because there is not a lot of documentation/comments and because the code is written in a really imperative jump-around way.

Also, it seems like HTML generation is really baked into the core of highlight.js.

isagalaev:
Yes, it was deliberately done so in the past to squeeze the code size as much as possible. Currently this "optimization target" is much less desirable and that's why I'm all for refactoring this.

I'll do a short writeup about the architecture of the core to make it more comprehensible.

derhuerst:
> I'll do a short writeup about the architecture of the core to make it more comprehensible.

This sounds great, especially since there seems to be no render-agnostic snytax highlighter in JavaScript land right now.

isagalaev:
> I'll do a short writeup about the architecture of the core to make it more comprehensible.

[...]

PR #1094, opened 2016-02-26
Allow `className` to be configured via functions

izaakschroeder:
Previously `hljs` had fixed rules for injecting classes into markup – no longer! You can now control exactly what classes get generated based on language, mode or keyword. This functionality is useful if you use CSS modules and do not wish to pollute the global CSS namespace. It also means that `classPrefix` isn't strictly necessary anymore – you can generate the names however you like.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -208,3 +208,4 @@ Contributors:
 - Denis Ciccale <dciccale@gmail.com>
 - Michael Johnston <lastobelus@gmail.com>
 - Taras <oxdef@oxdef.info>
+- Izaak Schroeder <izaak.schroeder@gmail.com>
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -445,7 +445,7 @@ https://highlightjs.org/
           }
         }
         do {
-          if (top.className) {
+          if (options.className.mode(top)) {
             result += '</span>';
           }
           if (!top.skip) {
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -30,20 +30,11 @@ https://highlightjs.org/
 
   /* Utility functions */
 
-  function getClassNameForMode(mode) {
-    return mode.className ? options.classPrefix + mode.className : null;
-  }
-
-  function getClassNameForLanguage(language) {
-    return language ? language : null;
-  }
-
-  function getClassNameForKeyword(keyword) {
-    return options.classPrefix + keyword;
-  }
-
-  function getClassNameDefault() {
-    return 'hljs';
+  function getClassName(existing, type) {
+    if (type === 'mode' || type === 'keyword') {
+      return options.classPrefix + existing;
+    }
+    return existing;
   }
 
   function escape(value) {
@@ -355,7 +346,7 @@ https://highlightjs.org/
         var className;
         if (keyword_match) {
           relevance += keyword_match[1];
-          className = options.className.keyword(keyword_match[0], keyword_match[1]);
+          className = options.className(keyword_match[0], 'keyword', keyword_match[1]);
           if (className) {
             result += buildSpan(className, escape(match[0]));
           } else {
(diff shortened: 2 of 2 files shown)


isagalaev:
Hello Izaak! While I can certainly see the appeal of this idea, this is something that we deliberately don't want to do. The reason for this is explained in #348, which was a large refactoring we did recently to enforce a limited set of class names. And this how highlight.js is going to live from now on.

izaakschroeder:
@isagalaev I think this still fits in alright? My aim is mostly not to let _language authors_ pick whatever names they want, but to allow library consumers to transform the local set of available names.

izaakschroeder:
Happy to consider other strategies for making CSS modules work :smile: this seemed like the least intrusive.

izaakschroeder:
I'd be open to having a `transformClassName` if that sounds safer, where it just maps an existing class name to some new class name; the current strategy just seemed more general.

izaakschroeder:
```javascript
onmessage = ({data: {code, language, styles}}) => {
  hljs.configure({
    tabReplace: '  ',
    className: {
      default: () => null,
      language: () => null,
      mode: (mode) => {
        console.log('GOT MODE', mode);
        return styles[mode.className];
      },
      keyword: (keyword) => {
        console.log('GOT KEYWORD', keyword);
        return null;
      },
    },
  });
  require('bundle!highlight.js/build/node/lib/languages/' + language)((data) => {
    if (!hljs.getLanguage(language)) {
      hljs.registerLanguage(language, data);
    }
    const result = hljs.highlight(language, code, true);
    postMessage(result);
  });
};
```

is what I am using now.

isagalaev:
Ah, I've got your motivation this time :-) I'll have to think about it a bit more. It may be better to first refactor the library to separate tree building from rendering (something we're about to do in #1086) , and then let users to customize the HTML renderer in this way.

izaakschroeder:
I am totally happy with having a totally customizable renderer which sounds great! :smile: I think this is a reasonable stop-gap solution in the interim though? It doesn't break anything and keeps all existing behaviour as-is?

izaakschroeder:
I'm also super open and happy to making the API a little nicer.

isagalaev:
As a stop-gap, it's simpler to just keep it in your fork in whatever way you most like it. I wouldn't want to first merge it and then have to immediately rewrite it heavily during that refactoring. Our core is hairy enough, no need to make it more complex :-)

izaakschroeder:
Complexity reduced! It's actually pretty painful to keep externals in `npm` due to needing funny `postinstall` scripts and other such things to ensure source dependencies get built properly. :sob:

izaakschroeder:
À la:

```javascript
onmessage = ({data: {code, language, styles}}) => {
  hljs.configure({
    tabReplace: '  ',
    className: (item) => styles[item],
  });
  require('bundle!highlight.js/build/node/lib/languages/' + language)((data) => {
    if (!hljs.getLanguage(language)) {
      hljs.registerLanguage(language, data);
    }
    const result = hljs.highlight(language, code, true);
    postMessage(result);
  });
};
```

izaakschroeder:
Any other thoughts? :smile:

isagalaev:
I still haven't got around this issue. I can only dedicate about one day a week to highlight.js these days, and some weeks don't even get that. (My suggestion to use a fork in the mean time is very practical, the project simply can't change very fast.)

isagalaev:
Back to the issue! I've been thinking a lot about all the little things around this feature and I do find it useful, as there were a few requests about having different styling for different snippets on one page. Here are my current thoughts on the actual design:

- `options.classFunc` is a configurable callback returning a class name based on some context, which is by default just a base class name.

- By default `options.classFunc` should be `function(className) {return options.classPrefix + className}`. Having this default should allow us not to drop conditional logic in `buildSpan` and simply call the function.

- The distinction between language name classes and other classes is irrelevant by now and we can drop it (we may even drop generating language-specific `<span>`s in the future altogether).

- The distinction between keywords and modes should not concern style authors either.

@izaakschroeder do you think this would work with your use case with CSS modules (I think it should). If yes, do you think you could take it on and re-implement it this way?

isagalaev:
Have to close it due to inactivity and the lack of bandwidth from my side to work on it.

pa-nam:
@izaakschroeder Totally agree with you about customizing the className, especially with data like keyword and mode. Having this feature would be a lot of nicer. Will take a deeper look at your pull.

PR #1098, opened 2016-02-26
Enhance GAMS language parsing

mkouhia:
Previous version was quite basic.  Now highlighting supports
- keywords
- assignments
- comments
- symbols
- equation definitions.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -208,3 +208,4 @@ Contributors:
 - Denis Ciccale <dciccale@gmail.com>
 - Michael Johnston <lastobelus@gmail.com>
 - Taras <oxdef@oxdef.info>
+- Mikko Kouhia <mikko.kouhia@iki.fi>
--- a/AUTHORS.ru.txt
+++ b/AUTHORS.ru.txt
@@ -208,3 +208,4 @@ URL:   https://highlightjs.org/
 - Денис Сиккале <dciccale@gmail.com>
 - Майкл Джонстон <lastobelus@gmail.com>
 - Тарас <oxdef@oxdef.info>
+- Мікко Коухіа <mikko.kouhia@iki.fi>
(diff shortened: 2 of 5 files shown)


isagalaev:
Hi! This is a great enhancement to the language! While reviewing it I found one thing that seems rather odd: everything between `/ .. /` is marked up as numbers which catches entire sections like this one:

      /    NEW-YORK    325
           CHICAGO     300
           TOPEKA      275  /

I don't believe this is intentional?

Also, could you please modify the demo snippet to include some new things you defined?

mkouhia:
@isagalaev Highlighting is intentional, but CSS class selection is mostly due to not understanding which class would be a proper option. Would you would have better suggestion? Everything inside `/ .. /` in parameter definition is an assignment: we assign parameter values for these members. So in the example

    B(J)  demand at market j in cases
      /    NEW-YORK    325
           CHICAGO     300
           TOPEKA      275  / ;

`B('NEW-YORK')` gets value 325. Which class would you suggest using? Meta? Anyway, this is a major language feature and in my opinion would deserve some highlighting.

I'll update the snippet.

isagalaev:
This seems as important as, say, variable declaration in the beginning of a function or setting up named constants, but neither of those sections are highlighted as a single object in any language. I think parsing the contents of this section might be a better choice. Words could become 'attr' and numbers — 'number'.

mkouhia:
Hmm, you have a point there. In a project like this the coherence between languages should be the main driver, I guess? I was mainly focussing on what users expect when they come from IDE:

default.txt

I'll see if I have some time next week digging into hl.js again.

mkouhia:
@isagalaev How about now?

isagalaev:
Merged.

You're exactly right about the compromise between cross-language consistency and IDE emulating skewed towards the former, we even have it documented in [the style guide][sg]. Anyway, I was thinking during the week about the `/ ... /` construct and couldn't find a good match among our class names for it. If later we stumble across a similar construct in another language we'll probably define a new class name but I'm reluctant to do it for a single use case.

Thanks for the contribution!

[sg]: http://highlightjs.readthedocs.org/en/latest/style-guide.html

PR #1103, opened 2016-03-04
Added Tagger Script language definition.

phw:
Scripting syntax used by [MusicBrainz Picard](http://picard.musicbrainz.org).

The syntax is somewhat similar to what is used by foobar2000 (called titleformat), but not quite (handling of comments is different).


isagalaev:
JFYI: I adjusted class names to those with better semantics: 0c89400

phw:
@isagalaev Thanks, those changes make sense

PR #1118, opened 2016-03-18
Improve highlighting in Kotlin

temp-impl:
This commit does:

1. improve highlighting of the annotation
  ```kotlin
@file:JvmName("Foo")
// `@file:JvmName` will highlight
```

1. improve highlighting of the class definition
  ```kotlin
class Test @Funcy("Foo") public constructor()
// `@Funcy` will highlight as annotation
// `public constructor` will highlight as keyword
```

1. fix highlighting of the function parameter with annotation
  ```kotlin
fun foo(@Funcy("Foo") arg: Int)
// this will highlight correctly now
// previous version was breaking parameter area by ")"
```

1. highlight the number and string literal in the function
  ```kotlin
fun foo(arg: Int = 0) = "zero"
// the number and string literal will highlight
```

P.S.
There are still some problems.

1. In Kotlin, "break", "continue", "return" can have label like "break@label", and "this" can have class name like "this@Clazz".
These will highlight as annotation in this commit, and I think it shoud not.
But it is hard to exclude these by annotation regex.
Is there any good idea?

1. The type of the constructor parameters is not highlighted.
This may be difficult, because Kotlin has primary and secondary constructors, and "constructor" keyword is optional in primary constructor.

1. Multiple annotations with the same target is not highlighted.
  ```kotlin
@set:[Inject VisibleForTesting]
public var collaborator: Collaborator
```
"@set:" will highlight as annotation, but inside square brackets won't.

I will try them when I have a time...

### DIFF
--- a/src/languages/kotlin.js
+++ b/src/languages/kotlin.js
@@ -10,7 +10,7 @@ function (hljs) {
     keyword:
       'abstract as val var vararg get set class object open private protected public this noinline ' +
       'crossinline dynamic final enum if else do while for when break continue throw try catch finally ' +
-      'import package is in return fun override default companion reified inline ' +
+      'import package is in return fun override companion reified inline ' +
       'interface annotation data sealed internal infix operator out by constructor super ' +
       // to be deleted soon
       'trait volatile transient native default',
--- a/src/languages/kotlin.js
+++ b/src/languages/kotlin.js
@@ -31,6 +31,29 @@ function (hljs) {
     className: 'symbol', begin: hljs.UNDERSCORE_IDENT_RE + '@'
   };
 
+  // for string templates
+  var SUBST = {
+    className: 'subst',
+    variants: [
+      {begin: '\\$' + hljs.UNDERSCORE_IDENT_RE},
+      {begin: '\\${', end: '}', contains: [hljs.APOS_STRING_MODE, hljs.C_NUMBER_MODE]}
+    ]
+  };
+  var STRING = {
+    variants: [
+      {
+        className: 'string',
+        begin: '"""', end: '"""',
+        contains: [SUBST]
+      },
+      hljs.APOS_STRING_MODE,
+      hljs.inherit(
+        hljs.QUOTE_STRING_MODE,
+        {contains: [hljs.BACKSLASH_ESCAPE, SUBST]}
+      )
+    ]
+  };
+
   var ANNOTATION_USE_SITE = {
     className: 'meta', begin: '@(?:file|property|field|get|set|receiver|param|setparam|delegate)\\s*:(?:\\s*' + hljs.UNDERSCORE_IDENT_RE + ')?'
   };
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by temp-impl** [comment]
Solved the problem 1.

I found another problem.

- The annotation with parameter may break "params" area.
  Below is the example.
  ```kotlin
fun foo(
  @Funcy("Foo")
  arg: Int)

fun foo2(@Funcy("Foo") /*comment*/ arg: Int)
```
  Because I used  /\)(?=\s*[=:{\/]|\s*$)/ for detecting ends of the "params".

temp-impl:
I'm thinking this conversion to solve the annotation parameter problem.

```kotlin
@Funcy("Foo")
```

```html
<span class="hljs-meta">@Funcy<span class="hljs-????">(<span class="hljs-string">"Foo"</span>)</span></span>
```

I don't want to highlight "(" and ")", so what CSS class should I use?
Or need new CSS class?

isagalaev:
I haven't look at the code yet, but judging by comments it seems like it might be a little more complicated then it should. Please give us some time to look at it before going any further. Thanks!

isagalaev:
Hello again! I've merged the branch and made some additional changes that should address your questions:

- 6ca08ab280: cleaned up the definition of labeled keywords, so we don't end up with an empty span when there's no label.
- 1848ad710: parens after @annotations are now included in the annotation and strings within those are marked up as 'meta-string'. This is consistent with what we do in other languages
- 007a0722b: since @annotations would eat up all internal parens within params, we can simplify the definitions of `end` as simply being a closing paren

And I've got a few further question of my own:

- Currently things after `:` in params are marked up as 'type'. This class name is actually intended for user defined identifiers in languages with a strong structural concept of types (like Haskell). I don't know Kotlin to be sure but it seems to me that here 'type' is defined by context rather then by the identifier. In other words, if we have `arg: T`, it *doesn't* mean that `T` should be highlighted as 'type' in other places. If yes, I'd suggest we drop the definition of types in arguments here.

- Same goes for `<T>` everywhere: if it's the same thing as generics in languages like Java, C++, Rust, then we don't highlight those as 'type'.

Regardless of these questions, I'll leave everything as is for now as it is a definite improvement over what we had before and it'll be in the 9.3.0 release tomorrow (today in the most parts of the inhabited world :-) )

isagalaev:
Oh, and one more important thing: how do we attribute you in the contributor list? We prefer real names to nick names, so if you're okay with it, could you please name yourself :-)

temp-impl:
Hi,

> In other words, if we have arg: T, it doesn't mean that T should be highlighted as 'type' in other places

This is yes.

> Same goes for <T> everywhere: if it's the same thing as generics in languages like Java, C++, Rust, then we don't highlight those as 'type'.

Kotlin's generics is same things as Java. (keyword "reified" is difference)
I think CSS classes documentation makes little misunderstanding.
It says "type name used in a function signature, type parameter, generic template, etc".

> Oh, and one more important thing: how do we attribute you in the contributor list? We prefer real names to nick names, so if you're okay with it, could you please name yourself :-)

I'd like to add my name to the list in the next pull request.
(I want to fix a few more things)

isagalaev:
> I think CSS classes documentation makes little misunderstanding.
> It says "type name used in a function signature, type parameter, generic template, etc".

Yeah… Only shows how unclear I was about it myself when writing it. I've updated the wording right now, though I'm still not happy with it. Will keep trying :-)

I'm also thinking about introducing a new class, "generic", to mark up everything within `<..>`.

> I'd like to add my name to the list in the next pull request.
> (I want to fix a few more things)

Sure, your contribution is very welcome!

PR #1131, opened 2016-03-24
Improve vala code types

btkostner:
Changes
- types are now in `type`
- single addition of `Posix` to `built_in`


src/languages/vala.js
```diff
@@ -21,8 +18,11 @@ function(hljs) {
         'public private protected internal ' +
         // Other
         'using new this get set const stdout stdin stderr var',
+      type:
+        'char uchar unichar int uint long ulong short ushort int8 int16 int32 int64 uint8 ' +
+        'uint16 uint32 uint64 float double bool struct enum string void',
```
isagalaev: 'Type' is actually intended for user-defined types that are syntactically distinguishable from other identifiers in languages where they are important structural component (like Haskell). In the case of built-in language types we deliberately just call them 'keyword'.

We should definitely improve the description of the class name, as it often causes this confusion…

btkostner: Alright. I'll close this PR then considering there isn't much else besides this. Thanks for your time :)

isagalaev:
Ok :-) I added that "Posix" anyway, thanks for the suggestion!

PR #1132, opened 2016-03-24
Add `hljs` class to code tags even if considered nohighlight

Lance0312:
Without `hljs` class, code block not only is not highlighted,
but also lacks basic CSS decoration.


isagalaev:
This was done intentionally. "nohighlight" essentially means "this is not code, don't touch it", and we don't want apply any styles to those blocks because we don't know what they are for. If an author does want similar styling, they could easily do so in their own stylesheets.

PR #1135, opened 2016-03-30
Add `xsd` as alias for xml syntax

just-boris:
XSD documents are actually XML so it can be highlighted using this language.


baev:
Could you please add `xjb` as well?

Sannis:
And also actualize aliases list in CSS reference in docs.

just-boris:
done

Sannis:
Thanks!

PR #1139, opened 2016-04-06
Remove stray character in xt256

avdv:
CSS error reported by Firefox:

```
Error in parsing value for 'padding'.  Declaration dropped.
```


isagalaev:
Doh! That probably was my fat finger during merging… Thanks!

PR #1141, opened 2016-04-07
PureBASIC language

tajmone:
I've added PureBASIC language to highlight.js.
Tested, works well.  Sample code in `detect` dir.

Also added a stylesheet called PureBASIC, which emulates its native IDE colors, but covers all hljs classes, so it works with all languages.

All PureBASIC files are in a single commit. Changes to contributors lists (EN/RU) and CHANGES.MD come as individual commits (just in case...)

### DIFF
--- a/src/languages/purebasic.js
+++ b/src/languages/purebasic.js
@@ -13,13 +13,11 @@ function(hljs) {
 		className: 'string',
 		begin: '(~)?"', end: '"',
 		illegal: '\\n',
-		relevance: 0
 	};
 	var CONSTANTS = { // PB IDE color: #924B72 (Cannon Pink)
 		//  "#" + a letter or underscore + letters, digits or underscores + (optional) "$"
 			className: 'symbol',
 			begin: '#[a-zA-Z_]\\w*\\$?',
-			relevance: 0
 	};
 	var PROCEDURES_RHS = { // PB IDE color: #006666 (Blue Stone)
 		// ")"
--- a/src/languages/purebasic.js
+++ b/src/languages/purebasic.js
@@ -1,0 +1,64 @@
+/*
+Language: PureBASIC
+Author: Tristano Ajmone <tajmone@gmail.com>
+Description: Syntax highlighting for PureBASIC (v.5). No inline ASM highlighting. First release (v.1.0), April 2016.
+Credits: I've taken inspiration from the PureBasic language file for GeSHi, created by Gustavo Julio Fiorenza (GuShH).
+Category: misc
+*/
+
+// Base deafult colors in PB IDE: background: #FFFFDF; foreground: #000000;
+
+function(hljs) {
+	var STRINGS = { // PB IDE color: #0080FF (Azure Radiance)
+		className: 'string',
+		begin: '(~)?"', end: '"',
+		illegal: '\\n',
+		relevance: 0
+	};
+	var CONSTANTS = { // PB IDE color: #924B72 (Cannon Pink)
+		//  "#" + a letter or underscore + letters, digits or underscores + (optional) "$"
+			className: 'symbol',
+			begin: '#[a-zA-Z_]\\w*\\$?',
+			relevance: 0
+	};
+	var PROCEDURES_RHS = { // PB IDE color: #006666 (Blue Stone)
+		// ")"
+		className: 'function',
+		begin: '\\)',
+		relevance: 0
+	};
+
+  return {
+		aliases: ['pb', 'pbi'],
+		keywords: {
+			keyword: // PB IDE color: #006666 (Blue Stone)
+				// The following keywords list was taken and adapted from GuShH's PureBasic language file for GeSHi...
+				'And As Break CallDebugger Case CompilerCase CompilerDefault CompilerElse CompilerEndIf CompilerEndSelect ' +
+				'CompilerError CompilerIf CompilerSelect Continue Data DataSection EndDataSection Debug DebugLevel Declare ' +
+				'DeclareCDLL DeclareDLL Default Define Dim DisableASM DisableDebugger DisableExplicit Else ElseIf EnableASM ' +
+				'EnableDebugger EnableExplicit End EndEnumeration EndIf EndImport EndInterface EndMacro EndProcedure ' +
+				'EndSelect EndStructure EndStructureUnion EndWith Enumeration Extends FakeReturn For Next ForEach ' +
+				'ForEver Global Gosub Goto If Import ImportC IncludeBinary IncludeFile IncludePath Interface Macro ' +
+				'NewList Not Or Procedure ProcedureC ProcedureCDLL ProcedureDLL ProcedureReturn Protected Prototype ' +
+				'PrototypeC Read ReDim Repeat Until Restore Return Select Shared Static Step Structure StructureUnion ' +
+				'Swap To Wend While With XIncludeFile XOr'
+		},
+		contains: [
+			// PROCEDURES_LHS | PB IDE color: #006666 (Blue Stone)
+			{
+				className: 'function',
+				begin: '[a-zA-Z_]\\w*\\s*\\(',
+				relevance: 0,
+			},
+			STRINGS,
+			CONSTANTS,
+			PROCEDURES_RHS,
+			
+			{	// COMMENTS | PB IDE color: #00AAAA (Persian Green)
+				className: 'comment',
+				begin: ';', end: '$',
+				relevance: 10
+			}
+		]
+	};
+}
--- a/src/languages/purebasic.js
+++ b/src/languages/purebasic.js
@@ -12,12 +12,12 @@ function(hljs) {
 	var STRINGS = { // PB IDE color: #0080FF (Azure Radiance)
 		className: 'string',
 		begin: '(~)?"', end: '"',
-		illegal: '\\n',
+		illegal: '\\n'
 	};
 	var CONSTANTS = { // PB IDE color: #924B72 (Cannon Pink)
 		//  "#" + a letter or underscore + letters, digits or underscores + (optional) "$"
 			className: 'symbol',
-			begin: '#[a-zA-Z_]\\w*\\$?',
+			begin: '#[a-zA-Z_]\\w*\\$?'
 	};
 	var PROCEDURES_RHS = { // PB IDE color: #006666 (Blue Stone)
 		// ")"
@@ -42,6 +42,9 @@ function(hljs) {
 				'Swap To Wend While With XIncludeFile XOr'
 		},
 		contains: [
+			// COMMENTS | PB IDE color: #00AAAA (Persian Green)			
+			hljs.COMMENT(';', '$', {relevance: 0}),
+			
 			// PROCEDURES_LHS | PB IDE color: #006666 (Blue Stone)
 			{
 				className: 'function',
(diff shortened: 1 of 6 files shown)


isagalaev:
Hi! Thanks for the contribution. I've got a few comments on the code…

src/languages/purebasic.js
```diff
@@ -0,0 +1,64 @@
...
+*/
+
+// Base deafult colors in PB IDE: background: #FFFFDF; foreground: #000000;
+
+function(hljs) {
+	var STRINGS = { // PB IDE color: #0080FF (Azure Radiance)
+		className: 'string',
+		begin: '(~)?"', end: '"',
+		illegal: '\\n',
+		relevance: 0
```
isagalaev: No need to zero relevance of strings, it will artificially de-value this language relevance compared to the rest. Just drop the attribute so it gets a default value.

src/languages/purebasic.js
```diff
@@ -0,0 +1,64 @@
...
+	};
+	var CONSTANTS = { // PB IDE color: #924B72 (Cannon Pink)
+		//  "#" + a letter or underscore + letters, digits or underscores + (optional) "$"
+			className: 'symbol',
+			begin: '#[a-zA-Z_]\\w*\\$?',
+			relevance: 0
```
isagalaev: Same thing here, no need to zero relevance of a legitimate language construct without a particular reason.

src/languages/purebasic.js
```diff
@@ -0,0 +1,64 @@
...
+	};
+	var PROCEDURES_RHS = { // PB IDE color: #006666 (Blue Stone)
+		// ")"
+		className: 'function',
+		begin: '\\)',
+		relevance: 0
+	};
```
isagalaev: A single closing parentheses is certainly not a function declaration. I guess it was added to better match what the IDE does but in highlight.js consistency across the languages always beats IDE compatibility. In fact, it's even an explicit design choice that highlight.js doesn't try to emulate other highlighting engines. So let's not highlight that :-)

src/languages/purebasic.js
```diff
@@ -0,0 +1,64 @@
...
+				'PrototypeC Read ReDim Repeat Until Restore Return Select Shared Static Step Structure StructureUnion ' +
+				'Swap To Wend While With XIncludeFile XOr'
+		},
+		contains: [
+			// PROCEDURES_LHS | PB IDE color: #006666 (Blue Stone)
+			{
+				className: 'function',
+				begin: '[a-zA-Z_]\\w*\\s*\\(',
+				relevance: 0,
+			},
```
isagalaev: Same here: it will highlight every procedure *call*, but we only highlight function names at the place of declaration.

src/languages/purebasic.js
```diff
@@ -0,0 +1,64 @@
...
+			STRINGS,
+			CONSTANTS,
+			PROCEDURES_RHS,
+			
+			{	// COMMENTS | PB IDE color: #00AAAA (Persian Green)
+				className: 'comment',
+				begin: ';', end: '$',
+				relevance: 10
```
isagalaev: ;-comments are quite common, no need to set the relevance so high. In fact, it should probably be zero in order to languages with ;-delimited statements won't accumulate PureBasic score.

isagalaev: Oh, and the proper way to define comments these days is to use `hljs.COMMENT(';', '$', {relebance: 0})` (sorry, it seems we forgot to document it properly.)

src/languages/purebasic.js
tajmone: If I remember correctly, I ended up adding these `relevance: 0` during testing: I was testing with PureBASIC and Python (only 2 langs build for testing reasons) and hjs didn't distinguish well between the two, until I added the `relevance: 0`.

I'll remove them and update the commits!

src/languages/purebasic.js
tajmone: Right. Not sure I follow you here. The problem I faced was this: I couldn't get hjs to catch PB Procedures (functions) correctly — I've worked first on having hjs catch a function wholly, ie: from `xxx(` to `)`, but then I didn't manage to have it parse correctly strings, parameters, and other values passed within the braces.

I admint that I've worked on it with the intention to create (mainly) I highlighting script for a specific project oriented toward PureBASIC, trying to emulate the exact way it colors code in the language IDE — and the script works fine for that. But then I also worked to make it stick to general guidelines of being language-agnostic (in fact, the stylesheet was created to cover all possibile languages).

Mainly, the problem I couldn't solve was to have hjs reckognize a Function, excluding the braces (which I wanted to be kept in a separate colour), and then have it parse inside the braces for the expected parameters types.

How should I go on about it? I'm not quite sure if the problem has to do with "logical structure" of the final tags (ie: parameters should be a sub-element within the braces), or with the fact that splitting a function call (ie: "` xxx(`" and "`)`")as if they were 2 independent items is problematic.

src/languages/purebasic.js
tajmone: But this is exactly how PureBASIC highlights the code in its IDE. You mean I should have the parser provide different class names for procedure declaration and procedure calls? ie: the problem being at the semantica level, independently of how I decide to color it?

src/languages/purebasic.js
tajmone: Ok, I'll look for some examples into the implemented languages, and apply it this way.

Thanks a lot for the help, it's the first time I work with hjs, and I'm enjoying it. I've actually took quite some time to work on this syntax files, (a couple of week, no rushing at all, and reading the docs over and over.)

src/languages/purebasic.js
tajmone: **FIXED!**

src/languages/purebasic.js
tajmone: Ok. I've been studying some other langs sources, and I seem to gather that what you are telling me is that I should implement in the code to capture Procedure definitions by intercepting their keyword (ie: `Procedure` and `Declare`, in PureBASIC), and have it contain the Procedure `Title` (ie: the identifier that follows these keywords, before opening brace).

I can manage that, but I have a couple of questions:

* Since I will be defining the `Procedure` and `Declare` keywords in the RE that captures their declaration, should I remove them from the main keywords list?

* As for capturing and highlighting the Procedure _calls_, what tag should these calls go by?

Hjs documentation is good, but at times a bit tight in examples (especially regarding the HTML final tags), and I have trouble with the CSS classes reference page — I don't seem to find a class for functions/procedure class.

PureBASIC employs a very simple and minimalistic highlighting style, but I would like to be able to catch all the required case in order to be able to reproduce on my custom stylesheet the exact coloring of its IDE. So I need to catch Procedure calls and tag them somehow.

Any suggestions?

tajmone:
Ok. I think everything should be up to standards now:

Procedure declarations are now wrapped under `function` class, and followed by `keyword`, `type` (PB attaches return type declaration to the `Declare`/`Procedure` keyword themselves!).

`function` class ends at opening brace `(` — didn't include `params`, as this would complicate things, and PureBASIC has no special highlighting for Procedure parameters...

I also added a class called `function_call`, so that Procedure calls could be highlighted also.

PureBASIC has a tricky syntax, especially since types are specified by adding a dot followed by the type name (be it built-in type, or a custom type). In many ways it doesn't mirror other langs standards when it comes to syntax. According to its syntax highiliting library, a Procedure is a Procedure (whether it's being called or defined).

Hopefully, now the syntax files should be ready for merging.

Why does Travis complain? It mentions conflicts in the `README` file... I had a look at the conflicts details, but didn't make much sense of them...

isagalaev:
@tajmone just to let you know, I'm not ignoring your messages :-) I only can work on highlight.js about once a week, so I'll try to get back to you after the weekend.

tajmone:
No problem @isagalaev ! No hurry on this one.

As far as I am concerned, the syntax is working fine for PureBASIC. So if it's ok for you to merge it, and the chaos of multiple commits doesn't pose a problem, you could go ahead and merge it.

Else, if you need me to squash the commits, let me know — I don't master well Git, and squashing commits is a bit tedious for me unless I simply squash them all in one single commit (wouldn't know how to keep the README changes in separate commit at this point)

isagalaev:
> Else, if you need me to squash the commits, let me know

To be honest, I have no idea why one might want to do that! Individual commits are useful for resolving merging issues and reverting changes, so it's better to have them granular. When reviewing a set of changes you just look at the overall diff anyway.

isagalaev:
While I'm looking at the code, let me address some of your comments:

> As for capturing and highlighting the Procedure *calls*, what tag should these calls go by?

None. We don't highlight procedure/function calls at all. There are many highlighters out there that highlight much more types of tokens: punctuation, infix operators, CamelCased identifier, method calls starting with a dot (`.something()`). We try to keep our highlighting minimal and useful which results in some decisions that might seem arbitrary. Simply using an identifier in the code, be it an assignment, an operator or a function call does not deserve it being specially highlighted.

> but can [beginKeywords] contain pipe-seprated relevance values

No, only space-separated words. There's, in fact, very few reasons to tune relevance explicitly, especially for an outside contributor who can't possibly know all the patterns of dealing with cross-detection bugs. My usual advice is to get your own language correct and leave relevance magic for core developers. I'd love to have it documented somewhere but it's a) really hard to formulate and b) changes too fast.

isagalaev:
Okay. I've merged the code and made some [post-merge fixes][1] to both the language and the style. Feel free to raise any concerns. Thanks for the contribution :-)

[1]: https://github.com/isagalaev/highlight.js/compare/df65dd3...62ba6c9

tajmone:
Thanks for merging!

I've looked at the post-merging diff, so it looks like the function-call will not be part of the final package. I'll keep in my repo a separate branch in which I'll add it, because I need it for a mainly-PureBASIC project in which I'm striving to replicate IDE coloring — along with a custom sylesheet that employs different coloring for each language I'll be incorporating.

I guess that PureBASIC philosophy is a bit ad odds here: the language aims at highlighting Procedure names, always; and procedure declarations fall under the general keywords rules, because function definitions are always preceded by `Procedure` or `Define` keywords — somehow that's what PureBASIC users would expect since the language is tightly linked to it's IDE.

I'm really glad to see PureBASIC syntax in highlight.js — this is the first JS syntax highlighter reckognizing PureBASIC that I am aware of. There is an available syntax in GeSHi, but that's PHP, and server side; and there is an online highlighter via form pasting, and that's about all (apart from highlighter syntaxes for code editors).

So, I hope it will open up the path for PureBASIC code readability.

Hopefully, one day I'll add to it inline ASM highlighting also — by default, the IDE doesn't highlight it. But I'll need to look into it very closely, as it would mean also switching to case-sensitive mode, to avoid conflicts between `And` and `AND`, `Or` and `OR`, and so on...

Thanks again!

isagalaev:
Just to clarify, we prefer consistency across highlight.js languages to consistency within a particular language ecosystem simply because we'd never be able to maintain the latter. That would require having an active trusted maintainer familiar with highlight.js architecture for *every* language and we don't live in such a perfect Universe :-) Keeping a local fork is a good solution indeed.

> Hopefully, one day I'll add to it inline ASM highlighting also — by default, the IDE doesn't highlight it. But I'll need to look into it very closely, as it would mean also switching to case-sensitive mode, to avoid conflicts between And and AND, Or and OR, and so on...

If you do it as a sub-language, it will be parsed completely independently from outer rules.

tajmone:
True, but you might consider adding a single generic class for custom uses (like `.custom`, `.generic`, `.extra` or `.unspecified`) which could be employed for edge cases that don't fit anywhere else — this would circumvent the issue of maintainance.

I've now turned the PureBASIC branch into a forked subproject for PB syntax projects — added notice in the README, to make it clear.

PR #1143, opened 2016-04-12
Handle namespaced keywords in clojure.

mahinshaw:
For reference: http://clojure.org/reference/reader


Sannis:
Thanks for your contribution!

PR #1144, opened 2016-04-12
BNF language definition

Sannis:


src/languages/bnf.js
```diff
@@ -0,0 +1,29 @@
...
+        begin: /</,
+        end: />/
+      },
+      // Specific
+      {
+        begin: /::=/
+      },
+      {
+        className: 'string',
+        begin: /'"'/
```
isagalaev: Just a single quote? Why not the standard `hljs.QUOTE_STRING_MODE`?

isagalaev: Oh, I see it further down. But what this is for then?

src/languages/bnf.js
```diff
@@ -0,0 +1,29 @@
...
+
+function(hljs){
+  return {
+    contains: [
+      // Attribute
+      {
+        className: 'attr',
+        begin: /</,
+        end: />/
+      },
```
isagalaev: This will highlight those both on the left and on the right side of `::=`. How about defining two sides (like in 'ini.js' for example) and then only highlight `<..>` as attrs on the left and everything else on the right. This will be much more robust for detection, too.

isagalaev: Oh, and I think they are should be 'attribute', not just 'attr', as they define structure. In other words, 'attr' is for something small and simple, like HTML attributes inside tags, and 'attribute' is for something like properties in CSS.

src/languages/bnf.js
Sannis: Yes, we can use standard mode here.

src/languages/bnf.js
Sannis: > How about defining two sides (like in 'ini.js' for example) and then only highlight <..> as attrs on the left and everything else on the right?

Wikipedia highlight <attribute> similar on both sides, I'm just following that style.

> This will be much more robust for detection, too.
It works fine now, so I think we can change this in future if decide.

Sannis:
Unfortunately I have no ideas how to detect EBNF that have too weak syntax.

src/languages/bnf.js
isagalaev: > Wikipedia highlights similar on both sides, I'm just following that style.

Okay. It can always be improved later anyway.

PR #1145, opened 2016-04-15
Add string interpolation to C#

temp-impl:
String interpolation and nameof expression are new features in C# 6
https://github.com/dotnet/roslyn/wiki/New-Language-Features-in-C%23-6

C# string interpolation is complicated, because it can be nested and it can be combined with the verbatim string.
Inside normal interpolated string (not verbatim one), raw LF is not allowed even if in the nested verbatim string.

Examples:
```cs
// OK
var str1 = $"this is {@"OK"}";

var str2 = $@"this is {@"
OK"} too";

// ERROR (compilation error)
var str3 = $"this is {@"
ERROR"}";
```


isagalaev:
Thanks for a comprehensively implemented feature!

PR #1146, opened 2016-04-15
Add Ada support

captain-hanuta:
`node tools/build.js -n -t node $(ls src/languages/ | sed -e 's|\.js$||') && npm test`
runs straight through without problems.

Auto detection mistook  xml and elm files for ada, but I could fix that by
tweaking relevance a bit.
The same goes for vbscript-html, however, the race is still kinda close.


isagalaev:
Thanks, we'll look into it in the near future, as time allows. By the way:

> `node tools/build.js -n -t node $(ls src/languages/ | sed -e 's|\.js$||') && npm test`

You can omit the list of languages if you want all of them, and you can also omit `-n` (because it shouldn't affect tests and node builds aren't compressed anyway). So it becomes just this: `node tools/build.js -t node && npm test`

> Auto detection mistook xml and elm files for ada, but I could fix that by
tweaking relevance a bit.

Tweaking relevance is usually the last resort. Usually such different languages as Ada and XML shouldn't be close. I'll look into it too.

captain-hanuta:
>You can omit the list of languages if you want all of them

Thanks. `node tools/build.js --help`  doesn't mention this, I think it should.

>Tweaking relevance is usually the last resort. Usually such different languages as Ada and XML shouldn't be close. I'll look into it too.

It was the SOAP test thing in the test/detect/xml folder.
All those : made by ada definition fire quite often (local variable Foo : Bar).

Is there something like a debug mode where I can see which modes fired when and with what input?

PR #1147, opened 2016-04-16
Fix intraword builtins in objc

bgorven:


isagalaev:
Thanks!

(Test breakage seems to be caused by some intermittent problem with TravisCI. Works fine for me locally.)

PR #1152, opened 2016-04-19
Improve rules for PowerShell comments and strings

OiCMudkips:
The current PowerShell rules in highlight.js don't some of the rules for PowerShell [string escaping](https://technet.microsoft.com/library/hh847740.aspx) and [function documentation](https://technet.microsoft.com/en-us/library/hh847834.aspx), as well as multi-line comments in general.

This pull request adds the above rules to highlight.js.


Sannis:
Thanks!

PR #1154, opened 2016-04-20
Captured more verilog built_in keywords

BooneJS:
SystemVerilog has many built_in functions. I've captured the rest of them here. I'm done for a while after this. :)


src/languages/verilog.js
```diff
@@ -83,7 +93,11 @@ function(hljs) {
       {
         className: 'meta',
         begin: '`', end: '$',
-        keywords: {'meta-keyword': 'ifdef ifndef endif'},
+        keywords: 'define __FILE__ ' +
```
isagalaev: They should still be 'meta-keyword' to work well with all the styles.

BooneJS: Done.

isagalaev:
Thanks!

PR #1163, opened 2016-04-30
Accept trailing 'f' in C_NUMBER_RE

mkeeter:
When I highlight a floating-point constant like `0.0f`, I'd expect the trailing `f` to be highlighted with the rest of the number.

This patch changes [C_NUMBER_RE](https://github.com/isagalaev/highlight.js/blob/master/src/highlight.js#L683) to accept an optional trailing `f`.

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -680,7 +680,7 @@ https://highlightjs.org/
   hljs.IDENT_RE = '[a-zA-Z]\\w*';
   hljs.UNDERSCORE_IDENT_RE = '[a-zA-Z_]\\w*';
   hljs.NUMBER_RE = '\\b\\d+(\\.\\d+)?';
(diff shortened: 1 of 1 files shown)


isagalaev:
Changing the base regex is not a good idea as quite a few languages use it and they all have different rules about postfix letters (if any). If there's a problem with a particular language it's better to change only that.

mkeeter:
Fair enough.  I noticed this with GLSL, but it should also apply to C++ and C (among others – those are the three that I know for sure off the top of my head).

isagalaev:
C/C++ has an [even more extensive definition][1] at the moment. If you know that GLSL has something similar, could you make a patch against it?

[1]: https://github.com/isagalaev/highlight.js/blob/master/src/languages/cpp.js#L32

PR #1174, opened 2016-05-13
Add defprotocol and defimpl to Elixir highlighting

wjdix:
Add `defprotocol` and `defimpl` to Elixir and treat them similarly to `defmodule` and `defrecord`


Sannis:
@wjdix thanks!

PR #1177, opened 2016-05-15
Implemented an Awk syntax highlighter

matthewbdaly:
As there isn't yet a syntax highlighter for Awk, I've created one. It's pretty basic as I'm not the most skilled Awk user and ths is my first contribution to this projecti, but it does seem to work OK.


matthewbdaly:
Hmm... The build seems to fail when any environment variables get set. A tad frustrating...

matthewbdaly:
Fixed it - down to a superfluous semicolon.

isagalaev:
Thanks for your contribution, Matthew!

PR #1178, opened 2016-05-16
Elm code for rotation on highlightjs.org

jvoigtlaender:
Changed the test snippet in test/detect/elm/default.txt to fit into 54 columns and 16 rows of text, to enable it being showed on the highlightjs.org site, as per @isagalaev's suggestion in https://github.com/isagalaev/highlight.js/pull/1172#issuecomment-219316266.


isagalaev:
To be honest, I miscalculated the width somehow, it could only fit 55 characters. But I went ahead and made the box roomier as there was at least one other snippet that didn't actually fit in the old width.

Thanks!

PR #1179, opened 2016-05-16
More Objective-C keywords, prefixes, attributes

1ec5:
Added a number of Objective-C keywords, property attributes, and method attributes as keywords. Also added some common Apple class prefixes.

I didn’t have an official Objective-C grammar handy, so I consulted the following sources in addition to general knowledge of the language:

* [ParseObjc.cpp](https://github.com/llvm-mirror/clang/blob/3ae2ef0f0bbfcb5829ac2ac4f63ac555d94e3f78/lib/Parse/ParseObjc.cpp) and [TokenKinds.def](https://github.com/llvm-mirror/clang/blob/fa54202263d4a62bd3153cb9037d8801ebbdc15a/include/clang/Basic/TokenKinds.def) in Clang (I ignored anything that looked C++-only in the latter)
* NSObjCRuntime.h in the iOS SDK

The additional class prefixes are as follows:

Class prefix | Framework
----|----
CL | Core Location
CM | Core Motion
CN | Contacts
CT | Core Text
MTK | MetalKit
MTL | Metal
SCN | SceneKit
SK | SpriteKit, StoreKit
WK | WebKit, WatchKit

/cc @tmcw

### DIFF
--- a/src/languages/objectivec.js
+++ b/src/languages/objectivec.js
@@ -8,7 +8,7 @@ Category: common
 function(hljs) {
   var API_CLASS = {
     className: 'built_in',
-    begin: '\\b(AV|CA|CF|CG|CI|CL|CN|MK|MP|MTK|MTL|NS|SCN|SK|UI|WK|XC)\\w+',
+    begin: '\\b(AV|CA|CF|CG|CI|CL|CM|CN|CT|MK|MP|MTK|MTL|NS|SCN|SK|UI|WK|XC)\\w+',
   };
   var OBJC_KEYWORDS = {
     keyword:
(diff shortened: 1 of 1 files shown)


friedbunny:
Some other commonish/new class prefixes, if you’re feeling inclined to keep expanding. Even just on iOS, [there are a lot of public frameworks](https://github.com/nst/iOS-Runtime-Headers/tree/master/Frameworks).

Prefix | Framework
|---|---|
AB | AddressBook
CB | CoreBluetooth
CK | CloudKit
GC | GameController
GLK | GLKit
GK | GameplayKit
HK | HealthKit
HM | HomeKit
JS | JavaScriptCore
SL | Social

1ec5:
I figured the list could be restricted to the most commonly used frameworks (though I have no idea which ones they are), to minimize the risk of snagging non-Apple class prefixes that just happen to conflict with a slightly less common Apple class prefix.

Sannis:
@1ec5 thanks!

PR #1183, opened 2016-05-22
Improve coding style

marcobiedermann:
* sort properties alphabetically
* add space after color
* indent using two spaces
* shorten hex color notation
* use hex notation instead of keyword
* use longhand instead of shorthand
* remove trailing whitespace
* write color hex notation in lowercase

### DIFF
--- a/src/styles/agate.css
+++ b/src/styles/agate.css
@@ -17,11 +17,11 @@
  */
 
 .hljs {
+  background-color: #333;
+  color: white;
   display: block;
   overflow-x: auto;
   padding: 0.5em;
-  background: #333;
-  color: white;
 }
 
 .hljs-name,
--- a/src/styles/androidstudio.css
+++ b/src/styles/androidstudio.css
@@ -4,8 +4,8 @@ Author: Pedro Oliveira <kanytu@gmail . com>
 */
 
 .hljs {
+  background-color: #282b2e;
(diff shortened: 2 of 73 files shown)


isagalaev:
Frankly, I don't see how it improves anything. It's a purely stylistic change from one preference to another.

derhuerst:
Note that some of it aligns the code to [what's stated in the `.editorconfig`](https://github.com/isagalaev/highlight.js/blob/4c96012/.editorconfig).

marcobiedermann:
@isagalaev It helps to ensure a better coding style, which is easier to maintain and to extend.
Also it helps saving a few bytes when using GZIP compression.

derhuerst:
> @isagalaev It helps to ensure a better coding style, which is easier to maintain and to extend.

IMHO that's a matter of taste.

Sannis:
I have no objections to accept or decline this, hope this will not become code-style war :-)

isagalaev:
Sorry for taking so long to answer, I've been away from the project.

Ultimately, we don't want this change as it doesn't bring anything substantial. `#f0f` is not easier to maintain than `#ff00ff`, neither `#6a8759` vs `#6A8759`, nor `background-color` vs `background`. We simply don't have any tooling that might depend on it.

I appreciate the effort, though.

PR #1191, opened 2016-05-26
[LSL] Add Second Life's Linden Scripting Language

buildersbrewery:
fixes buildersbrewery/linden-scripting-language#6

---

* Adds [LSL](https://wiki.secondlife.com/wiki/LSL_Portal) for [Second Life](https://www.secondlife.com)
  * test file included
* Adds style of official Linden Lab viewer style for LSL code.
* As per request by @codeviolet for the [question2answer](http://question2answer.org) installation on [avsitter.com](https://avsitter.com)

---

Not sure about the CSS code and naming conventions to references the highlight.js tags from the language. Appreciate feedback to get this right!


buildersbrewery:
That error message on Travis doesn't help me. Any hints?

Makopo:
I totally don't know about highlight.js but, perhaps around these line?
```
                contains: [
                    { begin: '\\[\\\\nt"\\]' },
                    { subLanguage: 'json' },
                    { subLanguage: 'xml' }
                ]
```
I'm not sure the test case will work with the element with only "subLanguage" property or dual "subLanguage" elements in the list. You might want to try to commit without whole "contains" block, then without both subLanguage elements, then without one of subLanguage element to see what will happen.

buildersbrewery:
Neither of these worked:

* removed `subLanguages` from string
* and then whole `contains` block there within.

---

/cc @Makopo thanks for the suggestions, though.

isagalaev:
I've seen a flurry of activity here today :-) I'll try to look at it soon-ish. By the way, here's the reference doc about `subLanguage`: http://highlightjs.readthedocs.io/en/latest/reference.html#sublanguage

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
+/*
+    Language: Linden Scripting Language
+    Description: The Linden Scripting Language is used in Second Life by Linden Labs.
+    Author: Builder's Brewery <buildersbrewery@gmail.com>
+    Category: common, scripting
```
isagalaev: The category 'common' is special as it controls what goes into the default hosted build which we don't want to become too big. Also, LSL is hardly common by any definition of the word :-)

isagalaev: Also, please remove the indentation within the meta comment, it's not supposed to be there and may suddenly stop working if anyone decides to rewrite its parsing for any reason.

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+*/
+
+function(hljs) {
+    return {
+        case_insensitive: false,
```
isagalaev: `false` is the default, not need to specify it.

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+        aliases: ['lsl'],
```
isagalaev: No need to mention alias which equals the file names — it works as a canonical name already.

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+        illegal: ':',
+        contains: [
+            {
+                className: 'string',
+                begin: '"',
+                end: '"'
+            },
```
isagalaev: Any reason why you couldn't use hljs.QUOTE_STRING_MODE here?

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+            {
+                className: 'comment',
+                begin: '//',
+                end: '$'
+            },
+            {
+                className: 'comment',
+                begin: '/\\*',
+                end: '\\*/'
+            },
```
isagalaev: Please do use the helper to define comments: http://highlightjs.readthedocs.io/en/latest/language-guide.html#comments

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+            {
+                className: 'number',
+                begin: '(\\b0[xX][a-fA-F0-9]+|(\\b\\d+(\\.\\d*)?|\\.\\d+)([eE][-+]?\\d+)?)'
+            },
```
isagalaev: Looks like one of the standard modes may work here as well: https://github.com/isagalaev/highlight.js/blob/master/src/highlight.js#L726-L740

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+            {
+                className: 'state',
```
isagalaev: We use a pre-defined set of class names to ensure that all languages work well with all the styles. The philosophy is explained in the [style guide](http://highlightjs.readthedocs.io/en/latest/style-guide.html) but it boils down to simply using only class names from http://highlightjs.readthedocs.io/en/latest/css-classes-reference.html

On a related note, the CSS file you provided in this pull request is not needed per se, unless you plan to turn it into a general (as opposed to language specific) theme.

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+                        contains: [
+                          'self',
+                          'constant.float',
+                          'constant.integer',
+                          'constant.quaternion',
+                          'constant.string',
+                          'constant.vector',
+                          'string',
+                          'number'
+                        ]
```
isagalaev: `contains` should contains actual JavaScript objects, not names (with the exception of 'self', which is a syntactic sugar). To reuse numbers, strings, etc. assign them to regular variables beforehand. Look into, say, https://github.com/isagalaev/highlight.js/blob/master/src/languages/cpp.js for an example.

src/languages/lsl.js
```diff
@@ -0,0 +1,133 @@
...
+            { className: 'operator.bit-wise.xor', begin: '\\^' },
+            { className: 'operator.bit-wise.and', begin: '&' },
+            { className: 'operator.bit-wise.or', begin: '\\|' },
+            { className: 'operator.comparison.equal', begin: '==' },
+            { className: 'operator.comparison.not', begin: '!=' },
+            { className: 'operator.logical.not', begin: '!' },
+            { className: 'operator.comparison.less', begin: '<' },
+            { className: 'operator.comparison.greater', begin: '>' },
+            { className: 'operator.assignment.augmented', begin: '[%*+\\/\\-]=|=' },
+            { className: 'operator.arithmetic', begin: '[%*+\\/\\-]' }
```
isagalaev: By stylistic choice we don't style punctuation and operators in highlight.js.

test/markup/lsl/hello-avatar.expect.txt
```diff
@@ -0,0 +1,12 @@
...
+    <span class="hljs-event">state_entry</span>()
+    {
+        <span class="hljs-function">llSay</span>(<span class="hljs-constant.integer">PUBLIC_CHANNEL</span>, "Hello, Avatar!");
+    }
+
+    <span class="hljs-event">touch_start</span>(<span class="hljs-type">integer</span> num_detected)
+    {
+        <span class="hljs-function">llSay</span>(<span class="hljs-constant.integer">PUBLIC_CHANNEL</span>, "Touched.");
+    }
+}
```
isagalaev: Markup tests are intended to test specific features of the language. If it simply duplicates the detection test, it's not needed.

isagalaev:
Hello again! I've added quite a lot of inline comments, I hope it won't put you off the task :-) We could go back to test breakage after those are addressed, as it'd change a lot.

buildersbrewery:
* Is there a way to include within strings contained JSON and XML with the escape rules for LSL strings?
* Any plans regarding an `illegal` tag for keywords? Currently it's restricted to symbols and breaks highlighting.

isagalaev:
> Is there a way to include within strings contained JSON and XML with the escape rules for LSL strings?

Could you show an example with those escape rules? I don't think I'm quite getting the picture…

> Any plans regarding an illegal tag for keywords? Currently it's restricted to symbols and breaks highlighting.

What is it that you're trying to express with it? Illegals and keywords are completely different animals, they shouldn't depend on each other in any way.

src/languages/lsl.js
```diff
@@ -0,0 +1,90 @@
...
+Description: The Linden Scripting Language is used in Second Life by Linden Labs.
+Author: Builder's Brewery <buildersbrewery@gmail.com>
+Category: scripting
+*/
+
+function(hljs) {
+
+    var LSL_STRING_ESCAPE_CHARS = {
+        className: 'subst',
+        begin: '\\[\\\\nt"\\]'
```
isagalaev: This looks like it has too many back-slashes in it. In this form it will only match a single literal string `[\nt"]` which is probably not what it's supposed to mean? If LSL strings are simply backslah-escaped then we have a standard mode for that: `hljs.BACKSLASH_ESCAPE`.

src/languages/lsl.js
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    var LSL_STRINGS = {
+        className: 'string',
+        begin: '"',
+        end: '"',
+        contains: [
+            { begin: '/\\*', end: '\\*/', skip: true },
```
isagalaev: I'm pretty sure you don't need to use `skip` here, as it's a hack for some special case of sub-languages. But I don't get what this mode was supposed to do inside strings. Do `/* .. */` comments have a special meaning *inside* strings?

src/languages/lsl.js
```diff
@@ -0,0 +1,90 @@
...
+                begin: '\\b(?:EOF|JSON_(?:ARRAY|DELETE|FALSE|INVALID|NULL|NUMBER|OBJECT|STRING|TRUE)|NULL_KEY|TEXTURE_(?:BLANK|DEFAULT|MEDIA|PLYWOOD|TRANSPARENT)|URL_REQUEST_(?:GRANTED|DENIED))\\b'
+            },
+            {
+                begin: '\\b(?:ZERO_VECTOR|TOUCH_INVALID_(?:TEXCOORD|VECTOR))\\b'
+            }
+        ]
+    };
+
+    var LSL_FUNCTIONS = {
+        className: 'function',
```
isagalaev: The 'function' class name is for user-defined functions (like `function something()` in JS, for example), and these look like built-ins.

src/languages/lsl.js
```diff
@@ -0,0 +1,90 @@
...
+    var LSL_NUMBERS = {
+        className: 'number',
+        begin: hljs.C_NUMBER_RE
+    };
+
+    var LSL_CONSTANTS = {
+        className: 'built_in',
+        variants: [
+            {
+                begin: '\\b(?:PI|TWO_PI|PI_BY_TWO|DEG_TO_RAD|RAD_TO_DEG|SQRT2)\\b'
```
isagalaev: If `PI`, `TWO_PI` and `PI_BY_TWO` aren't functions (I suspect they aren't) then it makes sense to use the 'literal' class names for them.

src/languages/lsl.js
```diff
@@ -0,0 +1,90 @@
...
+            },
+            {
+                begin: '\\b(?:FALSE|TRUE)\\b'
```
isagalaev: These feel like 'literal' too. And probably some others too, I can't really say as I don't know the language :-) But the rule of thumb is simple: built-in constant values usually go as 'literal', and every other kind of built-in is 'built_in'.

src/languages/lsl.js
```diff
@@ -0,0 +1,90 @@
...
+            {
+                className: 'comment',
+                variants: [
+                    hljs.COMMENT('//', '$'),
+                    hljs.COMMENT('/\\*', '\\*/')
+                ]
+            },
+            LSL_NUMBERS,
+            {
+                className: 'class',
```
isagalaev: By the look of the code in the example they may be better served by the 'section' class name. 'Class' is for user class definitions, and they usually have some keywords and a title inside. 'Section' is for sections in structured config files, and here it seems close enough.

isagalaev:
As a general note, we're going to release a new version this Tuesday, and I think I'm going to merge this pull request even if we haven't worked out all the little thing by that time. It looks good enough for a start (definitely better than not having any definition :-) ) and we can continue working on it further on.

buildersbrewery:
I'll review this in two hours. The full caps words are constants, the ones with with leading `ll` and CamelCase are built_in functions.

isagalaev:
Then it'll probably make sense to use 'literal' for all uppercase constants.

src/languages/lsl.js
buildersbrewery: In LSL strings, only `\`, `"`, new line and tab characters are escaped.

src/languages/lsl.js
buildersbrewery: No special meaning. Was trying to make sure strings would override comments contained within.

src/languages/lsl.js
buildersbrewery: @isagalaev That leaves us with `\\`, `\"`, `\n` and `\t` as the only escape sequences for double-quoted strings in LSL. How to change the regex then?

buildersbrewery: Changed to `\\[tn"\\]` for now, that should do.

buildersbrewery: LSL strings (with `\\`, `\"`, `\n` and `\t` as  **only** escape sequences) can contain JSON and HTML code.

---

Illegals are generally not highlighted in LSL. They can however be categorized into `invalid.broken`, `invalid.deprecated`, `invalid.illegal`, `invalid.unimplemented` as well as (admin only) `reserved.godmode`, `reserved.log`.

isagalaev: Actually, it still makes sense to use the universal `hljs.BACKSLASH_ESCAPE` because it will properly highlight those four escapes which can legally be in an LSL string, while others simply won't be found in LSL code. It's a dirty secret of syntax highlighters: they don't need to actually check syntax, they can assume it's correct :-)

You regex is fine as well except that if you use a string instead of a literal regex syntax, you should double all the back-slashes, so either `/\\[tn"\\]/` or `'\\\\[tn"\\\\]'`.

buildersbrewery: So then I should change the other regexes to literal syntax as well?

isagalaev:
Merged finally. Thanks for your patience :-)

PR #1194, opened 2016-05-29
Add support for the Coq language

boyers:
[Coq](https://coq.inria.fr/) is a functional programming language and interactive theorem prover. Here is a screenshot of the syntax highlighting:

screenshot 2016-05-29 15 43 40

Thanks for reviewing!


src/languages/coq.js
```diff
@@ -0,0 +1,77 @@
...
+        'Scheme Scope Scopes Script Search SearchAbout SearchHead SearchPattern ' +
+        'SearchRewrite Section Separate Set Setoid Show Solve Sorted Step Strategies ' +
+        'Strategy Structure SubClass Table Tables Tactic Term Test Theorem Time ' +
+        'Timeout Transparent Type Typeclasses Types Undelimit Undo Unfocus Unfocused ' +
+        'Unfold Universe Universes Unset Unshelve using Variable Variables Variant ' +
+        'Verbose Visibility where with'
+    },
+    contains: [
+      hljs.QUOTE_STRING_MODE,
+      hljs.COMMENT('\\(\\*', '\\*\\)', { relevance: 2 }),
```
Sannis: It will be great if it will pass tests without relevance or with relevance=10.

boyers: Hmm, I'm having a bit of trouble making that happen without modifying other language definitions. It seems the "gams" language definition is too aggressive and matches any Coq code. If I bump this relevance to 10, it matches the Mathematica example (apparently Mathematica uses the same style of comments).

isagalaev:
Merged in a somewhat abridged form. Namely, I had to treat things like 'Inductive', 'Fixpoint', 'Theorem' as keywords, because they are definitely not 'meta'. We could change it later if someone comes up with a better name.

PR #1198, opened 2016-06-01
Protect against message-less exceptions

pkaminski:
In my app, I caught at least once instance where highlight.js crashed because `e.message` was undefined.  I can't be sure what kind of exception doesn't have a message (since it got pre-empted by the exception in trying to access `e.message`) but it might have been an out of memory error.  In any case, I figured it's better to propagate any such weird exception than to crash.

---
This change is [Reviewable](https://reviewable.io/reviews/isagalaev/highlight.js/1198)

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -488,7 +488,7 @@ https://highlightjs.org/
         top: top
       };
     } catch (e) {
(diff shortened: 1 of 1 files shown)


Sannis:
@pkaminski thanks!

PR #1203, opened 2016-06-03
TypeScript: Fix module.id being interpreted as module definition

eppsilon:
A reference to the built-in `module` object gets interpreted as the start of a `module ... { ... }` block.

Example of code that highlights incorrectly:

```ts
@Component({
  selector: 'my-example',
  directives: [SomeDirective],
  templateUrl: './my-example.component.html',
  moduleId: module.id,
  styles: [`
    .my-example {
      padding: 5px;
    }
  `]
})
export class MyExampleComponent {
  someProp: string = "blah";
}
```

### DIFF
--- a/src/languages/typescript.js
+++ b/src/languages/typescript.js
@@ -92,6 +92,11 @@ function(hljs) {
       {
         beginKeywords: 'constructor', end: /\{/, excludeEnd: true
       },
+      { // prevent references like module.id from being higlighted as module definitions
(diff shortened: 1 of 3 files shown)


Sannis:
@eppsilon can you please add markup test for this?

eppsilon:
@Sannis Sure, I've just added one.

Sannis:
@eppsilon thanks!

Issue #1193, opened 2016-05-29
C++14 binary literals and digit separators are not rendered properly


Morwenn:
C++14 added both binary literals and digit separators to the language, allowing to write expressions like the following one:

    auto var = 0b0110'1001'1001'0110;

In its current form, the expression above is rendered as follows:
* The first `0` is highlighted as an integer literal.
* The following `b0110` is highlighted as a variable name.
* The rest of the expression is highlighted as if it alterned between character literals and integer literals.

sourrust:
This was fixed with the merging in of pull request #1204, but I also added in the underscore digit separator in addition.

Morwenn:
Thanks! :)

Morwenn:
Hum, actually it still isn't displayed properly, see [this page][1] as an example with two problems:

       0b0110'1001'1001'0110
    //       ^ highlighted differently from the other ticks
       __builtin_parity(code.value)
    // ^^^ considered to be a digit literal

The second problem probably comes form the fact that `_` was added as a digit separator too in highlight.js. Digit separators are only allowed between symbols in C++ though. We might want to reopen the issue.


  [1]: http://morwenn.github.io/cpp-gray/gray-codes-and-parity/

Rapptz:
I could not get the first case to repro but I fixed the second one in #1304 since I wasn't sure why _ was put as a digit separator.

isagalaev:
Closing as it looks like #1304 has indeed fixed it.

PR #1204, opened 2016-06-05
Add digit separators and binary literals to C++

OiCMudkips:
Addresses #1193.

I'm not sure whether we want to have support for digit separators for all languages that use C_NUMBER_RE and BINARY_NUMBER_RE, considering that it's only a C++ feature as far as I can tell. I could change this to just apply digit separators for C++, but then I thought that it wouldn't be good to have too many definitions for C numbers lurking around... let me know what direction you'd like to take :)


Sannis:
@OiCMudkips thanks!

PR #1210, opened 2016-06-11
Add support for LDIF (LDAP Data Interchange Format).

braveulysses:
This adds support for LDIF, a very simple format used by LDAP directory servers to represent directory records and updates.


Sannis:
@braveulysses thanks for your contribution!

PR #1221, opened 2016-06-19
Fix TS support for array return annotations

bluepichu:
...and add a test.

This fixes instances like the following:

```ts
function getArray(): number[] {
  return [1, 2];
}
```

The array return-type annotation was preventing the function from highlighting properly, which in the node module prevented highlighting of this block altogether.  For some reason the `[` character was made illegal between the `function` keyword and the following `{` or `;` character within TypeScript.


Sannis:
@bluepichu thanks!

PR #1227, opened 2016-06-25
Update sql.js

aadu:
Add `comment` to beginKeywords for PostgreSQL: https://www.postgresql.org/docs/9.5/static/sql-comment.html


isagalaev:
Thanks for the contribution!

Issue #1140, opened 2016-04-06
Multi-stack parser


isagalaev:
## Problem

[Source](http://softwaremaniacs.org/blog/2014/07/26/highlight-js-what-next/en/):

> Our current syntax can't express things defined as a sequence of other things, like this:
> 
> ```
> function ::= <title> '=' <params> '->' <body>
> ```
> 
> The problem here is that you don't know that you're in a function definition until you get to the -> symbol. Our parser can start a new parsing mode based only on a single starting lexeme: an opening quote, a keyword, a number, etc.
> 
> To work around that we use a horrible kludge:
> 1. Match the whole body of the mode with a single regex.
> 2. Start a new mode at the beginning of the matched string.
> 3. Return the whole thing back to the parser.
> 4. Parse it again, by the rules of the new mode.
> 
> Not only it's ugly, it also works only when we're lucky that the whole body can actually be parsed by a regex.
## Solution design

We need to implement a "multi-stack" parser: instead of keeping track of a single stack of parsing modes the parser should be able to keep multiple stacks representing alternative parsing strategies to which it can go back if the current stack dead-ends for some reason. This effectively becomes a tree:

```
                                      stacks
                                      +----+
[root]---[ ]---[ ]---[ ]---[top]  <---|    |
          |           |               +----+
          |           +---[x]     <---|    | 
          |                           +----+
          +---[x]                 <---|    |
          |                           +----+
          +---[x]                 <---|    |
                                      +----+
```

This is an inverted tree with the `stacks` array keeping references to alternative parsing strategies (stacks) with each mode having a reference to its parent.

The topmost (or the rightmost) branch is the only long one and it represents the current alternative being parsed. Every other branch is exactly one level deep — a lead node representing a point from which to restart parsing if the current branch fails.

Operations:

**Discover**: upon finding lexemes (one or more) opening new possible modes add them to the beginning of the `stacks` array in the order they're defined. The old `top` is updated with the current index in the source so we could parse farther if all new strategies fail. `stacks[0]` becomes the new `top`.

**Cancel**: upon finding a canceling condition remove `top` from `stacks`, set `top` to the next available stack. If `stacks` is empty, treat as we currently do illegals — drop parsing.

**Commit**: if the mode successfully reaches its `end` lexeme, it is removed from the `top` stack and `stacks` is cleaned up from any items pointing to that mode
## Syntax

Current idea is to have something like this: 

``` js
var PARAMS = {
  className: 'params',
[...]

PR #1229, opened 2016-06-26
Fix "class" detection in protobuf syntax

connor4312:
Previously, if a protobuf file contained a field called `message` the highlighter would incorrectly detect that that started an actual message construct and abort the highlight. This fixes it so that it only classifies block statements when they are preceded exclusively on a line by whitespace and/or the end of another message block.

I also added some missing keywords/built-ins while I was there.

### DIFF
--- a/src/languages/protobuf.js
+++ b/src/languages/protobuf.js
@@ -8,9 +8,10 @@ Category: protocols
 function(hljs) {
   return {
     keywords: {
-      keyword: 'package import option optional required repeated group',
+      keyword: 'package import option optional required ' +
+        'repeated group oneof reserved',
       built_in: 'double float int32 int64 uint32 uint64 sint32 sint64 ' +
-        'fixed32 fixed64 sfixed32 sfixed64 bool string bytes',
+        'fixed32 fixed64 sfixed32 sfixed64 bool string bytes map',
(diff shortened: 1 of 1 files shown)


Sannis:
@connor4312 can you add example of incorrect highlighted code?

This snippet works well with master:
```protobuf
message Container {
  message Message {
    required int64 id = 1;
  }
  repeated Message messages = 1;
  optional int32 number = 2;
}
```
Result:
screen shot 2016-06-28 at 23 24 24

connor4312:
Sure, specifically [this message](https://github.com/WatchBeam/interactive-reference/blob/master/tetris.proto#L107-L111) was causing the highlighter to fail. It saw a `message` string, tried to look for a curly brace, but got a new line first and gave up.

isagalaev:
@Sannis JFYI: I'm getting ready to wrap the release, so if there's anything to fix here, I could hold up for a few hours :-)

Sannis:
@connor4312 thanks, but this should be fixed in more complex way to remain message keyword highlighted. I've raised issue for this.

isagalaev:
For what it's worth, I don't see a good way to fix this with the current parser either. It's the exact sort of parsing that we need #1140 implemented.

PR #1237, opened 2016-07-04
Specify custom elements in initHighlighting

lindell:
This will allow the use as follows:

````javascript
hljs.initHighlighting('div.code');
// or
hljs.initHighlightingOnLoad('div.code');
````

Instead of having to doing things like:
````javascript
$(document).ready(function() {
  $('pre code').each(function(i, block) {
    hljs.highlightBlock(block);
  });
});
````
Or with a lot more code when not using jQuery.

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -662,22 +662,27 @@ https://highlightjs.org/
 
   /*
   Applies highlighting to all <pre><code>..</code></pre> blocks on a page.
+  If elements parameter is specified, highlight those blocks instead.
   */
-  function initHighlighting() {
+  function initHighlighting(elements) {
     if (initHighlighting.called)
       return;
     initHighlighting.called = true;
 
-    var blocks = document.querySelectorAll('pre code');
+    var blocks;
+    if(typeof elements === 'string')
+      blocks = document.querySelectorAll(elements);
+    else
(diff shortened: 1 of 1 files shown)


lindell:
This might also be done via the configure like:

````javascript
hljs.configure({codeBlocks: 'div.code'});
````

isagalaev:
Thanks for the suggestion. It's been discussed before and we decide not to do it, partly because the custom initialization suggested in README is not "a lot" of code even without jQuery, as is evident in the current `initHighlighting()`.

lindell:
Too bad you don't think it's necessary. But I would still disagree. I can't see the

> not "a lot" of code

argument and still having `initHighlighting` and `initHighlightingOnLoad` since those thing could also easily be done by the user. Not having the setting only removes the option for users that don't want to use the `<pre><code>` tags to use a one line setup.

isagalaev:
Sorry, I still wanted to reply but my inbox keeps getting away from me :-)

`initHighlightingOnLoad()` is here to implement one single most common use-case. Going further is more or less arbitrary: do we allow only selectors in the arguments? Config options? A flag to re-highlight already highlighted code? There are just too many scenarios that one could express in a few lines of initializing code, so keeping things at the bare minimum is our (well, mine) choice of coping with that.

PR #1238, opened 2016-07-05
Add SubUnit support

ligurio:
SubUnit is a streaming protocol for test results.
Format specification: https://pypi.python.org/pypi/python-subunit


Sannis:
@ligurio thanks!

src/languages/subunit.js
```diff
@@ -0,0 +1,40 @@
...
+  var DETAILS = {
+    className: 'string',
+    begin: '\\[\n(multipart)?', end: '\\]\n'
+  };
+  var TIME = {
+    className: 'string',
+    begin: '\\d{4}-\\d{2}-\\d{2}(\\s+)\\d{2}:\\d{2}:\\d{2}\.\\d+Z'
+  };
+  var PROGRESSVALUE = {
+    className: 'string',
```
isagalaev: looks more like 'number' to me, what do you think?

PR #1239, opened 2016-07-06
Add Elm keyword: "subscription"

jvoigtlaender:
A new keyword that can appear (exactly) in module headers in Elm.

For example: https://github.com/elm-lang/navigation/blob/181ef742c755ccad3ce49b41b56f516a53a73628/src/Navigation.elm#L1.


isagalaev:
Thanks!

PR #1243, opened 2016-07-08
Adding new language PARI/GP

CRGreathouse:
I created a syntax highlighting file for [PARI/GP](http://pari.math.u-bordeaux.fr/). With the changes all tests pass, after a bit of tweaking with the relevance weights.

I have also included, as a separate commit, a style sheet I used in testing. It seems that none of the other styles will display a type or meta-keyword inside a string. You could also change the cascade on existing sheets but this seemed like a less invasive way to correct that.


CRGreathouse:
This code is ready to go as-is. But there is one thing that I can't quite figure out. I tried to highlight literal variables, which are a single apostrophe followed by a letter followed by zero or more letters, digits, or underscores. I defined it as
```
var GP_MONOMIAL_REGEX = /'[a-zA-Z][a-zA-Z0-9_]*\b/;
var GP_MONOMIAL = {className: 'literal', begin: GP_MONOMIAL_REGEX, relevance: 0};
```
but it doesn't highlight them in my test file. This isn't hugely important, but it would be nice to have a fix. I should mention that lexemes is `/\b[':.]?[a-zA-Z0-9_]+\b/` with this case in mind.

Actually, if I had limitless customizability I'd actually like to highlight `/'[a-zA-Z][a-zA-Z0-9_ ]*\b/` because whitespace is not significant in this language, but that's probably more trouble than it's worth.

CRGreathouse:
Let me know if there's anything I should add or edit.

For some code samples 'in the wild', see https://oeis.org/ which has about 50,000 snippets in this language.

isagalaev:
Catching up on email, sorry for the delay! 

Thank you for the contribution. It needs a bit more work, though:

- Balancing relevance is usually a means of last resort, it's not supposed to be used liberally. In fact, as I just found out, removing all non-zero relevances from keywords and modes doesn't break any detection tests and reliably detects the test snippet with a good margin. So please drop the relevances :-)

- In the test snippet the word `addhelp` gets only the "help" part highlighted as a keyword. I believe, it's not intentional?

- Define `GP_LINE_COMMENT` using the helper [`hljs.COMMENT`][1].

- `lexemes` are only relevant for keywords, and as I don't see anything there that starts with `[`:.]`, it's safe to drop the custom lexemes regexp altogether.

- `'something` literals work for me just fine with your definition. Can you show an example where it doesn't?

- Regarding "meta-keyword" and "type". In fact, all styles where they are defined will highlight them just fine inside "string", cascading won't affect it. However the real problem is that neither of those classes are designed to work within strings, so they're unlikely to look good. After looking at the code I think that the best way to name for both of them is "built_in", and it would make sense to capture them together with the quotes like this:
    ```js
    var GP_BUILT_IN = {
      className: 'built_in',
      begin: /"[te]_(INT|REAL| ... )"/
    }
    ```
    And make sure it's listed before strings.

- Strings definition can be safely replaced with the standard `hljs.QUOTED_STRING_MODE`.
[1]: http://highlightjs.readthedocs.io/en/latest/language-guide.html#comments

yyyc514:
@CRGreathouse  

**First of all, truly sorry for the long delay in anyone getting back to you regarding to this PR.**

Thank you so much for desiring to contribute and writing a language grammar.  Unfortunately, we currently are no longer accepting new languages into the core repository.  The existing maintainers simply don't have time to maintain new languages (fixing bugs, dealing with issues, etc).  **We'd still love to host your language as a 3rd party module though** - if you're willing to help maintain it.  

It's easy to turn your language into a module that anyone who likes can easily integrate with Highlight.js using a single line of code:

```html
<script src="/path/to/highlight.min.js"></script>
<!-- they'd simply add your language modules -->
<script src="/path/to/highlightjs-pari_gp.js"></script> 
```

It's also possible for users to rebuild the `highlight.js` library to include specific 3rd party modules, all in a single file (vs loading a bunch of 3rd party languages one at a time).

If you'd like to become the maintainer for this language then we're happy to create a new repository in the `highlightjs` organization and host the repository there.  That assumes you'd be open to reviewing issues and fixing bugs as they arise.  You're also welcome to host the repo yourself and we'll link to it from our README.

There isn't much in the way of how to docs on this yet, but there are sample repositories and it's honestly pretty easy.  We're still fleshing out an [official "spec" for 3rd party modules](https://github.com/highlightjs/highlight.js/issues/2328), but even after that's done it should require minimal changes from the examples we have already.  IE, if you wanted to get started now there isn't any big reason to wait.

My own `robots.txt` grammar repository is a great, simple example:

https://github.com/highlightjs/highlightjs-robots-txt

yyyc514:

As mentioned above if you're interested in becoming the maintainer of a 3rd party module, please respond [whenever you find time] and we'd be happy to help you get started.

PR #1244, opened 2016-07-13
Update lasso.js

EricFromCanada:
Shorten some regexes, allow dots in identifiers, update code sample


isagalaev:
Thank you for keeping it up to date!

PR #1251, opened 2016-07-17
Fix ABNF keyword highlighting.

mckibbenta:
Fix keywords not being highlighted when not preceded by a word boundary.


isagalaev:
Thanks! I've been thinking about doing this but it slipped out of my mind for some reason :-)

PR #1252, opened 2016-07-18
Add Atom One styles

danielgamage:
I added a couple of themes that I fancy.
https://github.com/atom/one-dark-syntax
https://github.com/atom/one-light-syntax

Also kept a hex color palette in the header of the src CSS files for future reference (the Atom themes use `hsl()` color functions so there was a little converting going on). I figure they get chopped off in the minified files anyways.

I also colorized `.hljs-subst` to better match Atom One's original styling. Let me know if this is problematic.

Otherwise I believe I covered all the bases for contributing... let me know if not.


isagalaev:
Everything is excellent! Thanks for your contribution!

PR #1254, opened 2016-07-19
Added C++ modules highlighting

alaingalvan:
Recently C++ modules were added to the language spec, with [Visual Studio adding support of it](https://channel9.msdn.com/Shows/C9-GoingNative/GoingNative-46-Why-you-should-be-Using-Cpp-Modules), here's a change that adds them to the project. ;)


Sannis:
Thanks @alaingalvan!

PR #1262, opened 2016-07-29
Fix for https://github.com/isagalaev/highlight.js/issues/1258

bulenkov:
1. Renamed darkula.css to darcula.css
2. Add @import of darcula.css in darkula.css and put deprecation warning into comments
3. Change foreground for strings to original Darcula foreground for strings


isagalaev:
Thanks! I wish all pull requests would be so clean and instantly mergeable :-)

PR #1273, opened 2016-08-18
tedjs Element

poryagrand:
hello
i added you library to tedjs to be accessable from my library.
https://tedjs.org

this a example source:
```html
<!doctype html>
<html>

<head>
    <title>asd</title>
    <link rel="stylesheet" title="Default" href="https://highlightjs.org/static/demo/styles/default.css">
    <script src="https://cdn.tedjs.org/tedjs/1.0.0/js/2ed.min.js"></script>
    <script>
        var hljs = include("system.highlight.core").hljs;
        hljs.configure({
            useBR: false
        });
    </script>
</head>

<body>
    <highlight lang="javascript">
        function test(){}
    </highlight>

    <highlight lang="c">
        int a = 2;
    </highlight>
    <highlight lang="x86asm">
         push    ebp       ; save calling function's stack frame (ebp)
         mov     ebp, esp  ; make a new stack frame on top of our caller's stack
         sub     esp, 4    ; allocate 4 bytes of stack space for this function's local variables
    </highlight>
</body>

</html>
```

all library are accessible from:
  https://tedjs.org/#page:libsearch

if you search highlight . it will be shown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,47 @@ detection.
 
 ## Getting Started
 
+### Use Tedjs Library
+
+you can use Highlight.js with tedjs library. to do this first add [`tedjs`](https://tedjs.org) library with your css style:
+```html
+<link rel="stylesheet" title="Default" href="https://highlightjs.org/static/demo/styles/default.css">
+<script src="https://cdn.tedjs.org/tedjs/1.0.0/js/2ed.min.js"></script>
+```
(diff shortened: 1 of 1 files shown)


isagalaev:
Thanks for using highlight.js! However we don't plan on mentioning every third-party code in our README, that'd be completely impractical.

PR #1302, opened 2016-09-18
IRC logs,lang

svbeon:
So, I'm not expecting this to be merged - at least not in it's current state, because errors.
1. the nick groups `begin: ' <', end: '> '` and `begin: ' -', end: '- '` should of course preferrably be merged with those that have `begin: '^<'` and `begin: '^-'` the reason they are splitted is to have spaces in the first two examples so dates won't get caught with the `-nick-` ones (notices) 
2. in some way  `begin: ' <', end: '> '` and `begin: ' -', end: '- '` shouldn't be caught if they don't appear after a `\n` or `]\s` (note the whitespace)

so this PR is more a request for help with the above two things


yyyc514:
IRC logs from what software?  Is there some sort of standard for these logs?

yyyc514:
Ping.

yyyc514:
Possibly related to #1509

svbeon:
Hey, sorry I started this and completely forgot about it... There isn't really a standard but most of them föllow a sort of convention. This was a while ago but I think it should be tailored to znc logs iirc. #1509 is probably a better point to work from

yyyc514:
> This was a while ago but I think it should be tailored to znc logs iirc. #1509 is probably a better point to work from

I will close this PR then.  We're no longer accepting languages into the core repository (lack of time to maintain and support them) but if you wanted to press forward we'd be happy for you to write a 3rd party language module that users could add to Highlight.js on their own if you're still interested.  Let me know and I can tell ya more.

PR #1304, opened 2016-09-19
Fix C++ digit separator parsing.

Rapptz:
The only digit separator allowed in C++ is the ' character, not the
_ character. This also means that identifiers such as __foo are not
incorrectly highlighted as a number.

Somewhat related to #1193


isagalaev:
Thanks a lot for taking care of it!

isagalaev:
And thanks even more for a comprehensive test!

PR #1305, opened 2016-09-19
Update cs.js - Added some contextual keywords

martinhodler:
# References
- https://msdn.microsoft.com/en-us/library/x53a06bb.aspx

# Notes
The only keyword which is not on this page is the "nameof"-Keyword.
But I never saw the keyword "when"

### DIFF
--- a/src/languages/cs.js
+++ b/src/languages/cs.js
@@ -8,16 +8,16 @@ function(hljs) {
   var KEYWORDS = {
     keyword:
       // Normal keywords.
-      'abstract as base bool break byte case catch char checked const continue decimal dynamic ' +
+      'abstract as base bool break byte case catch char checked const continue decimal ' +
       'default delegate do double else enum event explicit extern finally fixed float ' +
-      'for foreach goto if implicit in int interface internal is lock long when ' +
+      'for foreach goto if implicit in int interface internal is lock long ' +
       'object operator out override params private protected public readonly ref sbyte ' +
       'sealed short sizeof stackalloc static string struct switch this try typeof ' +
-      'uint ulong unchecked unsafe ushort using virtual volatile void while async ' +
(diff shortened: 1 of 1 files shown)


src/languages/cs.js
```diff
@@ -8,16 +8,16 @@ function(hljs) {
   var KEYWORDS = {
     keyword:
       // Normal keywords.
-      'abstract as base bool break byte case catch char checked const continue decimal dynamic ' +
```
martinhodler: Moved dynamic keyword to the contextual keywords section

src/languages/cs.js
```diff
@@ -8,16 +8,16 @@ function(hljs) {
...
+      'abstract as base bool break byte case catch char checked const continue decimal ' +
       'default delegate do double else enum event explicit extern finally fixed float ' +
-      'for foreach goto if implicit in int interface internal is lock long when ' +
```
martinhodler: Removed when keyword

src/languages/cs.js
```diff
@@ -8,16 +8,16 @@ function(hljs) {
...
       'object operator out override params private protected public readonly ref sbyte ' +
       'sealed short sizeof stackalloc static string struct switch this try typeof ' +
-      'uint ulong unchecked unsafe ushort using virtual volatile void while async ' +
+      'uint ulong unchecked unsafe ushort using virtual void volatile while ' +
       'nameof ' +
       // Contextual keywords.
-      'ascending descending from get group into join let orderby partial select set value var ' +
-      'where yield',
+      'add alias ascending async await by descending dynamic equals from get global group into join ' +
+      'let on orderby partial remove select set value var where yield',
```
martinhodler: Added some contextual keywords like by, global, on, etc.

isagalaev:
Thanks for updating the language!

PR #1306, opened 2016-09-20
fix(javascript): PARAMS_CONTAINS missing `var`

marcoscaceres:
I think this is breaking strict mode.


isagalaev:
Good catch, thanks!

PR #1307, opened 2016-09-25
Add the Flix programming language

magnus-madsen:
Add support for the Flix programming language.

The language website is here:
http://flix.github.io/
The GitHub page is here:
https://github.com/flix

Let me know if I did anything wrong.


isagalaev:
Thanks for the contribution! I'll have a closer look at the code, but one things I can say right away is that the relevance tuning in the last two commits is definitely not the way to fix cross-detection problems. Tuning relevance introduces unpredictable effects (the test suite can only catch so much), ans is basically the means of last resort. Could you revert the last two commits, even if that'd break the detection tests again. I'll have a look at how to fix them. Thanks!

isagalaev:
Merged! For the record, here's how I fixed the relevance issues: bb1cdfe05

PR #1316, opened 2016-10-09
Add vs15 themes

aarondandy:
I created these two themes based off of the VS15 preview 5 IDE for my website and wanted to share them with others as a part of this project. It may conflict with the existing "vs" theme, not really sure what to do about that. There are some slight differences between "vs15-light" and "vs" and there may be people that prefer it.


yyyc514:
Is this different from vs2015?

yyyc514:
Closing due to age and lack of response.

PR #1318, opened 2016-10-09
Update pattern

idleberg:


idleberg:
Thanks Ivan and sorry for flooding you, should've tested this locally!

isagalaev:
No problem at all :-) Let Travis work, it's a robot it doesn't care :-)

src/languages/nsis.js
```diff
@@ -8,26 +8,26 @@ Website: http://github.com/idleberg
...
-    begin: '\\$+{[a-zA-Z0-9_]+}'
+    begin: /\$+{[\w\.:-]+}/
   };
 
   var VARIABLES = {
     // $variables
     className: 'variable',
-    begin: '\\$+[a-zA-Z0-9_]+',
-    illegal: '\\(\\){}'
+    begin: /\$+[\w]+/,
```
isagalaev: You don't need square brackets for a single choice, so: `/\$+\w+/`

src/languages/nsis.js
```diff
@@ -8,26 +8,26 @@ Website: http://github.com/idleberg
...
+    illegal: /\(\){}/
   };
 
   var LANGUAGES = {
     // $(language_strings)
     className: 'variable',
-    begin: '\\$+\\([a-zA-Z0-9_]+\\)'
+    begin: /\$+{[[\w\.:-]+]+}/
```
isagalaev: One of the nested pair of `[..]+` is excessive and may even lead to exponential performance degradation if not optimized away by a regex engine. Please drop one!

isagalaev:
Merged! Thanks :-)

PR #1323, opened 2016-10-14
Add keywords for GLSL switch statement

tikhomirov:
`switch` statement is available in GLSL since version 1.30. Also fixes a problem with missing space after `while` keyword


Sannis:
@tikhomirov thanks!

PR #1326, opened 2016-10-16
Added umlauts for identifier recognition

manniL:
Because identifiers can also contain german "umlauts" (äöüß), the regex for identifiers should also match those.


isagalaev:
This is not true for all the languages that use these library regexes. If you had any particular language in mind, update only that one.

manniL:
Alright, will do so.

PR #1327, opened 2016-10-16
Improved Java highlighting

manniL:
Alright, so I made a few improvements for the Java highlighting.

1.  The "do" keyword was missing in the keyword list
2.  Because Java identifiers can contain any unicode letter, I've added them to the regex. Also, identifiers must not start with digits, which was also missing.

For more information how a Java identifier is specified (or to validate my statement),[ read page 23 of the specs](http://docs.oracle.com/javase/specs/jls/se8/jls8.pdf)


isagalaev:
Merged. Thanks!

PR #1330, opened 2016-10-22
Add valid unicode for swift classes and types

manniL:
Umlauts and several other characters are now also highlighted (correctly, [corresponding to this document](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/swift/grammar/identifier-character) to fix such highlighting issues:
fail4


isagalaev:
Thanks for the update!

PR #1337, opened 2016-10-31
Added basic LLVM IR language support

f0rki:
The LLVM compiler framework has a rich [intermediate representation](http://llvm.org/docs/LangRef.html), which is used for optimizations etc.


isagalaev:
Thank you for your contribution!

PR #1346, opened 2016-11-14
Change bad uses of substr to substring

logicplace:
String.prototype.substr(start, count)
String.prototype.substring(start, end)
end = count - start


isagalaev:
Oh, cool! Thanks for catching!

logicplace:
No problem, that was amazingly fast. lol

isagalaev:
A lucky coincidence + a straightforward patch :-) We have our fair share of horribly outdated issues we really really have to look at at some point…

PR #1347, opened 2016-11-14
Add callback to highlightBlock

logicplace:
I added this so I could do some postprocessing on the block, like adding interactivity (I added roll-over bracket matching).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -44,7 +44,7 @@ Post-processing of the highlighted markup. Currently consists of replacing inden
 Accepts a string with the highlighted markup.
 
 
-``highlightBlock(block)``
+``highlightBlock(block, callback)``
 -------------------------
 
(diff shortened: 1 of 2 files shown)


Sannis:
@logicplace can you please update API docs and function comment as well?
@isagalaev are you OK with this?

logicplace:
Added that, dunno how you'd best want it written but it's something.

isagalaev:
I'm not sure why it's needed. `highlightBlock()` is synchronous, if you want something called after it just do

```js
hljs.highlightBlock(block);
f(block);
```

instead of 

```js
hljs.highlightBlock(block, f);
```

No need to complicate things, unless I'm missing something.

logicplace:
oh my god it totally is. I can't believe I'm that used to everything being async. lol

PR #1348, opened 2016-11-14
Allow starts to be defined as 'self' much like one can do in contains

logicplace:
I used this for a weird linking system. Colon separated values as { end: /:/, endsWithParent: true, starts: 'self', ... }

It's not a necessary thing, but it's consistent with the feature for contains.


yyyc514:
I'm confused when you would use this exactly?

logicplace:
Since this was so long ago I hardly remember but going off of "colon separated values" I think I used it for like `1 : 1 : ...` but where 1 is a complex type (like could be a number or a string or whatever)

It's in [here](https://github.com/logicplace/Imperial/blob/gh-pages/scripts/languages/rpl.js), and while the grammar sorta works, it's slow in complex cases so I kind of have up.

yyyc514:
Seems like a good way to get stuck in an infinite loop.  How does it break out if after end it always starts self?

And typically you'd do this with a parent and then inside your contains you'd just have you matcher... so it would run over and over and over again as many times as needed until you left the mode... no need for this start hack.

logicplace:
It breaks out when the parent ends, as the example says. :p

yyyc514:
```
...
contains: [
  {
     begin: /\d\s:\s?/,
     endWithParent: true
  }
]
```

Doesn't that do exactly the same thing?

logicplace:
I don't know, I'm out of touch with the system. I'm pretty sure I only did it for reducing redundancy. I have no complaints if this is just closed.

yyyc514:
Ok, closing.  It would have needed extra documentation and I had some issues with the code also... and generally this type of feature we shouldn't just add "abstractly"... if there was a grammar included with it that required the new functionality that would make a much better case for why it's necessary/useful, etc...

PR #1360, opened 2016-11-22
Add negative Infinity

Vap0r1ze:


isagalaev:
A minus doesn't make it a separate keyword though, and I think it's best not to highlight the minus as a whole with `Infinity`. Same as we don't highlight `-true` as a whole (which is valid JavaScript too).

I'm closing this PR, but please don't be discouraged in the future!

PR #1361, opened 2016-11-23
Add more common shell commands

twang2218:
Add more commands based on the commands list <http://ss64.com/bash/>.

Signed-off-by: Tao Wang <twang2218@gmail.com>


isagalaev:
Thank you for this pull request, but we're not going to merge it, unfortunately. bash.js is included in the default build, and we are very reluctant to blow up their size. Also, the list of common commands is going to be pretty much arbitrary. I'd rather get rid even of the non-bash built-ins, but that ship has sailed, unfortunately.

PR #1373, opened 2016-12-04
SQF Update - Arma 3 v1.66

zgmrvn:
An update for Arma 3 SQF.

here is the command list :
https://community.bistudio.com/wiki/Category:Scripting_Commands_Arma_3

Here is exemples of what it looks like on my forum :
http://www.corp-arma.fr/topic/83/question-technique-%C3%A9v%C3%A9nements-et-d%C3%A9clencheurs/4
http://www.corp-arma.fr/topic/127/positionner-un-objet-al%C3%A9atoirement


isagalaev:
Thanks for the comprehensive update!

PR #1382, opened 2016-12-11
Add Hy support

profitware:
Pull request adds support for Hy language (aka lisp-flavoured Python): https://github.com/hylang/hy

/CC @isagalaev


Sannis:
Thanks @profitware.

PR #1386, opened 2016-12-14
Powershell highlighting - major overhaul.

g8tguy:
Расширенная подсветка.
Теперь подсветка на порядок лучше, чем в любой IDE!

As of now, the PowerShell has come a long way from being just a simple scripting language.
It currently has lots of features, like Classes, Enums, Strong Typing etc.

This commit is a major overhaul of the previous one.

Check out the before and after comparison:
[Previous](https://g8tguy.github.io/highlight.js/powershell/previous.html).
[Current](https://g8tguy.github.io/highlight.js/powershell/current.html).

**Rich highlighting enhancements were introduced to the following language structures:**

- Approved Verbs - [See.](https://msdn.microsoft.com/en-us/library/ms714428(v=vs.85).aspx)
  No more need to list a long sheet of built-in commands,
  If the cmdlet is valid it will be highlighted, and if not,
  well - it will be marked as such.
- Assemblies import.
- Types.
- Static classes reference.
- Methods invocation.
- Enums.
- Classes.
- Constructors.
- Methods.
- "$this" service keyword.
- Escaped $ variables.
- Objects creation.
- Comparison operators. Also,lots of new ones were added!
- Functions with arguments.
- Named function parameters.
- Function parameters validators.
- Hash tables.
- A bunch of small tweaks here and there.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -233,3 +233,4 @@ Contributors:
 - Kenton Hamaluik <kentonh@gmail.com>
 - Marvin Saignat <contact@zgmrvn.com>
 - Michael Rodler <contact@f0rki.at>
+- G8t Guy <g8tguy@g8tguy.com>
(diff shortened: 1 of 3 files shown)


oivindoh:
This is wonderful. Can we merge this PR?

g8tguy:
Hi, Are you going to merge this one?

isagalaev:
Eventually! We maintain the project as time allows, which is, frankly, not much.

yyyc514:
Seems time has moved on.  Is this still needed or have the changes made to powershelll since resolved things?

PR #1389, opened 2016-12-21
Add Kotlin Keywords 'lateinit' and 'init'

phauer:
Hi,
I just noticed, that the both Kotlin keywords `lateinit` and `init` are missing in highlight.js. 

You can find examples for these keywords in the official Kotlin reference:
- [lateinit](https://kotlinlang.org/docs/reference/properties.html#late-initialized-properties)
- [init](https://kotlinlang.org/docs/reference/classes.html#constructors)

Cheers, Philipp

### DIFF
--- a/src/languages/kotlin.js
+++ b/src/languages/kotlin.js
@@ -4,12 +4,12 @@
  */
 
 
-function (hljs) {
+function(hljs) {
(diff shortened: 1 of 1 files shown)


isagalaev:
Thanks for contributing!

PR #1393, opened 2016-12-29
New language: properties configuration files

RaffaeleSgarro:
I provided hightlight for Java property files

```
# example
foo.bar.baz=1
```

### DIFF
--- a/src/languages/properties.js
+++ b/src/languages/properties.js
@@ -1,0 +1,22 @@
+/*
+Language: Property file
+Contributors: Raffaele Sgarro <raffaeleSgarro@gmail.com>
+Category: config
+Description: Property files used in Java. A simple key value pair separated by '='
+*/
+function(hljs) {
+  return {
+    contains: [
+      hljs.HASH_COMMENT_MODE,
+      {
+        className: 'variable',
+        begin: /^[^=]+/,
+      },
+      {
+        className: 'literal',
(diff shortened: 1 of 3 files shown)


src/languages/properties.js
```diff
@@ -0,0 +1,22 @@
...
+Category: config
+Description: Property files used in Java. A simple key value pair separated by '='
+*/
+function(hljs) {
+  return {
+    contains: [
+      hljs.HASH_COMMENT_MODE,
+      {
+        className: 'variable',
+        begin: /^[^=]+/,
```
error418: Keys and values can also be separated by the character `:`.
See https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html

RaffaeleSgarro: Somehow the pull request broke the build and so won't be merged as-is. Can't find the failing test, though, so I'm not sure I want to keep on trying

error418: If you don't mind I would try to implement this in a separate pull request

paladox:
Can be abandoned as properties support was added in https://github.com/highlightjs/highlight.js/commit/d65651b8062fbc03f85fe4f73b17a02855f3b702

egor-rogov:
Close as already implemented.

PR #1395, opened 2016-12-30
Add Python 3.6 f-literals

### LINKED ISSUE #1299: Python f-strings
https://docs.python.org/3.6/reference/lexical_analysis.html#f-strings
https://www.python.org/dev/peps/pep-0498/

flying-sheep:
I also added them to the example.

You don’t seem to have unit tests for correct highlighting, am I correct?

fixes #1299

### DIFF
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -4,8 +4,22 @@ Category: common
 */
 
 function(hljs) {
+  var KEYWORDS = {
+    keyword:
+      'and elif is global as in if from raise for except finally print import pass return ' +
+      'exec else break not with class assert yield try while continue del or def lambda ' +
+      'async await nonlocal|10 None True False',
+    built_in:
+      'Ellipsis NotImplemented'
+  };
   var PROMPT = {
     className: 'meta',  begin: /^(>>>|\.\.\.) /
+  };
+  var SUBST = {
+    className: 'subst',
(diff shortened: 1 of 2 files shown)


Sannis:
Thanks @flying-sheep.

We have markup tests in test/markup/python folder, you can alter them in other pull-request if you want.

flying-sheep:
Sure, as soon as I'm back from my holidays. Why did you have to revert it?

isagalaev:
Hi! Sorry for dropping the conversation a bit… @flying-sheep: the thing that prompted me to revert your solution was the end regexp for `SUBST`, it looked way more complicated than it should be. Unfortunately I didn't find time to actually work with you on that, and today's a release day, so I just reworked your solution in a new commit. It looks similar to yours: 5ee7306. Thanks for taking it on!

flying-sheep:
cool!

my rationale for adding that regex was that it reliably highlights the formatting part at the end of format literals differently the from python code part.

that’s why it basically encodes the format syntax grammar.

isagalaev:
Yeah, that's a good point… Wouldn't it be better to define it as something following a `:` inside subst?

flying-sheep:
so the grammar is: `!r` or `!s` followed by `:…` (all optional)

the only problem with this is `f'{ (lambda: pass) }'` which is legal and contains a `:` that isn’t part of the formatting. and strings of course, but i guess they’re recognized anyway?

PR #1396, opened 2017-01-01
Selector option added to easily select blocks to highlight

timacdonald:
```
hljs.configure({
    selector: 'pre samp, pre code' // any css selector
});

hljs.initHighlightingOnLoad();
```

On my current site I am showing code, but I'm also showing the output and wanted like to utilise the `<samp>` html tag. I'm also not pulling in jQuery and wasn't real keen to re-do all the DOMContentReady stuff + highlightBlock() already in the package just to include another selector so I added a configure option.

[samp html tag on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/samp)

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -48,7 +48,8 @@ https://highlightjs.org/
     classPrefix: 'hljs-',
     tabReplace: null,
     useBR: false,
-    languages: undefined
+    languages: undefined,
+    selector: 'pre code'
   };
 
   // Object map that is used to escape some common HTML characters.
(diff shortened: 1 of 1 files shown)


timacdonald:
I can update the docs if this is of interest to the project.

timacdonald:
@isagalaev is this something you are interested in? I might close if your not keen on it 😀

isagalaev:
This looks like a slippery slope: once we support one bit of flexibility, it's hard to justify not supporting something else (like having a filtering callback in the parameters, for instance). The current contract is simple: there's automatic initialization that works for the majority of cases, and all the rest can be initialized by a short custom code.

timacdonald:
No worries - I totally understand that.

PR #1403, opened 2017-01-04
Added note on using turbolinks and highlight.js

timacdonald:
If you are using [TurboLinks](https://github.com/turbolinks/turbolinks) and highlight.js, you are required to re-call the highlight method, however because the state of the hljs object is maintained, you also need to set `hljs.initHighlighting.called = false` to ensure it is able to re-hightlight, otherwise it simply exists the initialization method, as it thinks it has already been initialized.

It is not enough to just call `hljs.initHighlighting();`;

Not sure if you really want this as a documentation thing, but I had some issues and found this worked well on [my site](https://timacdonald.me/) which I've implemented this on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,18 @@ $('div.code').each(function(i, block) {
 
 For other options refer to the documentation for [`configure`][4].
 
+### Additional initialization when utilising TurboLinks.
+
+If you are using highlight.js with the [TurboLinks](https://github.com/turbolinks/turbolinks) library you will need to ALSO include the following initilization:
+
+```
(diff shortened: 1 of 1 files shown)


timacdonald:
@isagalaev also just checking if you are interested in having this in the docs, otherwise I might close it also. Thanks.

isagalaev:
Hey! Sorry for neglecting this PR… No, we don't in fact want it as part of the doc due to being ju-u-ust a bit on the other side of "too specific". Totally subjective, call of course, but there it is.

Thanks for the effort in any case!

timacdonald:
No troubles, thanks for the package!

PR #1404, opened 2017-01-05
Fix qml var keyword

jf990:


isagalaev:
Thanks for the fix! Sorry it took so long to get to it…

PR #1416, opened 2017-01-13
Added some Rust built-ins

KasMA1990:
`assert_ne!` and `debug_assert_ne!` were added in Rust 1.12, and `Debug` has been there the whole time

### DIFF
--- a/src/languages/rust.js
+++ b/src/languages/rust.js
@@ -18,7 +18,7 @@ function(hljs) {
     'str char bool'
   var BUILTINS =
     // prelude
-    'Copy Send Sized Sync Drop Fn FnMut FnOnce drop Box ToOwned Clone ' +
+    'Copy Send Sized Sync Drop Fn FnMut FnOnce drop Box ToOwned Clone Debug ' +
     'PartialEq PartialOrd Eq Ord AsRef AsMut Into From Default Iterator ' +
     'Extend IntoIterator DoubleEndedIterator ExactSizeIterator Option ' +
     'Result SliceConcatExt String ToString Vec ' +
@@ -27,7 +27,7 @@ function(hljs) {
     'debug_assert! debug_assert_eq! env! panic! file! format! format_args! ' +
     'include_bin! include_str! line! local_data_key! module_path! ' +
     'option_env! print! println! select! stringify! try! unimplemented! ' +
-    'unreachable! vec! write! writeln! macro_rules!';
+    'unreachable! vec! write! writeln! macro_rules! assert_ne! debug_assert_ne!';
   return {
     aliases: ['rs'],
     keywords: {


KasMA1990:
I decided to do a more comprehensive PR :)

isagalaev:
Thanks! However it may be a good idea to first agree on whether a more comprehensive does make sense. We're notorious of having a rather restrained highlighting, so sometimes it happens that people spend a lot of effort on making something very complicated only to be faced with me saying "sorry, can't take it" :-)

Any hints on what you're up to?

KasMA1990:
Sure, and thanks for the heads up :) The `?` operator was made stable recently and the community consensus is to highlight it when possible since it's an important operator that can be hard to spot some times; additionally the `i128` and `u128` built-ins are going stable soon, and I think I can make more accurate relevance settings for a few of the existing rules (e.g. the `#[...]` annotations are fairly unique to Rust AFAIK, but carry the default relevance).

isagalaev:
Right… Of those, `i128` and `u128` is a pretty straightforward addition, so go ahead with those.

As for `?`, one of the design choices in highlight.js is not to highlight operators (except in very esoteric languages), and we don't actually have an appropriate class name for those.

The relevance is better to be kept at its basic value. The idea of boosting unique constructs that I came up with very early on proved to cause much more problems over time than it solves, so currently the rule is to employ relevance tuning only if everything else fails. I would say Rust is in a pretty distinguishable and does well in regards to detection as currently defined.

KasMA1990:
Good to know. :)

For the relevance, the whole reason I came here in the first place is because I had a piece of code that failed the auto detect on my website:

    #[derive(Debug)]
    struct MyData;

    let first_binding = MyData;
    let second_binding = first_binding;
    println!("{:?}", first_binding); //ERROR! first_binding was invalidated above

It's funny that if I run the detection test for Rust with this, it works fine, but the setup I have on my website gives up. I'm running hljs 9.9.0, and I have configured it to only check for Java and Rust (and it detects the Java code fine). Given that the test works, my idea for tuning might not be relevant, but I could use some help on figuring out how to fix this issue then.

I'll add the missing keywords and `i128`/`u128` for now :)

isagalaev:
I'll have a look into this mis-detection… It's usually a matter of adding more specificity to modes or zeroing an overzealous relevance of another language.

KasMA1990:
I made the new [PR](https://github.com/isagalaev/highlight.js/pull/1420)

Thanks for the help :)

isagalaev:
About that mis-detection… The usual way to look at these things is the little tool located in tools/developer.html: you build a browser package with whatever languages you want and see the detection scores and the generated markup.

Your snippet is pretty reliably detected as Rust even using the build with all the languages, so I would guess there's something happening in the way your site displays code in HTML or the way highlighting is applied to the page. A couple of guesses from the top of my head without seeing a live example would be either some stray character triggering one of the rust.js `illegal` expressions or simply not having rust.js hooked up in the build you're using (the CDN build doesn't include rust.js but does include java.js, for example).

KasMA1990:
Doh! I'm using the CDN build 😕

isagalaev:
Then you should add rust.js separately, [as per README](https://github.com/isagalaev/highlight.js#getting-the-library).

KasMA1990:
Thanks :)

PR #1423, opened 2017-01-16
Adds WebAssembly Linerar Bytecode textual representation

rsms:
https://github.com/WebAssembly/design/blob/master/TextFormat.md#linear-bytecode

Example (part of test/detect/wasm/default.txt):

```wasm
get_local 0    // push parameter #0 on stack.
i64.const 0    // push contant int64 "0" on stack.
i64.eq         // execute "eq" which pops two operands from stack
               //  and pushes int32 "1" or "0" on stack.
if i64         // pops one int32 from stack; if its not "0":
  i64.const 1  //   push contant int64 "0" on stack.
else           // else (if operand was "0"):
  get_local 0  //   push parameter #0 on stack. $1
  get_local 0  //   push parameter #0 on stack.
  i64.const 1  //   push contant int64 "0" on stack.
  i64.sub      //   execute "sub[tract]" which pops two operands
               //    from stack (parameter #0 and constant int64 "1")
               //    and finally pushes the result int64 on stack.
  call 0       //   call function #0 ("factorial") which pops one
               //    int64 from the stack and when it returns an
               //    int64 has been pushed on stack
  i64.mul      //   execute "sub[tract]" which pops two operands
               //    from stack ($1 and result from function call)
               //    and finally pushes the resulting int64 on stack
end            // ends function, returning one int64 result (on stack.)
               // Stack now contains one int64 value that's the result from one of
               // the two branches above.
```

screen shot 2017-01-15 at 19 52 08


rsms:
Ping

yyyc514:
If we were to consider this for inclusion does it need any updates for 2019?

yyyc514:
@rsms Ping.

yyyc514:
Closing due to lack of response.

PR #1453, opened 2017-02-17
Do not add "undefined" to the beginning of the result of .fixMarkup()

### LINKED ISSUE #1452: fixMarkup add "undefined" when "useBR" is enabled
The current version (9.9.0)  adds 'undefined' to the beginning of the output, if 'usBR' is enabled.

```js
var highlight = require('highlight.js')

highlight.configure({
  'useBR': true
})

var highlighted = highlight.highlightAuto(`{ "some": "json" }`).value
console.log("hightlighted",highlighted)

var fixMarkup = highlight.fixMarkup(highlighted);
console.log("fixmarkup", fixMarkup)
```

Output:

```
hightlighted { <span class="hljs-attr">"some"</span>: <span class="hljs-string">"json"</span> }
fixmarkup undefined{ <span class="hljs-attr">"some"</span>: <span class="hljs-string">"json"</span> }
```

I am working on a PR (unless you tell me not to).

nknapp:
Fixes #1452

- Added default return value (empty string) to the ".replace()" call.

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -593,6 +593,7 @@ https://highlightjs.org/
           } else if (options.tabReplace) {
             return p1.replace(/\t/g, options.tabReplace);
           }
+          return '';
       });
   }
 
--- a/test/api/fixmarkup.js
+++ b/test/api/fixmarkup.js
@@ -1,0 +1,21 @@
+'use strict';
+
+let should = require('should');
+let hljs   = require('../../build');
+
+describe('.fixmarkup()', function() {
+  after(function() {
+    hljs.configure({ useBR: false })
+  })
+
+  it('should not add "undefined" to the beginning of the result (#1452)', function() {
+    hljs.configure({ useBR: true })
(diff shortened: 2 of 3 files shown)


nknapp:
I tried to keep to the existing code style, and I added a test for the fix.

src/highlight.js
```diff
@@ -593,6 +593,7 @@ https://highlightjs.org/
           } else if (options.tabReplace) {
             return p1.replace(/\t/g, options.tabReplace);
           }
+          return "";
```
isagalaev: A tiny nit: could you use single quotes for strings as in the rest of the code? :-)

nknapp:
Oops... That's the thing with german keyboards. Typing `"` is almost easier than typing `'`. I always get it confused.

I've changed that and made a force-push to my branch to no pollute history...

isagalaev:
Thanks, merged! (Don't really care about history "pollution" myself :-) )

PR #1454, opened 2017-02-17
Update Rust highlighting.

eddyb:
* `NUM_SUFFIX` doesn't allow `f(8|16|128|size)` anymore
* removed `int`, `uint` and `float` (they haven't been in Rust for a long time)
* added `u16` (it was just missing)
* grouped together all builtin types, moved all types from keywords
  * e.g. `i32`, `str`, `char` etc. can be used as variable names, definitely not keywords
* split "prelude" into "functions", "types" and "traits"
* included `union` in the same `keyword name ... {` rule used for `trait`, `enum` and `struct`

### DIFF
--- a/src/languages/rust.js
+++ b/src/languages/rust.js
@@ -6,22 +6,26 @@ Category: system
 */
 
 function(hljs) {
-  var NUM_SUFFIX = '([uif](8|16|32|64|128|size))\?';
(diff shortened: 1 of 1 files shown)


isagalaev:
Thanks for the excellent update!

PR #1455, opened 2017-02-18
WIP: Configured simple eslint-rules and run into TravisCI

nknapp:
I tried to configure eslint-rules that match the code-style of the project and would have detected double quotes. 
I've set all failing rules to "warn" for the moment.
I have also excluded the  files in "src/languages", because appearantly, they are not valid javascript on its own (functions without name are only allowed as expression)
I have configured Travis to run eslint and when errors are detected.

This PR is to see, if the TravisCI configuration runs and to show you which rules are currently violated.
I would like to know, if you appreciate a linter at all and if I should fix the violations.


nknapp:
There are the warnings: https://travis-ci.org/isagalaev/highlight.js/jobs/202870095

marcoscaceres:
Closing as we will now probably use Github actions.

PR #1459, opened 2017-02-23
Fix highlight of quasiquote of Scheme

Kraks:
The quasiquote should be treated as same as quote.

For example, in current syntax schema, the code ```(add 1 2) and '(add 1 2) are highlighted differently. (Note there is only one quasiquote)
The `add` in quasiquote example is in bold style but should not.
This patch fixes this.


isagalaev:
Merged. Thanks for your contribution!

PR #1461, opened 2017-02-24
add highlight style css.

takemori-kondo:
this is style like visual studio 2015 default for c-sharp.
---
Dear team
Thank you for a cool project.
I created css of the same color setting as the default setting (dark) of Visual Studio 2015.
If you do not mind, please use it.

### DIFF
--- a/src/styles/vs-dark.css
+++ b/src/styles/vs-dark.css
@@ -1,0 +1,55 @@
+/*
+
+Visual Studio-like style based on original C# coloring by Mamoru Kondo <etherealgarden@msn.com>
+
+*/
+.hljs {
+  display: block;
+  overflow-x: auto;
+  padding: 0.5em;
+  background: #1e1e1e;
+  color: #dcdcdc;
+}
+
+.hljs-keyword,
+.hljs-literal,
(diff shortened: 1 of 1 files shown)


mariuszjamro:
I think you're missing color of the `.hljs-title` element. That's the green-one used for example for type names.

    .hljs-title {
        color: #45cab5;
    }

yyyc514:
Is there not already vs2015?  How is this different?

takemori-kondo:
> Yyyc 514
Thank you, thanks, I was able to notice.

> Mariuszjamro
Oh, sorry, I completely forgotten.

About .hljs-title seems to be highlighted elsewhere.
(Although it is originally a class name,  namespace & functions seem to be highlighted too.)

About .hljs-number, it is as pointed out. Thank you very much.

I'll apply only .hljs-number for now.
Thank you.

byvs2017

byhighlight js

takemori-kondo:
I was misunderstanding.
Nicolas 01 was already creating vs 2015 dark style at https://github.com/highlightjs/highlight.js/pull/1517.
This pull requiest can be closed.

PR #1466, opened 2017-03-03
catch single line comments in SQL language

solugebefola:
This is to catch single line comments in the SQL language when the code text is a multiline string

### DIFF
--- a/src/languages/sql.js
+++ b/src/languages/sql.js
@@ -5,7 +5,7 @@
  */
 
 function(hljs) {
-  var COMMENT_MODE = hljs.COMMENT('--', '$');
+  var COMMENT_MODE = hljs.COMMENT('--', '\n|$');
   return {
     case_insensitive: true,
     illegal: /[<>{}*#]/,


isagalaev:
Do you have a test that shows it not working before your fix and working after? I'm asking because `$` already means the end of line, and it works throughout pretty much all the languages in highlight.js.

solugebefola:
I realized that the issue was with the use of a hashmark in my text prompts which is picked up by the sql language as illegal.  I am withdrawing the PR.  Maybe there needs to be a 'psql' language to allow putting the appropriate prompts in the code?

Anyway, sorry to take up your time with this.  Thanks for responding so fast.

isagalaev:
There can be two different approaches to handling repl prompts. One is adding a mode that is then inserted everywhere, like in Python: https://github.com/isagalaev/highlight.js/blob/master/src/languages/python.js#L7-L9. Another is a helper language that embeds the parent language, like in Clojure: https://github.com/isagalaev/highlight.js/blob/master/src/languages/clojure-repl.js

PR #1470, opened 2017-03-09
Haxe: remove the "callback" keyword

Gama11:
`callback` used to be a keyword in Haxe 2, which is fairly old at this point (last 2.x release was 2012, see http://haxe.org/download/list/). See http://old.haxe.org/manual/haxe3/migration#callback.

Since `callback` is a fairly common identifier, it's annoying to see it highlighted as a keyword.


isagalaev:
Merged. Thanks for the contribution!

PR #1475, opened 2017-03-17
Escape quotes

13rentgen:
Hi,
The quotes escaping is necessary for the correct display of additional markup inside codeblocks, especially for the JSON data in data attributes.


isagalaev:
Could you demonstrate what doesn't work with the current code?

13rentgen:
Sure,
https://jsfiddle.net/13rentgen/52vjg2km/1/ - "data-attr" is broken because of highlight.js

isagalaev:
Thanks for the report and for the test case! However, I ultimately went ahead with a more local change.

PR #1476, opened 2017-03-18
Add a callback function to highlightBlock()

binford2k:
Sometimes when I'm manually highlighting blocks of text, I need to
manipulate the block after highlighting is complete. For example, if I
needed to apply alternating line background colors, I could do that in a
callback function very easily.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -44,7 +44,7 @@ Post-processing of the highlighted markup. Currently consists of replacing inden
 Accepts a string with the highlighted markup.
 
 
-``highlightBlock(block)``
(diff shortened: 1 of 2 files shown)


isagalaev:
`highlightBlock()` is synchronous, you can simply call your function after `highlightBlock()`.

On a related note, highlight.js preserves the original user markup, so you can add those background colors even before you call `highlightBlock()`

binford2k:
Huh. It failed when I tried that, but now seems to work ok. Must have been user error :)

isagalaev:
Happens all the time, no worries :-)

PR #1477, opened 2017-03-21
[Lua language def] Built in keywords are now listed properly

sicher:
Removed commas and added spaces at the end of each concatenated string so these keywords are recognised as built-ins.


isagalaev:
Well caught, thanks!

PR #1483, opened 2017-03-31
Add or, and, not operators to C++

jordi-petit:


isagalaev:
Merged, thanks! Also: 98aae60

PR #1487, opened 2017-04-05
Added support for jboss-cli

rparree:


rparree:
will add relevance: 0

rparree:
I am getting a unresponsive page when i try to use within the browser.

Works fine when i access `build/demo/index.html` when building with only `node tools/build.js -n jboss-cli`. When building with `node tools/build.js` my browser hangs.

rparree:
In fact it works fine in the browser as long as i use the `--no-compress` option. After compressing the page becomes unresponsive

src/languages/jboss-cli.js
```diff
@@ -0,0 +1,67 @@
...
+ Category: config
+ */
+
+function (hljs) {
+  var PARAMS = [{
+    className: "attribute",
+    begin: /[\w-]+\s?=/,
+    starts: {
+
+      ends: /$/,
```
isagalaev: Should be `end` I guess?

src/languages/jboss-cli.js
```diff
@@ -0,0 +1,67 @@
...
+      className: "string"
+    },
+    relevance: 1
```
isagalaev: 1 is the default value, so shouldn't be set explicitly (JavaScript being a weird language where we care about code size). Same for all the other occurrences of it down the file.

src/languages/jboss-cli.js
```diff
@@ -0,0 +1,67 @@
...
+  var COMMAND_PARAMS = {
+    className: 'params',
+    begin: "--",
+    end: "$",
+    relevance: 1
+
+  };
+  return {
+    aliases: ['wildfly-cli'],
+    case_insensitive: false,
```
isagalaev: This one is also unnecessary, since `false` is the default.

rparree:
fixed the problems, but now i have a problem that the `PARAMSBLOCK` (`begin: /\(/,end: /\)/`) which contains `PARAMS` is very eager and consumes all content until the last closing parenthesis. I will look into this today (working off another branch)

isagalaev:
Merged with a few addition tweaks. Thanks!

PR #1489, opened 2017-04-06
makefile: improve support

joel-porquet:
Improve support for makefile syntax:
- simple variables (eg '$(var)') and special variables (eg '$@')
- variable inside strings (eg "blabla $(var) blabla")
- functions (eg '$(patsubst arg,arg,$(var))')
- variable assignement (eg 'var=toto', 'var+=toto', 'var:=toto')
- targets (eg 'all: dependency $(var)')
- language keywords (define, endef, ifeq, etc.)

Signed-off-by: Joel Porquet <joel@porquet.org>

### DIFF
--- a/src/languages/makefile.js
+++ b/src/languages/makefile.js
@@ -1,50 +1,87 @@
 /*
 Language: Makefile
 Author: Ivan Sagalaev <maniac@softwaremaniacs.org>
+Contributors: Joël Porquet <joel@porquet.org>
 Category: common
 */
 
 function(hljs) {
+  /* Variables: simple (eg $(var)) and special (eg $@) */
   var VARIABLE = {
     className: 'variable',
-    begin: /\$\(/, end: /\)/,
-    contains: [hljs.BACKSLASH_ESCAPE]
+    variants: [
+      {
+        begin: '\\$\\(' + hljs.UNDERSCORE_IDENT_RE + '\\)',
+        contains: [hljs.BACKSLASH_ESCAPE],
+      },
+      {
+        begin: /\$[@%<?\^\+\*]/
+      },
+    ]
+  };
+  /* Quoted string with variables inside */
+  var QUOTE_STRING = {
+    className: 'string',
+    begin: /"/, end: /"/,
+    contains: [
+      hljs.BACKSLASH_ESCAPE,
+      VARIABLE,
+    ]
+  };
+  /* Function: $(func arg,...) */
+  var FUNC = {
+    className: 'variable',
+    begin: /\$\([\w-]+\s/, end: /\)/,
+    keywords: {
+      built_in:
+        'subst patsubst strip findstring filter filter-out sort ' +
+        'word wordlist firstword lastword dir notdir suffix basename ' +
+        'addsuffix addprefix join wildcard realpath abspath error warning ' +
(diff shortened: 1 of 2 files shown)


src/languages/makefile.js
```diff
@@ -1,50 +1,89 @@
...
+    returnBegin: true,
+    contains: [
       {
-        begin: /^\t+/, end: /$/,
-        relevance: 0,
-        contains: [
-          hljs.QUOTE_STRING_MODE,
-          VARIABLE
-        ]
+        className: 'variable',
```
isagalaev: We usually highlight things as "variable" when they are a part of a string or, more generally, appear in a context where they should be distinguished from their surroundings. Here it feels to me like highlighting the left sign of the assignment might be unnecessary. I'll let you make the final call, but I'm -0 on having the class name. (The mode itself should stay even without a class name since to improve detection.)

src/languages/makefile.js
```diff
@@ -1,50 +1,89 @@
...
+        begin: '^' + hljs.IDENT_RE, end: '(:=|=|\\+=)',
```
isagalaev: `end` also needs a variant for `?=`. Our default example (test/detect/makefile/default.txt) has it, and it's not recognized with this change.

src/languages/makefile.js
```diff
@@ -1,50 +1,89 @@
...
+    begin: /^[^\s]+:/, end: /$/,
+    contains: [VARIABLE,]
+  };
+  return {
+    aliases: ['mk', 'mak'],
+    keywords: {
+      keyword:
+      'define endef undefine ifdef ifndef ifeq ifneq else endif ' +
+      'include -include sinclude override export unexport private vpath',
+    },
```
isagalaev: When there's only one set of keywords, it can be simply `keywords: '....'` instead of `keywords: {keyword: '...'}`.

isagalaev: And because there's `-inlcude` it won't be picked up by the default lexeme regexp, so it needs `lexemes: /[\w-]+/` alongside `keywords`.

src/languages/makefile.js
isagalaev: Oh, and the similar regexp above in `VAR_ASSIG`'s `begin` also needs to be changed accordingly. (Something like `\\s*[:+?]?=`)

src/languages/makefile.js
joel-porquet: OK, sounds good.

FYI, I sort of followed what vim was highlighting, and vim highlights the left part of assignments. That's why.

src/languages/makefile.js
joel-porquet: Forgot about this one, thanks!

src/languages/makefile.js
joel-porquet: fixed

joel-porquet: I was working on an example which included a "-include" but I didn't notice that the '-' wasn't recognized. Fixed now.

joel-porquet:
Hi Ivan,
I just updated my pull request based on your review. Let me know if there is anything else that you would like to change.
Thanks,
Joël

isagalaev:
Looks perfect! Merged for the upcoming release.

PR #1491, opened 2017-04-10
Correctly highlight negative floating point values in C++

NuclearCookie:
Negative floating point numbers in C++ did not highlight correctly.
From: 
```html
<span class="hljs-keyword">float</span> negative = <span class="hljs-number">-123.0</span>f; <span class="hljs-comment">// negative floating point numbers</span>
```
To:
```html
<span class="hljs-keyword">float</span> negative = <span class="hljs-number">-123.0f</span>; <span class="hljs-comment">// negative floating point numbers</span>
```

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -1,7 +1,7 @@
 /*
 Language: C++
 Author: Ivan Sagalaev <maniac@softwaremaniacs.org>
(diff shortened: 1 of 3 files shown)


isagalaev:
Looks reasonable, thanks! It's a shame really how bugs such as this crop up after 10+ years in the wild…

PR #1496, opened 2017-04-18
Add missing dockerfile instructions

dnephin:
Add all the new Dockerfile instructions to the dockerfile language definition. 

Updated the example to include the new instructions, and flags on instructions.


isagalaev:
Thanks a lot for the update!

PR #1498, opened 2017-04-21
Update Julia definitions for version 0.6

ararslan:
Previously the Julia definitions were based on the 0.3 and 0.4 series of releases. While 0.5 brought many changes, 0.6 has brought sweeping changes that require more significant updates. Definitions only relevant to older versions of Julia have been removed.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -246,3 +246,4 @@ Contributors:
 - Jordi Petit <jordi.petit@gmail.com>
 - Raphaël Parrëe <rparree@edc4it.com>
 - Joël Porquet <joel@porquet.org>
+- Alex Arslan <ararslan@comcast.net>
--- a/AUTHORS.ru.txt
+++ b/AUTHORS.ru.txt
@@ -246,3 +246,4 @@ URL:   https://highlightjs.org/
 - Джорди Петит <jordi.petit@gmail.com>
 - Рафаэль Паре <rparree@edc4it.com>
 - Джоель Порке <joel@porquet.org>
+- Алекс Арслан <ararslan@comcast.net>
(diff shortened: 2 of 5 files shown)


ararslan:
@isagalaev I don't mean to rush, but the official Julia documentation, as well as the documentation for many Julia packages, uses this library as part of [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl). These definitions haven't been updated in quite a while, so the documentation looks kind of weird (e.g. new syntax is not highlighted, old syntax is, etc.). If possible, it would be great if this could be merged soon and a new version of highlight.js tagged so that Documenter.jl can be updated and the Julia documentation rebuilt before the upcoming release.

isagalaev:
Hey there! Sorry for the wait… I merged the branch with one additional change that we may want to discuss further: f0a88f2. The `::` and `<:` were highlighted all by themselves, which is not what we traditionally do in highlight.js, and they definitely weren't what "type" is supposed to be. And since most actual types in our default examples are highlighted as built-ins anyway I decided for now to simply remove those definitions (leaving `<:` as a nameless relevance booster). Do you think it's sensible enough from the language point of view?

ararslan:
Yep that makes total sense to me. Thank you, much appreciated!

ararslan:
P.S. I don't suppose you'd be willing to tag a new version so we can update Documenter.jl? 🙂

isagalaev:
Hello :-) Nice weather today, for a change! For now, at least…

We're doing releases every 6 weeks, and the last one happened just last week, so the next is scheduled on May, 30th. But you can just use the master branch, in a pinch.

ararslan:
Sounds good. Thanks again! Let's enjoy the favorable weather while it lasts!

ararslan:
I can't use the CDN for master though, right?

isagalaev:
Nope… Gotta check out the source and build it with `node tools/build.js julia` (and whatever other languages you need.)

ararslan:
Okay, thanks for the info.

Issue #1502, opened 2017-04-23
(cpp) template type in function declaration


arximboldi:
Hi!

In general, highlight.js is highlighting C++ function declarations with a special syntax so the function names and arguments do stand out. However if there is a template anywhere in there, it fails. Example:
```
boost::optional&lt;application> handle_key(application state, key_code key, coord size);
```
Is highlighted all in the same basic color. A simpler declaration like this otherwise works:
```
int main(int argc, char* argv[])
```

yyyc514:
Shouldn't this really be a little tighter:

```
var IDENT_RE = '[a-zA-Z][<>\\w]*';
```

Ie detect `identifier<something>`.  Right now it's pretty open ended.  You could probably even reuse the existing ident:

```
var IDENT_RE = hljs.IDENT_RE + '(<...>)?;
```

yyyc514:
If this was tightened up a little and you added a `markup` test (or updated one if there is one that makes sense) and submitted a PR we'd be happy to look at it more closely.

yyyc514:
Might first confirm we haven't fixed this in the meanwhile already? :)

PR #1503, opened 2017-04-24
C++ highlighting improvements

arximboldi:
I hope you like this!

### Fix templates in function signatures

  #1502

This makes identifiers consume subsequent `<` or `>` symbols that are
there without spaces.  Because of how Highlight.js treats other
syntactic elements, I believe this is no problem, but may require
further testing.

### Mark using expressions as classes

This shows in "class/type" color the type alias in expressions of the
form:

```
    using my_type_alias = ...;
```

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -6,6 +6,9 @@ Category: common, system
 */
 
 function(hljs) {
+
+  var IDENT_RE = '[a-zA-Z][<>\\w]*';
+
   var CPP_PRIMITIVE_TYPES = {
     className: 'keyword',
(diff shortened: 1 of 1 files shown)


egor-rogov:
Hi @arximboldi,
any chances to revive this work? I think this PR needs some markup tests to cover improvements you've made.

arximboldi:
Hmmm... I am very busy this month let's see in a few weeks if I can take a look. Or maybe @ryb-ableton wants to take a look since I vaguely recall he used this ;)

marcoscaceres:
Closing due to inactivity. We've done a bunch of updates to cpp since this, so hopefully it matches.

PR #1510, opened 2017-05-03
Highlight NGINX config files which have .conf extension

bocharsky-bw:
As I see in the [reference](https://github.com/isagalaev/highlight.js/blob/master/docs/css-classes-reference.rst) `.conf` extension isn't reserved yet. I think we can add this one, but I wonder why it hasn't added yet? Will it cause any conflicts? I'm not sure whether Apache (or others) also uses `.conf` extension, but it would be nice to highlight `.conf` files somehow.

### DIFF
--- a/src/languages/nginx.js
+++ b/src/languages/nginx.js
@@ -68,7 +68,7 @@ function(hljs) {
   };
 
   return {
-    aliases: ['nginxconf'],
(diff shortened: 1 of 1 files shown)


yyyc514:
From what I know about Linux `conf` is not a filetype or alias unique to Nginx, therefore it's not a useful alias for us here.

PR #1519, opened 2017-05-13
Improve c#

Nicolas01:
Some improvement to the C# parser:

- add classes, interfaces and structures from the System assembly as keywords
- highlight type in new declaration
- highlight type in [Attributes("")]
- highlights type in typeof()
- highlights types of parameters in methods declarations
- highlights the first type in using block
- highlights the first type in foreach block
- highlights the first type in assignation
- comment the function mode, because 'else if()' is considered as a function
- add more elements to the example


Nicolas01:
Could someone explain me which check fails and what does the following log means:
```
hljs.highlightAuto() should be detected as flix:
      AssertionError: expected 'cs' to be 'flix'      
      actual expected
      csflix
```

src/languages/cs.js
```diff
@@ -1,25 +1,37 @@
...
+  }
```
isagalaev: There's one big unsolvable problem with this: as soon as a language gets into the "common" set (C# is one of them) we really don't want to blow up its size because "common" is what comprises our CDN build. So we simply can't have an exhaustive list of builtins from a big standard library.

src/languages/cs.js
```diff
@@ -1,25 +1,37 @@
...
+
```
isagalaev: There's one big unsolvable problem with this: as soon as a language gets into the "common" set (C# is one of them) we really don't want to blow up its size because "common" is what comprises our CDN build. So we simply can't have an exhaustive list of builtins from a big standard library.

src/languages/cs.js
```diff
@@ -1,25 +1,37 @@
...
-      'nameof ' +
       // Contextual keywords.
       'add alias ascending async await by descending dynamic equals from get global group into join ' +
       'let on orderby partial remove select set value var where yield',
     literal:
-      'null false true'
+      'null false true',
+    built_in:
+      SYSTEM.classes,
+    class:
```
isagalaev: `class` is reserved for definitions of user classes, not built-in class names. We clump together all the built-ins regardless of their nature.

src/languages/cs.js
```diff
@@ -111,15 +124,40 @@ function(hljs) {
...
+            // for inherited interface in class declaration
+            className: 'class',
+            begin: 'I[A-Z]\\w*',
+            relevance: 0
+          },
+          {
+            className: 'type',
+            begin: hljs.IDENT_RE,
+            relevance: 0
+          },
```
isagalaev: Same thing here, "type" and "class" are reserved for different things, and here it really should be "title". This is just a design decision made in highlight.js and it has to be consistent across similar languages.

src/languages/cs.js
```diff
@@ -134,15 +172,79 @@ function(hljs) {
         ]
       },
       {
+        // Type in new declaration. Ex: new MaClasse();
+        // FIXME in 'new object();' object should be considered as keyword
+        className: 'type',
+        beginKeywords: 'new', end: '[\\(<\\[\\{]', excludeEnd: true,
+      },
```
isagalaev: This would lead to inconsistencies in code: the same name will be highlighted after `new` but not highlighted in other cases. The way we do it is: highlight titles of use classes at the place of declaration and not highlight when used in the code.

Nicolas01:
I removed everything related to the usage of "type".
So it remains

- the highlight of Attributes
- a fix so that 'else if' not be taken as a function
- a richer example

src/languages/cs.js
```diff
@@ -134,15 +137,23 @@ function(hljs) {
...
         // recognized as a function definition
         beginKeywords: 'new return throw await',
         relevance: 0
       },
       {
         className: 'function',
-        begin: '(' + TYPE_IDENT_RE + '\\s+)+' + hljs.IDENT_RE + '\\s*\\(', returnBegin: true, end: /[{;=]/,
-        excludeEnd: true,
+        begin: '(' + TYPE_IDENT_RE + '\\s+)+[A-Z]\\w*\\s*\\(', returnBegin: true,
+        end: /[{;=]/, excludeEnd: true,
```
isagalaev: Looks like this change broke this use case:

    int[] f(int val = 0) { .. }

It's not recognized as a function declaration anymore, as it does in master.

test/detect/cs/default.txt
```diff
@@ -1,11 +1,48 @@
...
+                return _myList;
+            }
+        }
+
+        private async Task<int, int> AccessTheWebAsync()
+        {
+            string urlContents = await getStringTask;
+            return urlContents.Length;
+        }
+    }
```
isagalaev: The problem with a bigger example is that it won't fit on the front page of https://highlightjs.org/, and C# is currently there. Other reasons to have shorter snippets is they make better detection tests and also that there's no point in displaying a big chunk of code to simply demonstrate highlighting. If you wanted to use it as a test for language features, they should go as separate files in test/markup/cs/ (http://highlightjs.readthedocs.io/en/latest/building-testing.html#testing-markup).

src/languages/cs.js
Nicolas01: This change fix the `else if()` taken as a function.
In your use case, the function name starts with a lower-case letter.
In C#, naming guideline advises to use pascal case for method names.
So it seemed to me a good way to fix the `else if` issue.

isagalaev:
Merged with additional tweaks (I changed `[attributes]` to be "meta" as we do for this sort of annotations in other languages). Thanks for you contribution!

PR #1523, opened 2017-05-16
Added and fixed AutoHotkey highlighting

nnnik:
Fixed variable highlighting ( collision with forced expression mode )
Added command highlighting
Added block comments
Added some keywords
I don't really know how to test those changes. I just know that the previous code broke with forced expression mode. Some help would be appreciated.


isagalaev:
Thanks for the update!

PR #1525, opened 2017-05-19
feat: allow hljs to be run from command line

marcoscaceres:
This allows the ability for hljs's build script to be run from
the command line.

This is useful for sofware that requires hljs as part of a build
chain.

It disables generation of docs and demo by default.

The patch adds the following options to "tools/build.js":

 * `--output`: path to write build files to. Defaults to "build".
 * `--docs`: tells builder to include the docs and demo.

Build just for Web, don't include docs:

```Bash
hljs --output "some/path xml" javascript
```

Build for Web, include docs:

```Bash
hljs --docs --output "some/path" xml javascript
```

Build for NodeJS:

```Bash
hljs -t node --output "some/path" xml javascript
```

Build for everything:

```Bash
hljs -t all --output "some/path" xml javascript
```


marcoscaceres:
I'm now using a fork of this in production for [ReSpec](https://github.com/w3c/respec/), a tool that is used at the W3C for creating specifications. 

Here is our setup (via package.json) if people want to see what it looks like to use in practice:  
https://github.com/w3c/respec/blob/develop/package.json#L63

marcoscaceres:
Rebased.

PR #1535, opened 2017-06-02
Powershell

dschuermans:
Powershell has an "approved verbs list" found [here](https://msdn.microsoft.com/en-us/library/ms714428(v=vs.85).aspx)

The initial script had several built-in commands hardcoded (e.g. Add-Computer, Add-Content, ...) but highlight wouldn't pick up on custom commands created in 3rd party modules (e.g. New-ACMEIdentifier)

I removed the list of words for built_in and added a contains with a list of all possible verbs.

No clue what to put in CHANGES.md though, not that familiar with github processes :)


dschuermans:
Whoops, didn't see #1386 

@isagalaev Might be beter to include that one then :)

PR #1536, opened 2017-06-05
New language: Structured Text (IEC 61131-3 ST)

bjosa:
New language added: Structured Text (IEC 611-3 ST). 
Structured text is one of the five languages supported by the IEC 61131-3 standard, designed for programmable logic controllers (PLCs).  (https://en.wikipedia.org/wiki/Structured_text)


yyyc514:
@bjosa 

**First of all, truly sorry for the long delay in anyone responding to this PR.**

Thank you so much for desiring to contribute and writing a language grammar.  Unfortunately, we currently are no longer accepting new languages into the core repository.  The existing maintainers simply don't have time to maintain new languages (fixing bugs, dealing with issues, etc). 

And right here is where I'd usually mention how easy it is to turn your language contribution into a 3rd party module so our users could still easily use it, but someone else has already done that:

https://github.com/highlightjs/highlightjs-structured-text

Could you take a look there and if it's missing anything perhaps you could contribute your fixes/improvements there instead? Of course it's also possible for you to publish your own repository separately (and then we'd have competing grammars for a single language), but hopefully (for the benefit of all our users) we can consolidate everything into a single grammar - and I think in most cases that's the way to go.

Please let us know if you have any questions.

yyyc514:

If you have any questions or concerns you can always respond and we'll still see them.

PR #1537, opened 2017-06-05
Document running inside Web workers

tchernomax:
New javascript code for running highlight.js on web worker.

* use highlight instead of highlightAuto
* can highlight multiple code blocks


marcoscaceres:
Planning to rewrite that part of the README, but somewhat differently to this. But this is good because it game me some ideas.

PR #1542, opened 2017-06-12
updated package.json

EsrefDurna:
updated package glob from 7.0.3 to 7.1.2
 Npm link :  https://www.npmjs.com/package/glob
 Repository link : https://github.com/isaacs/node-glob/releases


isagalaev:
What is the reason for upgrading it?

EsrefDurna:
we use glob 7.1.2,  at other projects, there isn't so clear release notes probably bug fixes,

isagalaev:
I still don't understand why highlight.js dependencies need to be touched simply because other projects use other versions?

EsrefDurna:
yes, I'm trying to minimize my project dependencies,

isagalaev:
But it can't work this way, can it? There are many downstream projects, and they can have different ideas about versions of any of their upstream projects. An upstream project can't satisfy it for everyone, obviously.

Besides, highlight.js does not have any dependencies for running. Those are dev dependencies which don't affect you at all.

PR #1547, opened 2017-06-14
adding brief instructions for building the demo

CatherineH:
So that I don't ask questions like https://github.com/isagalaev/highlight.js/issues/1546 again.


isagalaev:
Comments of this kind is a bit of a slippery slope (as in, someone might have the same question in a different place, and then you end up with a flurry of seemingly unrelated comments everywhere). However I think in this case I think it's more useful than potentially harmful. Merging.

Thanks! :-)

PR #1554, opened 2017-06-21
✨ - add cli tool to quickly check relevance conflicts

ellismarkf:
Rather than having to individually check for relevance score conflicts with the `developer.html` UI, this PR introduces a Node program to check for relevance score conflicts programmatically, all at once, displaying results in the terminal.

To run the check: `node test/detect/checkAutoDetectResults.js`

If there are **no conflicts**:
screen shot 2017-06-21 at 4 30 01 pm

The tool distinguishes between moderate and severe conflicts, where moderate conflicts are when the expected language is returned as the `second_best` result, and severe conflicts are when the expected language was out-scored by both the primary and `second_best` results.

**Moderate conflicts**:
screen shot 2017-06-21 at 4 31 37 pm

**Severe conflicts**:
screen shot 2017-06-21 at 4 36 40 pm

The goal of this tool is to make it easier to understand relevance, as well as give faster feedback to developers about where their language is over-relevant.

The tool should be invoked _after_ building the project, and works best when doing a full, uncompressed build, ie `node tools/build.js -n`.  In compressed mode, `relevance` doesn't seem to be returned, or at least not in the same way, so I coded around that.

(To capture above screenshots, I arbitrarily spiked relevance in Typescript, Javascript, and Clojure, then discarded the changes.  Don't worry 😉 )


marcoscaceres:
Wonder if it we should combine superpowers: https://github.com/isagalaev/highlight.js/pull/1525

yyyc514:
@ellismarkf This is pretty great. Thanks!

PR #1557, opened 2017-06-23
Update AUTHORS.en.txt

revosw:
Added myself


revosw:
Never done a pull request before, but now I've learned

Issue #1563, opened 2017-07-05
sql - FOREIGN is not highlighted 


pleerock:
It looks very weird that `FOREIGN` keyword in not highlighted, probably you should add it in the keyword list that needs to be highlighted in sql:

![sql](https://www.dropbox.com/s/xhhf5ro9msn5r1g/Screenshot%202017-07-05%2016.21.41.png?dl=1)

DavidSimner:
This is a duplicate of https://github.com/isagalaev/highlight.js/issues/1488

DavidSimner:
For Microsoft SQL Server's Transact-SQL (T-SQL), `foreign` is added as a keyword in https://github.com/isagalaev/highlight.js/pull/1524

pleerock:
I sent a PR for this fix. I hope this will be merged soon because its a very simple fix.
![sql](https://www.dropbox.com/s/cwh3zq4pztlvsq2/Screenshot%202017-07-05%2017.36.22.png?dl=1)

PR #1564, opened 2017-07-05
fix for #1563 - FOREIGN key should be highlighted in SQL.

pleerock:
`FOREIGN` key should be highlighted in SQL.


DavidSimner:
For Microsoft SQL Server's Transact-SQL (T-SQL), `foreign` is added as a keyword in https://github.com/isagalaev/highlight.js/pull/1524

PR #1581, opened 2017-07-31
Count matching parenthesis in cpp's params mode.

davidben:
In some cases, a function's parameters will contains parentheses themselves. For instance, one might tag a parameter with `__attribute__((unused))`. Or, if writing code using OpenSSL, `STACK_OF(foo)` is a common type name.

The immediate motivation is https://bugs.chromium.org/p/gerrit/issues/detail?id=6051, though that case's real root cause is `__asm__ volatile (...)` being mis-identified as a function. Resolving that would require carefully counting blocks, but this does at least improve it.


davidben:
Oh, actually, a better use case is function pointer parameters which will also have inner parentheses.

davidben:
Friendly ping.

PR #1608, opened 2017-10-03
Fix filterByQualifiers: fileInfo can be null

schra:
I came across this issue when I tried to use `tools/build.js`.


schra:
*ping*

schra:
Thanks for merging :)

PR #1613, opened 2017-10-10
Add 'with' as an SQL operator starting keyword.

unhandled-exception:
The 'with' keyword is using for CTE:
https://dev.mysql.com/doc/refman/8.0/en/with.html#common-table-expressions-syntax
https://www.postgresql.org/docs/10/static/queries-with.html


ldct:
included with https://github.com/highlightjs/highlight.js/pull/1711/files - thanks though!

PR #1623, opened 2017-11-01
Remove unused variable in javascript.js

vjeux:
This was never used.


marcoscaceres:
Prefer to just leave this as undefined for now. Eventually, we will lint all these files.

Issue #1268, opened 2016-08-05
How to include punctuations/operators for custom highlighting?


igauravsehrawat:
I know this goes against philosophy of highlight.js, but I have been trying to customize it.
I put the following key-value in `keywords` object.

`built_in: /= \+= -= \*= \= %= == != <= >= < > \+ - \* % \^/`

But they aren't getting any `span` class, seems like they are being skipped. Is there a way to this?
Any help would be appreciated.

Thanks

isagalaev:
No, not keywords. You want a separate mode like this:
```js
{
  className: 'built_in',
  begin: /=|+=|-=/
}
```

PR #1629, opened 2017-11-07
Add support for some built-in R operators

yihui:
@isagalaev I'm opening this PR mainly to see if you like the idea before I spend more time on it. Basically I want to highlight some operators in the R language. I'm using your suggestion in #1268.

The code is borrowed from my colleague @jcheng5 a few years ago: https://github.com/rstudio/highlight.js/blob/0710243a8ea53f2b2d34cebd58b7fa82b7760863/src/languages/r.js


yyyc514:
Sorry, but generally we do not highlight operators or punctuation.  Is there some reason R might be exceptional in this regard?  Or if there are some exceptions in our other grammars that I'm not aware of please let me know.

yihui:
If that is the general rule, then please feel free to ignore this PR. Thank you!

PR #1642, opened 2017-11-20
add 'node' literal to xquery

### LINKED ISSUE #1641: xquery language is missing node literal
Github catches the ``node()`` literal of first operand, 
```XQuery
declare function app:helloworld($node as node(), $model as map(*), $name as xs:string?) {
    if ($name) then
        <p>Hello {$name}!</p>
    else
        ()
};
```

highlight correctly catches xs:string, but not ``node()``

screenshot 2017-11-20 14 26 03

duncdrum:
closes  #1641


duncdrum:
see [the specs](https://www.w3.org/TR/xmlschema-2/#boolean) for xml datatypes

PR #1659, opened 2017-12-05
Add z80 assembly language

ISSOtm:
The z80 is a relatively old 8-bit CPU, used for example in some TI graphing calculators, and in the ZX Spectrum. A somewhat modified version is also present in the Game Boy and Game Boy Color consoles.
The syntax picked here is from the TASM assembler (http://www.cpcalive.com/docs/TASMMAN.HTM), for its simplicity and widespread use.

Auto-detect works, Mocha tests returned 522 passing. Nothing seems broken.


morozov:
Does this PR need any extra work in order to be accepted? IMO it could use some improvements (e.g. keywords are highlighted as instructions and the `xor` keyword is missing). But it's still a good start and could be improved later.

ISSOtm:
If it needs to be improved, I'd like to know how. It's been a solid while, though.
Fixing `xor` should be simple enough, but I'm less sure about how to handle the keywords part?

morozov:
@ISSOtm is possible to put the keywords under `keywords.keyword`?. E.g. like it's done in `x86asm`:
https://github.com/highlightjs/highlight.js/blob/160a57192755d6b620d896e9877abd54522c8ca7/src/languages/x86asm.js#L12-L15

yyyc514:
@morozov I was going to agree, but they went this way so they could account for the whitespace requirement:

```
// TASM's spec explicitely says an instruction must start after a whitespace
```

Do we just want to ignore that and make the grammar a little more general? I take it by using keyboards we get the official automated relevant boost and probably other future benefits from standardizing?

yyyc514:
Now that the plan is to keep new languages in their own repositories and have language maintainers I'm also not sure how "strongly" core wants to be about enforcing rules like this (if it was/is even a rule)...? If it was going to be merged into the main codebase I imagine the "requirements" would likely be more stringent.

morozov:
IIRC, when I tested this patch, the instructions like `add`, `xor`, etc. were not highlighted in any special way, so I tried to compare this language definition with the other existing ones and figured out that those tokens would be highlighted there because they would be parsed as keywords.

Maybe the fact that this definition requires a leading space while my code didn't have it was the cause but I don't see a reason why a generic `z80asm` definition should enforce the TASM syntax.

Whether those are keywords or instructions is the matter of the terminology/semantics which I'm not really strong in but the same token `#endif` that is declared here as a keyword/directive is a keyword/meta in `x86asm`. Wouldn't it make sense to have the same tokens parsed the same in various dialects?

yyyc514:
> but I don't see a reason why a generic z80asm definition should enforce the TASM syntax.

Ah, yes now that's a good point. :-)

>  but the same token #endif that is declared here as a keyword/directive is a keyword/meta in x86asm

Could be neither is doing it properly. :-)  `#endif` is definitely a directive if it's being used as like a macro... (like in C, etc)

> Wouldn't it make sense to have the same relatively generic tokens parsed the same in various dialects?

In generally I agree.

It's also possible to do `keywords: { instruction: ` if the flagging was truly useful.

morozov:
@yyyc514 I think the point of my original https://github.com/highlightjs/highlight.js/pull/1659#issuecomment-494107174 wasn't taken as I intended:

> Does this PR need any extra work in order to be accepted? 

If it does, I can help and make the needed changes. If not, why isn't it merged yet?

> But it's still a good start and could be improved later.

My concern is minor, I can live with it. Let's get it merged.

yyyc514:
> My concern is minor, I can live with it. Let's get it merged.

I'm only a fellow contributor.  AFAIK we aren't merging any more (many?) languages - and instead languages are going into their own repositories (some at the highlightjs GitHub organization if we have maintainers for them).  I don't know if there is some backdating or retro-active magic for PRs that have been around a while or not - or if that also means they need to be split out into separate repos.  That is the question.

Either way I was just trying to push forward the syntax in general. :-)

marcoscaceres:
We’d prefer all new languages to go into their own repositories (so they don’t end up stuck in a review queue like this). No exceptions I’m afraid.

yyyc514:
@marcoscaceres Are there any docs on how that process works (to request a repo at `highlightjs` organization), who to ask, etc?  Or do we need that process documented?  If I knew the "correct" or "desired" process and the template I could probably help push some of these languishing PRs along...

Ie, is it possible/preferable to start with an external repo first and then `highlightjs` could just fork it and then assign write permissions to the original author?

marcoscaceres:
No docs yet... I've just been pointing people to the other repos and saying "copy that one, it's good :)". But yes... need to document the process.

yyyc514:
@ISSOtm  

**First of all, again truly sorry for the long delay before you got any response to this PR at all.**

Thank you so much for desiring to contribute and writing a language grammar. I just wanted to follow up on what @marcoscaceres already said earlier.  Unfortunately, we currently are no longer merging new languages into the core repository.  The existing maintainers simply don't have time to maintain new languages (fixing bugs, dealing with issues, etc).  **We'd still love to host your language as a 3rd party module though** - if you're willing to help maintain it.  

It's easy to turn your language into a module that anyone who likes can easily integrate with Highlight.js using a single line of code:

```html
<script src="/path/to/highlight.min.js"></script>
<!-- they'd simply add your language modules -->
<script src="/path/to/highlightjs-z80asm.js"></script> 
```

It's also possible for users to rebuild the `highlight.js` library to include specific 3rd party modules, all in a single file (vs loading a bunch of 3rd party languages one at a time).

If you'd like to become the maintainer for this language then we're happy to create a new repository in the `highlightjs` organization and host the repository there.  That assumes you'd be open to reviewing issues and fixing bugs as they arise.  You're also welcome to host the repo yourself and we'll link to it from our README.

There isn't much in the way of how to docs on this yet, but there are sample repositories and it's honestly pretty easy.  We're still fleshing out an [official "spec" for 3rd party modules](https://github.com/highlightjs/highlight.js/issues/2328), but even after that's done it should require minimal changes from the examples we have already.  IE, if you wanted to get started now there isn't any big reason to wait.

My own `robots.txt` grammar repository is a great, simple example:

https://github.com/highlightjs/highlightjs-robots-txt

yyyc514:
@morozov If @ISSOtm is busy or unavailable and you wanted to take this and run with it I'm sure we'd be happy to have you as the grammar maintainer also - or you could both collaborate.

yyyc514:

As mentioned above if you're interested in becoming the maintainer of a 3rd party module, please respond [whenever you find time] and we'd be happy to help you get started.

PR #1663, opened 2017-12-12
Add tsx (TypeScript + JSX) support

### LINKED ISSUE #1155: (typescript) Add JSX support to Typescript
There's a somewhat popular variant of TypeScript that incorporates JSX, and replaces TypeScript's normal angle-bracket casts with the `as` keyword.  Source files typically end in `.tsx`.  It would be nice to have a language definition for this -- it's not hard to make one, but not clear how to cleanly share the bulk of the definition with `typescript`.

Related issues: #878, reviewable/reviewable#285

ddevault:
This uses the same rules as JavaScript's JSX support.

Fixes #1155


zoran5:
@SirCmpwn Could you please not put the return statement in the constructor, but instead create a separate function?

SirCmpwn:
Does it matter? This is just for testing the highlighitng kicks in right.

zoran5:
@SirCmpwn I think the same examples are also used on https://highlightjs.org/static/demo/  
And I'm pretty sure the users who would see that typescript code there would get annoyed.

mmartinson:
Anything I can do to help move this one along? Testing, changes?

SirCmpwn:
https://github.com/isagalaev/highlight.js/issues/1678

paladox:
Hi, this needs rebasing.

paladox:
@ddevault would you be able to rebase please?

rhigdon:
https://github.com/highlightjs/highlight.js/pull/1966
I rebased Drew's patch and updated the CHANGES.md and AUTHORS.en.txt

PabloSzx:
any news on this?

test/markup/typescript/jsx.expect.txt
```diff
@@ -33,9 +33,10 @@
         <span class="hljs-keyword">return</span> state;
     }
     <span class="hljs-keyword">else</span> <span class="hljs-keyword">if</span> (node.kind === SyntaxKind.ModuleDeclaration) {
-        <span class="hljs-keyword">return</span> getModuleInstanceState((&lt;ModuleDeclaration&gt;node).body);
+        <span class="hljs-keyword">return</span> getModuleInstanceState((<span class="xml"><span class="hljs-tag">&lt;<span class="hljs-name">ModuleDeclaration</span>&gt;</span>node).body);
     }
-    <span class="hljs-keyword">else</span> {
```
yyyc514: Why is the syntax highlighting on else and return breaking here?

test/detect/typescript/default.txt
```diff
@@ -2,6 +2,7 @@ class MyClass {
   public static myValue: string;
   constructor(init: string) {
     this.myValue = init;
+    return <div className="test">jsx syntax</div>;
```
yyyc514: Please also add a test with a self-closing tag as that seems to be a big issue with our React/JSX matching feature for this...

yyyc514:
Closing this as it's out of date now. 

I think we're open to this, but it should be re-reviewed after 10.0 when we can make a single tiny JSX module and then import that exact module into both JS and TS, otherwise this is going to diverge and get very messy over time (as copy and paste tends to do).

PR #1693, opened 2018-02-13
Fixed syntax highlighting for floating points in C#

NuclearCookie:
Numbers regexp copied from c++.

Before:
```
<span class="hljs-keyword">float</span> test = <span class="hljs-number">1.0</span>f;
```
After:
```
<span class="hljs-keyword">float</span> test = <span class="hljs-number">1.0f</span>;
```


NuclearCookie:
@isagalaev are you still maintaining this repo?

PR #1700, opened 2018-02-22
Add some keywords newly appended in Kotlin v1.1 & v1.2

saplf:
The new keywords in Kotlin 1.1
- [suspend](https://kotlinlang.org/docs/reference/coroutines.html#suspending-functions)
- [typealias](https://kotlinlang.org/docs/reference/whatsnew11.html#type-aliases)
- [external](https://kotlinlang.org/docs/reference/whatsnew11.html#the-external-modifier)

And the ones in Kotlin 1.2
- [expect & actual](https://kotlinlang.org/docs/reference/whatsnew12.html#multiplatform-projects-experimental)

Moreover, I have updated the number mode of Kotlin as the description [here](https://kotlinlang.org/docs/reference/whatsnew11.html#underscores-in-numeric-literals). I copied the code from [java.js](https://github.com/isagalaev/highlight.js/blob/master/src/languages/java.js) since I found the both's modes are the same.

### DIFF
--- a/src/languages/kotlin.js
+++ b/src/languages/kotlin.js
@@ -9,7 +9,7 @@ function(hljs) {
     keyword:
       'abstract as val var vararg get set class object open private protected public noinline ' +
       'crossinline dynamic final enum if else do while for when throw try catch finally ' +
-      'import package is in fun override companion reified inline lateinit init' +
+      'import package is in fun override companion reified inline lateinit init ' +
       'interface annotation data sealed internal infix operator out by constructor super ' +
       // new from kotlin 1.1 & 1.2
       'suspend typealias external expect actual ' +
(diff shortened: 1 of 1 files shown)


src/languages/kotlin.js
```diff
@@ -9,8 +9,11 @@ function(hljs) {
     keyword:
       'abstract as val var vararg get set class object open private protected public noinline ' +
       'crossinline dynamic final enum if else do while for when throw try catch finally ' +
-      'import package is in fun override companion reified inline lateinit init' +
+      'import package is in fun override companion reified inline lateinit init ' +
       'interface annotation data sealed internal infix operator out by constructor super ' +
+      'tailrec where const inner ' +
+      // new from kotlin 1.1 & 1.2
```
ldct: I think this comment is unnecessary. We do not classify keywords by which language version introduced them.

saplf: I marked this just for the contributor later to tell these keywords apart more easily. I would move it if necessary.

PR #1708, opened 2018-03-05
enh: adds `data-language` attribute to highlighted blocks

canvasplay:
Based on options.langAttribute (boolean), adds a data-lang attribute to highlighted blocks. The idea behind this attribute is to provide an easy way to retrieve the language string literal from CSS.
So it can be used from stylesheets like the following to add a nice language label.
hljs:before{ content: attr(data-lang); }


src/highlight.js
```diff
@@ -48,7 +48,8 @@ https://highlightjs.org/
     classPrefix: 'hljs-',
     tabReplace: null,
     useBR: false,
-    languages: undefined
+    languages: undefined,
+    langAttribute: true
```
marcoscaceres: We should probably default to false.

marcoscaceres:
Ok, cool. We need to figure out where to document the feature.

canvasplay:
This section looks the right place to document it:
https://highlightjs.readthedocs.io/en/latest/api.html#configure-options

canvasplay:
Is this mergable or not? It was when I submited it.
I am able to the required changes if this PR have a happy ending :)

marcoscaceres:
Sorry this fell off the radar. We’d probably want to add some tests to make sure it doesn’t break anything in non-browser environments.

canvasplay:
All right! Just tell me if anything is required from my part.

src/highlight.js
```diff
@@ -53,7 +53,8 @@ https://highlightjs.org/
     classPrefix: 'hljs-',
     tabReplace: null,
     useBR: false,
-    languages: undefined
+    languages: undefined,
+    langAttribute: false
```
yyyc514: `addLanguageAttr`

And could we add docs also `api.rst` would need to be updated to mention this.

src/highlight.js
```diff
@@ -707,6 +708,8 @@ https://highlightjs.org/
 
     block.innerHTML = result.value;
     block.className = buildClassName(block.className, language, result.language);
+    if(options.langAttribute)
+       block.setAttribute('data-lang',result.language);
```
yyyc514: `data-language` please.

src/highlight.js
canvasplay: Changed. Better, more explicit.

src/highlight.js
canvasplay: Done.

src/highlight.js
canvasplay: Now on docs also :)

yyyc514:
Hmmm.... now that we have this should we consider allowing it to be used to SET the language also?

yyyc514:
Can you add a quick test for this in the browser test suite?  Other than that this looks good!

canvasplay:
> Hmmm.... now that we have this should we consider allowing it to be used to SET the language also?

Looks interesting! but... I can't afford this challenge right now.

canvasplay:
> Can you add a quick test for this in the browser test suite? Other than that this looks good!

Ive had quick to test/browser/ and can't figure out where to fit an specific test for this enhancement. Additionally this feature is disabled for the default configuration...

yyyc514:
`plain.js` would be better than nothing.  We don't have much there yet so it's not as well organized as the other tests grouping.

yyyc514:
@egor-rogov Can you think of a reason not to just do this by default assuming there was no data-language attribute already present?

egor-rogov:
> @egor-rogov Can you think of a reason not to just do this by default assuming there was no data-language attribute already present? And then we'd just make this default behavior instead of an option.

+1 for default behavior. Can't see any harm in setting this attribute.

yyyc514:
@canvasplay Lets change this then to just make it the default behavior, remove the config option... if `data-language` isn't already set, set it.

I'd also like this to be a way someone could tell Highlight.js WHICH language to highlight (as a new option vs using class), but if you aren't up for that we can make a separate issue for it.  Sorry to keep changing the goalpost. :)

canvasplay:
Done! :)

yyyc514:
I think we only want to do this if the attribute doesn't already exist though... i don't see that you added any code to check that?

src/highlight.js
```diff
@@ -845,8 +844,7 @@ https://highlightjs.org/
 
     block.innerHTML = result.value;
     block.className = buildClassName(block.className, language, result.language);
-    if(options.addLanguageAttr)
-       block.setAttribute('data-language',result.language);
+    block.setAttribute('data-language', block.getAttribute('data-language') || result.language);
```
canvasplay: > I think we only want to do this if the attribute doesn't already exist though... i don't see that you added any code to check that?

Well, here I meant to set the attribute value to `result.language` except the attribute already has a value, in that case the attribute will remain with the initially defined value.
Maybe you mean something more like:
`if(!block.hasAttribute('data-language')) block.setAttribute('data-language', result.language);`
What do you think?

yyyc514:
@canvasplay update changelog then we can get this merged?

yyyc514:
> We’d probably want to add some tests to make sure it doesn’t break anything in non-browser environments.

@marcoscaceres  This code is only running in browser environments, it's inside `highlightBlock`  - which only runs on the browser.  Can we get your approval?

marcoscaceres:
I think this would benefit from a couple of tests.

yyyc514:
tests to add:

- [ ] that it successfully tests the attribute
- [ ] that it does not overwrite an existing attribute

yyyc514:
@marcoscaceres Better?  If insufficient please let me know exactly what else you'd like to see tested.

src/highlight.js
```diff
@@ -854,6 +854,7 @@ https://highlightjs.org/
 
     block.innerHTML = result.value;
     block.className = buildClassName(block.className, language, result.language);
+    block.setAttribute('data-language', block.getAttribute('data-language') || result.language);
```
marcoscaceres:
```suggestion
    block.dataset.language = block.getAttribute('data-language') || result.language;
```

test/browser/plain.js
```diff
@@ -33,6 +33,21 @@ describe('browser with html with quotes in attributes', function() {
...
+describe("highlighBlock and data-language attribute", function() {
+  before(async function() {
+    await buildFakeDOM.bind(this)();
+  })
+  it('should set if unset', async function() {
+    this.hljs.highlightBlock(this.block);
+    this.block.dataset.language.should.equal("javascript")
+  });
+  it("should not alter if already set", async function() {
+    this.block.dataset.language="existingData";
```
marcoscaceres:
```suggestion
    this.block.dataset.language = "existingData";
```

CHANGES.md
```diff
@@ -7,6 +7,7 @@ New styles:
   none.
 
 Improvements:
+- enh(parser): adds `data-language` attribute to highlighted blocks (#1708) [canvasplay][]
```
marcoscaceres: I'm worried that "data-language" is overly generic. maybe we namespace this as "data-hljs-language"? The will reduce the risk of conflict, and allows us to more easily add more attributes in the future while minimizing the risk of collisions with existing attributes.

test/browser/plain.js
yyyc514: Lol

src/highlight.js
yyyc514: Ah I thought dataset was newer. Didn’t realize it was so old. Sure I can use newer api.

CHANGES.md
yyyc514: I say “wait and see” and we can add a prefix option later if necessary.

test/browser/plain.js
marcoscaceres: We should consider applying [prettier](http://prettier.io) via ESLint to our tests and code to maintain a consistent coding style.

yyyc514: Perhaps. You could open an issue. 😉. I’ve been using JSLint some but it doesn’t yell at that level I don’t think. I try but don’t get hung up on tiny things like that too much. 

How would you make that automatic? Like add it as a CI check?

marcoscaceres: Yeah, CI check. We would just add on TravisCI `eslint .` and few other dev dependencies to package.json ... a few style rules.

src/highlight.js
yyyc514: Ah I was right. It’s not fully supported on older browsers. I know we can’t actively support all those older browsers but I’ve been trying not to PURPOSELY break them until 10.0. there are all sorts of little things to change once we jump over that fence. 

I’ll add this stuff to the 10.0 list.

test/browser/plain.js
yyyc514: I’ll play around with it locally first.

CHANGES.md
marcoscaceres: This might catch people off tho... imagine some random application that depends on our library already does the following:

```
document.querySelectorAll(['data-language']).forEach(doStuff)
```

(You can see just on [github search](https://github.com/search?q=document.querySelectorAll%28%5B%27data-language%27%5D%29&type=Code) examples)

By adding just 'data-language', we've just unexpectedly got added to someone's unrelated code potentially breaking someone's application. The risk of that is reduced by the "hljs-"  prefix. 

Again, there are [zero hits](https://github.com/search?q=document.querySelectorAll%28%5B%27hljs-data-language%27%5D%29&type=Code) on Github search for `data-hljs-language`. 

And I only found 2 hit on Google: 

https://github.com/pc035860/angular-highlightjs/issues/33

And: 
Screenshot 2019-11-14 14 32 04

test/browser/plain.js
marcoscaceres: Ok, if you need want to see how it's used in a different project:
https://github.com/w3c/respec/

See the .travis file and package.json... you can also try `eslint --fix` to see how many changes there would be.

CHANGES.md
yyyc514: Well, now I thought people would actually look at the pieces they were putting together and make sure they were all compatible.  I'm not sure the best thing is for every program to preemptively assume a namespace is needed... that sounds very ugly.  Though I understand the reasoning behind it.

But you've persuaded me that perhaps the whole idea here was accepted a bit too quickly... Things already in-flight (PRs) before I started doing triage got a lot lighter review than new requests... I think this is perhaps better handled with a plugin, and makes a good use case for a post highlightBlock plugin.

https://github.com/highlightjs/highlight.js/pull/2285

yyyc514:
Closing this now in favor of #2285 and discussion can continue there.  Instead of making this a default feature, we can make it a plugin... then someone can configure it however THEY think is best, whether with a namespace or not.

CHANGES.md
marcoscaceres: > Well, now I thought people would actually look at the pieces they were putting together and make sure they were all compatible.

As you’ve eloquently stated previously: lol. 

> I think this is perhaps better handled with a plugin, and makes a good use case for a post highlightBlock plugin.

I think this still has merit/utility when automatic language detection occurs. 

Re namespace: Note also we use the namespace in our CSS. This follows the same principle to minimize the chance of collisions.

yyyc514:
> I think this still has merit/utility when automatic language detection occurs.

Well, it's already easy to get the language via JS (just not CSS)... but if someone else wanted to finish this and champion it, I won't stand in the way.  I've wasted too much time on this already though.

I think ultimately it comes down to the philosophy the library wants to take on things like this.  Since we seem to be very low on maintainers and time I think a plugin philosophy to make it easy for people to extend it themselves (and those extensions to continue to work moving forward).  And I also like the idea of "recipes"... esp. for things that seem helpful, but not GLOBALLY so.

Of course both approaches can also exist concurrently.

canvasplay:
Ok. I understand your point. Plugin phylosophy sounds good.
Anyways I just wanted to make little contribution. Sorry for making you to waste your time ;)

yyyc514:
>  Sorry for making you to waste your time

Oh it definitely wasn't personal at all... you just went somewhere and I'm impatient so I spent time on this and then Marco said more tests and then I did that and then he thought the namespace was an issue... and then I lost interest plus realized this is really a plugin thing.  The cool part to me was "data-language" and it being purposely general vs namespaced. :-)

I tend to think anything that could be a few lines of code should start as a plug-in and then MAYBE if it becomes popular it could be built-in... or use we just have an internal plug-in library and turn everything into plugins (that's possible).

I was already looking at useBR and convertTabs and thinking those could really be a small plugin keyed into just the two new hooks I added in the PR.

@canvasplay Did you see my suggestion over in the PR?

```js
    hljs.addPlugin( {
      afterHighlightBlock: ({block, result}) => {
        // move the language from the result into the dataset
        block.dataset.language = result.language }
    })
```

Issue #1713, opened 2018-03-20
Can anyone help me document .properties syntax


bostko:
I'd like to be able to support https://en.wikipedia.org/wiki/.properties format.
Could you point me at similar `language/*.js` file for reference?

jolkdarr:
Hi

Have you studied `ini` syntax?
Doc at: https://en.wikipedia.org/wiki/INI_file

yyyc514:
@marcoscaceres Another merged item we can close out.

yyyc514:
Thanks!

PR #1717, opened 2018-03-24
#1713 Support for .properties syntax

bostko:


bostko:
Can anyone review this simple PR, thanks :)

egor-rogov:
Hi @bostko, thanks for your effort! It is a bit more complicated than it seems, but finally I sort of managed to do it right. I've oped new PR #1858 for it.

PR #1719, opened 2018-03-25
Add missing Swift keywords

hartbit:


ldct:
All but one of the additions look like preprocessor directives; can we use something like https://github.com/highlightjs/highlight.js/blob/master/src/languages/cpp.js#L43 instead of marking them as keywords?

marcoscaceres:
Oops, I missed 
https://github.com/highlightjs/highlight.js/pull/1719#issuecomment-408458134

Seems like a better solution.

hartbit:
@marcoscaceres I've been very busy lately but still plan to fix those issues.

marcoscaceres:
Thanks for the update, @hartbit!

marcoscaceres:
Closing due to inactivity. Happy to reopen if updated tho.

PR #1721, opened 2018-03-29
Add missing "var" keyword in Java 10

lc-soft:
Java 10 has introduced a new "var" keyword.

[JEP 286: Local-Variable Type Inference](http://openjdk.java.net/jeps/286)


ldct:
thanks!

PR #1741, opened 2018-04-30
Update SQF Language to Arma 3 v1.82

dedmen:
Last update was #1373 

The code used to generate this update can be found in https://github.com/dedmen/highlight.js_SQFgen
This update adds all the commands that were added since 1.66.

Additionally some "commands" that produce constant output were moved to `literals`.
And the `private` keyword was moved to `keywords`

Here is a more easy to view overview of all changes https://github.com/dedmen/highlight.js_SQFgen/commit/62cd50c6b72bea04ca91c9562a575dc38c89fb2b

### DIFF
--- a/src/languages/sqf.js
+++ b/src/languages/sqf.js
@@ -1,7 +1,7 @@
 /*
 Language: SQF
 Author: Søren Enevoldsen <senevoldsen90@gmail.com>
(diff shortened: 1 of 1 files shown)


dedmen:
Might aswell wait another week or two. 1.84 will be out soon. But there won't be alot of important changes as they were in 1.82

PR #1749, opened 2018-05-24
Fix Elm characters

jdmnd:
The line
```elm
cons 'C' <| cons 'h' "ars"
```
currently renders as
```html
cons '<span class="hljs-type">C'</span> &lt;| cons 'h' <span class="hljs-string">"ars"</span>
```

This change renders it as
```html
cons <span class="hljs-string">'C'</span> &lt;| cons <span class="hljs-string">'h'</span> &lt;| <span class="hljs-string">"ars"</span>
```

The same change could be copied over to Haskell, since that suffers from the same issue.

### DIFF
--- a/src/languages/elm.js
+++ b/src/languages/elm.js
@@ -38,6 +38,12 @@ function(hljs) {
     contains: LIST.contains
   };
 
+  var CHARACTER = {
+    className: 'string',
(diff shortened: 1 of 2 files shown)


ldct:
could you use `APOS_STRING_MODE` instead? seems like the intention here is to highlight things like `'abc'` as well.

PR #1765, opened 2018-07-22
Revert "Add `apache` alias"

isagalaev:
Reverts highlightjs/highlight.js#1751

The change is unnecessary since the file name works as the main language name. Aliases only hold additional names.

/cc @ldct


ldct:
ack! sorry for that!

isagalaev:
There's a lot of these undocumented gotchas, unfortunately…

PR #1768, opened 2018-07-22
Ask travis to test against stable and LTS node.js releases

ldct:
The currently specified node.js versions that travis tests against are 4 and 5, which are quite old. This PR adds the latest stable version of node to that list. 

https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions

I am not sure what build infra or other things might be running node 4 and 5; we can delete those versions at a later time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-  - "4"
-  - "5"
   - "node"
 env:
   -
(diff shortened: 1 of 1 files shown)


isagalaev:
My local node says it's 8.10, and it builds and runs tests just fine. So I'm +1 on removing 4 and 5.

marcoscaceres:
Might be good to simply target node LTS, unless there is a good reason/feature not to do so. That’s usually least disruptive down stream.

ldct:
@marcoscaceres makes sense to me. now we have LTS (currently 8) and stable (currently 10)

PR #1769, opened 2018-07-22
Dart: string interpolation without curly braces

Dorumin:
Adding support for string interpolation without curly braces.


ldct:
thanks!

PR #1771, opened 2018-07-22
Fix Dockerfile sublanguage keywords which produce no highlighting

l-vo:
When using a Dockerfile sublanguage keyword, all the following instructions are not colored. This PR fix this bug.


gusbemacbe:
@ldct and @SirCmpwn

l-vo:
@gusbemacbe @ldct @SirCmpwn Seems to not work only in a specific environment (Hexo Blog), sorry for the inconvenience.

gusbemacbe:
OK, no problems.

PR #1792, opened 2018-08-09
chore: update deps, alias mocha as script

marcoscaceres:
This updates everything except jsdom, as latest breaks the build. We can investigate that one separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,24 +22,25 @@
   },
   "main": "./lib/index.js",
   "scripts": {
-    "test": "./node_modules/.bin/mocha test/",
-    "test-browser": "./node_modules/.bin/mocha test/browser/"
+    "mocha": "mocha",
+    "test": "mocha --globals document test",
+    "test-browser": "mocha --globals document test/browser"
   },
   "engines": {
     "node": "*"
   },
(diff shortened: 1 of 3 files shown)


marcoscaceres:
folks who signed up a as maintainers, @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll, @sangdth ping.

package.json
```diff
@@ -22,24 +22,25 @@
   },
   "main": "./lib/index.js",
   "scripts": {
-    "test": "./node_modules/.bin/mocha test/",
-    "test-browser": "./node_modules/.bin/mocha test/browser/"
+    "mocha": "mocha",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: what is this script for?

marcoscaceres: Otherwise, the script points to "./node_modules/.bin/mocha", which is bad. It let's npm or yarn work out which mocha binary to use.

marcoscaceres:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll do you have enough rights to do a release?

PR #1795, opened 2018-08-14
Fix Dockerfile definition when using highlight continuation parameter

l-vo:
Let's consider this case:

```javascript
const str1 = 'FROM ubuntu:trusty';
const str2 = 'RUN echo "foo"';
const str3 = 'USER bar';

var html = '';
var result = hljs.highlight('docker', str1);
html += result.value;
result = hljs.highlight('docker', str2, false, result.top);
html += result.value;
result = hljs.highlight('docker', str3, false, result.top);
html += result.value;
console.log(html);
```

Before this fix, the previous chunk of code returns:

```xml
<span class="hljs-keyword">FROM</span> ubuntu:trusty<span class="hljs-keyword">RUN</span><span class="bash"> <span class="hljs-built_in">echo</span> <span class="hljs-string">"foo"</span></span><span class="bash">USER bar</span>
```
As we can see, the USER dockerfile directive isn't detected as a Docker keyword.

So, after the fix:

```xml
<span class="hljs-keyword">FROM</span> ubuntu:trusty<span class="hljs-keyword">RUN</span><span class="bash"> <span class="hljs-built_in">echo</span> <span class="hljs-string">"foo"</span></span><span class="hljs-keyword">USER</span> bar
```

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -272,3 +272,4 @@ Contributors:
 - Gustavo Costa <gusbemacbe@gmail.com>
 - Antoine Boisier-Michaud <aboisiermichaud@gmail.com>
 - Alejandro Isaza <al@isaza.ca>
+- Laurent Voullemier <laurent.voullemier@gmail.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,11 +16,13 @@ Improvements:
 - Added support for multiline strings to *Swift*, by [Alejandro Isaza][]
 - Fixed issue that was causing some minifiers to fail.
 - Fixed `autoDetection` to accept language aliases.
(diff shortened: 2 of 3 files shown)


l-vo:
@egor-rogov I updated `CHANGES.md` and `AUTHORS.md` but not the markups since markups don't use continuation parameter, and the bug that this PR fixes only occurs when continuation parameter is set to `true`.

egor-rogov:
@l-vo Ok, merged!
I see you didn't change the markup. The reason I ask is that there are no markup tests at all for dockerfile right now. We'll appreciate another PR for tests.

l-vo:
Ok! No problem, I'm going to propose a PR for that soon :-)

Le jeu. 18 oct. 2018 à 22:03, Egor Rogov <notifications@github.com> a
écrit :

PR #1818, opened 2018-09-04
docs(AUTHORS): add Gustavo Costa

gusbemacbe:
Ready. I revised the CHANGES. No errors.

But I think the AUTHORS list should be in alphabetical order because certains persons have difficulty to locate us to contact us.


marcoscaceres:
> But I think the AUTHORS list should be in alphabetical order

Sure, but remember we should always update those things as new pull request and usually right before release. Otherwise, it causes further merge conflicts for all other pull requests (no big deal right now... but please keep that in mind).

PR #1755, opened 2018-07-06
Update stan

jrnold:
This is more-or-less a rewrite of the current Stan language definition by @brendan-r . This version has improved coverage of the language and brings it up to date with the current defintion of the Stan modeling language (2.17.1).

PR #1829, opened 2018-09-09
enh(stan) updated with improved coverage of language keywords and patterns. 

jrnold:
-  Almost complete rewrite. Corrected most patterns and updated function 
definitions.
-  Add "stanfuncs" as an alias

This PR fixes the botched rebase in #1755.


yyyc514:
@jrnold You still out there?  Any thoughts about being maintainer for this?

src/languages/stan.js
```diff
@@ -1,89 +1,227 @@
...
+  ]
+
   return {
+    aliases: ['stanfuncs'],
+    keywords: {
+      'title': BLOCKS.join(' '),
+      'keyword': STATEMENTS.concat(VAR_TYPES).concat(SPECIAL_FUNCTIONS).join(' '),
+      'built_in': FUNCTIONS.join(' ')
+    },
+    lexemes: hljs.IDENT_RE,
```
yyyc514: Was UNDERSCORE_IDENT_RE incorrect?

egor-rogov:
This language is already in the core. Why should we move it away?

yyyc514:
Moving this to the language discussion.

yyyc514:
Since this is already in core though I was moving forward with reviewing it as if it was mergeable since it looks good to me... what to do long-term is a separate discussion.

src/languages/stan.js
yyyc514: Ping.  If we figure this out I think we could get this merged into the next release. 

Did @jrnold disappear?  The auto-complete doesn't work. Strange.

yyyc514:
Bumping this from 9.15.11.

yyyc514:
@jrnold You still around?  Would love to work on this with you and get it merged...

yyyc514:
@jrnold Ping.

src/languages/stan.js
yyyc514: Oh, so in looking very closely our underscore is for identifiers that can START with underscore... boy it could use a better name IMHO. :-)

Issue #1822, opened 2018-09-05
Custom No-op Plain Text Syntax exhibits odd Behavior


RickStrahl:
In order to handle plain text I've created a custom `test.js` syntax which is intended as a no-op. I realize that I can use `class="nohighlight"` for no-ops that just render the code, but that involves intercepting the tags, which currently is handled natively by a Markdown parser which generates `langauage-syntax` tags.

To get around this I created a no-op syntax that doesn't do anything which looks like this:

```js
function(hljs) {
  return {
    aliases: ['txt','plain','none'],
    case_insensitive: true,
    keywords: '',
    contains: []
  }
}
```

and rebuild the package that includes this `text` syntax.

This sort of works - for `txt` and `none`, but not for `text` and plain:

![image](https://user-images.githubusercontent.com/1374013/45072671-da5b5d00-b090-11e8-9a9a-6d2d320d7fc5.png)

The rendered output is this.

**Doesn't work:**

```html
<pre><code class="language-text">text
one
two
three
</code></pre>
```

**Works:**

```html
<pre><code class="language-txt">txt
one
two
three
</code></pre>
```

when using this code to highlight:

```js
function highlightCode() {
    $("pre code")
        .each(function (i, block) {
            hljs.highlightBlock(block);
       });
}
```

If I remove the text syntax completely none of them work, so it seems the no-op syntax definition is working but somehow not hitting all the aliases.

Two questions:

* Is there an easier way to handle plain text - ideally I'd like to be able to just use a fence without auto-detection **without checking the entire document** (ie. `highlight()` - I'm handling each block individually using `highlightBlock()`).

* Why is the `text.js` syntax working for some of the aliases and not for others?

egor-rogov:
It would be helpful to see a complete example which can be reproduced (all files necessary to build the package).
I was thinking of something like this myself (to render `select` results), but I can't figure out what to do about `detect/text/default.txt`. This "language" won't autodetect any document and hence never will pass tests.
AFAIK currently there is no way to disable autodetection. Maybe it's worth discussing such feature.

RickStrahl:
Yes it would be super nice if there was a way to specify that highlightBlock() wouldn't auto-detect. The docs seem to imply that `highlight()` doesn't auto-detect, so it seems that this is possible? Or am I reading that wrong?

IAC the behavior that I would love to see is a `noAutoDetection` flag which when set just applies the `.hljs` style and doesn't do anything else.

[...]

PR #1834, opened 2018-09-17
Fix for #1822

egor-rogov:
### Problem statement

Sometimes you need to give the plain text a "console" look. Think of output of a program or results of a SQL query for an example. The text should look like code, but without any highlighting.

Highlightjs has several language names (defined by `noHighlightRe`: `no-highlight`, `nohighlight`, `plain`, `text`), one of which is described in `README.md`:

> To disable highlighting altogether use the `nohighlight` class:
> ```html
> <pre><code class="nohighlight">...</code></pre>
> ```

Unfortunately in this case `hljs` class is not added to `code`'s classes, hence the text doesn't look like other (highlighted) code.

### Solution proposed

The proposal is to change behavior of `plain` / `text` language: to make it act just like a normal language which highlights nothing. Any attempt of autodetection for this language will result in zero relevance for any input text. Hence `plain` / `text` language must be specified explicitly by the user, and autodetection tests must be skipped for it.

Specifically, this pull request adds:
1. New attribute `disableAutodetect`.
2. New language `plain` (alias `text`).
3. Changes for skipping autodetection and the corresponding test.
4. Removes `plain` and `text` from `noHighlightRe`; `no-?highlight` remains intact.


src/highlight.js
```diff
@@ -36,7 +36,7 @@ https://highlightjs.org/
       aliases   = {};
 
   // Regular expressions used throughout the highlight.js library.
-  var noHighlightRe    = /^(no-?highlight|plain|text)$/i,
+  var noHighlightRe    = /^no-?highlight$/i,
       languagePrefixRe = /\blang(?:uage)?-([\w-]+)\b/i,
```
marcoscaceres: Will removing plain and text affect existing users?

marcoscaceres: Hmm 🤔 We might need to search source files in other projects via Github search.  I suspect there will be some, even if not document. We should strive to remain backwards compatible.

egor-rogov: I'm skeptical about the idea of searching... Too easy to miss something.
Let's choose another name. How about `output` or `plaintext`?

marcoscaceres: `plaintext` works for me!

egor-rogov: Fine, I'll make changes.

marcoscaceres: Appropriate it. I’ll give others a chance to weigh in, but would like to merge by end of the week. Thanks for your hard work here and for thinking this through.

marcoscaceres:
I’ll need to get the original maintainer to give me npm publish rights.

qubyte:
Just dropping in to say thanks for this!

RickStrahl:
So I'm back looking at this today and I still am not seeing the behavior I expected based on this discussion.

Adding the `plaintext` provider into my custom build the following works:

````
```plaintext
This is plain text
```
````

However, using something like this:

````
```text
This is plain text
```
````

or using `plain` or `''` still will render a plain tag without the `.hljs` style applied.

I thought I might be able to quick fix this by changing the `plaintext.js` like this and adding the aliases:

```js
function(hljs) {
    return {        
        aliases: ['text','txt','plain','plaintext'],
        disableAutodetect: true
    };
}
```

but that has no effect.

What I really need for my scenario is a `text`/`txt` handler that shows plain text, but does apply the `.hljs` class.

Am I missing something or is that still not possible even with this fix?

egor-rogov:
Hi @RickStrahl,
it's still not possible and unlikely be possible in the future. For a quick-and-dirty fix you can find and edit this line in `highlight.js`:
```js
   var noHighlightRe    = /^(no-?highlight|plain|text)$/i,
```
Remove `|plain|text` from it, and together with your change to `plaintext.js` this should work.

RickStrahl:
Thank you so much. The workaround works, but not really a long term solution given that this will change with every release unfortunately and I see no good way to automate that.

It would be nice to have some way to pass in the 'exclusion-list'.

For now I added a note to my custom build script with steps to manually change after updates which hopefully I'll remember.

Anyway - thank you for confirming. I remember when I followed the original discussion we had it wasn't clear that `text/plain` would not work and at least now I can either fix it

PR #1793, opened 2018-08-09
Added Support for ReasonML

gmmorris:
Added support for ReasonML including some new css classes for ReasonML specific syntax.

Also added a variant of the One Dark Syntax theme with support for the additional reason syntax (chose to add a new style as I've changed some of the styles in the existing Atom One Dark to work nicer with the new classes)

PR #1837, opened 2018-09-18
enh(reasonml) Improve highlighting of modules and content, remove function annotation

### LINKED ISSUE #1836: Pull out reasonml integration?
Hello! I'm a maintainer of Reason. @gmmorris has created to the reasonml highlight recently, and in order to iterate on it faster, I've pulled it out into https://github.com/reasonml-editor/reason-highlightjs.  I've been using it in place of the first-party one, and it's been solid.
Pulling the plugin out also makes it so that folks don't have to build highlightjs master to obtain it (which isn't always easily doable since it might be a transitive dependency). Last release was a year ago. Additionally, it seems that highlightjs would like to externalize the language plugins anyway; this would already be done.
So I'm wondering if we should just remove the reasonml plugin from this repo; if not, then I can sync that plugin back in if you'd like.

Thanks!

chenglou:
Fixes #1836

Building up to the work of #1793:

- Change reasonml to reason
- Improve highlighting of modules and content
- Temporarily remove the annotation for functions (many false positives)


chenglou:
Not sure why these two tests are failing. Works fine locally

marcoscaceres:
Build for target  “browser” seems to be failing.

chenglou:
Worked locally

yyyc514:
@chenglou Are you still around to work on this if I review it?  First it would need to be updated and resolve the conflicts with master.  And please don't rename the language file... that's a breaking change because people of CDN usage, etc...

chenglou:
@yyyc514 so what would be the way to go to move toward `reason` instead of `reasonml`? For markdown code snippets etc.

yyyc514:
> @yyyc514 so what would be the way to go to move toward reason instead of reasonml? For markdown code snippets etc.

For now just add `reason` as another alias... that way we don't break all the people already using `reasonml`.  If the ML is truly `wrong` now we could add this to the 10.0 list to "break" it officially.

yyyc514:
You still out there?

yyyc514:
Closing for inactivity (cleaning out old PRs for the new year).  If you return and have time to work on this again we could pick up the discussion at that point.  Happy new Year!

PR #1839, opened 2018-09-20
Soft illegals

brunophilipe:
Instead of ignoring illegals completely, using "soft illegals" will cause highlight.js to use the illegal expressions as an alternate end expression. This way, it will still refuse to highlight illegal expressions, but without bailing out completely.

Please note my experience with JavaScript is limited. Perhaps there is a better way to achieve this behavior. However I tested it and it works very well! This is interesting for my use case because I want to be able to highlight small chunks of code at a time in my code editor, which are not necessarily correct, as the user might still be composing them.

---

For example. This illegal multi-line string is currently highlighted when using "ignore_illegals":

```c
char *string = "foo\
bash
baz";
```

screen shot 2018-09-20 at 01 25 57

---

Using soft illegals, the string expression is highlighted up to the "bash" word, after which the expression ends. The resulting highlight is better than not having any highlight at all, and actually matches exactly what Xcode highlights for the same expression:

screen shot 2018-09-20 at 01 22 52

In Xcode:

screen shot 2018-09-20 at 01 22 42

Cheers!

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -224,7 +224,7 @@ https://highlightjs.org/
     return mode.cached_variants || (mode.endsWithParent && [inherit(mode)]) || [mode];
   }
 
-  function compileLanguage(language) {
+  function compileLanguage(language, soft_illegals) {
 
     function reStr(re) {
         return (re && re.source) || re;
@@ -278,7 +278,9 @@ https://highlightjs.org/
           mode.end = mode.begin;
         if (!mode.end && !mode.endsWithParent)
           mode.end = /\B|\b/;
-        if (mode.end)
+        if (mode.end && soft_illegals && mode.illegal)
+          mode.endRe = langRe('(' + mode.end + '|' + mode.illegal + ')');
+        else if (mode.end)
           mode.endRe = langRe(mode.end);
         mode.terminator_end = reStr(mode.end) || '';
         if (mode.endsWithParent && parent.terminator_end)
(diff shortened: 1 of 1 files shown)


src/highlight.js
```diff
@@ -314,15 +319,19 @@ https://highlightjs.org/
...
+  - continuation:
+
+  Returns an object with the following properties:
 
   - relevance (int)
   - value (an HTML string with highlighting markup)
-
   */
-  function highlight(name, value, ignore_illegals, continuation) {
+  function highlight(name, value, soft_illegals, continuation) {
```
marcoscaceres: so, instead of changing the function signature in this manner, I suggest you instead change it to accept an an object as the last argument:

``` JS
function highlight(name, value, ignore_illegals, continuation, options) {
}
``` 

Then pass the `options` to `compileLanguage`:

```JS
compileLanguage(language, options);
``` 

And then check the `options.soft_illegals`... I would also change  `soft_illegals` to `treat_illegals_as: "soft"`, that way, it's more extensible in the future.

src/highlight.js
```diff
@@ -278,7 +278,12 @@ https://highlightjs.org/
           mode.end = mode.begin;
         if (!mode.end && !mode.endsWithParent)
           mode.end = /\B|\b/;
-        if (mode.end)
+        if (soft_illegals && mode.end && mode.illegal) {
+          var endSource = (mode.end instanceof RegExp ? mode.end.source : mode.end);
```
yyyc514: Isn't this why we have the `reStr` helper?

src/highlight.js
yyyc514: Definitely agree.

yyyc514:
Having two different settings does seem messy... I wonder if it wouldn't be clearer to change the parameter to "handleIllegals": ["ignore", "endMatch", "error"]... we could grandfather in `true` and `false` for ignore/error...

yyyc514:
This needs to be held until after https://github.com/highlightjs/highlight.js/issues/2277 in which case we can just come up with the perfect clear name and then add it as a boolean flag.

yyyc514:
Actually this code base would have to be entirely reworked, so I'm closing this PR for now.  It would likely need to interface with `processLexeme` (at run-time) not hack the modes at compile time.  _We only compile modes a single time, so the mode itself needs to be static (post-compile) and cannot be dependent upon run-time config constraints, like it is in this patch._  

> for my use case because I want to be able to highlight small chunks of code at a time in my code editor

Also, just FYI this is not really a supported use-case. Of course you're free to do whatever you want with the library, but I'm hesitant to add features just to support a use-case we actually don't encourage or support.  

**I'd suggest we first consider how to possibly solve this with a plugin rather than adding it to the core highlighter.**   Perhaps an `on:Illegal` hook/callback of some kind that plugins could tap into... OR I've also been considering mode compile callbacks... so a plugin could perhaps mess with the mode compilation and turn illegal into an end variant (like this PR does)... or course this would have the same caveats I mention above, but that might not be a concern for the users of the plugin.

If we want to continue the discussion I'd suggest opening a new issue thread and pointing to this PR for context.

PR #1856, opened 2018-10-08
Updated cmake.js to contain new keywords/commands.

Bagira80:
Added the latest commands and keywords


egor-rogov:
@Bagira80 Thanks, pushed!

Issue #1859, opened 2018-10-11
Typescript decorator factories in constructor


Aboisier:
Hi!

It seems like there is a problem with decorator factories in constructors.

```ts
class Foo { 
    constructor(@decorator() bar: Bar) { }
}
```

![image](https://user-images.githubusercontent.com/6698529/46777973-a0105b80-ccdf-11e8-8988-4e3b94c2cd69.png)

It seems to be working just fine with plain decorators:

```ts
class Foo { 
    constructor(@decorator bar: Bar) { }
}
```

![image](https://user-images.githubusercontent.com/6698529/46777989-aacaf080-ccdf-11e8-8486-0bab43c5b07f.png)

PR #1870, opened 2018-10-14
Typescript decorator factories

Aboisier:
Fix for #1859. 

Currently, having a decorator factory in a constructor breaks the highlighting. 

Example:

```ts
constructor(@decorator() private bar: Bar) { }
```

Is rendered as (notice the `</span>` after `@decorator(` )
```html
<span class="hljs-keyword">constructor</span>(<span class="hljs-params">@decorator(</span>) private bar: Bar) { }
```

It should be rendered as 
```html
<span class="hljs-keyword">constructor</span>(<span class="hljs-params"><span class="hljs-meta">@decorator</span>() <span class="hljs-keyword">private</span> bar: Bar</span>) { }
```

There are also a couple more scenarios to take into account:
```ts
constructor(@decorator('Hello') private bar: Bar) { }
constructor(@decorator(123) private bar: Bar) { }
constructor(@decorator(true) private bar: Bar) { }
constructor(@decorator(someFunction('Hello')) private bar: Bar) { }
function(@decorator() param: Param) { }
```

This pull request adds decorators and arguments. These constructs may be contained in the root, in functions definitions, and in constructors. I added a test with different scenarios. I reduced duplication by creating the PARAMS object. Finally, the current decorators definition did not allow decorators names to start with `$` or `_`, which is a valid. I also fixed that.

Thanks for considering this fix!
Antoine

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,9 +8,11 @@ Improvements:
 
 - *C#* function declarations no longer include trailing whitespace, by [JeremyTCD][]
 - Added new and missing keywords to *AngelScript*, by [Melissa Geels][]
+- *TypeScript* decorator factories highlighting fix, by [Antoine Boisier-Michaud][]
 
 [JeremyTCD]: https://github.com/JeremyTCD
 [Melissa Geels]: https://github.com/codecat
+[Antoine Boisier-Michaud]: https://github.com/Aboisier
 
 ## Version 9.13.0
 
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -270,3 +270,4 @@ Contributors:
 - Harmon <Harmon.Public@gmail.com>
 - Eric Bailey <eric.w.bailey@gmail.com>
 - Gustavo Costa <gusbemacbe@gmail.com>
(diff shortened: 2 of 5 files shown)


marcoscaceres:
@Aboisier, while we wait, would you be so kind as to add your “improvement:” to CHANGES.md just under `## Master`. Thanks in advance!

egor-rogov:
Looks just fine!

CHANGES.md
```diff
@@ -42,6 +42,7 @@ Improvements:
 - *Matlab*: transpose operators and double quote strings, by [JohnC32][] and [Egor Rogov][]
 - Various documentation typos and improvemets by [Jimmy Wärting][], [Lutz Büch][], [bcleland][]
 - *Cmake* updated with new keywords and commands by [Deniz Bahadir][]
+- *TypeScript* decorator factories highlighting fixed by [Antoine Boisier-Michaud][]
```
marcoscaceres: You might need to rebase, as 9.13.0 already shipped.

marcoscaceres:
Will get this out in a few weeks as part of 9.14.0.

Issue #1868, opened 2018-10-14
`highlightAuto` with language subset doesn't appear to be working


Martii:
Hello,

Re: 
* [/highlightjs/highlight.js/blob/`0651014`/docs/api.rst#highlightautovalue-languagesubset](https://github.com/highlightjs/highlight.js/blob/06510146c3130532f920bbb609e1e66347c80760/docs/api.rst#highlightautovalue-languagesubset) *(we use this... and thank you and your team for the longevity and expertise with this package :)*
* [https://openuserjs.org/&hellip;#comment-16615f8e2df](https://openuserjs.org/scripts/eileen12/HaremHeroes_Automatic/issues/Anyone_working_on_this_If_not,_then_Ill_try_to_figure_it_out#comment-16615f8e2df) *(one target where it's currently not auto highlighting since it's not a named fence)*
* [/OpenUserJS/OpenUserJS.org/pull/1528/files#diff-`b9cfc7f`](https://github.com/OpenUserJS/OpenUserJS.org/pull/1528/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L28) *(Dep upgrade from *highlight.js*@9.12.0 to *highlight.js*@9.13.0)*
* [/OpenUserJS/OpenUserJS.org/blob/`c5cb49c`/libs/markdown.js#L225-L260](https://github.com/OpenUserJS/OpenUserJS.org/blob/c5cb49cdf0bdb4f0df2611c824ac5da9bef4424f/libs/markdown.js#L225-L260) *(our code where it currently works in *highlight.js*@9.12.0 but __not__ *highlight.js*@9.13.0 which is currently on production. Using a limited Array of Strings.)*

Spec:
``` sh-session
$ node -v
v10.12.0
```

NOTES:
* [https://openuserjs.org/&hellip;#comment-16627b793f6](https://openuserjs.org/scripts/eileen12/HaremHeroes_Automatic/issues/Anyone_working_on_this_If_not,_then_Ill_try_to_figure_it_out#comment-16627b793f6) *(fixed this one to be a named fence to ensure correct library is being loaded and it is along with the CSS we inject... so it is currently working with named ones just not automatic detection)*.
* All languages are available for named fences however we have had to limit unnamed fences since auto-detection doesn't always pin the right language. Been through a few revisions and landed on this methodology as the best available for our use cases.
* Same issue in development. Clean package installs, etc.
* We use ES5 syntax as a default and minimal to none ES6+.

[...]

PR #1873, opened 2018-10-16
Fix autoDetection to accept language aliases

egor-rogov:
Bug introduced by #1834, found in #1868.

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -707,7 +707,7 @@ https://highlightjs.org/
   }
 
   function autoDetection(name) {
-    lang = getLanguage(name);
+    var lang = getLanguage(name);
     return lang && !lang.disableAutodetect;
   }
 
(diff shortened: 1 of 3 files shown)


src/highlight.js
```diff
@@ -707,8 +707,8 @@ https://highlightjs.org/
   }
 
   function autoDetection(name) {
-    name = (name || '').toLowerCase();
-    return !languages[name].disableAutodetect;
+    lang = getLanguage(name);
+    return lang && !lang.disableAutodetect;
```
marcoscaceres: Does lang have to be a global or closure? Could it be a local var?

egor-rogov: Yep, you're right of course.

alazurenko:
Hello guys,
Do you know how soon this fix could be shipped with a new version?

egor-rogov:
@alazurenko I guess pretty soon, as @marcoscaceres has plans to issue a minor release (for #1872).

marcoscaceres:
Yep, will get out tomorrow. It’s a little after midnight here, so want to be sure I’m around when we release in case anything goes wrong 🙏

PR #1875, opened 2018-10-16
Update crystal.js

bew:
- Adds new `annotation` keyword

Edit:
- Adds new `verbatim` keyword

/cc @crystal-lang

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@ New styles:
 
 Improvements:
 
+- Added tests & new `annotation` and `verbatim` keywords to *Crystal*, by [Benoit de Chezelles][]
+
+[Benoit de Chezelles]: https://github.com/bew
+
 ## Version 9.13.1
 
 Improvements:
--- a/src/languages/crystal.js
+++ b/src/languages/crystal.js
@@ -6,7 +6,7 @@ Author: TSUYUSATO Kitsune <make.just.on@gmail.com>
 function(hljs) {
   var NUM_SUFFIX = '(_[uif](8|16|32|64))?';
   var CRYSTAL_IDENT_RE = '[a-zA-Z_]\\w*[!?=]?';
-  var RE_STARTER = '!=|!==|%|%=|&|&&|&=|\\*|\\*=|\\+|\\+=|,|-|-=|/=|/|:|;|<<|<<=|<=|<|===|==|=|>>>=|>>=|>=|>>>|' +
+  var RE_STARTER = '!=|!==|%(?!})|%=|&|&&|&=|\\*|\\*=|\\+|\\+=|,|-|-=|/=|/|:|;|<<|<<=|<=|<|===|==|=|>>>=|>>=|>=|>>>|' +
     '>>|>|\\[|\\{|\\(|\\^|\\^=|\\||\\|=|\\|\\||~';
   var CRYSTAL_METHOD_RE = '[a-zA-Z_]\\w*[!?=]?|[-+~]\\@|<<|>>|=~|===?|<=>|[<>]=?|\\*\\*|[-/+%^&*~`|]|\\[\\][=?]?';
   var CRYSTAL_PATH_RE = '[A-Za-z_]\\w*(::\\w+)*(\\?|\\!)?';
@@ -49,9 +49,6 @@ function(hljs) {
       {begin: '%w?\\[', end: '\\]', contains: recursiveParen('\\[', '\\]')},
       {begin: '%w?{', end: '}', contains: recursiveParen('{', '}')},
       {begin: '%w?<', end: '>', contains: recursiveParen('<', '>')},
-      {begin: '%w?/', end: '/'},
-      {begin: '%w?%', end: '%'},
-      {begin: '%w?-', end: '-'},
       {begin: '%w?\\|', end: '\\|'},
       {begin: /<<-\w+$/, end: /^\s*\w+$/},
     ],
@@ -81,14 +78,6 @@ function(hljs) {
         variants: [
           {begin: '//[a-z]*', relevance: 0},
           {begin: '/', end: '/[a-z]*'},
-          {begin: '%r\\(', end: '\\)', contains: recursiveParen('\\(', '\\)')},
-          {begin: '%r\\[', end: '\\]', contains: recursiveParen('\\[', '\\]')},
-          {begin: '%r{', end: '}', contains: recursiveParen('{', '}')},
-          {begin: '%r<', end: '>', contains: recursiveParen('<', '>')},
-          {begin: '%r/', end: '/'},
-          {begin: '%r%', end: '%'},
-          {begin: '%r-', end: '-'},
-          {begin: '%r\\|', end: '\\|'},
         ]
       }
     ],
@@ -102,9 +91,6 @@ function(hljs) {
       {begin: '%r\\[', end: '\\]', contains: recursiveParen('\\[', '\\]')},
       {begin: '%r{', end: '}', contains: recursiveParen('{', '}')},
       {begin: '%r<', end: '>', contains: recursiveParen('<', '>')},
-      {begin: '%r/', end: '/'},
-      {begin: '%r%', end: '%'},
-      {begin: '%r-', end: '-'},
       {begin: '%r\\|', end: '\\|'},
     ],
     relevance: 0
--- a/src/languages/crystal.js
+++ b/src/languages/crystal.js
@@ -11,9 +11,9 @@ function(hljs) {
   var CRYSTAL_METHOD_RE = '[a-zA-Z_]\\w*[!?=]?|[-+~]\\@|<<|>>|=~|===?|<=>|[<>]=?|\\*\\*|[-/+%^&*~`|]|\\[\\][=?]?';
   var CRYSTAL_KEYWORDS = {
     keyword:
-      'abstract alias as as? asm begin break case class def do else elsif end ensure enum extend for fun if ' +
+      'abstract alias annotation as as? asm begin break case class def do else elsif end ensure enum extend for fun if ' +
       'include instance_sizeof is_a? lib macro module next nil? of out pointerof private protected rescue responds_to? ' +
-      'return require select self sizeof struct super then type typeof union uninitialized unless until when while with yield ' +
+      'return require select self sizeof struct super then typeof union uninitialized unless until when while with yield ' +
       '__DIR__ __END_LINE__ __FILE__ __LINE__',
     literal: 'false nil true'
   };
--- a/src/languages/crystal.js
+++ b/src/languages/crystal.js
@@ -13,7 +13,7 @@ function(hljs) {
     keyword:
       'abstract alias annotation as as? asm begin break case class def do else elsif end ensure enum extend for fun if ' +
       'include instance_sizeof is_a? lib macro module next nil? of out pointerof private protected rescue responds_to? ' +
-      'return require select self sizeof struct super then type typeof union uninitialized unless until when while with yield ' +
+      'return require select self sizeof struct super then type typeof union uninitialized unless until verbatim when while with yield ' +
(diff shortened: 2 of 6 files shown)


straight-shoota:
Why remove `type`?

bew:
Yeah, i completely forgot about it being a keyword, i'll change

straight-shoota:
@egor-rogov There is no comprehensive and complete list of current Crystal keywords, unfortunately.

`annotation` has been introduced in the latest release, see crystal-lang/crystal#6063

Also `verbatim` has been added as a macro keyword in the latest release (see crystal-lang/crystal#6108). @bew maybe you could add this, too?

bew:
I'm going to add a simple markup test too!

There's already this detect test file: https://github.com/highlightjs/highlight.js/blob/master/test/detect/crystal/default.txt
What is it for? (how does it work?)

egor-rogov:
@bew `default.txt` is for autodetection test. The language must be detected as crystal when given this file. Just don't touch this file (:
What you need is to create `test/markup/crystal` directory and put two files there: `something.txt` (test case) and `something.expect.txt` (expected markup for that case). Replace `something` with, well, something meaningful. You can look into any existing directory for an example.
Than run tests locally:
```
nodejs tools/build.js -t node
npm test
```

bew:
I can't get it to work..
It looks like `annotation` is not recognized as a keyword, and the path following is recognized as a weird mix of symbols...
```cr
annotation JSON::Field; end
```
Should be:
```html
<span class="hljs-keyword">annotation</span> <span class="hljs-title">JSON::Field</span>; <span class="hljs-keyword">end</span>
```
But it's recognized as:
```html
annotation <span class="hljs-symbol">JSON:</span>:Field; <span class="hljs-keyword">end</span>
```
@MakeNowJust @straight-shoota maybe you can help?

egor-rogov:
@bew Hmm, I can see that `annotation` is correctly recognized as a keyword (both in Travis log here: https://travis-ci.org/highlightjs/highlight.js/jobs/442665826 and locally with the patch applied).
The difference is in handling of the following `JSON::Field`. You expect it to be
```
<span class="hljs-title">JSON::Field</span>
```
but it's processed as
```
<span class="hljs-symbol">JSON:</span>:Field;
```
I guess you need something very similar to the definition of `class` style on lines 126-145.
I tried to simply add `annotation` keyword next to `lib enum union`. It kinda works but that way you get all the line wrapped in `<span class="hljs-class">...</span>`.

bew:
> I can see that annotation is correctly recognized as a keyword

Weird, it didn't work locally..

> I guess you need something very similar to the definition of class style on lines 126-145.

Yes I suspected that, i've been trying to find a way that won't wrap it in `hljs-class` but can't find how..

I guess it's still better than nothing though

egor-rogov:
@bew Try to add another block:
```javascript
{
      // className: 'class',    <--- comment this out
      beginKeywords: 'annotation', end: '$|;',
      illegal: /=/,
      contains: [
        hljs.HASH_COMMENT_MODE,
        hljs.inherit(hljs.TITLE_MODE, {begin: '[A-Za-z_]\\w*(::\\w+)*(\\?|\\!)?'}),
      ],
      relevance: 10
},
```
As for local changes... Make sure to rebuild the library (`nodejs tools/build.js -n`) and _reload page_ in web-browser.

bew:
Oh that might explain it, I was just running `npm test` thinking that the build was just a one time thing to setup some files.. My bad

bew:
Should be good now, thank you for your help!

egor-rogov:
Looks fine! Could you also add a test for `verbatim`?
And one more thing: please add a line into `CHANGES.md`, to Improvements section under Master.

bew:
While testing simple macros (between `{{ .. }}` or `{% ... %}`, I have a strange behavior:
```cr
{{ bla }}

{% bla %}
```
Gives the _almost_ expected behavior:
```html
<span class="hljs-template-variable">{{ bla }}</span>

<span class="hljs-template-variable">{% bla %}
</span>
```
Note the last `</span>` on a new line.

A contrario, this shows the problem:
```cr
{% bla %}

{{ bla }}
```
Gives:
```html
<span class="hljs-template-variable">{% bla %}

{{ bla }}
</span>
```
So the template variable detects the start `{%` but not the end `%}` (the last closing span is probably put because this is the end of input and there is still an opened span in the stack or something like this).

The template variable is declared as:
```js
{
  className: 'template-variable',
  variants: [
    {begin: '\\{\\{', end: '\\}\\}'},
    {begin: '\\{%', end: '%\\}'}
  ],
  keywords: CRYSTAL_KEYWORDS
}
```
Is there something wrong here?

bew:
Note: I rebased and fixed conflicts with new master

egor-rogov:
Ok, thanks. I'll have a look at macros later.

egor-rogov:
@bew I've found what's wrong. This definition:
```javascript
  var REGEXP = {
    begin: '(' + RE_STARTER + ')\\s*',
```
matches `%` alone and hence prevents `%}` to be recognized as the end of a macro.
I can't get why there are two definitions for regexps (REGEXP and REGEXP2).
Am I right that regexps are denoted by either `/.../` or `%r(...)` (with some variants for braces)?
I would try to remove REGEXP2 completely and rewrite REGEXP in the following manner:
```javascript
  var REGEXP = {
    className: 'regexp',
    contains: [hljs.BACKSLASH_ESCAPE, SUBST],
    variants: [
      {begin: '//[a-z]*', relevance: 0},
      {begin: '/', end: '/[a-z]*'},
      {begin: '%r\\(', end: '\\)', contains: recursiveParen('\\(', '\\)')},
      {begin: '%r\\[', end: '\\]', contains: recursiveParen('\\[', '\\]')},
      {begin: '%r{', end: '}', contains: recursiveParen('{', '}')},
      {begin: '%r<', end: '>', contains: recursiveParen('<', '>')},
      {begin: '%r/', end: '/'},
      {begin: '%r%', end: '%'},
      {begin: '%r-', end: '-'},
      {begin: '%r\\|', end: '\\|'},
    ],
    relevance: 0
  };
```
This of course requires a test for regexps, too...

straight-shoota:
`%r//`, `%r--` and `%r%%` are not valid regex literals in Crystal. These three alternatives can be removed.

MakeNowJust:
`%r//` and variants are my fault. Sorry. Please remove them.

However `REGEXP`'s `begin` condition is important. Crystal and many languages have the *slash* problem: lexer cannot determine `/` denotes a division operator or regexp literal delimiter. So, this `begin` condition is one of the solutions on `highlight.js`. Actually JavaScript and Ruby use the same trick.
For example, `1 / 2 / 3` should be highlighted to:

```html
<span class="hljs-number">1</span> / <span class="hljs-number">2</span> / <span class="hljs-number">3</span>
```

However, it does not work if `begin` is dropped. (But now, I think it should use `hljs.RE_STARTERS_RE` instead of my own. It is not defined when I write `crystal.js`)

egor-rogov:
@MakeNowJust, thanks for your explanation. Can you give an example of a regex that follows`%`? Is it possible in Crystal?

straight-shoota:
`"foo %s" % /bar/"` calls `#%` on `"foo"` with argument `/bar/` (resulting in `"foo (?-imsx:bar)"`). White space is optional.

But I don't think that's actually the issue, because `%/.../` is not a regex literal. It's rather that two `/` with some identifier and whitespace in between could either delimit a regular expression or calls to `#/` (division operator). With the (new) `#//` operator, two consecutive `/` could be either an empty regex or an operator, depending on context.

bew:
I'm quite lost now..
I tried to add some tests for regex, but it simply doesn't recognize anything (even though on github regexes are recognized).

It's getting bigger than I originally thought, I won't have time to debug the whole thing. @straight-shoota and/or @MakeNowJust if you have time and want to give it a try, I can give you access to my fork to continue this branch

egor-rogov:
Hi @bew. It seams to me that the following will do the trick:

1. In `RE_STARTER` add negative lookahead for closing curly brace to `%` , i. e. replace `%` with `%(?!})`. This way we'll fix macros and won't touch regexps.
2. In `var REGEX` block remove variants starting with `%r` (they are defined in `REGEX2` and don't require the starter magic).
3. In var REGEX2` block remove invalid variants mentioned by @straight-shoota and @MakeNowJust.

Can you try it?

straight-shoota:
@egor-rogov Yes, that makes it pass.

```diff
diff --git c/src/languages/crystal.js i/src/languages/crystal.js
index 793fb41..c18d37a 100644
--- c/src/languages/crystal.js
+++ i/src/languages/crystal.js
@@ -6,7 +6,7 @@ Author: TSUYUSATO Kitsune <make.just.on@gmail.com>
 function(hljs) {
   var NUM_SUFFIX = '(_[uif](8|16|32|64))?';
   var CRYSTAL_IDENT_RE = '[a-zA-Z_]\\w*[!?=]?';
-  var RE_STARTER = '!=|!==|%|%=|&|&&|&=|\\*|\\*=|\\+|\\+=|,|-|-=|/=|/|:|;|<<|<<=|<=|<|===|==|=|>>>=|>>=|>=|>>>|' +
+  var RE_STARTER = '!=|!==|%(?!})|%=|&|&&|&=|\\*|\\*=|\\+|\\+=|,|-|-=|/=|/|:|;|<<|<<=|<=|<|===|==|=|>>>=|>>=|>=|>>>|' +
     '>>|>|\\[|\\{|\\(|\\^|\\^=|\\||\\|=|\\|\\||~';
   var CRYSTAL_METHOD_RE = '[a-zA-Z_]\\w*[!?=]?|[-+~]\\@|<<|>>|=~|===?|<=>|[<>]=?|\\*\\*|[-/+%^&*~`|]|\\[\\][=?]?';
   var CRYSTAL_PATH_RE = '[A-Za-z_]\\w*(::\\w+)*(\\?|\\!)?';
@@ -81,14 +81,6 @@ function(hljs) {
         variants: [
           {begin: '//[a-z]*', relevance: 0},
           {begin: '/', end: '/[a-z]*'},
-          {begin: '%r\\(', end: '\\)', contains: recursiveParen('\\(', '\\)')},
-          {begin: '%r\\[', end: '\\]', contains: recursiveParen('\\[', '\\]')},
-          {begin: '%r{', end: '}', contains: recursiveParen('{', '}')},
-          {begin: '%r<', end: '>', contains: recursiveParen('<', '>')},
-          {begin: '%r/', end: '/'},
-          {begin: '%r%', end: '%'},
-          {begin: '%r-', end: '-'},
-          {begin: '%r\\|', end: '\\|'},
         ]
       }
     ],
@@ -102,9 +94,6 @@ function(hljs) {
       {begin: '%r\\[', end: '\\]', contains: recursiveParen('\\[', '\\]')},
       {begin: '%r{', end: '}', contains: recursiveParen('{', '}')},
       {begin: '%r<', end: '>', contains: recursiveParen('<', '>')},
-      {begin: '%r/', end: '/'},
-      {begin: '%r%', end: '%'},
-      {begin: '%r-', end: '-'},
       {begin: '%r\\|', end: '\\|'},
     ],
     relevance: 0
```

I also have a few other fixes, but I can open a new PR for this.

bew:
Tests are passing now, thanks to all for your time :pray:

egor-rogov:
Finally merged. Thanks everybody!

bew:
If I understood correctly, this project is used for the highlighting on github, do you know when these changes will be visible on github? (at next release?) Or does github simply pulls this repo once in a while? thanks

MakeNowJust:
@bew I think GitHub uses TextMate and Atom syntax file. You can check it [here](https://github.com/github/linguist/tree/master/vendor). So if you want to improve Crystal syntax highlighting on GitHub, you should fix [this](https://github.com/crystal-lang-tools/language-crystal) instead of highlight.js.

bew:
Ah lol, grrr @straight-shoota !! 
At least highlightjs is fixed now ^^

PR #1876, opened 2018-10-16
Update css-classes-reference.rst

Martii:
Add default non-aliased language as the first item for XQuery to match the rest of the doc. e.g. `xquery`

Post #1868


marcoscaceres:
Thanks @Martii!

PR #1878, opened 2018-10-17
Changes markdown fix

isagalaev:
It's a one-line change, but my editor also realized that the file had Windows line endings, so I fixed that in a separate commit.


marcoscaceres:
My bad. I let this through. 

Would adding `* text=auto` to `.gitattributes` prevent this from happening again? 

https://help.github.com/articles/dealing-with-line-endings/

isagalaev:
Isn't it already auto be default? We have this setting https://github.com/highlightjs/highlight.js/blob/master/.editorconfig#L2, so I don't expect this problem to appear very often anyway.

EricFromCanada:
Add an [.editorconfig](https://editorconfig.org) file.

PR #1883, opened 2018-10-19
Add missing dockerfile markup tests

l-vo:
As requested in #1771  :)

PR #1884, opened 2018-10-19
Fix change added in the wrong version

l-vo:
In the PR #1771, I added my changes in the already released version 9.13.1 instead of master.

I'm sorry for that, this fixes that mistake.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,8 +6,10 @@ New styles:
 
 Improvements:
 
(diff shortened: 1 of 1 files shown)


egor-rogov:
Hi @l-vo, whould you mind changing #1883 instead of a separate PR?

l-vo:
@egor-rogov done :)

PR #1886, opened 2018-10-20
Additional fixes for Crystal language specification

straight-shoota:
This fixes a few issues with the language definition.

* Removes invalid string and regex literals like `%-string-`.
* Adds `%Q` string literals and `%i` symbol array literals.
* Fixes number literals can have underscores almost anywhere.
* Fixes number literals separate int and float suffixes.
* Fixes recognition of regex literals.

There are still a few issues, tough, that I can't find a fix for.

**Regex literals with `%r`**
```cr
%r((/))
%r[[/]]
%r{{/}}
%r<</>>
%r|/|
```
Should produce this:
```html
<span class="hljs-regexp">%r((/))</span>
<span class="hljs-regexp">%r[[/]]</span>
<span class="hljs-regex">%r{{/}}</span>
<span class="hljs-regexp">%r&lt;&lt;/&gt;&gt;</span>
<span class="hljs-regexp">%r|/|</span>
```
Actual result:
```html
%r((<span class="hljs-regexp">/))
%r[[/</span>]]
%r<span class="hljs-template-variable">{{/}}</span>
%r&lt;&lt;<span class="hljs-regexp">/&gt;&gt;
%r|/</span>|
```

/cc @bew @MakeNowJust


MakeNowJust:
@straight-shoota It is the general highlight.js (and regex based syntax highlighter) limitation. JavaScript, Ruby and some languages, which use slash as both meaning regexp and operator, have the same problem. Regexp literal at line beginning is rare case, so it is no problem.

src/languages/crystal.js
```diff
@@ -4,11 +4,12 @@ Author: TSUYUSATO Kitsune <make.just.on@gmail.com>
...
 
 function(hljs) {
-  var NUM_SUFFIX = '(_[uif](8|16|32|64))?';
+  var INT_SUFFIX = '(_*[ui](8|16|32|64))?';
+  var FLOAT_SUFFIX = '(_*f(32|64))?';
   var CRYSTAL_IDENT_RE = '[a-zA-Z_]\\w*[!?=]?';
-  var RE_STARTER = '!=|!==|%(?!})|%=|&|&&|&=|\\*|\\*=|\\+|\\+=|,|-|-=|/=|/|:|;|<<|<<=|<=|<|===|==|=|>>>=|>>=|>=|>>>|' +
-    '>>|>|\\[|\\{|\\(|\\^|\\^=|\\||\\|=|\\|\\||~';
-  var CRYSTAL_METHOD_RE = '[a-zA-Z_]\\w*[!?=]?|[-+~]\\@|<<|>>|=~|===?|<=>|[<>]=?|\\*\\*|[-/+%^&*~`|]|\\[\\][=?]?';
+  var RE_STARTER = '!=|!==|%(?!})|%=|&|&&|&=|\\*|\\*=|\\+|\\+=|,|-|-=|/=|:|;|<<|<<=|<=|<|===|==|=|>>>=|>>=|>=|>>>|' +
```
MakeNowJust: Removing this and using `hljs.RE_STARTERS_RE` are preferred. Ruby grammar uses it.

egor-rogov:
@straight-shoota That's because `%` in `%r...` is recognized as a starter for REGEX.
Change `%(?!})` to `%(?![}r])`, this seems to fix the problem.

src/languages/crystal.js
MakeNowJust: Q&A:

> Why `hljs.RE_STARTERS_RE` is preferred?

It is standard way and good enough.

> However, it reproduce missing macro ending problem like previous PR.

It can prevent by following:

```javascript
    begin: '(?!%})(' + hljs.RE_STARTERS_RE + ')\\s*',
```

- - -

I believe it is no problem using `hljs.RE_STARTERS_RE` instead.

egor-rogov:
> `REGEXP2` puts before `REGEXP` then get right result. I think it is more efficient than @egor-rogov opinion. JavaScript regexp alternation operator works ordered from left to right.

Agree.

src/languages/crystal.js
straight-shoota: `hljs.RE_STARTERS_RE` includes `%` which messes up macro expansions. We need to have `%(?!})` instead. See https://github.com/highlightjs/highlight.js/pull/1875#issuecomment-431299162

egor-rogov: Doesn't the solution proposed by @MakeNowJust work?

> It can prevent by following:
>
>        begin: '(?!%})(' + hljs.RE_STARTERS_RE + ')\\s*',

egor-rogov:
@straight-shoota Looks fine!
One more question: what is `literals.out.txt`file for?
And please add a line in `CHANGES.md`.

straight-shoota:
Oh, I commited that file accidentally. I just used it for diffing against the expected result because the test output is really bad for this.

Changes will be added.

egor-rogov:
Thanks for the fixes and tests! Merged.

PR #1887, opened 2018-10-20
Added missing semicolons

reviewher:


marcoscaceres:
Thanks @reviewher! Eventually we may end up prettying up the code a bit more.

reviewher:
@marcominetti definitely would make sense to add a linter to help standardize (something like ESLint).  The missing semicolons showed up when running JSHint on `src/highlight.js`.

PR #1888, opened 2018-10-21
Support external language files in minified version of highlight.js

romanresh:
**Background**
The external language files are incompatible with the minified version of the highlight.js library. It happens because of the "replace" stage of the "minify" task. It globally replaces a few API members in the highlight.pack.js file (using the REPLACES dictionary in the tools/utility file). It leads to 2 problems which don't allow to use external language files with the minified library:

**Problem 1. Null Reference Exception**
The null reference exception occurs if an external language file calls the hljs API members which are minified in the hljs object (hljs.COMMENT(), hljs.NUMBER_MODE, etc). To prevent the exception the PR changes the regex and "replace" function used by the minify task. It allows to save both minified and unminified versions of API members in the hljs object. For instance, the following code:
```js
hljs.COMMENT = function (begin, end, inherits) { ... }
```
was replaced with (without PR fix):
```js
hljs.c = function (begin, end, inherits) { ... }
```
it will be replaced with (with PR fix):
```js
hljs.c = hljs.COMMENT = function (begin, end, inherits) { ... }
```

**Problem 2. An external language cannot be compiled because its properties aren't minified**
The compileLanguage() function of the minified hljs library cannot process a language object contained unminified properties names (begin, end, returnBegin, variants, etc). The PR introduces the restoreLanguageApi function which preprocesses the language file and copies unminified properties to corresponding minified properties. It uses the API_REPLACES variable which is assigned to the utility.REPLACES object via the minify task of the build tool.


romanresh:
Hi @marcoscaceres @isagalaev @egor-rogov . Any news with this PR? Would you review it?

marcoscaceres:
Sorry for the delay... trying to get to it.

marcoscaceres:
I’m willing to go with this but I guess eventually we need to rewrite the core library as an ES6 module.

isagalaev:
For historical perspective, the idea for this customized minification with changing objects property names stems from the time when I was trying to keep the entire library with `:common` languages under 30K. By now we're past that point, and I think we could remove this complexity entirely and sacrifice the final package size for the sanity of the API.

marcoscaceres:
Agree - and that's not to say that what the library did originally is not amazing. But now that we have great packagers/minifiers and a module system for JS, it would be great to do this. However, it's going to be a lot of work so need to figure out some way to do it incrementally.

PR #1891, opened 2018-10-25
Make clean.js less greedy

isagalaev:
- Removed very common cross-language symbols from the list of
  relevance boosters (`::`, `..`, back-tick).
- Added built-in types to increase relevance in a natural way.

(Found it when a random Rust fragment got auto-detected as Clean.)

### DIFF
--- a/src/languages/clean.js
+++ b/src/languages/clean.js
@@ -14,6 +14,8 @@ function(hljs) {
         'implementation definition system module from import qualified as ' +
         'special code inline foreign export ccall stdcall generic derive ' +
         'infix infixl infixr',
+      built_in:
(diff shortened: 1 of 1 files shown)


isagalaev:
Good idea. However it's tricky because the snippet in question was not a fully valid Rust as it used `...` to fill in the blanks:

```
unsafe fn foo_avx(...) {
    let ... = _mm256_add_ps(..., ...);
}
```

Clean.js grabbed all of the dots for itself as a relevance booster and won. Since using the dots is an extremely common practice, I usually try not to make it invalid. However putting something like that into a default language detection snippet feels wrong.

I'll leave it as is for now and may be someone will come up with a working idea later.

PR #1893, opened 2018-10-27
Improve Crystal language to highlight regexes after some keywords

MakeNowJust:
Highlight following as Crystal:

```
if /foo/ =~ foo
end
```

Before:

```html
<span class="hljs-keyword">if</span> /foo/ =~ foo
<span class="hljs-keyword">end</span>
```

After:

```html
<span class="hljs-keyword">if</span> <span class="hljs-regexp">/foo/</span> =~ foo
<span class="hljs-keyword">end</span>
```

/cc @straight-shoota @bew


egor-rogov:
Starters again...
Merged!

PR #1897, opened 2018-10-30
cpp: Fully support C++11 raw strings.

davidben:
See https://en.cppreference.com/w/cpp/language/string_literal for the syntax. This requires a fix in highlight.js itself.

mode.terminators joins each node's begin regexps with |. This breaks if one of the begin regexps has backreferences. Backreferences count capturing parenthesized groups, and adding new groups in front will change that count.  Thus far, the only language that uses backreferences is Rust (also for raw strings), which happens to be the first in the list and avoids this bug. C++ cannot as easily avoid this because, even were raw strings the first option in STRINGS, STRINGS itself is included in other lists.

Rather than carefully order things, rewrite the regularly expressions to fix the backreferences.


egor-rogov:
> The backreferences trick only lets you mark one element, so you can't highlight interior bits.

Well, actually you can with help of `returnBegin` and nested `contains`... but I agree that each method have its own drawbacks. I need some time to think about it. Maybe we can come up with something more general and clear.

egor-rogov:
It seems that actually we need a way to use backreference from the `end` regexp to a capturing group in the `begin` regexp.
I'm doubtful about the syntax, but lets assume that `\0n` denotes a backreference to n-th group in `begin`. Then for C++ you'll write something like
```javascript
{ begin: /(?:u8?|U|L)?R"([^()\\ ]{0,16})\(/,
  end: /\)\01"/
}
```
and for PostgreSQL
```javascript
{ begin: '(' + DOLLAR_STRING + ')',
  end: /\01/
}
```
This looks like a general enough mechanism that suits both our cases well. What do you think?

davidben:
Inventing our own regexp syntax while still fitting into JS regexps feels off. `endSameAsBegin` is a fairly concrete understandable feature. `joinRe` is rewriting regexps, but it's arguably just a bugfix to retain existing JS semantics.

TBH, I doubt use cases that don't fit in the mechanisms' intersection will ever come up. Requiring duplicated tokens like this isn't common.

davidben:
(Also I believe \0 matches a nul character.)

egor-rogov:
So be it.
I've spent some time with `joinRe` and it seems quite solid to me now, but its logic is not the simplest one. I think its worth adding some tests for it.

marcoscaceres:
@davidben, could you please add a couple of tests?

egor-rogov:
@marcoscaceres It is useful and it doesn't break things, so I guess we can.

yyyc514:
@davidben Thanks, this code saved me a LOT of time when I was fixing the look-ahead functionality. :-)

Issue #1722, opened 2018-04-04
Warning about line continuation when using Google Closure compiler


p-himik:
https://github.com/isagalaev/highlight.js/blob/master/src/languages/tp.js#L20 uses line continuation instead of direct string concatenation, and Google Closure doesn't like it:

```
WARNING - String continuations are not recommended. See https://google.github.io/styleguide/jsguide.html#features-strings-no-line-continuations
```

cjsauer:
I am also seeing this warning. Using this library from ClojureScript.

```
node_modules/highlight.js/lib/languages/tp.js:14: WARNING - String continuations are not recommended. See https://google.github.io/styleguide/jsguide.html#features-strings-no-line-continuations
```

PR #1916, opened 2018-11-28
String continuations are not recommended by google closure compiler

### LINKED ISSUE #1770: String continuations are not recommended.
Style issue. `String continuations are not recommended.` warning thrown by Google Closure Compiler

https://github.com/highlightjs/highlight.js/blob/b18af64d744e067cce0469d283d6b7c089b0162d/src/languages/tp.js#L20

Only file it complains about (TP language), so I would think easy "fix" in case you'd want to comply. But as it is a style thing, figured I'd post it here instead of making a pull request. Shoot down if you feel this lacks importance.

souenzzo:
https://google.github.io/styleguide/jsguide.html#features-strings-no-line-continuations
Fixes:  #1770 #1722 .

### DIFF
--- a/src/languages/tp.js
+++ b/src/languages/tp.js
@@ -17,8 +17,8 @@ function(hljs) {
   };
   var TPDATA = {
     className: 'built_in',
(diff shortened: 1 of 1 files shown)


souenzzo:
@marcoscaceres please close #1722

marcoscaceres:
Done! Thanks

PR #1913, opened 2018-11-25
Add SuperCollider language

brianlheim:
For whatever reason adding these files caused Swift to be misdetected as Kotlin when I ran tests locally, so I tried to boost relevance for Swift by emphasizing optional-using types (!, ?). This means Swift is highlighted differently. If that's not acceptable please let me know how you'd like to resolve that.

PR #1919, opened 2018-11-30
Swift: be eager on optional-using types (!/?)

brianlheim:
Swift and Kotlin are currently very close on detection, this commit makes Swift a little more eager to recognize optional types.

See also \#1913


brianlheim:
Thanks!

PR #1934, opened 2018-12-25
Correct for Fortran literal number

kookma:
In modern fortran you can define the **kind** of intrinsic data type. It means you can define additional intrinsic data type. When you do so, for literal constants you have to use the kind name attached to literal consonant using under score (like a prefix `_newtype`).
Example:
 a = 12.23_dp

# How the current code works
When using the current highlight.js to highlight a Fortran code contains such named constant, highlight.js fails to do the job! It highlights the remaining code (from the position of named constant) in a wrong way!
current_bug

Note to line 4. The literal number has a `_DBL` suffix and it makes the current code breaks. The remaining highlighting is wrong.

# The Corrected code
The below image shows the corrected code output. As you can see the literal constant is highlighted correctly and there is no more error in the following texts.

corrected_code

Note to line four and following lines. The literal constants are highlighted correctly. Also note to line 5 the `WRITE` keyword in two image.

### DIFF
--- a/src/languages/fortran.js
+++ b/src/languages/fortran.js
@@ -66,9 +66,9 @@ function(hljs) {
         contains: [hljs.UNDERSCORE_TITLE_MODE, PARAMS]
       },
       hljs.COMMENT('!', '$', {relevance: 0}),
-      {
+{
         className: 'number',
-        begin: '(?=\\b|\\+|\\-|\\.)(?=\\.\\d|\\d)(?:\\d+)?(?:\\.?\\d*)(?:[de][+-]?\\d+)?\\b\\.?',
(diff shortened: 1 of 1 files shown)


kookma:
@egor-rogov 
I revised the fortran.js. Now the last period is part of number and starting plus/minus signs are not

kookma:
@egor-rogov  
Please let me know if it fine now. I have added many tests.

fortran-highlight-test

egor-rogov:
@kookma 
Probably you did't push tests? I can't see them in this PR.
Also, why change `className` to `cN` etc?

kookma:
@egor-rogov 
It seems something goes wrong! Can I delete this PR and create a new?
I am not very familiar with github system.

By the way I did not changed the cN class, but as you said in the `files changed` panel I see it has been changed! I just changed the content of regexp string.

egor-rogov:
@kookma Sure you can create a new PR.

PR #1955, opened 2019-01-27
Perhaps it would be better.

iliyaZelenko:


marcoscaceres:
I appreciate this, but I'm worried about moving into this structure because there might be some use pf the `this` within the `highlightBlock` method  (which would break). So prefer to keep current code.

Eventually, if we modernize the code into an ES Module, what you propose will make total sense... we can cleanly `import highlightBlock from "hljs";` and use it like you propose. 

Hope that's ok!

PR #1932, opened 2018-12-24
Add missing "object" highlighting to Pony definition

SeanTAllen:
The original Pony highlighting definition that Joe did is missing a Pony
keyword for highlighting like a class. After consulting with other Pony
committers and comparing results, we feel this small change is a big
improvement when highlighting Pony's object literals.

PR #1958, opened 2019-01-28
Pony markup tests

gregcline:
Based on the discussion in #1932 I added some markup tests for Pony.

I also made a couple adjustments to highlighting for classes and functions. I added `primitive` to the list of beginning keywords for class highlighting and made the class name highlight the same as other type names. `is` is also now highlighted as a keyword in class definitions.

The only change I made to functions was including `be` in the beginning keywords.


egor-rogov:
Thanks @gregcline, this looks fine for me.
@SeanTAllen, could you also have a look before I merge?

egor-rogov:
Okay then. Thanks, merged!

PR #1972, opened 2019-02-08
Rename lambda.expect.xt to lambda.expect.txt

allejo:


marcoscaceres:
well spotted! thanks @allejo.

PR #1979, opened 2019-02-23
Fix a few highlighting errors and remove two modes

gregcline:
Fixes a few highlighting bugs such as keyword lists not having spaces at
the end causing the last keyword of one line and the first keyword of
the next to be combined. Also fixed highlighting of `iso` in class
definitions and function heads without bodies in traits and interfaces.

This also removes the FUNCTION and CLASS modes until there is a more
clear benefit to using them. In a couple cases such as an interface
with multiple function heads without an `=>` indicating a body they
broke the highlighting of the second function head e.g.
```pony
interface Iterator[A: A]
  fun has_next(): Bool
  fun next(): T?
```


marcoscaceres:
Thanks for fixing those. Will try to do a release with the fixes today or tomorrow.

PR #1981, opened 2019-02-25
fix(package): upgrade jsdom to 11.12.0

marcoscaceres:
This is the last version to support the "old API".


marcoscaceres:
It's only used by the test suite, so it should be fine. I tested all version up to this point.

PR #2000, opened 2019-03-05
Update JSX detection to match XML syntax match

kynatro:
Updates the JSX detection regex pattern in the JavaScript language file to use the same regex pattern as the XML language file. This is to accommodate for scenarios where JSX is implementing Web Components which may have tag name structures including special characters that `\w` cannot match.


marcoscaceres:
This is great! but would you mind adding a few tests? I don’t have much experience with jsx, so unsure what to expect.

kynatro:
Added a couple tests

PR #2002, opened 2019-03-07
add new lenguage - terraform

WinOpsDBA:
A new language, please have a look and let me know if anything is missing.


WinOpsDBA:
I could do with some help.
JS is not my cup of tea and I have stuck on testing.
Please help.

marcoscaceres:
Hi @WinOpsDBA, we are in the process of moving new languages to their own repos. If I made you a repository under the hljs organization, would you be willing to maintain this language?  

Here is what a typical repository would look like: 
https://github.com/highlightjs/highlightjs-solidity

And then you can publish the language on npm for people to use.

WinOpsDBA:
Hi @marcoscaceres, I can do language syntax etc. however I am new to js and will require lots of learning and help ;)
I am up for the challenge. Let's give it a go and will see.

marcoscaceres:
Looks like what you have there is ok so far. All you really need in the repository in the terraform.js file. Testing is, of course, nice to have. 

Doing a testing setup should be fairly straight forward. I'll see if I can find time to do it from one of the other languages, and the you can just reuse everything you've already got here. 

I'd suggest to start just moving everything over... then we put all the bits together.

WinOpsDBA:
Sounds good, please send me details for a repository and advise which one should I use as a reference.

marcoscaceres:
Done: https://github.com/highlightjs/highlight-terraform/invitations

WinOpsDBA:
Hi @marcoscaceres I have done my 1st commit when you have a second could you have a look, please?
Could you let me know how to test it etc.
I am on uncharted waters here so any feedback is much appreciated.

marcoscaceres:
Sent :) Have another PR for linting the code, but will send that one after as it depends on the first.

PR #2012, opened 2019-03-31
Fix continuation for JS/TS regexps.

### LINKED ISSUE #1962: (javascript) syntax highlighter is confused by regexp literals
js like:

```js
      hash += (hash === '#' ? '' : '&').replace(
          /(^#|&)test=[^&]*(&test=[^&]*)*&?/g,
          '$1') + 'test=' + encodeURIComponent(testName);
```

is not correctly being highlighted under highlight.js see https://chromium-review.googlesource.com/c/chromium/src/+/594890/22/third_party/WebKit/LayoutTests/external/wpt/async_cookies/resources/testharness-helpers.js#35 (line 35) and see bug report at https://bugs.chromium.org/p/gerrit/issues/detail?id=7296

rehmsen:
Verified this works in `tools/developer.html`, but did not manage to runs the tests with `npm test` (complains about ../../build not being there even though it is there...). Hopefully CI can check if the other tests and the new one I added pass.

Fixes #1962.

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -91,6 +91,7 @@ function(hljs) {
       },
       { // "value" container
         begin: '(' + hljs.RE_STARTERS_RE + '|\\b(case|return|throw)\\b)\\s*',
+        endsWithParent: true,
         keywords: 'return throw case',
         contains: [
           hljs.C_LINE_COMMENT_MODE,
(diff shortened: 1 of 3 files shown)


rehmsen:
The failing test is for auto-detection of JS. I tried to understand how my change could have influenced that, but it's not very clear to me: The mode I am changing has `relevance: 0`. I read the documentation at https://highlightjs.readthedocs.io/en/latest/building-testing.html and it says to "refer to the discussion group", which I have been unable to find.

I will have a closer look later - any advice and ideas welcome.

rehmsen:
I took a closer look and my PR breaks the javascript/default.txt. I will try something else to fix the issue I was looking into and reopen.

PR #2017, opened 2019-04-13
Adding aliases

RachamimYaakobov:
New aliases: 'hh', 'hxx', 'cxx'

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -95,7 +95,7 @@ function(hljs) {
   ];
 
   return {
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Could you help me understand the rationale for adding these? Are they part of c++ standard or common file extensions?

RachamimYaakobov:
There is a lot of C ++ code written in these files
When I extract from such a file several lines highlight.js does not recognize the language
And you have to tell him explicitly that this is C ++, so I just do it through the file extension,
Therefore, highlight.js must know all the extensions.

marcoscaceres:
Ok, did some additional research and indeed they are commonly used.

PR #2020, opened 2019-04-16
Adding c++ keywords

RachamimYaakobov:
New keywords: __stdcall, __try, __finally.

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -73,7 +73,7 @@ function(hljs) {
       'atomic_bool atomic_char atomic_schar ' +
       'atomic_uchar atomic_short atomic_ushort atomic_int atomic_uint atomic_long atomic_ulong atomic_llong ' +
       'atomic_ullong new throw return ' +
-      'and or not',
+      'and or not __stdcall __try __finally',
     built_in: 'std string cin cout cerr clog stdin stdout stderr stringstream istringstream ostringstream ' +
       'auto_ptr deque list queue stack vector map set bitset multiset multimap unordered_set ' +
       'unordered_map unordered_multiset unordered_multimap array shared_ptr abort abs acos ' +


src/languages/cpp.js
```diff
@@ -73,7 +73,7 @@ function(hljs) {
       'atomic_bool atomic_char atomic_schar ' +
       'atomic_uchar atomic_short atomic_ushort atomic_int atomic_uint atomic_long atomic_ulong atomic_llong ' +
       'atomic_ullong new throw return ' +
-      'and or not',
+      'and or not __stdcall __try __finally',
```
marcoscaceres: I think any __* is classified as a keyword, so you might need to update the keywords regex at the top of the file instead. Please also be sure to include a test.

RachamimYaakobov: What do you mean? Turn me to a more specific place. I not found keywords regex.

egor-rogov: The point is that (according to https://en.cppreference.com/w/cpp/keyword):

> all identifiers that contain a double underscore __ in any position and each identifier that begins with an underscore followed by an uppercase letter is always reserved

Hence it's make sense to highlight all such identifiers as keywords rather than add them one by one.

Technically you'll need to add another submode for such keywords, see CPP_PRIMITIVE_TYPES for an example.

RachamimYaakobov: @egor-rogov Thank you for the explanation, but unfortunately I do not know how to do what you said. If you can do this, thank you.

yyyc514:
I took a quick look at this and it's tricky because I think to do this with a regex you'd need to not only match the variants (_Capital, __all_lower) [easy] but you'd also need to make sure that everywhere KEYWORDS can currently be matched (a bunch of child scopes) that you also add your custom matchers to `contains`...

Maybe it's just that simple, but I'd be worried about engineering a proper test case... or maybe you just cover the most common cases, and "trust" that the rest work?  That sounds scary though.

Highlight.js currently makes this easy to do with KEYWORDS, but a lot more difficult (by comparison) for custom matchers.  Unless I'm missing something.

yyyc514:
We also already have precedent for _Cap keywords just begin added to keywords:

```
_Bool complex _Complex _Imaginary
```

If we were going to do that though this PR feels far too minimal.  Someone would have to actually make an attempt to gather a real list of __ keywords and then add them all, not just the 3 we noticed were missing... and perhaps now we're right back to figuring out how to make this a role. :-)

yyyc514:
Oh, boy with `xl.js` we already have a precedent for using regex inside of keywords... would this be another case where that might make sense:

```js
    keyword: 'int float while private char catch import module export virtual operator sizeof ' +
      ...
      'noexcept static_assert thread_local restrict _Bool complex _Complex _Imaginary ' +
```

After (or something similar).  We might want to use constants then add them on the end, but the effect would be the same.

```js
    keyword: 'int float while private char catch import module export virtual operator sizeof ' +
      ...
      'noexcept static_assert thread_local restrict _[A-Z][a-z\\d_]+ __[a-z\\d_]+' +
```

Not sure we want to encourage this but AFAICT it would work, as long as we stay away from `|` which is reserved for keyword relevance.

Thoughts?

egor-rogov:
I'd prefer a separate mode, something like
```js
{ className: 'keyword',
  begin: '_[A-Z][a-z\\d_]+ __[a-z\\d_]+' 
}
```
not to depend on how keyword flattener works.

yyyc514:
> I'd prefer a separate mode, something like

If so someone now has to repeat that within every scope in the syntax file, yes?  So if KEYWORDS is used 50 times, then they'd need to add that rule to all 50 place also, correct?

That's why I wondered if there wasn't some better way.

egor-rogov:
> this seems like kind of a deficiency of the keyword system.

I think so.
Not very severe deficiency I'd say, 'cause you still have to repeat `keywords` 50 times, right?

>>    not to depend on how keyword flattener works.

>    Is this the only concern? So if we resolved that concern we'd be good? :)

If you can come up with something, It'd be great.

yyyc514:
Since we have no plans to merge this as-is I'm going to close this in favor of opening up a new issue to track the discussion.

PR #2029, opened 2019-04-30
Make sure exports is not a HTML element

jorik90:
Having an object with the id `exports` (like `<h1 id="exports">`) will make `exports` defined ([see HTML spec](https://html.spec.whatwg.org/#named-access-on-the-window-object)), thereby breaking the check to see if the library is running inside nodeJS and not assigning `hljs` to the window object.

This fixs checks if `exports` is a HTML-node. If it is, it'll act like `exports` is `undefined`. This is the same solution as [underscorejs](https://github.com/jashkenas/underscore/blob/master/underscore.js#L51-L54) and [observejs](https://github.com/googlearchive/observe-js/pull/82/files) have.


src/highlight.js
```diff
@@ -11,7 +11,8 @@ https://highlightjs.org/
 
   // Setup highlight.js for different environments. First is Node.js or
   // CommonJS.
-  if(typeof exports !== 'undefined') {
+  // `nodeType` is checked to ensure that `exports` is not a HTML element.
+  if(typeof exports !== 'undefined' && !exports.nodeType) {
```
marcoscaceres: I wonder if there is a more robust brand check we can do? 

Object.prototype.toString.call(exports).includes("HTML")

Maybe not... not sure if anyone is using this with XML nodes. 🤔

jorik90: I think this is pretty robust (and it's the same as underscorejs is using). In any case where `exports` is defined and has the `nodeType` on the object, it should not be detected as being nodejs or commonJS. What do you mean with XML nodes? Higlighting XML is not affected, as this is only the initialising of the script. Running this inside XML should be fine as well, as that way the nodeJS way of initialising shouldn't be used. Does that answer your question?

Issue #2032, opened 2019-05-08
styles/school-book.css applies style to all <pre> (element selector, no classes)


tincho:
Hi there!
I'm facing a case where my bundle un-intendenly included **all** the css files from hljs/styles (that's something I ought to fix, yes, but...)

Unlike all other .css files, `school-book.css` [has a block defining styles for all `pre` elements, without any specificity](https://github.com/highlightjs/highlight.js/blob/160a57192755d6b620d896e9877abd54522c8ca7/src/styles/school-book.css#L15)

That's causing that, without initializing hljs in any way, all `<pre>` blocks are getting that yellow background image.

marcoscaceres:
closed via https://github.com/highlightjs/highlight.js/pull/2034

marcoscaceres:
thanks for your help on this @tincho!

PR #2034, opened 2019-05-13
fix(styles/school-book): don't style all the pre, use .hljs instead

tincho:
Regarding issue #2032 
What do you think about this fix? I tried changing the CSS from browser DevTools in the HLJS demo and seemed to work just as fine.


src/styles/school-book.css
```diff
@@ -10,9 +10,12 @@ School Book style from goldblog.com.ua (c) Zaripov Yura <yur4ik7@ukr.net>
   padding: 15px 0.5em 0.5em 30px;
   font-size: 11px;
   line-height:16px;
-}
+
```
marcoscaceres: Might be best to just delete instead of using a comment here.

PR #2041, opened 2019-05-27
Bash escaped double quotes

### LINKED ISSUE #1930: Escaping quotes in bash broken
When using escaped quotes in bash, the highlighting breaks. This applies to latest version (currently 9.13.1).

### Live Example
URL: https://plnkr.co/edit/u0diCW1oxmKhLOEwKo8W

### Example bash script
```
#!/bin/bash
DATA=$(curl json | cut -d\" -f4)
echo "$DATA"
```

### Sample Screenshot 

screen shot 2018-12-20 at 4 25 27 pm

Inzeppelin:
This should fix #1930 

### Before:
hljs matches `\"` symbol as start position of string. And because of there is no ending quote, markup brakes after it.

Screenshot 2019-05-27 at 13 06 42

### After:
Changing regexp helps to identify escaped quote and not use it as a string starter symbol.

Screenshot 2019-05-27 at 13 06 21

Speciall markup tests and example are also provided.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -276,3 +276,4 @@ Contributors:
 - Sean T. Allen <sean@monkeysnatchbanana.com>
 - Greg Cline <gregrcline@gmail.com>
 - Sejin Jeon <jinaidy93@gmail.com>
+- Taif Alimov <inzeppelin@gmail.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ New languages:
 New styles:
 
 Improvements:
+- Issue #1930 - fix markup for escaped quotes in Bash
 
 ## Version 9.15.6
 New languages:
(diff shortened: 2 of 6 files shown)


marcoscaceres:
@Inzeppelin, just a heads up that the RegExp used in this PR seems to only be supported in Chrome. Unfortunately, I'm going to have to back this out of the latest release as it's crashing Firefox and Safari (#2043). 

> SyntaxError: Invalid regular expression: invalid group specifier name

Could you please take a look. Could we use some kind of alternative?

Inzeppelin:
@marcoscaceres yep, will try to resolve this one, sorry

Inzeppelin:
@marcominetti please check this one #2048

marcoscaceres:
Thanks @Inzeppelin! And no worries about the breakage before :) it happens. We need more robust cross browser testing on our side.

PR #2046, opened 2019-05-30
fix(clojure): top-level symbols/quoted forms

mhuebert:
Currently, if a Clojure source string contains a quoted form like `'()` or a symbol like `x` at the top-level, the entire string is not highlighted. This commit adds `SYMBOL` to the accepted top-level forms.


yyyc514:
Now you have an autodetect conflict with go.  Could you take a look at that?

yyyc514:
Ping.

yyyc514:
@mhuebert Closing for inactivity.  If you have time to follow-up on this in the future we can re-open.  This would also need a markup test to be accepted so we can see what symbols look like and see that this change works as expected.

Regarding the failing test.  This match rule looks very broad... I'm guessing it's matching something that previously was matching another rule and therefore reducing the overall relevancy. You'd think it being last would avoid that, but I'm just guessing.

PR #2049, opened 2019-05-31
Support for ruby's squiggly heredoc

skyfmmf:
Ruby has the squiggly heredoc notation to insert indented multiline
strings. This should be supported in addition to the existing heredoc
notation.


egor-rogov:
Thanks, merged!

PR #2056, opened 2019-06-13
remove FIX language

bmintz:
this language is apparently dead. The commit that added it references http://fixprotocol.org
but that site no longer exists.


marcoscaceres:
So, we probably don't want to remove dead languages... as people might still want to highlight them for historical reasons.

PR #2060, opened 2019-06-22
fix(go): correctly highlight hex numbers, rather than stopping at last 'd' or 'f'.

pkaminski:
Also had to lower number relevance to 0 to avoid misidentifying JS as Go in the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/highlightjs/highlight.js/2060)

### DIFF
--- a/src/languages/go.js
+++ b/src/languages/go.js
@@ -36,7 +36,7 @@ function(hljs) {
       {
         className: 'number',
         variants: [
-          {begin: hljs.C_NUMBER_RE + '[dflsi]?', relevance: 0},
+          {begin: hljs.C_NUMBER_RE + '[i]', relevance: 1},
           hljs.C_NUMBER_MODE
         ]
       },
--- a/src/languages/go.js
+++ b/src/languages/go.js
@@ -36,7 +36,7 @@ function(hljs) {
       {
         className: 'number',
         variants: [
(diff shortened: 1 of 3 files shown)


egor-rogov:
Hi @pkaminski,
some clarification is required. AFAIK, `i` suffix is for imaginary part of complex numbers. But what do other suffixes mean (in `[dflsi]`)?

pkaminski:
I don't know what all these suffixes mean and didn't add them in this PR.  Rather, in a [previous commit](https://github.com/highlightjs/highlight.js/commit/8ce35bcd4a38568a69e5e926772e08c7827cb1a8) the suffixes were (accidentally?) made compulsory by dropping the `?`, which resulted in hex numbers stopping parsing at the last `d` or `f`.

Note that I don't understand how `variants` work or why the previous commit dropped the `?` so there may be a better fix, but whatever it ends up being it needs to fix the newly added test line.

pkaminski:
Weird, `d` and `f` are typically used to indicate double and single precision float literals, but I concur that it appears that Go doesn't use them.  I tried to figure out when the suffixes were added to the regex but they were there from the very first commit in 2011.  Perhaps they used to be in some early version of the language and have been removed since?  In any case your fix works for me, thanks!

egor-rogov:
Thank you!

PR #2068, opened 2019-07-03
HL completion event

Hectopascal:
Add custom event that fires after highlighting is complete


marcoscaceres:
This doesn't feel appropriate to be honest. Highlighting happens synchronously here so dispatching an event isn't really needed.

Issue #2072, opened 2019-07-08
TypeError: Cannot read property 'length' of null.


rama291041610:
## For Bug
This library is used in [hexo](https://github.com/hexojs/hexo). When I execute `hexo s`, sometimes it throws a error:  
```
PS>hexo s
INFO  Start processing
FATAL Something's wrong. Maybe you can find the solution here: http://hexo.io/docs/troubleshooting.html
TypeError: Cannot read property 'length' of null
    at subMode (E:\software\blog\hexo\node_modules\highlight.js\lib\highlight.js:388:42)
    at processLexeme (E:\software\blog\hexo\node_modules\highlight.js\lib\highlight.js:497:22)
    at highlight (E:\software\blog\hexo\node_modules\highlight.js\lib\highlight.js:581:17)
    at E:\software\blog\hexo\node_modules\highlight.js\lib\highlight.js:627:21
    at Array.forEach (<anonymous>)
    at Object.highlightAuto (E:\software\blog\hexo\node_modules\highlight.js\lib\highlight.js:626:62)
    at E:\software\blog\hexo\node_modules\hexo-util\lib\highlight.js:117:25
    at highlight (E:\software\blog\hexo\node_modules\hexo-util\lib\highlight.js:120:7)
    at highlightUtil (E:\software\blog\hexo\node_modules\hexo-util\lib\highlight.js:22:14)
    at data.content.data.content.replace (E:\software\blog\hexo\node_modules\hexo\lib\plugins\filter\before_post_render\backtick_code_block.js:59:15)
    at String.replace (<anonymous>)
    at Hexo.backtickCodeBlock (E:\software\blog\hexo\node_modules\hexo\lib\plugins\filter\before_post_render\backtick_code_block.js:14:31)
    at Hexo.tryCatcher (E:\software\blog\hexo\node_modules\bluebird\js\release\util.js:16:23)
    at Hexo.<anonymous> (E:\software\blog\hexo\node_modules\bluebird\js\release\method.js:15:34)
    at Promise.each.filter (E:\software\blog\hexo\node_modules\hexo\lib\extend\filter.js:63:65)
    at tryCatcher (E:\software\blog\hexo\node_modules\bluebird\js\release\util.js:16:23)
    at Object.gotValue (E:\software\blog\hexo\node_modules\bluebird\js\release\reduce.js:155:18)
    at Object.gotAccum (E:\software\blog\hexo\node_modules\bluebird\js\release\reduce.js:144:25)
[...]

PR #2073, opened 2019-07-08
fix bug on TypeError: Cannot read property 'length' of null #2072

rama291041610:
#2072 fix bug on TypeError: Cannot read property 'length' of null.


src/highlight.js
```diff
@@ -384,13 +384,14 @@ https://highlightjs.org/
 
     function subMode(lexeme, mode) {
       var i, length;
-
-      for (i = 0, length = mode.contains.length; i < length; i++) {
-        if (testRe(mode.contains[i].beginRe, lexeme)) {
-          if (mode.contains[i].endSameAsBegin) {
-            mode.contains[i].endRe = escapeRe( mode.contains[i].beginRe.exec(lexeme)[0] );
+      if(mode.contains){ //Determine if the attribute exists
```
marcoscaceres: Maybe just: 

```suggestion
      if (!mode.contains) return;
```

rama291041610: Yes, it is better.

egor-rogov:
Hmm... I think this doesn't fix the root cause, but just hiding the problem.

Error stack trace looks like this: `highlightAuto` -> `higlight` -> `processLexeme` -> `subMode`.
Now, `highlight()` function starts with `compileLanguage()`, which calls `compileMode()`, and looking at the latter I wonder how can `mode.contains` be `null`?

```
      if (!mode.contains) {
        mode.contains = [];
      }
```
It seems to me that some language somehow didn't get compiled in time. And this is what we should fix.

@rama291041610, you wrote that the error occurs sometimes. Does it happen often? Have you notice any patterns? What language do you highlight? Can you reproduce the error with some diagnostic messages added to the highlighting code?

yyyc514:
Every mode will have a `contains`, that's setup when the language is parsed.  Would need a reproducible example here as already requested.

```
 if (!mode.contains) {
        mode.contains = [];
      }
```

It's possible this is a dependency issue (X depends on Y, but Y hasn't been compiled), but if so it should be 100% reproducible.

yyyc514:
Closing this as we still have the open issue.

PR #2076, opened 2019-07-18
Adds a language definition for GitHub Workflow.

qubyte:
This language definition is designed to highlight GitHubs automation language. Confusingly, the feature is referred to as "Actions", but the file suffix is `.workflow` and [the documentation](https://developer.github.com/actions/managing-workflows/creating-and-cancelling-a-workflow/) refers to the format as a workflow.


qubyte:
GH actions have changed! The new beta format (which will go out to everyone in November is plain old .yml, so there’s nothing to do here.

PR #2082, opened 2019-07-31
Support css custom properties

colomolo:
Support custom properties syntax:
```
.selector {
  --heading-1: 30px/32px Helvetica, sans-serif;
}
```

Since css vars have the same naming rules, digits could be used in naming.

Proofs:
https://drafts.csswg.org/css-variables-1/#custom-property
https://drafts.csswg.org/css-syntax-3/#identifier
https://drafts.csswg.org/css-syntax-3/#ident-token-diagram


egor-rogov:
Thanks, merged.

PR #2083, opened 2019-07-31
Add css vars example

colomolo:
Add example for css var.


egor-rogov:
@colomolo thanks for these improvements!

PR #2090, opened 2019-08-05
Updated the list of Twig filter and tags

javiereguiluz:
Updated according to the latest Twig documentation which can be found here: https://twig.symfony.com/doc/2.x/

### DIFF
--- a/src/languages/twig.js
+++ b/src/languages/twig.js
@@ -27,17 +27,17 @@ function(hljs) {
   var FILTER = {
     begin: /\|[A-Za-z_]+:?/,
     keywords:
-      'abs batch capitalize convert_encoding date date_modify default ' +
-      'escape first format join json_encode keys last length lower ' +
-      'merge nl2br number_format raw replace reverse round slice sort split ' +
-      'striptags title trim upper url_encode',
+      'abs batch capitalize column convert_encoding date date_modify default ' +
+      'escape filter first format inky inline_css join json_encode keys last ' +
+      'length lower map markdown merge nl2br number_format raw reduce replace ' +
+      'reverse round slice sort spaceless split striptags title trim upper url_encode',
     contains: [
       FUNCTIONS
     ]
   };
 
-  var TAGS = 'autoescape block do embed extends filter flush for ' +
-    'if import include macro sandbox set spaceless use verbatim';
+  var TAGS = 'apply autoescape block deprecated do embed extends flush for from' +
+    'if import include macro sandbox set use verbatim with';
 
   TAGS = TAGS + ' ' + TAGS.split(' ').map(function(t){return 'end' + t}).join(' ');
 


marcoscaceres:
@javiereguiluz could you please check why TravisCI is failing?

marcoscaceres:
Looks like this change causes the related test for this language to fail... @javiereguiluz or @weaverryan could you please update the test to reflect the new output.

javiereguiluz:
I'll need some help trying to fix the tests. I can't even run the test on my machine: `node tools/build.js` shows some errors and `npm run test` fails too 😢  Thanks!

egor-rogov:
@javiereguiluz,
what (exact) error messages you got?

You can also look into Travis job log (Details link to the right of "The Travis CI build failed" message). It shows that `template_tags` markup test is failed and gives you found/expected strings.

yyyc514:
@javiereguiluz Can you look into the failing tests and provide an updated PR?

yyyc514:
@javiereguiluz Still around? Ping.

src/languages/twig.js
```diff
@@ -27,17 +27,17 @@ function(hljs) {
   var FILTER = {
     begin: /\|[A-Za-z_]+:?/,
     keywords:
-      'abs batch capitalize convert_encoding date date_modify default ' +
-      'escape first format join json_encode keys last length lower ' +
-      'merge nl2br number_format raw replace reverse round slice sort split ' +
-      'striptags title trim upper url_encode',
+      'abs batch capitalize column convert_encoding date date_modify default ' +
+      'escape filter first format inky inline_css join json_encode keys last ' +
```
weaverryan: now inky_to_html

src/languages/twig.js
```diff
@@ -27,17 +27,17 @@ function(hljs) {
...
+      'length lower map markdown merge nl2br number_format raw reduce replace ' +
+      'reverse round slice sort spaceless split striptags title trim upper url_encode',
     contains: [
       FUNCTIONS
     ]
   };
 
-  var TAGS = 'autoescape block do embed extends filter flush for ' +
```
weaverryan: The failing test is either being caused by the removal of filter here or the addition above. In the interest of getting the PR merged, try re-adding it here or removing it above. Afterall, filter IS still a valid tag, just deprecated.

yyyc514: Oh yes is it’s only deprecated I would leave it for sure. That would be an easy fix.

src/languages/twig.js
yyyc514: And does inky stay as deprecated?

weaverryan: Good point. We could do either: inky *is* deprecated... but it was from such a new feature, that it only existed for some months before being deprecated. Having it or removing it - both are probably fine.

yyyc514:
@weaverryan Any chance you have the motivation to jump in here and push this to competition?  You could branch from this branch and just lay the tiny changes you need to fix the relevancy on top and make a new PR.

javiereguiluz:
@yyyc514 I'm really sorry for not having replied to your messages. I did the the changes proposed by Ryan. Let's see if it fixes the issues. Thanks.

src/languages/twig.js
```diff
@@ -27,17 +27,17 @@ function(hljs) {
...
-    'if import include macro sandbox set spaceless use verbatim';
+  var TAGS = 'apply autoescape block deprecated do embed extends filter flush for from' +
```
yyyc514: Missing a space here.

yyyc514: So you broke the if endif tags...

yyyc514:
@javiereguiluz Thanks so much for contributing!  Sorry it took so long to get to this!

yyyc514:
This should go out in 9.15.11 over the weekend.

javiereguiluz:
@yyyc514 thanks a lot for merging this! I apologize again for not having replied to your requests on time. Cheers!

Issue #2055, opened 2019-06-11
Update the language contribution documentation and procedure


jf990:
By way of reading issues and comments the language definition procedure is undergoing change. The request is now to define the language in a separate repo from this one. Nothing in the current documentation supports this. However there are several things we need to do in order to make this work.

1. Define a clear language contributing procedure and update [the documentation](https://highlightjs.readthedocs.io/en/latest/language-requests.html).
2. All test harness and data should also be in the language repo: build and test procedures should also be updated and clearly defined.
3. CI should apply to the languages, so would-be consumers of language definitions can see the current status of the languages they want to use.
4. a maintenance effort to update all existing language definitions to abide by these new rules.

unfortunately making major structural changes like this comes at a cost, but in order to keep this package well maintained and relevant we need to put in this effort.

PR #2096, opened 2019-08-08
Move the list of languages to README

egor-rogov:
An attempt to move the list of supported languages to README.

I extended the list with the new languages and their corresponding repos. Probably we should also add some words to "Getting the Library" section?


marcoscaceres:
Thanks for doing this @egor-rogov. Some words around getting the library would be great. Separate PR?

egor-rogov:
> Some words around getting the library would be great. Separate PR?
The problem is I'm not quite sure of the wording. First I should try to use some of the external packages myself.

egor-rogov:
@marcoscaceres Can you update the docs at https://highlightjs.readthedocs.io/?
It's terribly outdated, and we reference it at https://highlightjs.org/ and in the README.

marcoscaceres:
We might need to check with @isagalaev how to refresh the docs. It's not mentioned in the maintainers guide.

isagalaev:
Riiight… This is because I forgot to update readthedocs when we moved the project over to the org, so readthedocs still looks at my fork at `isagalaev`. I updated it now.

Building process is completely automatic: it picks up the docs from `docs/` on github and builds it with Sphinx. It can also be done locally with `make html` in the `docs/` directory.

You guys probably need access to the readthedocs admin page for cases like this. If you want to register on readthedocs and tell me your usernames, I'll add them to maintainers.

isagalaev:
P.S. I picked Sphinx and readthedocs back in the day purely due to personal familiarity, there's no reason against switching the docs to a different system, if needed.

gitpitch:
Hi guys, I just wanted to share a quick bit of feedback about this recently merged change to the location of the language aliases table. As a frequent user of this library by chance I found myself going to the docs today to find the appropriate alias for a block of code.

First up, it's great to see the list updated.

But there might also be some small steps that could be taken to make these changes a little more user friendly.

First item, you might want to consider adding a *moved-notice* or a *you-can-find-notice* to the original doc page. At the original location on the page indicated by the `#language-names-and-aliases`. I suggest this because when I got there I had no idea where the table had gone. Only by reading the commit history on this repo did I realize where I could find the table now. 

Second item, is really a question. Is it possible to make the new table on the README more easily discovered? I missed it when I went looking for it. That Details/Summary syntax results in a pretty subtle bit of UI magic that is not very obvious. To me at least.

Third and final item is again a question. Is there any way to specify/activate the same kind of `#language-names-and-aliases` page anchor on that README widget so people can create a link directly to the language aliases table? I'm asking this because it might go a long way to helping with the discovery problem mentioned above.

All this is just food for thought. Thanks for a great library.

marcoscaceres:
@isagalaev pinged you on email.

@gitpitch the suggestions sound good... would you be willing to send a PR with improvements? About your 3rd question... I'm not sure. Might have to be done manually somehow.

gitpitch:
@marcoscaceres sure, I'm willing to create a new PR once I understand how best to proceed.

I'm not entirely sure what inspired this original PR. Specifically, why was the alias table was moved out of the docs in the first place? And not, for example, to its own page in the docs? After all, this info is pretty critical, especially when displaying code blocks that include *markup* content such as HTML, XML, Python, React snippets, etc.

I suspect there are also a lot of historical links to this content hanging around. So the *moved-notice* in particular seems important...or at least worthwhile.

And now that this table is part of the **README** and not the docs,  I'm honestly not sure how to improve discovery of this table given the Details/Summary syntax used. I do understand how displaying the complete table in the **README** might be too verbose. Albeit this approach would make it easier to find this information. But I'm wide open to your thoughts on this.

marcoscaceres:
> I'm not entirely sure what inspired this original PR.

We wanted to put the list of supported languages "front and center". And also make it possible for people to find the languages that are not in the main repository.

gitpitch:
> We wanted to put the list of supported languages "front and center". 

Oh! :smile: I feel bad to raise my *discovery* issue now. I do like the idea/motivation. But perhaps not 100% the current execution.

So does this mean adding the table back to the docs is not an option? If not, I think a simple notice in those docs about the move still makes sense. What do you think?

marcoscaceres:
> So does this mean adding the table back to the docs is not an option? If not, I think a simple notice in those docs about the move still makes sense. What do you think?

Agree. That would be good.

egor-rogov:
@gitpitch, I also agree with your points and thank you for the PR. In fact I'm not very happy with the summary/detail approach myself, but a giant table in the middle of the README looks even worse to me. Any thoughts on how to improve this are welcome!

gitpitch:
@egor-rogov I think you're 100% right...a giant table in the middle of the README would not be nice at all. The only alternative to summary/detail that I can think of would be to create a dedicated markdown document for the aliases table.

For example, in a dedicated sub-directory like `aliases/README.md`. Then simply link to that table document from the top-level README. Perhaps something like this:

```
The list of supported languages and corresponding class names are detailed in the [Language Names and Class Aliases](aliases/README.md) table.
```

This approach may have the following benefits over summary/detail:

1. The clearly named hyperlink would make the aliases table stand-out, so easier to discover.
2. And the new standalone aliases table would now have it's own unambiguous URL.

I'm not yet 100% convinced of this approach myself. But it might be worth considering. What do you think?

gitpitch:
Btw, perhaps a top-level `lang`, `langs`, or `languages` directory would be more appropriate to `aliases` in this case. But hopefully the general idea is clear.

marcoscaceres:
FWIW, I'm happy with any direction we go with.

isagalaev:
> but a giant table in the middle of the README looks even worse to me. Any thoughts on how to improve this are welcome!

But what was wrong with having a link to a separate document? If it was discoverability then I'd say simply moving this link in its own section titled `## Supported languages` would have solved it…

In fact, the README has grown too big, and that's the actual reason it's hard to notice important things in there. Moving more stuff into it won't help, and the correct way of dealing with it (I think) is to move more things into separate documents and link to them from README.

egor-rogov:
Okay, lets start from the beginning, i.e. from the site (https://highlightjs.org/).
First of all a newbie looking for the language list will see "185 languages and 89 styles" link, which will lead him to a demo page. On that page languages are grouped by not-so-obvious categories, so it's relatively hard to find a language of choice on it.
To get to the actual supported language list, one had to navigate to Usage page (which is README in fact), then to "the API and other topics" (highlightjs.readthedocs.io), then to "CSS classes reference". Which seems to be highly counter-intuitive, at least to me.

By this PR we brought the language list closer to the user, but still it's far from ideal:

- Still the list is not very easy to discover.
- Summary/detail works inside github, but doesn't work on the site (see https://highlightjs.org/usage/), which is unacceptable.

So I think we should do the following:

- Move the list of languages and aliases to a separate page and make a link to it from README  (as it was suggested by @gitpitch and @isagalaev),
- Make a direct link to the list from https://highlightjs.org/. Probably split "185 languages and 89 styles" link into two: one for the list of languages, another for the list of styles (btw do we have it somewhere?), and make another link for the demo page.

Thoughts?

egor-rogov:
> You guys probably need access to the readthedocs admin page for cases like this. If you want to register on readthedocs and tell me your usernames, I'll add them to maintainers.

@isagalaev I've registered as `egorius`.

jf990:
I also find the docs a bit difficult to navigate, and agree with @egor-rogov comments about discoverability of the languages (and the styles too). Part of the problem I see is we have docs living in 3 separate places with 3 different doc systems and a lot of it is redundant causing us difficulty keep it all in sync and reader difficulty finding the right information quick and easy. The languages and styles is particularly critical info for users and maintainers, so we should be considerate to both audiences. I prefer the proposal of moving the list of languages out of the README and instead linking to a page in the docs.

We have the README and https://highlightjs.org/usage/ that are both mostly the same information. We also have https://highlightjs.readthedocs.io/en/latest/api.html. So there's 3 places we are asking users to go read to figure out how to use the API. It would be better if there was just one place. The README should be the _Getting Started_ and _Getting The Library_, and link to the API doc for everything else. highlightjs.org should be the demo and the build tool and link to API doc for everything else. This will also make our life easier keeping all this stuff up to date and accurate.

With the introduction of the language definitions in separate repos I'm a little uncertain how we are going to maintain the list of languages and https://highlightjs.org/download/. In fact, I can't find where that download page is currently defined.

I also prefer the suggestion to split languages and styles into two. they are separate concerns.

Let me know your concerns on this. I had doc updates pending and I was going to work on #2055 and I could help on some of this if we agree on a path forward.

PR #2105, opened 2019-08-16
Support highlighting inline HTML and CSS tagged template strings in JS and TS

rictic:
Also copies over the JavaScript parser's support for nested template strings into the TypeScript parser.

Includes tests for all three scenarios.


rictic:
Thanks for the quick review!

marcoscaceres:
Will try to publish a new release in the next few days.

PR #2109, opened 2019-08-20
Docs(move-notice): lang names and aliases table (#2096)

gitpitch:
As an update to #2096 this PR adds a simple *Moved-Notice* to the original docs page. Also, by re-using the original section header text old hyperlinks referencing the original page with the `#language-names-and-aliases` page-anchor should work again.


gitpitch:
> Thanks for the speedy PR!

You're welcome. Let me know if you need me to make any changes ahead of any merge.

PR #2112, opened 2019-08-22
fix(kotlin): Support substitutions with nested substitutions

### LINKED ISSUE #2111: Kotlin nested string substitution
Strings that contain substitutions with a nested substitution causes highlight.js to give up.

Example: `"a ${  "string $var ${subs}" } b"`

chuckjaz:
Fixes #2111


chuckjaz:
Updated the commit with what I believe you wanted. Let me know if you need something else.

egor-rogov:
Thanks, merged!

PR #2120, opened 2019-09-07
fix(tools/node): treat css/scss as UTF-8

### LINKED ISSUE #2118: Invalid UTF-8 in Sass generated file
After installing with `yarn add highlight.js` and trying to import the Sass file for the **dracula** theme with this:

```sass
@import '~highlight.js/scss/dracula.scss'
```

I get this error:

```bash
SassError: Invalid UTF-8.
   ╷
12 │ @author �verton Ribeiro <nuxlli@gmail.com>
   │         ^
   ╵
  node_modules/highlight.js/scss/dracula.scss 12:9  @import
```

Looking at the CSS files, it looks like the character that should be there is an: `É`.

So the full line should read:

```
@author Éverton Ribeiro <nuxlli@gmail.com>
```

marcoscaceres:
Fixes #2118

### DIFF
--- a/tools/node.js
+++ b/tools/node.js
@@ -95,7 +95,7 @@ function copyMetaFiles() {
 function buildSCSSStyles() {
   let input   = path.join(directory.root, 'src', 'styles', '*.css'),
       output  = path.join(directory.build, 'scss'),
-      options = { encoding: 'binary', dir: output }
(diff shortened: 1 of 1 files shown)


marcoscaceres:
actually, there are a few binary files (images) in that directory... so need to deal with those too.

PR #2123, opened 2019-09-10
Added missing (contextual) keywords for VB.NET.

jroessel:
Most notably, the `Async` and `Await` contextual keywords were missing. Proper handling of those would require parsing, so using them as identifiers will be broken, as every other contextual keyword, but I don't think this can be improved considering how highlight.js handles parsing.


egor-rogov:
Thanks, merged!

PR #2124, opened 2019-09-10
Added missing (contextual) keywords for C#.

jroessel:


jroessel:
I'd not classify Wikibooks as the authoritative source. The official list is here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/

`var` is a contextual keyword, as it loses its keyword meaning as soon as you have a type named `var`. I missed `nameof` in the original list admittedly.

Not that the distinction matters much; as noted in the VB.NET keyword PR I doubt hljs is going to do that much actual parsing to distinguish keywords from contextual keywords, so for current intents and purposes they're pretty much the same.

egor-rogov:
I agree, thanks for the right list.

> I doubt hljs is going to do that much actual parsing to distinguish keywords from contextual keywords, so for current intents and purposes they're pretty much the same.

Right, it's solely to simplify further support. As long as the keywords were originally divided into two categories, it makes sense to keep them in order.

egor-rogov:
Thanks, merged.

PR #2141, opened 2019-09-26
SQF should have it's own preprecesssor definition

yyyc514:
- relying on the one from C++ caused unpredictable results
  depending on which language you would parse first because
  the pre-processor regexs would either be compiled case 
  insensative (SQF) or case-sensative (C++), depending on 
  which language you highlighted first


yyyc514:
Do I need to rebase this onto master?

src/languages/sqf.js
```diff
@@ -40,6 +40,30 @@ function(hljs) {
     ]
   };
 
+  // list of keywords from:
+  // https://community.bistudio.com/wiki/PreProcessor_Commands
+  var PREPROCESSOR =       {
```
marcoscaceres:
```suggestion
  var PREPROCESSOR = {
```

yyyc514:
Do we not care about "merge branch 'master' into whatever" commits polluting the master branch?  I really don't mind rebasing if that would help.

marcoscaceres:
Nope, we use squash merge for PRs, so we never get merge commits in history.

yyyc514:
Well that's useful to know, I guess now I can worry less about the exact sequencing of tiny commits for a larger PR if they are never going to see master anyways.

egor-rogov:
Done with it. Thanks!

yyyc514:
Thanks!

PR #2150, opened 2019-10-04
feat(javascript): Add big number syntax (n postfix)

### LINKED ISSUE #1797: JavaScript BigInt / arbitrary precision number support
Per TC39: https://github.com/tc39/proposal-bigint

Though technically a proposal, this is shipped in Chrome stable and Node.js. The syntax change is an optional `n` postfix on literal numbers, like

```js
let x = 10n;
let y = 0x10n;
let z = 0b0101n;
```

I think the change might be something like this ↓, and I'll tinker with rolling a PR to implement.

```diff
diff --git a/src/languages/javascript.js b/src/languages/javascript.js
index 24d4ae9a..d6a824a3 100644
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -29,9 +29,9 @@ function(hljs) {
   var NUMBER = {
     className: 'number',
     variants: [
-      { begin: '\\b(0[bB][01]+)' },
-      { begin: '\\b(0[oO][0-7]+)' },
-      { begin: hljs.C_NUMBER_RE }
+      { begin: '\\b(0[bB][01]+)n?' },
+      { begin: '\\b(0[oO][0-7]+)n?' },
+      { begin: hljs.C_NUMBER_RE + 'n?' }
     ],
     relevance: 0
   };
```

tmcw:
Fixes #1797 - allows a `n` postfix for the JavaScript big integer syntax: https://github.com/tc39/proposal-bigint

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -28,9 +28,9 @@ function(hljs) {
   var NUMBER = {
     className: 'number',
     variants: [
-      { begin: '\\b(0[bB][01]+)' },
-      { begin: '\\b(0[oO][0-7]+)' },
-      { begin: hljs.C_NUMBER_RE }
+      { begin: '\\b(0[bB][01]+)n?' },
+      { begin: '\\b(0[oO][0-7]+)n?' },
+      { begin: hljs.C_NUMBER_RE + 'n?' }
     ],
     relevance: 0
   };
--- a/src/languages/typescript.js
+++ b/src/languages/typescript.js
@@ -61,9 +61,9 @@ function(hljs) {
   var NUMBER = {
     className: 'number',
     variants: [
-      { begin: '\\b(0[bB][01]+)' },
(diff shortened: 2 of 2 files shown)


yyyc514:
Does this affect typescript or livescript or any of the other JS derivative languages?  Do you have any idea? :-)

marcoscaceres:
I don’t think it would. The others would eventually need to support this.

yyyc514:
> The others would eventually need to support this.

That's what I'm asking if we should go ahead and figure tha tout now vs later. :)

marcoscaceres:
I think we should go ahead.

yyyc514:
@tmcw Could you by any chance help with that?

tmcw:
I've added support to TypeScript. It doesn't affect LiveScript or any derivative languages afaict.

yyyc514:
@tmcw Thanks so much.

PR #2151, opened 2019-10-05
New style: Night Owl

cdbax:
Adds a port of Sarah Drasner's Night Owl Theme for VS Code

Here's a sample of what it looks like for Elixir:
image


cdbax:
Ahh - I was unsure of how to properly handle the attribution. Thanks heaps for the fix!

marcoscaceres:
I'm not 100% sure either, to be honest... licenses are scary, so I just paste them in as that's what they say to do.

PR #2153, opened 2019-10-05
fix(Elixir): improve regex for numbers

cdbax:
Elixir has a lot of different syntax for representing number literals:

image

...some of which aren't properly detected by the existing regex for numbers in the Elixir language definition:

image

I tweaked the regex and added some tests to check that they are being detected properly:

image


egor-rogov:
Merged. @cdbax thanks for fixing it!

PR #2154, opened 2019-10-06
fix(Pony): improve regex for numbers

cdbax:
I noticed that the syntax highlighting for numeric literals wasn't working properly in the [Pony Tutorial](https://tutorial.ponylang.io/expressions/literals.html):
image

This PR switches Pony from using `C_NUMBER_MODE` to a custom number mode with a tweaked regex based on `C_NUMBER_RE` to allow for underscores and the binary literal syntax.

**BEFORE:**
image

**AFTER:**
image

The example with an underscore on the end wouldn't actually compile, but I wanted to make sure it was only highlighting the part that was legit 😬- please let me know if this is best taken out.


cdbax:
I've just noticed that I've got a bug in the binary syntax here. I'm allowing digits instead of just `[01]`. I'll push up a fix.

egor-rogov:
Thanks, merged!

PR #2155, opened 2019-10-06
Allow JSON to handle comments

yyyc514:


yyyc514:
Github calls this "jsonc" in one place but I'm not sure if that is actually a real "thing" or not.  If we want I can add it as an alias, but I didn't so far.

test/markup/index.js
```diff
@@ -26,7 +26,7 @@ function testLanguage(language) {
 
           actual.trim().should.equal(expected.trim());
           done();
-        });
+        }).catch(function(err) { return done(err) });
```
egor-rogov: Wonderful (:

yyyc514:
Awesomeness.

egor-rogov:
Thanks, merged!

PR #2156, opened 2019-10-06
[cpp] improve single quote string highlighting

yyyc514:


sylveon:
In C++, single quotes are for single character literals, not string literals. This code fixes the issue by treating them as if they where string literals, which is fine, but the example should probably not use them like so.

sylveon:
While we're at it, it doesn't seems like wide character raw strings are handled.
```cpp
LR"(hello
world)"
```
doesn't highlights the L even though it should be.

yyyc514:
Ah, duh, so that explain the weirdness of the existing regex.  Ok, back to the drawing board on this, not read to merge yet. :-)

> doesn't highlights the L even though it should be.

Is it highlighting the rest?

Do you understand the regex for raw strings:

```js
{ begin: /(?:u8?|U|L)?R"([^()\\ ]{0,16})\((?:.|\n)*?\)\1"/ },
```

I'm not sure what we're checking for at the beginning with the {0,16}, do you know?

yyyc514:
> In C++, single quotes are for single character literals, not string literals.

Can a character literal be MORE than one literal character wide then if it's encoded as u8 or U or L?

sylveon:
Even without being Unicode it can, due to being able to create character
literals for escapes, eg '\n'

sylveon:
And you also have unicode character escapes, like `L'\u202e'` or hexadecimal character escapes like `'\xFF'`

yyyc514:
I meant OTHER than escapes. :-)    YES, those other ones are the more interesting ones.

sylveon:
It'd be easier to handle character literals as a normal string, they're the same escapes and even normal strings don't seem to highlight those anyways.

As for raw strings, it does highlight the text, only the `L` isn't. The regex seems to be doing some things to correctly detect custom prefixes for raw strings:
```cpp
LR"foobar(hello
world)foobar"
```
Those custom delimiters can be nothing or at most 16 characters and must not contain parentheses, spaces or backslashes, hence the `([^()\\ ]{0,16})`. See https://en.cppreference.com/w/cpp/language/string_literal

sylveon:
As for L not being highlighted it's probably just because of an out of date version, I used hastebin (which uses highlight.js to highlight its stuff) to test out of laziness, the regex seems to be doing the correct thing now but has no associated test case.

yyyc514:
Better?

test/markup/index.js
```diff
@@ -26,7 +26,7 @@ function testLanguage(language) {
 
           actual.trim().should.equal(expected.trim());
           done();
-        });
+        }).catch(function(err) { return done(err) });
```
sylveon: Not sure if this was intended to be left there.

egor-rogov: It's already merged in fact.

egor-rogov:
Merged, thanks everybody!

test/markup/index.js
yyyc514: Yes, this is intentional.  It allows the test runner to properly colorize errors for debugging failing tests.

PR #2159, opened 2019-10-06
[diff] fix context lines matching

yyyc514:


yyyc514:
https://www.gnu.org/software/diffutils/manual/html_node/Example-Context.html#Example-Context

egor-rogov:
Merged, thanks!

PR #2160, opened 2019-10-06
varchar2 support

ar0ra1:
varchar2 support for SQL (Oracle DB)


yyyc514:
Looks like auto-cancellation stopped the CI?  can we kick it off again?

PR #2164, opened 2019-10-07
Add svg as alias in xml

lukasoppermann:


yyyc514:
@lukasoppermann Don't we need to update the README also?

lukasoppermann:
Good point: #2165

PR #2167, opened 2019-10-07
Adding node support to readme

### LINKED ISSUE #1284: How to use highlight.js on nodejs
Hello,
I see on the homepage, hightlight.js can be used with nodejs as well. However I am not able to find any documentation or short sample which I can follow to use it on nodejs.
Would be really great if someone can point to right direction for this.

Thank you,

lukasoppermann:
This fixed #1284

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ const highlightedCode = hljs.highlight('html', '<span>Hello World!</span>').valu
 {
   language: … // detected language
   relevance: … // integer value
-  value: HTML … // string with highlighting markup
+  value: … // HTML string with highlighting markup
   second_best: … // object with the same structure for second-best heuristically detected language, may be absent
 }
 ```
--- a/README.md
+++ b/README.md
@@ -289,19 +289,19 @@ onmessage = (event) => {
 
 ## Node.js
 
-You can run highlight.js on node to prepare content before sending it to the browser.
+You can use highlight.js with node to highlight content before sending it to the browser.
 
 
 ```js
 // require the highlight.js library including all languages
-const hljs = require('./highlight');
+const hljs = require('./highlight.js');
 const highlightedCode = hljs.highlightAuto('<span>Hello World!</span>').value
 ```
 
 ```js
 // require the highlight.js library without languages
 const hljs = require("highlight.js/lib/highlight.js");
-// seperatly require languages
+// separately require languages
 hljs.registerLanguage('html', require('highlight.js/lib/languages/html'));
 hljs.registerLanguage('sql', require('highlight.js/lib/languages/sql'));
 // highlight with providing the language
@@ -309,6 +309,7 @@ const highlightedCode = hljs.highlight('html', '<span>Hello World!</span>').valu
 ```
 
 ### highlight() return value
+
 ```js
 {
   language: … // language name, same as the one passed into a function, returned for consistency with highlightAuto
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -287,6 +287,46 @@ onmessage = (event) => {
...
 ```
 
+## Node.js
+
+You can run highlight.js on node to prepare content before sending it to the browser.
+
+
+```js
+// require the highlight.js library including all languages
+const hljs = require('./highlight');
```
yyyc514: I know it's an example but should this be `highlight.js` just for consistency with the below example?

README.md
```diff
@@ -287,6 +287,46 @@ onmessage = (event) => {
 };
 ```
 
+## Node.js
+
+You can run highlight.js on node to prepare content before sending it to the browser.
```
yyyc514: - run on or "use with"?  
- prepare vs "highlight"?

README.md
```diff
@@ -287,6 +287,46 @@ onmessage = (event) => {
...
+```js
+// require the highlight.js library including all languages
+const hljs = require('./highlight');
+const highlightedCode = hljs.highlightAuto('<span>Hello World!</span>').value
+```
+
+```js
+// require the highlight.js library without languages
+const hljs = require("highlight.js/lib/highlight.js");
+// seperatly require languages
```
yyyc514: spelling: separately

README.md
```diff
@@ -287,6 +287,46 @@ onmessage = (event) => {
...
+hljs.registerLanguage('html', require('highlight.js/lib/languages/html'));
+hljs.registerLanguage('sql', require('highlight.js/lib/languages/sql'));
+// highlight with providing the language
+const highlightedCode = hljs.highlight('html', '<span>Hello World!</span>').value
+```
+
+### highlight() return value
+```js
```
yyyc514: Should there be a line break here or does it really not matteR?

lukasoppermann: You mean above the code block, correct? I added one.

egor-rogov:
I'm a bit concerned with "highlight() return value" and "highlightAuto() return value" subsections under "Node.js" section:

1. This information is not specific to Node.js
2. Such details are probably not intended for README, they are described in devveloper's docs : https://highlightjs.readthedocs.io/en/latest/api.html

What do you think?

lukasoppermann:
Hey @egor-rogov, I understand the concern, however I feel that somehow we need to describe, why you have to use `.value` to get the result.

yyyc514:
@egor-rogov Would this work better for you if we just dropped everything below:

"### highlight() return value"

I agree that's really for the API docs.  If a single example shows the use of `.value` I think that's clear enough at a glance, then if you want to know more, look at the docs.

README.md
```diff
@@ -287,6 +287,48 @@ onmessage = (event) => {
 };
 ```
 
+## Node.js
+
+You can use highlight.js with node to highlight content before sending it to the browser.
```
lukasoppermann: How about adding the info here?

> You can use highlight.js with node to highlight content before sending it to the browser. Make sure to use the `.value` property to get the formatted html. For more info about the returned object refer to the api docs https://highlightjs.readthedocs.io/en/latest/api.html

egor-rogov: That's fine with me.

lukasoppermann: I updated it now.

egor-rogov:
Great! Merged, thanks.

Issue #1681, opened 2018-01-26
Several quirks in YAML rendering (colon in attribute name)


marians:
I'd like to report a few problems we are seeing in YAML rendering.

Here is my reference document:

```yaml
endpoints:
  http://localhost:9000:
    alias: local
    email: developer@example.com
    token: 7d5b1161-87e6-4dfd-8017-5ab9f4e1cb14
  https://api.g8s.cloud.europe.gigantic.io:
    alias: cloud
    email: developer@example.com
    token: 103dc680-1b53-4ad3-b5cb-ac9b2f86134d
  https://api.g8s.onprem.example.com:
    alias: onprem
    email: developer@example.com
    token: ""
selected_endpoint: https://api.g8s.cloud.europe.gigantic.io
last_version_check: 2018-01-25T13:50:25.353937+01:00
updated: 2018-01-26T09:17:24+01:00
```

That YAML is generated using [go-yaml v2](https://github.com/go-yaml/yaml/tree/v2).

This is how highlight.js version 9.12.0 renders it with solarized-dark style:

![image](https://user-images.githubusercontent.com/273727/35444839-10d384a0-02b0-11e8-8444-3663983b4238.png)

The HTML code for the above:

```html
<pre><code class="language-yaml hljs"><span class="hljs-attr">endpoints:</span>
<span class="hljs-attr">  http:</span><span class="hljs-string">//localhost:9000:</span>
<span class="hljs-attr">    alias:</span> <span class="hljs-string">local</span>
<span class="hljs-attr">    email:</span> <span class="hljs-string">developer@example.com</span>
<span class="hljs-attr">    token:</span> <span class="hljs-number">7</span><span class="hljs-string">d5b1161-87e6-4dfd-8017-5ab9f4e1cb14</span>
<span class="hljs-attr">  https:</span><span class="hljs-string">//api.g8s.cloud.europe.gigantic.io:</span>
<span class="hljs-attr">    alias:</span> <span class="hljs-string">cloud</span>
<span class="hljs-attr">    email:</span> <span class="hljs-string">developer@example.com</span>
<span class="hljs-attr">    token:</span> <span class="hljs-number">103</span><span class="hljs-string">dc680-1b53-4ad3-b5cb-ac9b2f86134d</span>
<span class="hljs-attr">  https:</span><span class="hljs-string">//api.g8s.onprem.example.com:</span>
<span class="hljs-attr">    alias:</span> <span class="hljs-string">onprem</span>
[...]

PR #2170, opened 2019-10-07
fix(YAML): improve matching for keys

### LINKED ISSUE #2058: yaml not rendered correctly
Hi, 

I'm having a problem with my yaml files.

Example yaml file:
```yaml
sidebar:
  Basic Tutorial:
  - Getting Started: docs/getting-started.md
  - Customizing The Sidebar: docs/customizing-the-sidebar.md
  - Customizing The Public Api: docs/customizing-the-public-api.md
  - Discussions: docs/discussions.md
  - Themes and CSS: docs/themes-and-css.md
  More Info:
  - Autodoc Settings: docs/autodocSettings-full-list.md
  - Building Version Documentation: docs/building-version-documentation.md
  - Markdown Cheat Sheet: docs/markdown-cheat-sheet.md
  - Changelog: docs/changelog.md
  - Contact Us: docs/contact-us.md
  - Feature Requests: https://github.com/erez-o/doxiz/issues
autodocSettings: {}
```

rendered as:
yaml highlight

Using the latest version (9.15.8), also made a fiddle http://jsfiddle.net/f261xec4/

Thx

cdbax:
Fixes #2058 #1681 

I had a look at the YAML spec as part of fixing this, and what little I understood was horrifying (did you know JSON is valid YAML!?) so I opted for what is hopefully a nice middle-ground that seems to me to be an improvement on the existing detection.

**BEFORE**
image

**AFTER**
image


src/languages/yaml.js
```diff
@@ -8,14 +8,15 @@ Category: common, config
...
+  // ...and ending with a colon followed immediately by a space, tab or newline.
+  // The YAML spec allows for much more than this, but this covers most use-cases.
   var KEY = {
     className: 'attr',
     variants: [
-      { begin: keyPrefix + keyName + ":"},
-      { begin: keyPrefix + '"' + keyName + '"' + ":"},
-      { begin: keyPrefix + "'" + keyName + "'" + ":"}
+      { begin: '\\w[\\w :\\/\\.]*:(?=[ \t]|$)' },
+      { begin: '"\\w[\\w :\\/\\.]*":(?=[ \t]|$)' } //quoted keys
```
yyyc514: What happened to single quoted keys?

test/markup/tap/yaml-block.expect.txt
```diff
@@ -8,10 +8,10 @@
...
 <span class="hljs-symbol">ok</span>
   <span class="yaml"><span class="hljs-meta">---</span>
-<span class="hljs-attr">  message:</span> <span class="hljs-string">"Board layout"</span>
-<span class="hljs-attr">  severity:</span> <span class="hljs-string">comment</span>
-<span class="hljs-attr">  dump:</span>
-<span class="hljs-attr">     board:</span>
+  <span class="hljs-attr">message:</span> <span class="hljs-string">"Board layout"</span>
+  <span class="hljs-attr">severity:</span> <span class="hljs-string">comment</span>
+  <span class="hljs-attr">dump:</span>
+     <span class="hljs-attr">board:</span>
```
yyyc514: Ha, this is an improvement me thinks :)

test/markup/yaml/keys.txt
```diff
@@ -0,0 +1,22 @@
...
+
+some_key:
+  another_key: value
+
+some key:
+  another key: value
+
+"some key":
+  "another key": value
+
```
yyyc514: Would be nice to see "some-key" and "some.key" since those were points of issue from before.

cdbax: 👍 Added

src/languages/yaml.js
cdbax: Whoops! Thanks for the save.

yyyc514:
If not it's ok, but it'd be nice to solve that also if you're game.

src/languages/yaml.js
yyyc514: That's what I'm here for.

cdbax:
Yeah I planned to go looking for some of the other YAML issues next. Happy to try and do it in the one PR if that's preferred.

yyyc514:
No, lets split it up if you don't mind.

cdbax:
@yyyc514 can you explain the `|$` hack to me? I could not get lookaheads to work properly until I noticed that other lookaheads in the code seemed to be doing this...but I have no idea why it works.

yyyc514:
It's how the parser works... it's two stages and the second stage doesn't have access to the FULL content, only the match itself... and matches don't include look ahead... so the second stage will fail and the whole thing falls to pieces, unless you say look-ahead or "end of string" because in the 2nd stage all you'v got is the end of string.

And in many cases this hack is "good enough".

I'm currently working on fixing the parser so this isn't necessary and true look-ahead "just works" without caveats, but I've gotten a bit distracted as of late with the larger Highlight.js project. :-)

I plan to return to that work soon though.

yyyc514:
Can you fix the merge conflict?

@egor-rogov  @marcoscaceres You think there is any use in a two reviewer policy for PRs or if one of us reviews (who didn't also write) and then thinks it's good, then merge it?

marcoscaceres:
One reviewer is fine as long as you have a high degree of confidence (and there are no crypto-miners 🤣).

yyyc514:
@marcoscaceres Ok I'll follow that rule then and if I think it needs extra eyeballs I'll ping you all.

PR #2172, opened 2019-10-07
improve auto-detection of accesslog

yyyc514:
- weights slightly higher for popular user-agent strings
  (Googlebot, Mozilla/*) +3
- weights higher for IP address starting the line +5
- spreads the previous 10 points for verb differently
  (+5 relevance points for verb, +5 got HTTP/x.x in same string)
- [\d...] strings (likely date) of at least 8 characters long are +1
- [...] strings are worth nothing now
- "..." strings are worth nothing now


yyyc514:
This is just me playing around a bit on a simple syntax based on my syntax relevancy research.  Here was the performance of access log before:

```
( everything that matches down to potentially 80% less relevancy)
accesslog 15 0.07
  5 cpp 0.02
  8 awk 0.02
  11 bnf 0.02
  13 clean 0.02
  5 coffeescript 0.02
  9 cos 0.02
  37 ebnf 0.07
  2 excel 0.04
  0 json 0
  8 mathematica 0.03
  7 mel 0.01
  8 mercury 0.02
  16 moonscript 0.03
  12 n1ql 0.03
  4 openscad 0.02
  14 parser3 0.02
  15 prolog 0.06
  27 rib 0.04
  17 ruleslanguage 0.02
  25 scala 0.02
  5 scilab 0.02
  11 sql 0.04
  18 stan 0.02
  127 tp 0.03
```

And after:

```
(everything that matches down to an order of magnitude less relevancy)
accesslog 20 0.09
```

egor-rogov:
Is it OK that IP address is matching only from the beginning of the line?
And it lacks tests, really. Just one line.

yyyc514:
> Is it OK that IP address is matching only from the beginning of the line?

IP address always starts the line AFAIK... Accesslog (as a standard) is a pretty simple format...

> And it lacks tests, really. Just one line.

Can you be more specific?

- Would you like to see a larger detects test?
- Or just a larger markup sample set?  I can do that pretty easily.
- Some sort of test for the relevancy improvements (that's a bit harder right now I think, but something we should have long-term I'd think).

The tooling I'm working with myself locally is still very raw. :-)

yyyc514:
https://en.wikipedia.org/wiki/Common_Log_Format

I've been staring at these kind of files my whole life (it seems). :-) It is one of the "languages" I'm more familiar with as far as such things can go. :-)  It's a pretty boring format and easy format. I'm pretty sure I'm not breaking it - and our one sample (despite being few) is actually very representative.

I have no idea how IP6 might be encoded, but our existing format didn't support that either. :-)

If you want to read some more samples:

http://www.almhuette-raith.at/apache-log/access.log

:-)

yyyc514:
The one thing we don't support (but I wasn't in a hurry to add without request) are more uncommon verbs like for WebDAV and other such uses... but other than that again, pretty boring format. :)

yyyc514:
I can steal a few lines from other things and make the `detect` and `markup` samples longer if that'd make you happier. :-)

But if the detect tests for other languages start screaming  then I'm going to stop and limit myself to just `markup`...

You tell me. :)

yyyc514:
I'll circle back tomorrow and add another sample or two and perhaps a link to the Wikipedia article on the format.  Then we'll merge. :)

egor-rogov:
Some more (markup) tests would be great, though...

yyyc514:
> Some more (markup) tests would be great, though...

Yeah, I'll focus on that prolly.

PR #2175, opened 2019-10-09
fix(Handlebars): add support for raw-blocks and triple-mustaches

nknapp:
I have added support for raw blocks (quadruple mustaches) inside which template Handlebars expressions are not evaluated.

```handlebars
{{{{raw}}}} {{verbatim parts}} {{{{/raw}}}} {{handlebars again}}
```

I also added support for triple mustaches (which mean that the result is not HTML-escaped).

```handlebars
 {{{epxression that is not HTML escaped}}}
```

I tried to extract the `MUSTACHE_CONTENTS` into a variable first, but there seemed to be side-effects cause by sharing the same object in multiple mode-definitions, so I extracted it into a function.


yyyc514:
> I tried to extract the MUSTACHE_CONTENTS into a variable first, but there seemed to be side-effects cause by sharing the same object in multiple mode-definitions, so I extracted it into a function.

The proper solution to this is to use `hljs.inherits` that does a copy/clone.  There are tons of examples of it being used.

Please remove `testbench.html`.

src/languages/handlebars.js
```diff
@@ -5,35 +5,57 @@ Author: Robin Ward <robin.ward@gmail.com>
...
+      },
+      {
+        className: 'template-variable',
+        begin: '{{{', end: '}}}',
+        keywords: BUILT_INS
       },
       {
         className: 'template-variable',
-        begin: /\{\{/, end: /\}\}/,
+        begin: '{{', end: '}}',
```
yyyc514: Reason this changed?

yyyc514:
@nknapp Looks pretty nice.  Thanks for contributing!  I was going "WTF, 16 files changes" then I saw it was all a bunch of test cases. ;-)

src/languages/handlebars.js
nknapp: I thought there was a different, possibly more efficient implementation for matching strings directly. Now, I have looked at the code and seen that everything is converted to regexp anyway. I will change this back to a regular expression.

I'm not sure escaping the curly brackets is really necessary. My IDE tells me that it isn't, but I'm not sure if this is true for every browser out there.

yyyc514: Our consistency on // vs ‘’ is a bit all over the map but I just saw this changed but it didn’t really look like the actual content changed, so it jumped out at me.

nknapp:
> The proper solution to this is to use hljs.inherits that does a copy/clone. There are tons of 
> examples of it being used.

I just tried around a lot with `hljs.inherit` but I can't get it to work. The problem is that `inherit` only creates a shallow copy, but my problem seems to be caused by reusing the same instance of the `starts` in multiple rules. If I define my `MUSTACHE_CONTENTS` with a getter creating a new instance of the object,

```js
  var MUSTACHE_CONTENTS = {
    className: 'name',
    begin: /[a-zA-Z\.-]+/,
    keywords: BUILT_INS,
    get starts() {
      return {
        endsWithParent: true, relevance: 0,
        contains: [
          hljs.inherit(hljs.QUOTE_STRING_MODE)
        ]
      }
    }
```

and then use `contains: [hljs.inherit(MUSTACHE_CONTENTS)],` in the rules, then it works.
If I define

```js
 var MUSTACHE_CONTENTS = {
    className: 'name',
    begin: /[a-zA-Z\.-]+/,
    keywords: BUILT_INS,
    starts: {
      endsWithParent: true, relevance: 0,
      contains: [
        hljs.inherit(hljs.QUOTE_STRING_MODE)
      ]
    }
  }
```
the source 

```hbs
{{#block}}block content{{/block}}
```
compiles to 

```html
<span class="xml"></span><span class="hljs-template-tag">{{#<span class="hljs-name">block</span>}}block content{{/block}}
</span>
```
instead of the expected
```html
<span class="xml"></span><span class="hljs-template-tag">{{#<span class="hljs-name">block</span>}}</span><span class="xml">block content</span><span class="hljs-template-tag">{{/<span class="hljs-name">block</span>}}</span><span class="xml">
</span>
```

nknapp:
I can rebase and squash when we are done, but I think the review is easier if I keep the commits seperate at the moment.

yyyc514:
I need to dig into this when I get home.  I’m not sure why we need starts for this at all.   Did you look at how heredocs are done in say Ruby?  Does it not matter when we find the end that we actually check to see if we’re closing the same block we opened?

I’m not super familiar with starts yet, but I should fix that.  Seems it has some potential caveats.  I think you’re right... the problem is that it saves state during the parsing so if you use it in multiple places you need multiple copies of it.  

But I wonder if we just shouldn’t fix inherits to deal with that automatically...

nknapp:
From what I tried yesterday, "starts" in this context makes sure that only the first word in a multi-word-expression (i.e. a helper call with parameters) is highlighted as `hljs-name`. I think the idea is to highlight the helper name more prominently than its parameters.

Thanks for looking into it. 

I have just added another change to display comments like `{{!-- some {{expression}} in a comment --}}` correctly. The comment goes on after `expression}}`.

nknapp:
I just wanted to say that this is not urgent for me. I just stumbled accross the deficits, because I use  highlight.js to highlight in the examples on the new Handlebars homepage and some (like this one https://handlebars-draft.knappi.org/guide/#html-escaping) are not correctly displayed.

But the site is not ready yet anyway and even then, those are minor problems.

src/languages/handlebars.js
```diff
@@ -5,32 +5,58 @@ Author: Robin Ward <robin.ward@gmail.com>
...
     subLanguage: 'xml',
     contains: [
-    hljs.COMMENT('{{!(--)?', '(--)?}}'),
+      hljs.COMMENT(/\{\{!--/, /--\}\}/),
+      hljs.COMMENT(/\{\{!/, /\}\}/),
+      // raw block {{{{raw}}}} verbatim xml {{{{/raw}} {{handlebars}}
+      {
+        className: 'template-tag',
+        begin: /\{\{\{\{\//, end: /\}\}\}\}/,
+        contains: [MUSTACHE_CONTENTS()]
```
yyyc514: I think I'd prefer to see the open tag come first.  I was starting at this a while trying to figure out which item was doing what.  You can use a negative look-ahead to make sure there isn't a "/" immediately after so you don't flag the end tag by mistake.

Perhaps `// open/close ` comments might help also?

Our grammars don't tend to have  a ton of comments but I think they'd likely be more accessible if they did.

src/languages/handlebars.js
```diff
@@ -5,32 +5,58 @@ Author: Robin Ward <robin.ward@gmail.com>
...
   var BUILT_INS = {'builtin-name': 'each in with if else unless bindattr action collection debugger log outlet template unbound view yield'};
+  function MUSTACHE_CONTENTS() {
+    return  {
+      className: 'name',
+      begin: /[a-zA-Z\.-]+/,
+      keywords: BUILT_INS,
+      starts: {
+        endsWithParent: true, relevance: 0,
+        contains: [
+          hljs.QUOTE_STRING_MODE
```
yyyc514: Is this for literal params?  Should we have a test for that?

src/languages/handlebars.js
```diff
@@ -5,32 +5,58 @@ Author: Robin Ward <robin.ward@gmail.com>
...
 Category: template
 */
-
 function(hljs) {
   var BUILT_INS = {'builtin-name': 'each in with if else unless bindattr action collection debugger log outlet template unbound view yield'};
+  function MUSTACHE_CONTENTS() {
+    return  {
+      className: 'name',
+      begin: /[a-zA-Z\.-]+/,
+      keywords: BUILT_INS,
```
yyyc514: Do any of the tests include a built-in?  Might that be a good idea?

src/languages/handlebars.js
nknapp: Yes we should. And it is not working correctly (https://jsfiddle.net/abyejr70/), neither in the latest version nor in my PR. 

But I would like to create another PR for that.

src/languages/handlebars.js
nknapp: done

yyyc514:
Looking good, you happy?

nknapp:
Yes. Should I squash?

yyyc514:
No need. We squash everything before merging anyways. I’ll review one last time and get  this merged soon. Thanks for all the effort.

nknapp:
Thanks

test/markup/handlebars/built-ins.expect.txt
```diff
@@ -0,0 +1,12 @@
+<span class="xml"></span><span class="hljs-template-tag">{{#<span class="hljs-name"><span class="hljs-builtin-name">if</span></span> test}}</span><span class="xml">yes</span><span class="hljs-template-tag">{{/<span class="hljs-name"><span class="hljs-builtin-name">if</span></span>}}</span><span class="xml">
```
yyyc514: Hmmm, any idea why the `<span class="xml"></span>`  that starts all of these?  Perhaps a bug in the sublanguage parsing?

nknapp: so far, I have only written the language after try-and-error, using my tests and looking at the result in my `testbench.html`.

The spans where there from the start, so I haven't looked at where they come from...

yyyc514: Ok, Twig is the same way (it has existing markup tests), so this is just "how it works" now I guess, though it's strange... also not sure the value of raw text being supported by xml class just because it bounces in and out of that sublanguage... but oh well. :-)  

An issue for another time and place.

nknapp: I think it is the sublanguage processing:
 
*In [line 583](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L583), `match.index` is 0, because the token `{{{` is found directly at the beginning of the input.
*In [line 492](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L492), the `mode_buffer` is and remains empty because of that.
*[highlight()](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L464) will then be called with an empty string as value.
* buildSpan will then be called with an empty value.

I think `buildSpan` could also return and empty string instead of an empty span, if the input is empty, but I don't know all the effects, of course.

yyyc514: Would you mind pasting all that into a new issue perhaps? :-)

nknapp: Adding 

```
if (!leaveOpen && insideSpan==='') {
        return ''
}
```

to the top of `buildSpan` will cause the `twig`-tests to fail as well, but the only change I see in the result is a disappearing empty `<span class="xml">`-tag. I can create another PR for that.

yyyc514: I'd be happy to review it.  Need to look into if there are any deeper implications.  I'd tend to think that empty tags don't have much value though off the top of my head.

nknapp: There are other tests failing, that I haven't noticed before.

PR #2176, opened 2019-10-10
fix(YAML) Improve matching for blocks, numbers

### LINKED ISSUE #1417: (yaml) Fails on Literal Blocks followed by Comments
YAML highlighter doesn't handle well literal blocks followed by comments.

You can see the problem live at this link:

https://cdn.rawgit.com/tajmone/gh-themes-magick/baf5927/gh-themes/leapday/index.html#configuration

where the comments are treated as part of the literal block scalar even though indentation has ended:

```yaml
header-includes: |
  <!-- anything you put here will go verbatim in the header section
       Useful for adding javascripts, custom metadata, and so on... -->
# ******************************************************************************
# *                             INSERT BEFORE BODY                             *
# ******************************************************************************
```

In the [yaml.js source file (53-59)](https://github.com/isagalaev/highlight.js/blob/master/src/languages/yaml.js#L53) we read:

```javascript
      { // multi line string
        className: 'string',
        begin: '[\\|>] *$',
        returnEnd: true,
        contains: STRING.contains,
        // very simple termination: next hash key
        end: KEY.variants[0].begin
      },
```

so the problem seems that the criterion by which a liter block is considered ended is not by checking the indentation but by capturing a new key definition — therefore comments are ignored in this process. The author states in the comment: `// very simple termination: next hash key`, so it's not a bug but just the way it was implemented.

I've attempted to fix it butI don't know how to handle checking the indentation level — need to keep some record of the initial indentation level? Possibly the easiest soultion would be to add an `end` check that also considers comments as a sign that the literal block is over. 

I'd love to fix it myself, but this particular task is beyond my reach right now; so all I can do is open an issue for it.

cdbax:
Fixes #1417

This PR adds a few more improvements to the YAML matching:
* Multi-line strings no longer depend on keys for termination, but use indentation instead
* Bullets no longer include whitespace
* Numbers will only be classified as such if they're isolated from other characters

These changes fix the remaining YAML highlighting issues (that we know about), but the tests are now failing as it is detecting `ldif` as `yaml`. I had a look at the detect example for `ldif`, and the problem is immediately clear - it _is_ YAML.

This could be a problem for languages other than `ldif`, as YAML has such a permissive syntax, it has potential to match a lot of things.


yyyc514:
https://en.wikipedia.org/wiki/LDAP_Data_Interchange_Format

It seems the format supports `-` for record separators and a few other things... any chance adding those and improving our sample a bit might luckily disqualify it from matching as YAML?

cdbax:
I couldn't get LDIF to be sufficiently relevant with the record separators, but identifying a keyword and some common attributes did the trick. It feels messy to change all this in the 1 PR, but at least we can discuss it here.

After LDIF I had a new conflict with LiveScript. I increased it mostly with new examples, but had to increase the relevance of a keyword as well. This gets it scoring 1 higher than YAML.

yyyc514:
> After LDIF I had a new conflict with LiveScript. I increased it mostly with new examples, but had to increase the relevance of a keyword as well. This gets it scoring 1 higher than YAML.

I find it's usually better to turn relevance down rather than up.  Up tends to create the same problems long-term as you're just making things more sensitive.

> I couldn't get LDIF to be sufficiently relevant with the record separators

Not what I was suggesting.  I was wondering if there was a way to add them to YAML as illegal... meaning "YAML does not have record separators"... so as soon as the YAML parser saw them it would bail, leaving LDIF to clean up the mess.

Not always possible (and perhaps not even always  good idea, but it tends to solve the problem more permanently than increasing relevancy).

src/languages/yaml.js
```diff
@@ -54,12 +54,12 @@ function(hljs) {
...
+        //
+        // Indentation of subsequent lines must be the same to
+        // be considered part of the block
         className: 'string',
-        begin: '[\\|>] *$',
-        returnEnd: true,
-        contains: STRING.contains,
-        // very simple termination: next hash key
-        end: KEY.variants[0].begin
+        begin: '[\\|>][ ]*\\n( *)[\\S ]+\\n(\\1[\\S ]+\\n?)*',
```
yyyc514: Is this matching the whole block now? Do we not need STRING.contains anymore?

src/languages/livescript.js
```diff
@@ -14,7 +14,7 @@ function(hljs) {
       'in if for while finally new do return else break catch instanceof throw try this ' +
       'switch continue typeof delete debugger case default function var with ' +
       // LiveScript keywords
-      'then unless until loop of by when and or is isnt not it that otherwise from to til fallthrough super ' +
+      'then unless until loop of by when and or is isnt not it that otherwise|10 from to til fallthrough super ' +
```
yyyc514: 10?  If any number would accomplish the same for less I'd use it instead of 10.  "otherwise" is also a plain english word and you're going to make this auto-trigger for comments in other languages that mention the word otherwise... such as :

```cpp
// below should be clear, otherwise see the docs
```

src/languages/yaml.js
```diff
@@ -86,15 +86,21 @@ function(hljs) {
       },
       { // array listing
         className: 'bullet',
-        begin: '^ *-',
+      // TODO: remove |$ and |^ hacks when we have proper look-ahead support
+      begin: '((?<= )|^)?\\-(?=[ ]|$)',
```
yyyc514: What is the first part of this regex doing again?

src/languages/yaml.js
```diff
@@ -86,15 +86,21 @@ function(hljs) {
...
       {
         beginKeywords: LITERALS,
         keywords: {literal: LITERALS}
       },
-      hljs.C_NUMBER_MODE,
+      // numbers are any valid C-style number that
+      // sit isolated from other words
+      {
+        className: 'number',
+        begin: hljs.C_NUMBER_RE + '\\b'
```
yyyc514: Do we not need a boundary before also?  The last match might not always have been a \b.

src/languages/livescript.js
cdbax: Yeah fair enough - I hadn't considered that scenario, just thought it was an unusual keyword.
The value of 10 came from the language author docs:

> The default value for relevance is 1. When setting an explicit value it’s recommended to use either 10 or 0.

src/languages/yaml.js
cdbax: Yeah it's matching the whole block, as it was the only way I could find to enforce the consistent indentation rule.

src/languages/yaml.js
cdbax: It's a positive look-behind. It's saying the bullet must be preceded by a space, or be at the start of the line. That's my intent anyway, and the tests seemed to indicate it was working.

src/languages/yaml.js
cdbax: I _think_ the first boundary is included in the `C_NUMBER_RE` but I'll check and confirm.

src/languages/yaml.js
cdbax: I'll check this again actually - I'm not sure why I've made that look-behind optional...which seems like it defeats the purpose of it being there...

src/languages/livescript.js
yyyc514: Those docs should be updated. :-)

src/languages/yaml.js
yyyc514: Yeah, right now the engine really has no other way to save state like that. :-/

src/languages/yaml.js
yyyc514: I didn't see it, but I only glanced quickly.

yyyc514:
> I don't believe this will work, as it's also valid YAML. 

Ah, will no then that won't work. :-)

> My first thought was to reduce the relevance of most things in YAML to zero, as it's such a permissive syntax...but it's also a very popular format, so I think it follows that people using autodetect are more likely to be expecting YAML than LDIF. 

*nods* I just think right now too many things are all vying for relevance overall - and that makes the system very fragile, but that's a much larger problem.

> Hence I went down the path of trying to make LDIF more relevant.

By adding things unique to it, yes, that's a good call.

> I saw some tricks in some other language definitions last night where begin blocks were added just to increase relevance without having any affect on the markup. I might try using that if you aren't against it.

No, I think it can be a useful trick when it works.

src/languages/yaml.js
cdbax: Yeah it's in there, but isn't in front of the `-`...unsure what the impact of that would be:
`'(-?)(\\b0[xX][a-fA-F0-9]+|(\\b\\d+(\\.\\d*)?|\\.\\d+)([eE][-+]?\\d+)?)'`

I'll add a few tests to check it's working as expected.

yyyc514: Ah, no that's prolly ok. :-)

cdbax:
I've made some further tweaks based on our discussion. Let me you know if you think it needs further work...and thanks for all the feedback and reviews!

src/languages/ldif.js
```diff
@@ -9,7 +9,10 @@ function(hljs) {
       {
         className: 'attribute',
         begin: '^dn', end: ': ', excludeEnd: true,
-        starts: {end: '$', relevance: 0},
+        relevance: 10
+      },
+      {
+        begin:'(cn|CN|ou|OU|dc|DC)=',
         relevance: 10
```
yyyc514: Are these not attributes also?  Did we not need that starts/ends $ match?

src/languages/livescript.js
```diff
@@ -109,13 +109,18 @@ function(hljs) {
...
+  };
+
   return {
     aliases: ['ls'],
     keywords: KEYWORDS,
     illegal: /\/\*/,
     contains: EXPRESSIONS.concat([
       hljs.COMMENT('\\/\\*', '\\*\\/'),
       hljs.HASH_COMMENT_MODE,
+      SYMBOLS,
```
yyyc514: Please add a "boosts relevance" comment is that is what this is for.  That's tradition.

test/detect/livescript/default.txt
```diff
@@ -36,3 +36,21 @@ const t2 =
...
+
+class A
+  (num) ->
+    @x = num
+  property: 1
+  method: (y) ->
+    @x + @property + y
+
+a = new A 3
+a.x        #=> 3
```
yyyc514: Feels a little contrived, is this the only way you could make the detects happy?

src/languages/ldif.js
yyyc514: Oh now that's a good idea, would have been useful to see it with the tests so I had some idea what your'e trying to accomplish. :)

cdbax: Sorry - it's purely an effort to get the autodetect to not match LDIF as YAML. I figured this was closer to the original, and by adding a lot of extra `dn` entries to the LDIF default test, it seems to do the trick.

yyyc514: Yeah, sorry.  The autodetect stuff does make contributing a real pain.  I'm trying to think of some way to improve it.

test/detect/livescript/default.txt
cdbax: Yeah unfortunately that's what this is about. I just found a likely looking snippet on the LiveScript site, and copied it in here to boost the relevance.

src/languages/ldif.js
yyyc514: So we’re end and begin meant to swap here then?

cdbax: Nope that was a stuff-up. I've just pushed a fix.

src/languages/yaml.js
yyyc514: So does this mean we lose escaping and variables inside these strings?

yyyc514: Sorry I guess I asked that earlier but is your answer mean you think that losing this is a reasonable trade-off?

cdbax: I'm not super confident here, as I find the YAML spec really hard to understand...but I believe that this would be the right way to treat these block literals. My understanding is that everything inside is just that block, and it shouldn't be marked up further...but I could be wrong.

Relevant part of the spec: https://yaml.org/spec/1.2/spec.html#Block

yyyc514: wow, yeah.

Things that jumped out as pertinent to me:

> Inside literal scalars, all (indented) characters are considered to be content, including white space characters.

So in the literal form sounds like it's 100% content for sure.  And really there is nothing you could really even want to escape I suppose other than line breaks and whitespace.

Looking at the block form again now.

yyyc514: After reading that I'm tempting to say we should add Indentation indicators (0-9) and chomping indicttors (+/-) just to parse them, but I'm fine if we don't. :-)

I don't see anything about escape sequences or anything... I think I was just hung up that before we were using STRING.contains.

src/languages/yaml.js
```diff
@@ -60,7 +60,7 @@ function(hljs) {
         // Indentation of subsequent lines must be the same to
         // be considered part of the block
         className: 'string',
-        begin: '[\\|>][ ]*\\n( *)[\\S ]+\\n(\\1[\\S ]+\\n?)*',
+        begin: '[\\|>]([0-9]*[+-])?[ ]*\\n( *)[\\S ]+\\n(\\2[\\S ]+\\n?)*',
```
yyyc514: Nice!!!

src/languages/yaml.js
cdbax: They were easy enough to add, so I've put them in there.

Yeah I don't know how correct it is to be having any `contains` things inside the YAML strings. This grammar looks to have been built to cater for a specific use-case (something with Ruby + Ansible syntax inside) so I haven't gone and changed any of that, but I also haven't treated it as a constraint for fixing the blocks.

cdbax:
Very much ready for this to be done. 😅

yyyc514:
Ha, damn livescript.  I bet you lost some relevance because of blacklisted keywords?

cdbax:
Oh that's weird. The tests are still passing for me locally. 🤔

yyyc514:
Old build probably? not on the exact codebase?

cdbax:
Yeah I just merged master in here, so I'll sync everything up and get to the bottom of it.

cdbax:
Ahh I see what you're saying. Yeah the common keywords lost me 8 points which was enough.
I'll tweak the livescript default a bit more.

PR #2182, opened 2019-10-12
omit empty span tags in the output

nknapp:


src/highlight.js
```diff
@@ -418,6 +418,9 @@ https://highlightjs.org/
     }
 
     function buildSpan(classname, insideSpan, leaveOpen, noPrefix) {
+      if (!leaveOpen && insideSpan !== '') {
```
yyyc514: Wait wouldn't it be `insideSpan === ''`?  What am I missing?

nknapp: I made it wrong, changed it in the generated code, ran the tests and everything worked ;-)

nknapp:
I moved the other if-statement to the top of the method because the rest of the method does not need to be executed if not className is set.

yyyc514:
> I moved the other if-statement to the top of the method because the rest of the method does not need to be executed if not className is set.

Yeah I saw that the other day but wasn't worth a PR on it's own.  Good catch.

yyyc514:
@egor-rogov Unless you see some reason we should leave these empty tags around this seems pretty clear cut to me.

yyyc514:
@nknapp Can you update the changelog, please?

nknapp:
#2175  as well?

yyyc514:
Yes, we should always try to remember to update the changelog. :-)

PR #2184, opened 2019-10-12
(handlebars)  support for segment-literals and escaped mustaches

nknapp:
This PR adds support for rather exotic expressions like 

```handlebars
{{[property name containing spaces]}}
{{"property name containing spaces and [] square brackets"}}
{{'property name containing spaces and "" quotes'}}

{{helper [property name containing spaces]}}
{{property.[property name containing spaces]}}
```
and most importantly

```handlebars
{{"property containing a closing }} mustache"}} followed by xml {{and another mustache}}
```

I have set the `relevance` values such that no changes in auto-detection heuristics takes place.

Update: Added support for escaped mustaches as well:

```handlebars
\{{no-expression}} {{expression}}

\{{{no-expression}}} {{expression}}

\{{#no}} block \{{/no}} {{expression}}

\{{\{{no}}}} block \{{\{{/no}}}} {{expression}}
```

Update 2: looks like GitHub's syntax highlighter isn't supporting that.

### DIFF
--- a/src/languages/handlebars.js
+++ b/src/languages/handlebars.js
@@ -6,8 +6,12 @@ Description: Matcher for Handlebars as well as EmberJS additions.
 Website: https://handlebarsjs.com
 Category: template
 */
-function(hljs) {
+function (hljs) {
   var BUILT_INS = {'builtin-name': 'each in with if else unless bindattr action collection debugger log outlet template unbound view lookup'};
+
+  var IDENTIFIER_PLAIN_OR_QUOTED = {
+    begin: /("[^"]+"|'[^']+'|\[[^\]]+\]|\w+)/
+  };
 
   function BLOCK_MUSTACHE_CONTENTS() {
     return hljs.inherit(EXPRESSION_OR_HELPER_CALL(), {
@@ -23,7 +27,7 @@ function(hljs) {
   }
 
   function EXPRESSION_OR_HELPER_CALL() {
-    return  hljs.inherit(EXPRESSION(), {
+    return hljs.inherit(IDENTIFIER_PLAIN_OR_QUOTED, {
       keywords: BUILT_INS,
       starts: HELPER_PARAMETERS()
     });
@@ -34,16 +38,10 @@ function(hljs) {
       endsWithParent: true,
       relevance: 0,
       contains: [
-        hljs.inherit(EXPRESSION(), {
+        hljs.inherit(IDENTIFIER_PLAIN_OR_QUOTED, {
           relevance: 0
         })
       ]
-    };
-  }
-
-  function EXPRESSION() {
-    return  {
-      begin: /("[^"]+"|'[^']+'|\[[^\]]+\]|\w+)/
     };
   }
 
--- a/src/languages/handlebars.js
+++ b/src/languages/handlebars.js
@@ -48,36 +48,6 @@ function (hljs) {
   var ESCAPE_MUSTACHE_WITH_PRECEEDING_BACKSLASH = {begin: /\\\{\{/, skip: true};
   var PREVENT_ESCAPE_WITH_ANOTHER_PRECEEDING_BACKSLASH = {begin: /\\\\(?=\{\{)/, skip: true};
 
-  var OPEN_RAW_BLOCK = {
-    className: 'template-tag',
-    begin: /\{\{\{\{(?!\/)/, end: /\}\}\}\}/,
-    contains: [BLOCK_MUSTACHE_CONTENTS()],
-    starts: {end: /\{\{\{\{\//, returnEnd: true, subLanguage: 'xml'}
-  };
-  var CLOSE_RAW_BLOCK = {
-    className: 'template-tag',
-    begin: /\{\{\{\{\//, end: /\}\}\}\}/,
-    contains: [BLOCK_MUSTACHE_CONTENTS()]
-  };
-  var STANDARD_BLOCK = {
-    className: 'template-tag',
-    begin: /\{\{[#\/]/, end: /\}\}/,
-    contains: [BLOCK_MUSTACHE_CONTENTS()],
-  };
-  var UNESCAPED_OUTPUT = {
-    className: 'template-variable',
-    begin: /\{\{\{/, end: /\}\}\}/,
-    keywords: BUILT_INS,
-    contains: [BASIC_MUSTACHE_CONTENTS()]
-  };
-  var HTML_ESCAPED_OUTPUT = {
-    className: 'template-variable',
-    begin: /\{\{/, end: /\}\}/,
-    keywords: BUILT_INS,
-    contains: [
-      BASIC_MUSTACHE_CONTENTS()
-    ]
-  };
   return {
     aliases: ['hbs', 'html.hbs', 'html.handlebars'],
     case_insensitive: true,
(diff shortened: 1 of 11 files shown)


nknapp:
The PR is based on the commits of #2182 so we should probably wait until that one is merged.

yyyc514:
Lets not use functions if we don't need them. I think almost most of those cases can be normal variables. Only prefer the function when you need a unique object to make the parser happy.

For example no reason for EXPRESSION to be a function.

src/languages/handlebars.js
```diff
@@ -5,19 +5,44 @@ Author: Robin Ward <robin.ward@gmail.com>
 Description: Matcher for Handlebars as well as EmberJS additions.
 Category: template
 */
-function(hljs) {
-  var BUILT_INS = {'builtin-name': 'each in with if else unless bindattr action collection debugger log outlet template unbound view yield lookup'};
-  function MUSTACHE_CONTENTS() {
-    return  {
-      className: 'name',
-      begin: /[a-zA-Z\.-]+/,
+function handlebars(hljs) {
```
yyyc514: Lets not change the intro, some of the existing build process is very picky about this... so just `function(hljs)` please.

nknapp: What I like about functions, is that you can write them top-down in the order in which you would read the code. 

```js
function main() {
  if(isThis()) {
    doThat()
  }   
}

function isThis() {...}

function doThat() {...}
```

Mostly, functions are used the other way around, 

```js
function isThis() {...}

function doThat() {...}

function main() {
  if(isThis()) {
    doThat()
  }   
}
```
but I think it makes the code less readable, because when you start reading at the top, you get all the helper functions first and come to the actual calling code last.

But in that case, I actually started replace some functions by variables, and had all tests failing. I can try some more if you like.

nknapp:
To be honest, I'm not happy with changing this, because it really messes up the reading order.
I actually would move all functions to the bottom (behind the return) and even split more out of them.

Maybe I'll try what it does to the minified code.

yyyc514:
> To be honest, I'm not happy with changing this, because it really messes up the reading order.

To me it's a consistency thing (with every other grammar) and the functions add a lot of boilerplate.   @egor-rogov Any strong opinions on this?

I do think it's a little overkill if you LITERALLY have just a variable and that then you should just deal with the order not being perfect. :-)

Are there any implications here about minification and what's possible or not (whether they are variables vs functions)?

nknapp:
I just started writing a really long, but it is too long. Maybe I'll write a blog post about this. 
The short version is: I try to apply principles from the book "Clean Code" by Robert C. Martin. But in this particular case it does not work because "Clean Code" is mostly about Java and byte size of the compiled program is not critical there.

**So I agree with your request**.

I would extract even more functions, but I what I have tried would result in 300 bytes more code.

In fact, if we find out why reusing the "starts" object is a problem, and fix it, then we should remove the other functions as well.

If you are interested: The minified code shows why that is is the case

```js
hljs.registerLanguage("handlebars", function (e) {
  var t = {"builtin-name": "each in with if else unless bindattr action collection debugger log outlet template unbound view lookup"},
      a = {b: /("[^"]+"|'[^']+'|\[[^\]]+\]|\w+)/};

  function n() {
    return e.inherit(r(), {cN: "name"})
  }
...
```
The `function`, `()`, `{`, `return`, `}` of `function n()...` cannot be removed, whereas `a={...}` just requires a `,` before the `a` and a `=`.

yyyc514:
I agree in general.  I love Clean Code and clean code.  Guess right now I think of the grammars more as data than code, though.

I also think we've been too paranoid in the past about size and `gzip` solves most of these problems.  So I think mostly I was coming from a style and consistency standpoint when I made the original request to stick with variables.  We'll see what Egor says. :-)

src/languages/handlebars.js
```diff
@@ -5,19 +5,42 @@ Author: Robin Ward <robin.ward@gmail.com>
...
-  var BUILT_INS = {'builtin-name': 'each in with if else unless bindattr action collection debugger log outlet template unbound view yield lookup'};
-  function MUSTACHE_CONTENTS() {
-    return  {
-      className: 'name',
-      begin: /[a-zA-Z\.-]+/,
+function (hljs) {
+  var BUILT_INS = {'builtin-name': 'each in with if else unless bindattr action collection debugger log outlet template unbound view lookup'};
+
+  var IDENTIFIER_PLAIN_OR_QUOTED = {
+    begin: /("[^"]+"|'[^']+'|\[[^\]]+\]|\w+)/
```
yyyc514: Wouldn't back references be a better use here rather than repeating yourself 3-4 times?

nknapp: If at all, only the first two alternatives could be combined using back-referencens.

Non-greedy matching would make it better readable (and it must be `.*`, not `.+`, because `''`, `""` and `[]` are also allowed (in theory):

```js
 /".*?"|'.*?'|\[.*?\]|\w+/
```

But I don't think that this is intuitive:

```js
/(['"]).*?\1|\[.*?\]|\w+/
```
I actually wanted to use: 

```js
  var IDENTIFIER_PLAIN_OR_QUOTED = {
    begin: /\w+/,
    variants: [
      {begin: /"/, end: /"/},
      {begin: /'/, end: /'/},
      {begin: /[/, end: /]/}
    ]
  };
```
but that didn't work. The main begin-pattern `/\w+/` seemed to be ignored in this case.

test/markup/handlebars/escaped-mustaches.txt
```diff
@@ -0,0 +1,7 @@
+\{{no-expression}} {{expression}}
+
+\{{{no-expression}}} {{expression}}
+
+\{{#no}} block \{{/no}} {{expression}}
+
+\{{\{{no}}}} block \{{\{{/no}}}} {{expression}}
```
yyyc514: What if I want to skip a comment like snippet?  Won't those rules still match?

src/languages/handlebars.js
yyyc514: Non-greedy is more readable.

Variants are alternatives to the main block, not "in addition to" or "after"... If you just removed `begin` would the variants not work?

test/markup/handlebars/escaped-mustaches.txt
nknapp: No. You are right, that one is still missing. I'll add that.

nknapp: A `\` in front of the mustache is escaped by another `\` in front of it.
This means that the expressions in 

```handlebars
\\{{expression}}

\\\{{expression}}

\\\\{{expression}}
```
are all evaluated. I have added a commit that skips the escaping if another backslash appears in front of the backslash.

yyyc514: I mean don’t the escape rules need to come before the comment rules to allow escaping a comment?

yyyc514:
Sorry I might have just messed ya up :) You can just force push without my commit if it's no good.

test/markup/handlebars/escaped-mustaches.txt
nknapp: I have added a test for escaped comments and it passed without modification of the "handlebars.js" file.

I think, since the match of `\{{` has a lower index than the matching `{{`, it will always be evaluated first.

Also, after writing my own review, I felt I had to extract the rules into there own variables with descriptive names. Sorry for all the changes that you didn't ask about.

I also added consistent escaping of `{` and `}` in the regular expressions of the language. I don't think that is really necessary and I thought about removing all backslashes at first. But since the regex are assembled into larger once, we cannot be sure. There ARE contexts where a `{` needs to be escaped.

yyyc514: > I think, since the match of \{{ has a lower index than the matching {{, it will always be evaluated first.

Yeah, I thought I saw the comment regex coming FIRST, which is why i was curious.

yyyc514: > Also, after writing my own review, I felt I had to extract the rules into there own variables with descriptive names. 

Do the names have meaning if you know Handlebars really well?  I'm actually not sure it's a win (while normally I'd call it a win)... but to me I SEE the patterns and that  means something to me, but I see the labels you've assigned them and that means nothing to me.

yyyc514: If it were me I'd add comment examples beside them.

```
  STANDARD_BLOCK, // {{ ... }}
```

nknapp:
I'm think its good like that

nknapp:
Thanks for all the review work you put in there.

nknapp:
Argh no. I noticed that I removed the yield keyword. It should be added again. Its not important for Handlebars but it was in there and if keyword cleanup is something we want to do, of should be a separate PR

nknapp:
I have reverted the keyword list and rebased on master. If the tests pass, from my point of view, you can merge.

src/languages/handlebars.js
```diff
@@ -6,63 +6,91 @@ Description: Matcher for Handlebars as well as EmberJS additions.
 Website: https://handlebarsjs.com
 Category: template
 */
-function(hljs) {
+function (hljs) {
```
yyyc514: This kind of change just bit me today. :)  Although a few of the other languages are also so different.

yyyc514:
@nknapp Thanks for all the hard work!

PR #2188, opened 2019-10-13
fix(Go): improve function matching

### LINKED ISSUE #2177: Go highlighting bug with `func`
This piece of code:

```go
TLSNextProto: make(map[string]func(*http.Server, *tls.Conn, http.Handler), 0),
```

Seems to cause issues with further parsing. A good example of this in action can be seen here:
https://ssl-config.mozilla.org/#server=golang&server-version=1.13.1&config=intermediate

Where the comment about ECDHE is red and bolded, instead of muted and gray. If you replace `func` with `foo`, everything seems to work correctly.

cdbax:
Fixes #2177 

The existing implementation ended function matching at the opening of the function body `{`.
This approach didn't work for function type declarations.

Neither implementation is perfect, as the function body should probably be wrapped by the function tags. A shortcoming of this implementation is that it doesn't wrap the return type in the function tag, and for some reason it isn't detecting the return type as 'params'.

I'll keep trying things to improve it further. Suggestions would be welcome.


src/languages/go.js
```diff
@@ -45,14 +45,15 @@ function(hljs) {
       },
       {
         className: 'function',
-        beginKeywords: 'func', end: /\s*\{/, excludeEnd: true,
+        beginKeywords: 'func', end: /\B\b/,
```
yyyc514: If this is still goin to be a single line you’d just do {|$

Bracket or end of line.

src/languages/go.js
```diff
@@ -45,14 +45,15 @@ function(hljs) {
...
         contains: [
           hljs.TITLE_MODE,
           {
             className: 'params',
             begin: /\(/, end: /\)/,
             keywords: GO_KEYWORDS,
-            illegal: /["']/
+            illegal: /["']/,
+            endsParent: true
```
yyyc514: This will end the parent mode after you’d see a single set of parents. So that’s why it will never match a second set of parents.

cdbax:
That's much nicer. Thanks!

src/languages/go.js
yyyc514: I’d still gobble up the white space first though.

cdbax: Would you still use `\s*`? I'm guessing it won't matter that that includes newlines, as the `$` would presumably match first.

yyyc514: I’m not sure it does. Try it and see what the tests do.

cdbax:
Updated. It's still not ideal, as I'd prefer to get it so the function tag wrapped the body as well, but fixing that is a much deeper rabbit hole. I think this is an improvement for now that fixes the original issue, and I'll keep the bigger fix as an exercise for another day.

Also - I discovered the question about newlines was moot as that would be invalid Go.

yyyc514:
> Updated. It's still not ideal, as I'd prefer to get it so the function tag wrapped the body as well, but fixing that is a much deeper rabbit hole.

None of the other grammars (that I'm aware of do this) and I'm not entirely sure whether it's "in scope"... I'm afraid it would make the parsers much more complex if added to all of them and therefore increase the size significantly.  Take a look at the grammars for something like VS Code and you'll see they are usually very large.  That's the price of trying actually parse the full grammar vs just looking for useful patterns to highlight.

@egor-rogov Do you have any thoughts on this?  Surely if the size/complexity was small I think it'd be great if our parsers were smarter... so I wouldn't think we'd object... but if the code/patterns got **significantly** more complex, or the size became 500% larger, would that be a reason to reject a change a long these lines?

yyyc514:
Does my last commit still look good to you?

test/markup/go/functions.txt
```diff
@@ -0,0 +1,9 @@
+func name()
+
+func name(int32, *somePointer)(string)
+
+map[string]func(int32, string)
+
+func name (x int32) int32 {
```
yyyc514: Good addition.

cdbax:
Yeah that looks great. 👍

yyyc514:
Merging after tests run again..  @cdbax Thanks!

PR #2199, opened 2019-10-14
Fix TypeScript function issue

### LINKED ISSUE #2191: (javascript) Naive highlighting of keywords
In Typescript (and most likely in some other scenarios in Javascript) the following snippet is erroneously highlighted:

```typescript
type Foo = {
    functionInFoo(): void;
};
```

image

The library is naively highlighting `function` since it's a keyword but not considering it might be part of a different identifier (even though it is arguably a bad name anyways).

cdbax:
Fixes #2191 

The problem here seems to be that TypeScript is using `begin` whereas JavaScript is using `beginKeywords` and therefore doesn't suffer the same problem. The solution seems self-evident.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New styles:
 
 Improvements:
 - fix(typescript): constructor in declaration doesn't break highlighting
+- fix(typescript): only match function keyword as a separate identifier (#2191)
 - feature(arduino) make arduino a super-set of cpp grammar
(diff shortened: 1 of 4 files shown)


yyyc514:
This is cause beginKeywords adds word boundaries:

```
        if (mode.beginKeywords) {
          mode.begin = '\\b(' + mode.beginKeywords.split(' ').join('|') + ')\\b';
        }

```

PR #2200, opened 2019-10-14
fix(python) if incorrectly identified as f-string

### LINKED ISSUE #1809: (python) f-string incorrectly matches "if" followed by string
For Python, `if` followed by a string, without a space in between, highlights as an f-string.
For example:
```py
if"text"=="text":
    pass
```
becomes
```html
i<span class="hljs-string">f"test"</span>==<span class="hljs-string">"test"</span>:
    <span class="hljs-keyword">pass</span>
```
rather than
```html
<span class="hljs-keyword">if</span><span class="hljs-string">"test"</span>==<span class="hljs-string">"test"</span>:
    <span class="hljs-keyword">pass</span>
```

cdbax:
Fixes #1809 

Putting this up in the hope that someone can help with the auto-detect issue.

Per the suggestion from @yyyc514 this adds a requirement for a word boundary at the start of Python's string literals. This suggestion does indeed fix the f-string issue.

The weird part that I don't understand, is the auto-detect now thinks the Python sample is `awk`. Even after adding to the sample, and increasing the Python score to 52, the auto-detect still comes up with awk (34) and ruby (24). If it isn't highest-score wins, then what is the logic?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,8 @@ New styles:
   none.
 
 Improvements:
-- enh(powershell) major overhaul, huge improvements (#2224) 
+- fix(python) fix `if` getting confused as an f-string (#2200) [Josh Goebel][] and [Carl Baxter][]
+- enh(powershell) major overhaul, huge improvements (#2224)
 - enh(css) Improve @rule highlighting, including properties (#2241) [Josh Goebel][]
 - enh(css) Improve highlighting of numbers inside expr/func `calc(2px+3px)` (#2241)
 - enh(scss) Pull some of the CSS improvements back into SCSS (#2241)
@@ -16,6 +17,7 @@ Improvements:
 - fix(parser): Better errors when a language is missing (#2236) [Josh Goebel][]
 
 [Josh Goebel]: https://github.com/yyyc514
+[Carl Baxter]: https://github.com/cdbax
 
 
 ## Version 9.16.2
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -33,43 +33,43 @@ function(hljs) {
     contains: [hljs.BACKSLASH_ESCAPE],
     variants: [
       {
-        begin: /\b(u|b)?r?'''/, end: /'''/,
+        begin: /(u|b)?r?'''/, end: /'''/,
         contains: [hljs.BACKSLASH_ESCAPE, PROMPT],
         relevance: 10
       },
       {
-        begin: /\b(u|b)?r?"""/, end: /"""/,
+        begin: /(u|b)?r?"""/, end: /"""/,
         contains: [hljs.BACKSLASH_ESCAPE, PROMPT],
         relevance: 10
       },
       {
-        begin: /\b(fr|rf|f)'''/, end: /'''/,
+        begin: /(fr|rf|f)'''/, end: /'''/,
         contains: [hljs.BACKSLASH_ESCAPE, PROMPT, LITERAL_BRACKET, SUBST]
       },
       {
-        begin: /\b(fr|rf|f)"""/, end: /"""/,
+        begin: /(fr|rf|f)"""/, end: /"""/,
         contains: [hljs.BACKSLASH_ESCAPE, PROMPT, LITERAL_BRACKET, SUBST]
       },
       {
-        begin: /\b(u|r|ur)'/, end: /'/,
+        begin: /(u|r|ur)'/, end: /'/,
         relevance: 10
       },
       {
-        begin: /\b(u|r|ur)"/, end: /"/,
+        begin: /(u|r|ur)"/, end: /"/,
         relevance: 10
       },
       {
-        begin: /\b(b|br)'/, end: /'/
+        begin: /(b|br)'/, end: /'/
       },
       {
-        begin: /\b(b|br)"/, end: /"/
+        begin: /(b|br)"/, end: /"/
       },
       {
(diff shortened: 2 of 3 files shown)


yyyc514:
This doesn't work because the boundary between a space and a quote is not a word boundary.

`a'` has a word boundary... `[space]'` has no word boundary... 

So the APOS_STRING_MODE rule is matching instead... which gives you weird behavior. Ie, you're maybe solving one thing, but breaking a lot more.  Seems my thought wasn't quite sufficient here.  Thinking about this.....

yyyc514:
This is probably why Prism wrote their own poor mans look-behind feature. :)

yyyc514:
Ok here is how I think you'd do it.  You'd write a string matcher function (that returns the full mode) to make this easier... so for every variant you need a nested matcher:

The outside the matcher matches the prefix AND the string... so it  needs to deal with whitespace, word boundaries, start of content... then it will `returnBegin`(to get all the content back into the queue) and you will have `contains` to do the actual work.

Your first contains item will simply eat the prefix again (silently), then your second one is the ACTUAL string matcher.

Does that make sense?

yyyc514:
IE what you'd want to generate would look like this for a single string type (off the top of my head):

```
      {
        // match the whole string
        begin: /(\b|\s+|^)(u|b)?r?'''.*?'''/,
        
        returnBegin: true,
        contains: [
          { begin: /(\b|\s+|^)/, relevance: 0 }, // skip the prefix
          { begin: /(u|b)?r?'''/, end: /'''/, relevance: 0,
          contains: [hljs.BACKSLASH_ESCAPE, PROMPT],
        }
        ]
        relevance: 10
      },
```

cdbax:
I haven't fully grasped everything you've explained yet - but you've been super-detailed with your guidance so I'm sure it will make sense when I spend some time on it. This feels like a good step into some more complex work, so I'm absolutely up for tackling this.

cdbax:
I haven't started testing this, but in the interest of early feedback, is this close to what you were thinking?

```js
  function advMode(args){
    return {
      // match the whole string
      begin: reStr(args.lookbehind) + reStr(args.begin) + '.*' + reStr(args.end),
      
      returnBegin: true,
      contains: [
        { begin: args.lookbehind, relevance: 0 }, // skip the prefix
        { 
          begin: args.begin, 
          end: args.end, 
          relevance: 0,
          contains: args.contains,
        }
      ],
      relevance: args.relevance
    };

    function reStr(re) {
      return (re && re.source) || re;
    };
  };
```

yyyc514:
That's the rough shape of things yes!

Technically now that we're making it more generic you'd lose the ` + '.*' + reStr(args.end),``... we're not guaranteeing the future, just the past.

```js
  function advMode(mode){
    if (!mode.lookbehind)
        return mode
    return {
      // match the whole string
      begin: reStr(mode.lookbehind) + reStr(mode.begin) 
      
      returnBegin: true,
      contains: [
        { begin: mode.lookbehind, relevance: 0 }, // skip the prefix
        { 
          begin: mode.begin, 
          end: mode.end, 
          relevance: mode.relevance,
          contains: mode.contains,
        }
      ],
      relevance: 0
    };

    function reStr(re) {
      return (re && re.source) || re;
    };
  };
```

And its probably best to actually move the relevance to the deepest match that way we only get it if the full rule applies.

yyyc514:
What you were trying to do with the "look ahead" stuff that I just said we should remove could probably be done better on the `begin` expression itself with a REAL look-ahead, once my branch supporting those is merged... but right now we're find without it since none of our rules really look ahead to see if they can be completed anyways. :-)

yyyc514:
@isagalaev You might find this interesting.  Along the lines of your syntactic sugar idea from elsewhere. :-)  If it goes well of course eventually we might wrap it into the library proper.

yyyc514:
@cdbax Any luck?  I was hoping this might go into .11 which I was planning to release over the weekend, but maybe we push it to .12?

cdbax:
@yyyc514 probably best to push it out. I'm afraid I haven't been able to get back to this for the past week or so. I'm getting back to it now, but I'm equally happy for someone else to run with it if I'm holding things up.

yyyc514:
It can wait until .12 if it most.  No rush, me thinks. I was just wrapping things up for .11 and thought I'd check in.  You still have a little time. :)

cdbax:
@yyyc514 I'm not making any headway with this. I keep getting into an infinite loop when I try to skip over the lookbehind. I can't tell if it's just me not knowing enough, or if what we're trying to do with the existing setup isn't possible.

I haven't pushed anything new up, as I've just tried variations on your code which haven't helped.

yyyc514:
What about this?  I gave up on making it super abstract... that gets complex fast because of the fact that "variants" are virtual... so very quickly you'd run into the case where the variant didn't "inherit" from the parent properly because technically we're rearranging the tree and inside extra layers of parents...

So I just focused on the lookahead stuff itself.

yyyc514:
Except this doesn't work because it breaks inheritance on `contains`...

yyyc514:
Sorry, I may have sent you down an impossible rabbit hole... now I'm not sure it was solvable that way even if you got it perfect... I tried a few different things, but when I'd get it all working I'd actually still be left with the original problem, lol...

Because the problem is there might be a word boundary, or there might not be... really it turns out to be far simpler to just handle `if` specially...

yyyc514:
Throwing over to egor for review.

PR #2201, opened 2019-10-14
endWithParent inside a starts block needs it's own mode and starts mode deep cloned 

yyyc514:


yyyc514:
@nknapp I gave you credit on the test commit (though it got expanded a little).

The problem is the parser knew it needed to clone modes when endWithParent was used in general but that knowledge didn't extent to when it was used inside `starts`.  That still creates a dependency issue but now it requires "deep cloning" so that both the mode and the start mode are cloned (so they have individual instances so that they end matchers can be unique for each instance).

yyyc514:
I'm open to other suggestions on `deregisterLanguage`, but we need a way to unhook test languages that are only used for testing individual items so they don't break the larger detection tests, etc.

@egor-rogov This is a little complex but I think the tests (and my explanation above) show what the problem is.

nknapp:
Can't you reset the whole `hljs`-global to a new instance of highlight.js before running the detection tests?

Or maybe something create something like the `Handlebars.create()` method, that creates a new instance of highlight. That instance could be used in the "bugs"-tests

yyyc514:
> Can't you reset the whole hljs-global to a new instance of highlight.js before running the detection tests?

Maybe.  I don't know how to do that.  It was imported.

> Or maybe something create something like the Handlebars.create() method, that creates a new instance of highlight. That instance could be used in the "bugs"-tests.

Adding extra code just to create multiple instances of the library sounds a little annoying.  It would definitely be better if import could just handle that somehow.  I thought perhaps the import would be isolated per file but it seems somehow the state is global.

> Deregistering the language seems error-prone to me, because it is just an incremental step, not a complete reset.

There is really no other shared state though... so it's pretty safe.

yyyc514:
Maybe @egor-rogov Knows the magic sauce. :-)

nknapp:
In Node, you can clear the require-cache by using `delete require.cache[require.resolve('../../build')]'.
This should enforce the re-evaluation of the module code, and would create new instance of hljs.

Maybe you could try that.

yyyc514:
No interesting in adding that to every other file:

```
describe("bugs", function () {
  after(function() {
    // hljs.deregisterLanguage("test-language")
    delete require.cache[require.resolve('../../build')];
  })
```

Does not work.

yyyc514:
If you were being paranoid you'd insist on a new HLJS for every test, but we don't do that.  We (from what it appears) use ONE shared HLJS for ALL tests... and mostly that just works - and has for a long time. ;-) Until now.

nknapp:
You could clear the cache prior to running the detection tests. It seems to be most important in this place.

yyyc514:
```
'use strict';

delete require.cache[require.resolve('../../build')];

let bluebird = require('bluebird');
let fs       = bluebird.promisifyAll(require('fs'));
let hljs     = require('../../build');
let path     = require('path');
let utility  = require('../utility');

function testAutoDetection(language) {
```

No effect.  If you want to try to get it working, be my guest? :-)

yyyc514:
> https://github.com/mochajs/mocha/issues/536

Looks like mocha simply doesn't handle this type of isolation desire as well as some other frameworks...

yyyc514:
> Could you come up with a more descriptive test case which shows the actual use of this feature? I mean, the current test case can be easily rewritten without starts.

How would you improve it?  It's a spot on reproduction of the issue, paired with the fix.  This isn't a feature, it's a bug fix.  I don't think the relevancy of the test case in the real world is important here, it's that it clearly reproduce the problem the simplest way possible (so we aren't testing too much).

It's definitely a bug.

> Do we need deregisterLanguage at all? Why not simply use unique names?

The auto-detect stuff messes everything else up because these languages start claiming relevancy. So you really need a way to "remove" them.

nknapp:
@egor-rogov the bug came up when I tried to reuse the [https://github.com/highlightjs/highlight.js/blob/master/src/languages/handlebars.js#L29](EXPRESSION_OR_HELPER_CALL) expression in multiple rules. You can fix it by using function to create new instances of the rule (like I did), but if you use a variable instead of any of the functions that you EXPRESSION_OR_HELPER_CALL, the tests will fail.

@yyyc514 I can try to check why the require-cache is not clear, but I don't know if I can do it today. It bugs me, but I have a family upstairs ;-)

egor-rogov:
Hm. Maybe simply get rid of functions in handlebars then? Instead of making a separate test case.

yyyc514:
We'll do that too, but I don't consider handlebars a sufficient test for a parser bug.  Handlebars could change again in the future... it's not EXPLICITLY testing this.  This test is.

nknapp:
@yyyc514 This worked for me.

```
delete require.cache[require.resolve('../../build')]
delete require.cache[require.resolve('../../build/lib/highlight')]
```

yyyc514:
@nknapp Would you mind rejiggering handlebars and laying a commit on top of this PR, just so we keep things together.  That might be the best of all worlds - to give a real life example of what we're fixing.  But the explicit test case needs to stay.

nknapp:
Can I commit to your branch?

yyyc514:
Closing in favor of https://github.com/highlightjs/highlight.js/pull/2207

PR #2203, opened 2019-10-14
Add future to cpp keyword list. Fixes #1610

### LINKED ISSUE #1610: (cpp) [cpp11] std::future not highlighted
> Highlight.js 9.12.0

Compare the two following lines:

std future

The highlighting of `std::shared_ptr` is correct, but `std::future` is wrong. `future` should be colored as well.

externl:


yyyc514:
Please update changelog also.  Looks great!

yyyc514:
@externl Thank you do much!

Issue #1119, opened 2016-03-18
Reconsider :common set


isagalaev:
Cc: @Sannis, @sourrust 

Hi guys,

I've had a few ideas about updating the :common set of languages to better match reality and expectations. 

Here's the download stats from the site, by language:

```
 824 xml
 817 javascript
 776 json
 773 css
 690 sql
 680 php
 673 http
 657 markdown
 642 java
 638 bash
 611 python
 601 cpp
 592 apache
 587 diff
 587 cs
 585 ini
 582 ruby
 572 makefile
 571 coffeescript
 567 nginx
 561 objectivec
 556 perl
 551 dts
 131 yaml
 131 go
 100 swift
  96 less
  92 scss
  85 lua
< 80 — not interesting
```

The 500+ top are the current :common set (with "dts" being there by mistake, we've missed it during review).

I've got two alternative ideas:
- Drop the least used languages: coffeescript, objectivec, nginx, perl — and add in a few of the contenders: yaml, go, swift. This will freshen it up and leave it about the same size of ~40K.
- Extend the set considerably to include most non-niche languages. It'll probably move it into ~200K territory (uncompressed) but the Internet has changed considerably, and may be it's not that big of a deal these days. I could probably run a quick poll about it on Twitter.

What do you think?

Sannis:
I'm on second varian. But it looks we need to increase major version to change this set?

derhuerst:
Please introduce a new major version and create standalone definitions for almost all languages. :package: :tada:

isagalaev:
Versions changing is cheap, it's a technicality :-)

@derhuerst we already have all the languages available separately on CDNs at `languages/<name>.min.js`

derhuerst:
> @derhuerst we already have all the languages available separately on CDNs at languages/<name>.min.js

Yes, but to have them as standalone [CommonJS-compatible](http://wiki.commonjs.org/wiki/CommonJS) NPM modules. (;

isagalaev:
@derhuerst the NPM build already includes all the languages, there's no point in packaging them individually.

derhuerst:
> @derhuerst the NPM build already includes all the languages, there's no point in packaging them individually.

There is, for example in reducing the bundle size when using browserify, which seems to be the standard frontend workflow nowadays. It would also help keeping major version bump down to a minimum in the core.

Moving the core detection into a separate module would also be nice, as many non-browser tools could directly use it (see #1086).

[...]

PR #2204, opened 2019-10-14
add languages to the :common list

yyyc514:


yyyc514:
Discussion here.  

https://github.com/highlightjs/highlight.js/issues/1119

Once we're decided I can type up a nicer changelog message.

yyyc514:
> And what about coffeescript?

I say leave it, it's small. These files are all so small. :-)  Powershell is the only pseudo-popular annoying sized package, but even it blends in decently.

> I am worried a bit by lack of a clear criterion though. Maybe we can/should get access to some statistics from github or something like that?

Sure, whatever. :-)  Suggestion?

yyyc514:
Screen Shot 2019-10-14 at 7 12 15 PM

yyyc514:
We don't even have SQLPL or HCL. :-)

egor-rogov:
> > I am worried a bit by lack of a clear criterion though. Maybe we can/should get access to some statistics from github or something like that?
> 
> Sure, whatever. :-) Suggestion?

Just a thought for the future. It requires some research and I don't feel like doing it now. I'm personally fine with the list you suggested.

yyyc514:
Would you want to go ahead and add Powershell (because it's popular) and bump the minimal build from 70kb to 105kb or so?

egor-rogov:
Honestly I have no definite opinion on this. It's up to you (:

PR #2207, opened 2019-10-15
endWithParent inside a starts block needs it's own mode and starts mode deep cloned (with handlebars converted)

nknapp:
@yyyc514 I have rebased your branch onto the master and converted the function calls in the handlebars-language into variables. Works like a charm. 

Replaces #2201

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ New styles:
 Improvements:
 - blacklist super-common keywords from having relevance (#2179)
 - fix(swift): support for `@dynamicMemberLookup` and `@propertyWrapper` (#2202)
+- fix: `endWithParent` inside `starts` now always works (#2201)
 - fix(typescript): constructor in declaration doesn't break highlighting
 - fix(typescript): only match function keyword as a separate identifier (#2191)
 - feature(arduino) make arduino a super-set of cpp grammar
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -801,6 +801,7 @@ https://highlightjs.org/
     }
   }
 
+  // needed to unhook test languages after adding
   function deregisterLanguage(name, language) {
     var lang = languages[name]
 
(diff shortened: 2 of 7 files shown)


yyyc514:
I'll review later when I have more focus, but looks about right!  Thanks!  And thanks for the deregister fix also... did you also want to yank the API from hilightjs code?

yyyc514:
@egor-rogov There, now we have the change that this makes possible in the PR also, is that better for you?

egor-rogov:
> @egor-rogov There, now we have the change that this makes possible in the PR also, is that better for you?

Almost (:
We don't need `deregisterLanguage()` anymore, do we?
I see your point on the test case and agree. The only thing `test/bugs` doesn't seem a good name to me. This is all about bugs. Maybe `test/parser`?

yyyc514:
yes  It should be safe to remove now. That is a better name for sure

yyyc514:
@nknapp Do you want to finish up or you want me to take this to the finish line?

nknapp:
It would be great if you could finish this.

yyyc514:
There has got to be some easier way for two people to collaborate on one PR.

yyyc514:
Pushing a new PR of my own now that I'm taking this back.

nknapp:
As a collaborator of the repo, you can push on my branch. The other way around is the difficult way

yyyc514:
I didn’t see how.

nknapp:
You have to add the remote URL of my fork to your local clone. Then you can pushbto the PR branch. GitHub allows this when a PR author has checked "allow maintainers to push to my branch". But it only works if you have push access to the repo that the PR was created for, so I couldn't have pushed to your branch

nknapp:
At least I believe it works that way

yyyc514:
So you're saying I could have pushed to your master? Interesting.  I almost considered just trying it... I will next time. :)

nknapp:
To "nknapp:deep_clone...", yes. I think you still can. But I'm not sure

yyyc514:
At the time it said it was from your master IIRC.  But I could be mistaken.

PR #2209, opened 2019-10-16
(fix) endWithParent inside a starts block needs to trigger deep cloning

yyyc514:
Often rules (modes) can be static (ie, you can reuse the same rule over inside your grammar, making it a constant, etc. This is a very common pattern.  This works except for when `endWithParent` is used because then the rule becomes dependent on it's parent (due to the way we compile matching regexes).  If there was only a single static instance of the rule it couldn't possible have the correct "state" shared across many parents... so the parser detects these situations and clones the mode by default - so what might start out as a single constant becomes many different objects - each instance can then correctly inherit end matching rules from it's associated parent - and it all works smashingly.

Unfortunately, this logic did NOT apply when `endWithParent` was used inside a `start` block.

This resolves that issue.

- Renames `expand_mode` to `expand_or_clone_mode`
- Adds `dependencyOnParent` to abstract the concept of dependency


yyyc514:
@egor-rogov Removes deregister.  Renamed bugs to parser.

Issue #2149, opened 2019-10-04
Policy: Discuss extras repository for additional languages


yyyc514:
I see we have a lot of requests for new languages in the PRs... if the issue is time/maintenance over time, etc... might we perhaps consider an "extra" respository or something with community/unsupported syntaxes?  That way the criteria to be approved could be lessened a little and obscure languages that might not really make sense in Highlight.js proper could still have a home?

Or is the idea that eventually we'll get to them?

yyyc514:
Are we still moving languages into repositories?  Is that ALL the languges, or only "additional" ones?

marcoscaceres:
Only new ones because then we can make submitters the repo maintainers.

yyyc514:
So what is the policy on new (or very old) PRs already in the queue?  Is there some cut off date or pretty much the answer now is always "separate repo"?

marcoscaceres:
Yep, separate repo.

jf990:
The separate language repos concept addresses a maintenance problem that has caused consternation here since @isagalaev slowed down the regular maintenance. I think the separate repo idea is going to address that but we need to discuss the many problems it introduces. We should decide how to handle these things because they change the way this library has been managed since its inception.

1. **Discovery**. Available languages and how to use them are documented on [highlightjs.org](https://highlightjs.org/static/demo/). By putting new languages in separate repos we are effectively invalidating this entire page. No longer could we have a single source of truth as to what languages are available and how to get them. Even worse, old languages that were part of the original highlight js will be handled and documented differently than the new ones. this will make the library difficult to use and documentation will be all over the place as different language authors choose their own way of doing things without the oversight of the maintainers. We should figure out and document some minimum standards for language contributions. (To do this right) we should even move all the existing languages into this new format such that all languages are handled the same making using the library consistent. We should also come up with a meta data file that each language repo is responsible to maintain so that we can automate some aspects of the discovery, testing, detection, and packaging.

2. **Auto-detection**. I'm not sure we are going to easily have language auto-detection work as well as it does now. with the separate repos there's no easy way to run the test against all the other languages and determine if relevance is working.

3. **Testing**. Current testing methodology is well defined and there is a [process to handle it](https://highlightjs.readthedocs.io/en/latest/building-testing.html). This method is no longer valid and individual repos are on their own. So far with the new languages definitions we have not established any requirement for testing, and we see that some do and some don't.

[...]

PR #2213, opened 2019-10-17
Dockerfile for container development

vsoch:
This pull request will add a Dockerfile to build a container that will easily install dependencies, build highlight.js, and then the user can run the container to open the browser and see the tools/developer.html page easily.

An example logging syntax (default theme) running in the container is shown below:

image

Additionally, I updated the build and testing docs to have a section for how to do the same with the Docker container. I've checked the [language contribution](https://highlightjs.readthedocs.io/en/latest/language-contribution.html) to include proper metadata, and additions to authors and changelog markdown files.

Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>


yyyc514:
Sadly the contribution info is a little out of date.  Currently all new languages live in their own repositories (and we can set one up for you at the highlightjs organization if you'd like). 

You can follow the larger related discussion here if you're interested:
https://github.com/highlightjs/highlight.js/issues/2149

vsoch:
Why isn't there documentation for how to do this? I spent my morning working on this, and now I need to do it over, and not even with good instructions for doing so :(

vsoch:
Actually it looks fairly trivial - just moving the new language files into the same directories under a new repository? I'd be happy to do that. I'd still like to submit this PR here to add the Docker build option - installing npm/node on my host is never something I'm super excited to do, and I'd suspect others might want a container development option!

yyyc514:
> Why isn't there documentation for how to do this? I spent my morning working on this, and now I need to do it over, and not even with good instructions for doing so :(

Lack of time. I can point you to some examples, but it's not 100% agreed upon the best way to pull all these separate things back together yet - so that makes writing instructions a bit difficult.  It's something we need to sort out soon, for sure.

>need to do it over

It might be better to say package it differently.  As you're realizing.

> Actually it looks fairly trivial - just moving the new language files into the same directories under a new repository? 

That's the general idea.  Though again there isn't a 100% agreed upon way to put all the pieces back together again yet... some suggest `npm`, hence making your thingy a npm package... I'd like to see a simpler way (without requiring everyone publish 100 different npm packages).  We're still working on the pieces.

vsoch:
okey doke, I'm okay with things still being under development - I've updated the PR here to *just* be for the addition of the Dockerfile, and I'm ready to fork / PR the pythonlogging language to a repository, if you might be able to make one (locally I'd call it highlightjs-pythonlogging).

yyyc514:
I think if we went this way we'd want something more generic than `node tools/build.js :common`... there are many different ways and configurations someone might want to build Highlight, hence the varied options that build supports.

Or else clarify this is only for web use, but I don't see the reason to restrict it so narrowly.

yyyc514:
So to be clear: the goal here is someone who has a whole Docker environment setup but not Node.js?

vsoch:
> So to be clear: the goal here is someone who has a whole Docker environment setup but not Node.js?

Yes, and a "whole Docker environment" just means having Docker. For developers that use containers, that's likely a dependency they already have (for example, myself!)

> Or else clarify this is only for web use, but I don't see the reason to restrict it so narrowly.

The reason that I provide custom instructions for how to bind the src code and then execute commands is for exactly that reason. The default will build the common (a reasonable default) but the developer is totally free to use docker exec to run custom commands as he/she pleases.

yyyc514:
> Yes, and a "whole Docker environment" just means having Docker. 

True, it's easier than it used to be. :-) (I'm on Mac)

> The reason that I provide custom instructions for how to bind the src code and then execute commands is for exactly that reason.

Yes I was a bit hasty in my previous review.

---

Might it be useful to provide shims for some of the common things, running tests, building, etc?

vsoch:
What is a shim? You mean other examples of exec commands to do different kinds of builds and run tests?

vsoch:
Anyway here is the repo:

https://github.com/vsoch/highlightjs-pythonlogging

I'd be happy to transfer it here and maintain it as the framework for making this work is figured out.

yyyc514:
@egor-rogov Egor sets up the repos and stuff.  I’ll let him handle that.

yyyc514:
> You mean other examples of exec commands to do different kinds of builds and run tests?

I meant actual wrappers...

```
./tools/docker/build -t node
```

Or some such... haven’t given it much thought.  I will look at this more closely later and actually try it out and see how it works and then I might have some other ideas or doc tweaks.

vsoch:
Uou mean a script to build the container? Since the context is important, and relative to the Dockerfile, I'd suggest (as is standard for most repos with Dockerfiles) a clean, single Dockerfile with clear instructions for how to build, and then examples for exec is most appropriate. If you forced the Dockerfile itself into a subfolder it would make the build a lot harder, because you'd still need to build from the root of the repository, and then specify --f tools/docker/Dockerfile.

I've worked on containers for many years, implemented Docker workflows for many different types of applications, and this is the approach that I suggest. There are of course other possibilities, but they add unnecessary complexity.

yyyc514:
No. Scripts to launch the most common tools. Not talking about where the docker file is.

yyyc514:
Will elaborate when I’m back to my laptop.

vsoch:
okay sounds good, and no rush! I'm likely finishing up for the day.

egor-rogov:
Hi guys,
it looks to me that some misunderstanding has happened. This PR is not about new language. It's about docker container for highlightjs. @yyyc514 are you familiar with docker?

vsoch:
Yes that is correct!  I originally had added a new language along with the Dockerfile, but removed it per instruction here.  The new language is now at:

https://github.com/vsoch/highlightjs-pythonlogging

 And is considered separate to the work here.

egor-rogov:
> Yes that is correct! I originally had added a new language along with the Dockerfile, but removed it per instruction here.

Ah, so it's me who misunderstood (:
I've set up a repo for the language (https://github.com/highlightjs/highlightjs-pythonlogging), see an invite.

vsoch:
Thank you @egor-rogov! Ready to go! https://github.com/highlightjs/highlightjs-pythonlogging/pull/1

yyyc514:
Yes I’m aware what Docker is. Lol. I’m open to this but wanting to play around with it first to see if I think anything is missing or if we should/could improve the docs.

vsoch:
Sure! Happy to add more detail if you think it's needed. It serves a very basic purpose to build the source code, and serve (as entrypoint) a webserver for the development preview. The details that i added to the docs explain also how to optionally bind to the host so any commands can be executed (e.g., rebuild, build in a different way). These (I think) are the strongest use cases for having a container.

yyyc514:
@egor-rogov This was originally a language request, hence my response... and then it was altered to be just what it is now. :-)

egor-rogov:
> @egor-rogov This was originally a language request, hence my response... and then it was altered to be just what it is now. :-)

Yeah, I got it.

yyyc514:
Sorry for the delay... a few thoughts:

- Should the port really be 80?  Or should the example bind 80 to 80?  In development it's much more common to bind to a high port such as 8080 or 3000, 4000, etc...
- This seems to focus very solely on building a WEB version... surely building the node version and running it's tests would be a very key thing to do also...
- So I'm debating the value of having the Docker build itself build the web version (and with :common no less)... it seems we should just `npm install` (which is generally needed) but then stop, and then allow the user to do whatever they want, no?

vsoch:
> Should the port really be 80? Or should the example bind 80 to 80? In development it's much more common to bind to a high port such as 8080 or 3000, 4000, etc...

The nice thing about containers is that the internal port can be bound to any desired port on the host. For someone inexperienced with containers, it's easy to tell them to open up localhost without a port. For someone experienced (and that wants a different port) they just modify the command. This is basically up to the user.

> This seems to focus very solely on building a WEB version... surely building the node version and running it's tests would be a very key thing to do also...

The use case here is to provide a development environment. If you look at the Dockerfile, we are installing dependencies, and it would be trivial for a user to shell in and run tests.

>   So I'm debating the value of having the Docker build itself build the web version (and with :common no less)... it seems we should just npm install (which is generally needed) but then stop, and then allow the user to do whatever they want, no?

It feels more complete to provide the entire build, and then allow the user to shell in and interact with some subset of it. Most of the time the user won't know how to do the full build and deploy, and a docker container is a really nice way to show that. I would suggest that, instead of trying to predict user use cases and feedback, to release this full feature version and then get feedback from users and tweak documentation or the container build based on that. Decisions based on actual feedback make the most sense to me.

vsoch:
I'm not able to resolve conflicts either as I don't have write access.

yyyc514:
>  For someone inexperienced with containers, it's easy to tell them to open up localhost without a port. 

Can you even do that without being root?  Wouldn't you need root to bind it to a low port on your LOCAL machine?  That was my point, port 80 might be ok inside the container but outsider the container I think the example should perhaps be a high port.

> instead of trying to predict user use cases and feedback

You're making my point. :-)  I think you just don't fully understand the build process and what a full build is. :-)

> It feels more complete to provide the entire build, and then allow the user to shell in and interact with some subset of it. 

But you aren't providing an entire build... just a browser build.  It's impossible to even run most of the test suite with only a browser build.  I see a few choices:

- Do no build at all (the most generic option, making NO assumptions about the users needs) - and let the user follow the directions and learn how to build what type of build they need. (that's what I'm leaning towards)
- Do a node build, allowing someone to run ALMOST the full test suite... more of the suite seems one valid metric to measure "better" here...
- Do a FULL build, but then the paths are all wrong as none of the tools are actually intended to work that way... a full build does all 3 builds inside `build` in their own subfolder.

vsoch:
> Can you even do that without being root? Wouldn't you need root to bind it to a low port on your LOCAL machine? That was my point, port 80 might be ok inside the container but outsider the container I think the example should perhaps be a high port.

Someone using Docker (installed correctly) is already added to the docker group and yes, can bind any port that they please. It's common to use 8080 (or similar) for other development apps, but it's also fairly standard to use 80 for local development without https when a container is involved. Again, it's trivial because the user can bind to whatever port they want (e.g., `-p 80:8080`).

> You're making my point. :-) I think you just don't fully understand the build process and what a full build is. :-) You're suggesting a very specific use case, I'm suggesting something more broad.

Being condescending to your contributors is not a good way to keep them around... if you know best, then you can do it. Take care!

yyyc514:
**I wasn't being condescending.  I'm sorry if you choose to take it that way. I just speak plainly.**

That literally was the point I was making... I was suggesting a generic use case, you were suggesting a very specific use case (someone using the developer tool).  There is also:

- The demo tool
- Running the main test suite

For example, I contribute a great amount of time to Highlight.js and almost never use the web developer tool.  Mostly I work on the parser and grammars and run the Node test suites.

> Someone using Docker (installed correctly) is already added to the docker group and yes, can bind any port that they please

Interesting.

>  it's also fairly standard to use 80 for local development without https when a container is involved. 

It's common to do 80:80?  If so, then ok I guess... I'm more familiar with almost every other development server picking their own high port.  

> Again, it's trivial because the user can bind to whatever port they want (e.g., -p 80:8080).

Yeah, that's the example I was suggesting since it seemed more familiar to me.  But I'm willing to accept that perhaps my lack of knowledge might be the issue on this one.

yyyc514:
>  if you know best, then you can do it

Believe me, if I didn't want to discuss it I would have just done it. :-)  I opened up a discussion to see if we couldn't improve what we have here.  I'm not trying to "invent" use random cases.  I literally know of several real-life use cases because I'm a core contributor to the project. So part of the goal was to see if we could better address those.

Anyways, I'm sorry you were offended by what I said but if you'd like to come back and continue the discussion, we're here.

yyyc514:
> I'm not able to resolve conflicts either as I don't have write access.

Typically you'd do a rebase locally with your own repository and then force push a new commit history without conflicts.  Or a merge, then fix issues.  That's a sloppier commit history, but much simpler if you aren't familiar with rebasing.  And the commit history isn't critical since almost all PRs are squashed when merged anyways...

You don't need write access to another here to resolve conflicts.

vsoch:
I'd be happy to re-open and re-base because I think this is useful for the community, and I'd ask that you give specific feedback on the changes you would like to see.  Since you are the expert on this software, there isn't much discussion - just tell me the specific changes you want and I'll make them.

yyyc514:
> and I'd ask that you give specific feedback on the changes you would like to see.

That's certainly fair - and I think often I try to do exactly that, _but sometimes I actually don't know all the answers_.  

For example a specific change is I'd change the example to `80:3000` or `80:4000`... **BUT** if it's super common to do 80:80 bindings when working with Docker, then that's news to me and it would be a reasonable reason to just leave the docs as you have them now. :-)  

> Since you are the expert on this software, there isn't much discussion - just tell me the specific changes you want and I'll make them.

Well, sometimes there isn't a simple answer.  It could be this discussion is revealing a deeper problem with the build process (I think it may be) - and/or it could be that this whole PR is actually kind of targeted to a specific use case (and that it's not easily fixable), and maybe that's not so bad... but that should be understood going in.   


The whole fact that it's running `nginx` implies it needs to host something, which is already biased towards running the demo or developer tool.  Whereas, I do 95% of my development on the library without ever opening a browser... so most of the time the whole nginx component would be un-used and unneeded... is that a problem?  I'm not sure.

vsoch:
I don't think it would be a problem, because the user doesn't need to bind anything to the host at all. The nginx in the container can happily exist without the host caring.

yyyc514:
> I don't think it would be a problem, because the user doesn't need to bind anything to the host at all. The nginx in the container can happily exist without the host caring.

Sure, I get that.  I think the only big question here is just what to build by default... I think this is going to have me opening another thread.  Already I'm a little annoyed that there is no way to run the full test suite with a single build now.  You need a node build to run 95% of the suite and you need a browser build to run the browser tests.  Our CI is probably the only reason no one has cared too much since it runs both suites and most of the time there are no issues with the browser tests in any case.

Oh, one other point... is there a good reason not to just bind the directory in the first place?  I can't imagine the container would be very useful to me without the source files being bound... having to build a WHOLE new container just to run the tests again sounds like a big slowdown...

vsoch:
> Oh, one other point... is there a good reason not to just bind the directory in the first place? I can't imagine the container would be very useful to me without the source files being bound... having to build a WHOLE new container just to run the tests again sounds like a big slowdown...

You wouldn't need to rebuild the whole thing, very likely the most that might be needed is a restart. If there is a change in some core dependency and a rebuild is necessary, there is also caching so it's not as terrible as you make it out to be. :)

yyyc514:
> You wouldn't need to rebuild the whole thing, very likely the most that might be needed is a restart. If there is a change in some core dependency and a rebuild is necessary, there is also caching so it's not as terrible as you make it out to be. :)

Well, the way it is now it'd at least have to _[assuming the Docker build can start from a cached build (after apt-get update & install nginx)]_:

- recopy `.`
- rerun `npm install`
- rebuild
- then run your tests, etc.

Vs (bound): You'd only need the last step... rebuilt/run tests...

Yes?

The `npm install` part of that scares me. :-)  Though it MIGHT be possible to copy over package.json as a separate step... then do the `npm install` earlier? I'm not sure.  But I still think if you were doing a build/test cycle that you'd likely want to bind vs rebuilding the whole container each time.

I was thinking there was some way to specify the bind/mount inside the Dockerfile but now it seems perhaps there isn't? I guess you really wouldn't know the local side of the mount until runtime anyways, so now I'm not sure what I was thinking that.

yyyc514:
> Yes, for my use case (why I created this in the first place) I just needed the web interface, and wasn't interested in binding anything.

Were you not actively changing the library?  How did you deal with rebuilds?

vsoch:
If you'd like to make a specific suggestion to update the Dockerfile I'm happy to do that.

yyyc514:
I tried to push to your branch but seems you unchecked that ability perhaps when you made the PR?  First time I've tried that I've been unable to.

I was thinking wouldn't this make better use of caching:

```
FROM node:12
# docker build -t highlight-js .
RUN apt-get update && apt-get install -y nginx
WORKDIR /var/www/html
COPY ./package*.json /
RUN npm install
COPY . /var/www/html
RUN node tools/build.js :common
EXPOSE 80
CMD ["nginx", "-g", "daemon off;"]
```

Seems to work well here.

vsoch:
That would work, and I can update the PR. I think we would want to copy the package.json into the WORKDIR (and not the container root) however:

```
FROM node:12
# docker build -t highlight-js .
RUN apt-get update && apt-get install -y nginx
WORKDIR /var/www/html
COPY ./package*.json /var/www/html/
RUN npm install
COPY . /var/www/html
RUN node tools/build.js :common
EXPOSE 80
CMD ["nginx", "-g", "daemon off;"]
```

yyyc514:
Oh, duh. :-)

vsoch:
heyo! Just to ping you again, I realize you probably didn't receive a notification, but I updated the recipe to include the first add of the package.json, and then the install steps.

yyyc514:
Thanks, I saw.  I think I may change the build to CDN vs browser (since it's a little more flexible and also gives you the browser build) and I'm going to tweak the docs very slightly, but I do plan to get this merged in the near future.

AUTHORS.txt
```diff
@@ -20,6 +20,7 @@ Former maintainers:
 Contributors:
 
 - Peter Leonov <gojpeg@gmail.com>
+- Vanessa Sochat <@vsoch>
```
yyyc514: Are you opposed to adding your email?  This list is used in other places where it's expected that these are emails (npm contributor list, etc)... or perhaps NPM doesn't care?  I'm not sure.

vsoch: I’d prefer to use my GitHub alias, if it’s alright with you! I try to minimize the random public places where my email might be scraped, and with the alias you can still easily find and ping me anywhere on GitHub.

yyyc514: This gets split out later into NPM contributors... perhaps a URL would be better?

https://docs.npmjs.com/files/package.json#people-fields-author-contributors

Otherwise I think in NPM we'd be stuck with including nothing but your name...

vsoch: > Both email and url are optional either way.

Sounds ok to me!

yyyc514:
@vsoch Thanks for working on this!

vsoch:
Woohoo!

PR #2218, opened 2019-10-22
(shell) Fix formatting when '/' is in a shell path

z4ce:
Right now if you have a prompt like `bash /home/user#`
the syntax highlighter doesn't correctly detect the prompt.
This commit fixes that by adding / to the character set for
prompt detection


yyyc514:
Please include some test.

How does this differential between something like:

```
> /bin/true
```

Where the / is part of the command, not the prompt?  I think the ^ solves that but please add a `markup` test to show this working as expected.

yyyc514:
I'd love to get this in the next release if you could make the requested changes/additions.

z4ce:
I added a test, but I'm having trouble getting mocha to run with some of the relative require statements. Is some magic NODE_PATH that needs to be set? Here is an example of what I am seeing:

```
╰─ npm test

> highlight.js@9.15.10 test /git/highlight.js
> mocha --globals document test

/git/highlight.js/node_modules/yargs/yargs.js:1163
      else throw err
           ^

Error: Cannot find module '../../build'
```

yyyc514:
I'm not sure what you're doing... all you need to do is drop two files into.  You shouldn't need to add any actual JS code.

```
/test/markup/shell
```

Look at the other languages for examples.

z4ce:
Thanks! That was the issue. I had built for browser but not for node.

yyyc514:
And the changelog please. :-)

z4ce:
It looks like the CI system errored and no tests actually failed

yyyc514:
Yikes.

PR #2219, opened 2019-10-24
fix dropping characters if we choke on a bad regex

### LINKED ISSUE #1566: (fortran) number before underscore disappears
If an underscore follows a number in formatted Fortran code, the number disappears. The code
````
e = init_pair(1_c_short, COLOR_YELLOW, COLOR_BLACK)
````
will be shown as
````
e = init_pair(_c_short, COLOR_YELLOW, COLOR_BLACK)
````

### PR DESCRIPTION
**Pull request by yyyc514**
- This closes #1566
- Previously: If we would hit a 0 length regex match we would skip
  the parser forward by 1 charcter (necessary) but ALSO drop that
  character from the output.
- This makes sure the character is always included in the output
  and it appears as if whatever mode fired matched a single length
  match instead of a 0 length one

Before: (the 6 is silently dropped and we have an empty span)

```
  <span class="hljs-number"></span>_ikind
```

After:

```
  <span class="hljs-number">6</span>_ikind
```

It appears as if we matched 6, which is even sort of true. Future
fixes will be coming to actually fix theses regular expressions.

### CODE REVIEW

**Comment by yyyc514** [comment]
@egor-rogov This doesn't fix the grammar for Fortran (to behave better) but it fixes the bug in the parser, which I thought was worth splitting out into it's own commit.

I'll circle back with a separate PR for the other.

File: test/markup/fortran/numbers.txt [hunk 1]
```diff
@@ -12,4 +12,20 @@
...
+CALL testing(184756_ikind, 20, 10)
+CALL testing(184756_ikind, 10, -1)
+CALL testing(3268760_ikind, 25, 10)
+CALL testing(155117520_ikind, 30, 15)
+CALL testing(155117530_ikind, 30, -1)
+CALL testing(6540715896_ikind, 48, 10)
+CALL testing(6540715897_ikind, 48, -1)
+CALL testing(344867425584_ikind, 45, 15)
+CALL testing(118264581564861424_ikind, 60, 30)
+CALL testing(53194089192720_ikind, 60, 15)
```
egor-rogov: I don't think we need all of these. For Fortran number literals test, more different formats are needed. But here you want to test the parser bug, so maybe a separate test in `test/parser` would be better?

yyyc514: Ok you want a test that was red before and green now, that makes sense. I'll see what I can do.

I'll remove a few of these and see if I can add a little more variety from the other issue, not super worried about this though.  It can be improved when the actual issue is fixed.  So technically it could be removed here, but I like getting it committed.

PR #2221, opened 2019-10-24
bump to 9.15.11

yyyc514:


yyyc514:
Don't merge this yet, but is this all we need to do?

marcoscaceres:
This probably won't work. Make sure you follow: https://github.com/highlightjs/highlight.js/blob/master/docs/maintainers-guide.rst

marcoscaceres:
I.e., do the release on "master" - not as a pull request. And follow the instructions precisely.

yyyc514:
I don't think I can push directly to master... I can be more thorough and follow those notes, but why can't this be a PR (for review) and then tagged after it's merged?

yyyc514:
> The header in CHANGES.md (this is where the site looks for the latest version number)

Ugh.  Does this mean I need to rip out the next "Master" section I have and release with `## Version 9.15.11` leading the file?

marcoscaceres:
No, the changes that are in this PR is what it should look like on master.

yyyc514:
> No, the changes that are in this PR is what it should look like on master.

Not sure what you are answering.  And isn't that what we would have after the PR was merged?

marcoscaceres:
Yes, but the PR is unnecessary. You can just do the whole thing on master.

yyyc514:
> Yes, but the PR is unnecessary. You can just do the whole thing on master.

Well, the PR was necessary so you could point me to the right procedure. ;-)

Ok, when it's time I'll try it on master... but if that fails I'm going to reopen the PR. :-)

PR #2222, opened 2019-10-25
fix(crystal): correctly highlight `!~` after `def`

MakeNowJust:
It is simple edge case fix.

For example, this code:

```crystal
def !~
end
```

is currently highlighted as:

```html
<span class="hljs-function"><span class="hljs-keyword">def</span> !<span class="hljs-title">~</span></span>(other)
<span class="hljs-keyword">end</span>
```

But correct highlight is:

```html
<span class="hljs-function"><span class="hljs-keyword">def</span> <span class="hljs-title">!~</span></span>(other)
<span class="hljs-keyword">end</span>
```

Thank you.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New styles:
 - *Night Owl* by [Carl Baxter][]
 
 Improvements:
+- fix(crystal): correctly highlight `!~` method definition (#2222)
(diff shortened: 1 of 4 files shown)


yyyc514:
Oh Crystal doesn't have an param highlighting or anything, interesting.

@MakeNowJust Thanks for the contribution, looks good.

Issue #1899, opened 2018-11-02
error registering arduino language


AlbyIanna:
I am trying to register arduino language like this:
```js
import arduinoHljs from 'highlight.js/lib/languages/arduino';

hljs.registerLanguage('arduino', arduinoHljs);
hljs.configure({
  languages: [arduinoHljs]
});
```
But I am getting this error:
```
Uncaught TypeError: Cannot read property 'exports' of undefined
    at module.exports (arduino.js:2)
    at Object.registerLanguage (highlight.js:694)
...
``` 

Reading issue #1235 i understood that arduino relies on cpp language, and the order of the languages could mess things up, so I tried to register cpp too like this:
```js
import cppHljs from 'highlight.js/lib/languages/cpp';
import arduinoHljs from 'highlight.js/lib/languages/arduino';

hljs.registerLanguage('cpp', cppHljs);
hljs.registerLanguage('arduino', arduinoHljs);
hljs.configure({
  languages: [arduinoHljs]
});
```
and it went well.
But this still feels a little weird. Why do I have to import a language when it is already specified elsewhere that arduino relies on cpp? I believe this should be done automatically when registering a derived language. But maybe I am missing something, am I?

marcoscaceres:
No, you are right... what might be good fix is to make sure arduinoHljs is importing cpp.

yyyc514:
The problem is HighlightJS would have no way to know WHERE the CPP was imported at, right?  Thoughts?

```
hljs.registerLanguage('cpp', cppHljs);
```

This works because YOU know the CPP is available in the variable cppHljs.  When Highlight.JS loads the Arduino syntax file all it knows is that it depends on "cpp".

Thoughts?

yyyc514:
> But this still feels a little weird. Why do I have to import a language when it is already specified elsewhere that arduino relies on cpp?

Isn't import a keyword not a statement?  Can JS randomly import things dynamically?  I haven't played around too much with runtime modules, usually I compile everything down to one static file.

The whole "export/import" nature of the module system makes this kind of thing hard... We could perhaps make the API simpler though:

```js
import cppSyntax from 'highlight.js/lib/languages/cpp';
import arduinoSyntax from 'highlight.js/lib/languages/arduino';

hljs.registerLanguage(
  cppSyntax,
  arduinoSyntax)
hljs.configure({
  languages: ["arduino"]
});
```

So you just import them and register them, but you'd still be on your own to import them.

[...]

PR #2223, opened 2019-10-25
log error when requiring an unavailable language

AlbyIanna:
Solves #1899


yyyc514:
OMG, thanks for running with this but I imagined a little more general solution.  Hmmmm, though it seems that right now Arduino is the only language that depends on another (though that may not be true for 3rd party languages).  What I had imagined:

- Add `requireLanguage` API in addition to `getLanguage`.
- While getLanguage might or might not return a language
- requireLanguage always would, or would console.log that error (but generic to whichever langue was requested)...
- So we don't need to catch any errors... if the language requested isn't found, that's a hard error and should be logged.

Thoughts?

yyyc514:
https://github.com/highlightjs/highlight.js/pull/2236/

You might want to piggy back on this.

AlbyIanna:
@yyyc514 So what do you suggest?

yyyc514:
Same idea, just now you can reuse the error message.

yyyc514:
Closing this for inactivity.  I'll circle back around to this eventually myself or if you want to try your hand at implementing a generic solution (as suggested above) you can always simply open another PR (or we can reopen this one).  Thanks.

PR #2226, opened 2019-10-26
switch to fs.promises vs bluebird for tests

### LINKED ISSUE #1791: Testing: Drop Bluebird, use utils.promisify
As promises are now native, and promisify() ships with Node’s utils, we can remove our dependency on Bluebird. 

We can also use async/await to make code in the test directory more maintainable.

yyyc514:
Potentially closes #1791

### DIFF
--- a/test/detect/index.js
+++ b/test/detect/index.js
@@ -1,7 +1,6 @@
 'use strict';
 
-let bluebird = require('bluebird');
-let fs       = bluebird.promisifyAll(require('fs'));
+let fs       = require('fs').promises
 
 delete require.cache[require.resolve('../../build')]
 delete require.cache[require.resolve('../../build/lib/highlight')]
@@ -15,18 +14,19 @@ function testAutoDetection(language) {
   const languagePath = utility.buildPath('detect', language);
 
   it(`should have test for ${language}`, function() {
-    return fs.statAsync(languagePath)
+    return fs.stat(languagePath)
       .then(path => path.isDirectory().should.be.true);
   });
 
-  it(`should be detected as ${language}`, function() {
-    return fs.readdirAsync(languagePath)
+  it(`should be detected as ${language}`, async function() {
+    let dirs = await fs.readdir(languagePath)
+    let files = await Promise.all(dirs
       .map(function(example) {
         const filename = path.join(languagePath, example);
 
-        return fs.readFileAsync(filename, 'utf-8');
-      })
-      .each(function(content) {
+        return fs.readFile(filename, 'utf-8');
+      }))
+    files.forEach(function(content) {
         const expected = language,
               actual   = hljs.highlightAuto(content).language;
 
--- a/test/detect/index.js
+++ b/test/detect/index.js
@@ -1,26 +1,24 @@
 'use strict';
 
-let fs       = require('fs').promises
-
 delete require.cache[require.resolve('../../build')]
 delete require.cache[require.resolve('../../build/lib/highlight')]
 
-
-let hljs     = require('../../build');
-let path     = require('path');
-let utility  = require('../utility');
+const fs       = require('fs').promises
+const hljs     = require('../../build');
+const path     = require('path');
+const utility  = require('../utility');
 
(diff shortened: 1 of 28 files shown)


yyyc514:
@marcoscaceres I whipped this up quickly, but I'm not really sure I see the advantage of using the async stuff here when in almost all the cases we're doing things sequentially in a single thread... wouldn't it be easier and simpler to just use the Sync methods?

yyyc514:
Such as:

```
return fs.readdir(markupPath).then((list) => list.forEach(testLanguage));
// vs
return fs.readdirSync(markupPath).forEach(testLanguage);
```

test/detect/index.js
```diff
@@ -1,7 +1,6 @@
 'use strict';
 
-let bluebird = require('bluebird');
-let fs       = bluebird.promisifyAll(require('fs'));
+let fs       = require('fs').promises
```
marcoscaceres:
```suggestion
const fs = require('fs').promises
```

test/detect/index.js
```diff
@@ -15,18 +14,19 @@ function testAutoDetection(language) {
   const languagePath = utility.buildPath('detect', language);
 
   it(`should have test for ${language}`, function() {
```
marcoscaceres:
```suggestion
  it(`should have test for ${language}`, async () => {
```

test/detect/index.js
```diff
@@ -15,18 +14,19 @@ function testAutoDetection(language) {
...
-    return fs.statAsync(languagePath)
+    return fs.stat(languagePath)
```
marcoscaceres:
```suggestion
    const path = await fs.stat(languagePath);
```

test/detect/index.js
```diff
@@ -15,18 +14,19 @@ function testAutoDetection(language) {
...
       .then(path => path.isDirectory().should.be.true);
```
marcoscaceres:
```suggestion
      path.isDirectory().should.be.true;
```

test/detect/index.js
```diff
@@ -15,18 +14,19 @@ function testAutoDetection(language) {
...
   });
 
-  it(`should be detected as ${language}`, function() {
-    return fs.readdirAsync(languagePath)
+  it(`should be detected as ${language}`, async function() {
+    let dirs = await fs.readdir(languagePath)
```
marcoscaceres: Only use `let` for reassignment. 

```suggestion
    const dirs = await fs.readdir(languagePath)
```

test/detect/index.js
```diff
@@ -15,18 +14,19 @@ function testAutoDetection(language) {
...
+    let files = await Promise.all(dirs
```
marcoscaceres:
```suggestion
    const files = await Promise.all(dirs
```

test/special/index.js
```diff
@@ -1,10 +1,9 @@
 'use strict';
 
 let _        = require('lodash');
-let bluebird = require('bluebird');
 let hljs     = require('../../build');
 let { JSDOM } = require('jsdom');
-let readFile = bluebird.promisify(require('fs').readFile);
+let readFile = require('fs').promises.readFile
```
marcoscaceres:
```suggestion
let { readFile } = require('fs').promises;
```

test/tools.js
```diff
@@ -1,9 +1,8 @@
 'use strict';
 
 let utility = require('../tools/utility'),
-    bluebird = require('bluebird'),
     path = require('path'),
-    readFile = bluebird.promisify(require('fs').readFile);
+    readFile = require('fs').promises.readFile
```
marcoscaceres:
```suggestion
    { readFile } = require('fs').promises;
```

test/detect/index.js
yyyc514: Sure if we're doing that I'll change them all.  I was doing the minimum work, not rewriting. :-)  But I can.

yyyc514:
> Looks good. Is there more bluebird around the place?

Probably in build but I plan to redo all that, so not worth the time.

PR #2240, opened 2019-10-30
inherit to Object.assign

xp44mm:


yyyc514:
We'll revisit this kinda of thing when we look at moving the codebase to ES6.  Right now this is part of ES6/ES2015 and in general we're not using newer code like that (other than tests) because our current build system can't even support it.  This types of thing might be one exception because to the build system it appears to be just a function call, but still...

Also this PR wouldn't be usable anyways because of all the noise... `inherit` is also a public API, so we'd likely continue to use it internally and externally even if all it did was wrap Object.assign.

So if this did become a PR in the future likely it'd only be a 5-8 line changes of the `inherit` method itself.

There are lots of open issues if you'd like some idea of what would be accepted.

yyyc514:
Also pretty sure this resulted in grossly different behavior since `inherits` ALWAYS makes a new copy of the first passed object where `assigns` uses the existing value of the first argument, from how I'm reading it.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

About to make this optional for a related change and came back to review this quickly.

xp44mm:
if you need effect:
```js
let obj = object.assign(obj,{...})
```
if you keep original immutable:
```js
let b = object.assign({},b,{...})
```

yyyc514:
Right, the latter case is what `inherit` has always done.  :-)  You meant `Object` there not `object`, right?

xp44mm:
yes, i should be more serious.

PR #2241, opened 2019-10-30
enh(css) major CSS highlighting improvements

### LINKED ISSUE #2232: [css/scss] highlighting deficiencies
Current state:
BAD!
BAD

GOOD (if no prefix):
GOOD-NoPrefix
GOOD (prefix):
GOOD-Prefix

**Just do not introduce prefix notation because the themes do not support I think bad.**

If the new class were introduced I would be happy to adjust the themes.
Enough to add to css themes:
.hljs-prefix{
opacity:0.55
}

Unfortunately, there are other bugs:

- Numbers in parentheses are not highlighted. `color: rgba(51, 170, 51, .1);` or `width: calc(100% - 80px);` or transform: rotate(45deg); or `filter: grayscale(0.5) blur(10px);`
- In the parentheses of the at rules, properties are not highlighted. `@media only screen and (orientation: landscape)` or `@media screen and (max-width: 900px) and (min-width: 600px), (min-width: 1100px)`

yyyc514:
Should close #2232

### DIFF
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -68,7 +68,11 @@ function(hljs) {
       {
         className: 'selector-attr',
         begin: /\[/, end: /\]/,
-        illegal: '$'
+        illegal: '$',
+        contains: [
+          hljs.APOS_STRING_MODE,
+          hljs.QUOTE_STRING_MODE,
+        ]
       },
       {
         className: 'selector-pseudo',
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -5,45 +5,52 @@ Website: https://developer.mozilla.org/en-US/docs/Web/CSS
 */
 
 function(hljs) {
+  var FUNCTION_LIKE = {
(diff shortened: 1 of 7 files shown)


yyyc514:
The `scss` stuff isn't quite as nice... but I did port over a few things to make it a llittle nicer.

Long term if these built on top of each other that might be nice.

src/languages/css.js
```diff
@@ -5,45 +5,52 @@ Website: https://developer.mozilla.org/en-US/docs/Web/CSS
...
+        hljs.C_BLOCK_COMMENT_MODE,
+        {
+          className: 'number', begin: '#[0-9A-Fa-f]+'
+        },
+        {
+          className: 'meta', begin: '!important'
+        }
+      ]
+    }
+  }
```
yyyc514: 8-44 is pretty much just copy and paste IIRC.

Issue #1857, opened 2018-10-09
(javascript) highlight jsdoc annotations


jimmywarting:
like github's markdown dose:

```js
/**
 * @param  {Image} the image to rotate
 * @param  {Function} cb function to call when done
 */
function rotateFrame (img, cb) {...}
```

or atom
![skarmavbild 2018-10-09 kl 17 27 13](https://user-images.githubusercontent.com/1148376/46680108-9d7f0a80-cbe8-11e8-9cdd-c9df1bb51ef8.png)

right now the jsdoc is just grayed like a normal comment

yyyc514:
Closing this with #2245.

PR #2245, opened 2019-10-31
enh(javascript) Add support for jsdoc comments

milutin:
This should highlight jsdoc tags like @param, @return. Should solve a part of Issue #1857.


yyyc514:
Thanks for the commit, but I'm not sure this part of it is useful on it's own... 
Screen Shot 2019-10-31 at 2 46 48 AM

It kind of looks broken to me, but since we only have a single color I'm not sure what we should highlight... perhaps everything except the description?

yyyc514:
Screen Shot 2019-10-31 at 2 49 01 AM

Better? I'm not sure.

milutin:
Ok, I will try to extend it. I did it as javadoc and other implementations, but you are right, it will be better in your second screenshot

yyyc514:
And if we color it all then I think you'd completely close the issue. :-)

yyyc514:
Oh and you'll also need to update the changelog.

milutin:
Thanks for help. I added support for type. I will need more time to figure out how to do paramater name. I need to look how to do forward searching for - but not to include in result.

yyyc514:
Nice. What do you think about highlighting type and variable as `type` and `variable` classes to get back the differentiation?

milutin:
Sure. I added separation for type and variable and added a few new test cases

src/languages/javascript.js
```diff
@@ -113,7 +113,19 @@ function(hljs) {
...
+              contains : [
+                {
+                  className: 'type',
+                  relevance: 0,
+                  begin: '\\{',
+                  end: '\\}'
+                }
+                , {
+                  className: 'variable',
+                  begin: '\\s+' + IDENT_RE + '(?=\\s*(-)|($))',
```
yyyc514: Am I correct this would fail with:

```
 * @param {boolean} retArr If set to true, the function will return an array
```

A sample from:
https://devdocs.io/jsdoc/tags-returns

milutin: Yes, but do you have idea how to identify variable name? We don't want to have it for 
@return {Point} A Point object

I used - or end of line as information for variable name. The other solution is to parse @param and only there make that it can have variable.

yyyc514: Yeah, I'm torn.  I can just see someone filing an issue about this exact issue, but then if you fix @param you COULD end up down the rabbit hole of people expecting us to support ALL the tags:

https://devdocs.io/jsdoc-tags/

Which all seem to have their own set of rules...

yyyc514: Won't the variable rule misfire though anytime you have a:

```
@something word
```

Pattern?  Wouldn't word match as a variable? (or some variant of that)

milutin: I think as a starting point this is good enough. Maybe next feature/improvement will be to create a new jsdoc language which will be used in typescript, javascript. What do you think?

yyyc514: I could be persuaded.  I think I would change the variable rule to use word boundaries for begin though and add a smaller rule to eat the spaces... that way the extra “ “ isn’t inside the variable span.  Usually we try to be clean about little stuff like that.  

I think after that we could call this done for starters.

I’m not sure about a whole “jsdoc” language, but I take your point about abstracting it into a component we can reuse in the various languages.  What is the full list?  JS, TypeScript, Java? More?

milutin: Ok, it's now skipping spaces. 

Java is using JavaDoc which is a little bit different, for example you don't need to write types in JavaDoc. So JsDoc is important mostly for javascript, typescript, maybe in html. 

Let me know if this change is in good shape to be merged.

yyyc514:
Taking a quick swing at this myself.

yyyc514:
@milutin What do you think?  Still not perfect, but might be as good as we're going to get for a first pass without a lot more time invested...

milutin:
I agree. I am happy with current version. I just wanted to color jsdoc at least a little bit. This is more than I expected. If someone wants more, it can be easily enhanced.

yyyc514:
Do you want to add your name to the changelog?

milutin:
Sure, why not :). Should you edit yourself as well, since you helped me a lot?

yyyc514:
*shrugs* I don't need credit for every item. :-)

milutin:
Thanks for helping me!

PR #2247, opened 2019-10-31
Fix missing var declaration in arduino.js

EricMCornelius:
Fixes `ARDUINO_KW is not defined` errors when executing in strict mode.

PR #2251, opened 2019-11-01
automate a common build process to check for "use strict" safe code

yyyc514:
- Adds a test to the CI pipeline that builds a "typical" consolidated
  Highlight.js package from the full NPM source and then checks that
  that build actually executes without errors


yyyc514:
Related: #2247

@egor-rogov If you have thoughts on naming or a better approach here I'm all ears.  We could build the executable in `tmp` or something instead?

PR #2254, opened 2019-11-04
enh(gml) fix naming of keyword class (consistency fix)

liamnobel:
I believe the nested "keywords" variables are causing an issue. Since keywords must be returned to the hljs function, I am changing the inner "keywords" to "keyword" and the GML style to match.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New styles:
   none.
 
 Improvements:
+- enh(gml) fix naming of keyword class (consistency fix) (#2254) [Liam Nobel][]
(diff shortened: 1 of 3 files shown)


yyyc514:
Curious, What “issue” was this causing as is?

yyyc514:
Well accept this because of consistency.  This is definitely inconsistent with our other grammars (as it stands now).

Please add an entry to the changelog though.

liamnobel:
The initial issue made it so the highlighter was not highlighting keywords. I assume this is because the keyword class held a reference to itself.

I added the edit to the changelog.

PR #2257, opened 2019-11-05
enh(elixir) improve heredoc and sigil support

yyyc514:


yyyc514:
My attempt to save *something* from #927.

Turned out just to be a markup test that I then went on to expand.

I believe this solves points 1 and 2 from #730

src/languages/elixir.js
```diff
@@ -19,16 +19,78 @@ function(hljs) {
...
     variants: [
+      {
+        begin: /"""/, end: /"""/,
+      },
+      {
+        begin: /'''/, end: /'''/,
+      },
+      {
+        begin: /~S"""/, end: /"""/,
+        contains: []
```
egor-rogov: I don't get... What is the point in empty `contains`?

yyyc514: It's needed to override the contains on line 64... variants "inherit", so it's needed to break that.

egor-rogov: Ah.

Issue #2259, opened 2019-11-07
Incremental line-by-line highlighting breaks some languages


davidben:
highlight.js supports highlighting text incrementally via the `continuation` parameter to the `highlight` function. Gerrit uses this to highlight line-by-line. However, this causes some features, notably C++ raw string matching, to break. See the discussion [here](https://bugs.chromium.org/p/gerrit/issues/detail?id=7748#c27).

(Ironically, raw string matching partially worked in #1825 and then I broke it again for us when I tried to make it more featureful in #1897.)

I changed the tests to feed all the markup samples line-by-line to demonstrate the issue. Things mostly work, but when a begin or end element crosses a line boundary, things break.
https://github.com/highlightjs/highlight.js/compare/master...davidben:line-by-line-tests

This suggests that the strategy in #1897 was wrong and probably C++ raw strings should have separate `begin` and `end` units. But `endSameAsBegin` doesn't seem quite flexible enough, unless there's some trick I'm missing. One thought is `begin` matches `R"<blah>(` and then we have a `endFromBegin` attribute which takes a callback and will compute `)<blah>"` for the `end` pattern given a `begin`.

But it's not entirely clear what the general API contract here is. Supposing incremental highlight is a thing highlight.js wants to support (I assume yes, given the `continuation` parameter), it seems there's basically no hope in allowing a `begin` or `end` unit to span an input boundary. The API needs to return something for the text that was passed in, so you can't lookahead across chunks. That means highlight.js needs to have opinions about what input boundaries need to be supported by languages (One call per character obviously cannot do anything useful. Line-by-line seems reasonable?), or we change the incremental highlight API altogether.

Thoughts?

yyyc514:
>  But endSameAsBegin doesn't seem quite flexible enough, unless there's some trick I'm missing. One thought is begin matches R"<blah>( and then we have a endFromBegin attribute which takes a callback and will compute )<blah>" for the end pattern given a begin.

Currently you can do this by nesting, but yes it's annoying. You might also see my work on Elixir:

https://github.com/highlightjs/highlight.js/pull/2257/files

It doesn't use endSameAsBegin, but same concept with first matching on the prefix and then the child matchers can focus on just the begin/end syntax, not the prefix.

---

`continuation` really is NOT meant as a way to do line-by-line processing.  This would be better handled by a post-highlight or post-parse plugin as mentioned over in the plugin thread. 

[...]

PR #2261, opened 2019-11-08
(cpp) Fix highlighting of unterminated raw strings

davidben:
@yyyc514, thoughts?

PR #1897 switched C++ raw strings to use backreferences, however this breaks souce files where raw strings are truncated. Like comments, it would be preferable to highlight them.

Instead, go back to using separate begin and end regexps, but introduce an `endFilter` feature to filter out false positive matches. This internally works similarly to `endSameAsBegin`.

See also issue #2259.

(Also, I have to say, this C++ raw string syntax is a little absurd...)

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -208,6 +208,7 @@ In this case you can't simply specify the same regexp for ``begin`` and
 ``end`` (say, ``"\\$[a-z]\\$"``), but you can use ``begin: "\\$[a-z]\\$"``
 and ``endSameAsBegin: true``.
 
+
 .. _lexemes:
 
 lexemes
--- a/src/lib/modes.js
+++ b/src/lib/modes.js
@@ -117,3 +117,8 @@ export const METHOD_GUARD = {
   begin: '\\.\\s*' + UNDERSCORE_IDENT_RE,
   relevance: 0
 };
+
+export const END_FIRST_MATCH_SAME_AS_BEGIN = {
+  'after:begin': (m, resp) => { resp.data.heredoc = m[1]; },
+  'before:end': (m, resp) => { if (resp.data.heredoc !== m[1]) resp.ignoreMatch(); }
+};
--- a/src/lib/modes.js
+++ b/src/lib/modes.js
@@ -118,7 +118,10 @@ export const METHOD_GUARD = {
   relevance: 0
 };
 
-export const END_FIRST_MATCH_SAME_AS_BEGIN = {
-  'after:begin': (m, resp) => { resp.data.heredoc = m[1]; },
-  'before:end': (m, resp) => { if (resp.data.heredoc !== m[1]) resp.ignoreMatch(); }
+export const END_SAME_AS_BEGIN = function(mode) {
+  return Object.assign(mode,
+    {
+    'after:begin': (m, resp) => { resp.data._beginMatch = m[1]; },
+    'before:end': (m, resp) => { if (resp.data._beginMatch !== m[1]) resp.ignoreMatch() }
+    });
 };
--- a/src/lib/modes.js
+++ b/src/lib/modes.js
@@ -121,7 +121,7 @@ export const METHOD_GUARD = {
 export const END_SAME_AS_BEGIN = function(mode) {
   return Object.assign(mode,
     {
-    'after:begin': (m, resp) => { resp.data._beginMatch = m[1]; },
-    'before:end': (m, resp) => { if (resp.data._beginMatch !== m[1]) resp.ignoreMatch() }
+    'on:begin': (m, resp) => { resp.data._beginMatch = m[1]; },
+    'on:end': (m, resp) => { if (resp.data._beginMatch !== m[1]) resp.ignoreMatch() }
     });
 };
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,13 +1,26 @@
 ## Version 10.1.0 (in progress)
 
+Parser Engine:
+
+- (enh) Added `on:begin` callback for modes (#2261) [Josh Goebel][]
+- (enh) Added `on:end` callback for modes (#2261) [Josh Goebel][]
+- (enh) Added ability to programatically ignore begin and end matches (#2261) [Josh Goebel][]
+- (enh) Added `END_SAME_AS_BEGIN` mode to replace `endSameAsBegin` parser attribute (#2261) [Josh Goebel][]
+
+Deprecations:
+
+- (deprecation) `endSameAsBegin` is now deprecated. (#2261) [Josh Goebel][]
+
 Language Improvements:
 
+- fix(cpp) Fix highlighting of unterminated raw strings (#2261) [David Benjamin][]
 - fix(javascript) `=>` function with nested `()` in params now works (#2502) [Josh Goebel][]
 - fix(typescript) `=>` function with nested `()` in params now works (#2502) [Josh Goebel][]
 - fix(yaml) Fix tags to include non-word characters (#2486) [Peter Plantinga][]
 
 [Josh Goebel]: https://github.com/yyyc514
 [Peter Plantinga]: https://github.com/pplantinga
+[David Benjamin]: https://github.com/davidben
 
 
 ## Version 10.0.1
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,7 +13,7 @@ Contents:
 
    api
    language-guide
-   reference
+   mode-reference
    css-classes-reference
    style-guide
    plugin-api
--- a/docs/language-guide.rst
+++ b/docs/language-guide.rst
@@ -186,7 +186,7 @@ For such modes ``className`` attribute should be omitted so they won't generate 
 Mode attributes
 ---------------
 
-Other useful attributes are defined in the :doc:`mode reference </reference>`.
+Other useful attributes are defined in the :doc:`mode reference </mode-reference>`.
 
 
 .. _relevance:
--- a/docs/mode-reference.rst
+++ b/docs/mode-reference.rst
@@ -62,6 +62,19 @@ Regular expression starting a mode. For example a single quote for strings or tw
 If absent, ``begin`` defaults to a regexp that matches anything, so the mode starts immediately.
 
 
+on:begin
+^^^^^^^^^^^
+
+**type**: callback (matchData, response)
+
+This callback is triggered the moment a begin match is detected. ``matchData`` includes the typical regex match data; the full match, match groups, etc. The ``response`` object is used to tell the parser how it should handle the match. It can be also used to temporarily store data.
+
+- ``response.data`` - a simple object data store.  Can be used for building more complex rules where the end rule is dependent on the content of begin, etc.
+- ``response.ignoreMatch()`` - pretend as if this match never happened. The mode is not entered. Continues trying subsequent modes in the current mode's ``contains`` list
+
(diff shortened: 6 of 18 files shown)


yyyc514:
I'll play around with a bit.  Nice to see this was so easy to accomplish.  I'm going to explore going broader though:

```diff
         begin: /\B\?(\\\d{1,3}|\\x[A-Fa-f0-9]{1,2}|\\u[A-Fa-f0-9]{4}|\\?\S)\b/
       },
       { // heredocs
-        begin: /<<[-~]?'?(\w+)(?:.|\n)*?\n\s*\1\b/,
-        returnBegin: true,
-        contains: [
-          { begin: /<<[-~]?'?/ },
-          { begin: /\w+/,
-            endSameAsBegin: true,
-            contains: [hljs.BACKSLASH_ESCAPE, SUBST],
-          }
-        ]
+        begin: /<<[-~]?'?(\w+)/,
+        end: /(\w+)\b/,
+        onBegin: function(m, state) { state.heredoc = m[1] },
+        onEnd: function(m, state) { return state.heredoc === m[1] },
+        contains: [hljs.BACKSLASH_ESCAPE, SUBST],
       }
     ]
   };
```

yyyc514:
```
      {
        begin: /(?:u8?|U|L)?R"([^()\\ ]{0,16})\(/,
        end: /\)([^()\\ ]{0,16})"/,
        onBegin: function(m, state) { state.heredoc = m[1] },
        endWhen: function(m, state) { return state.heredoc === m[1] },
      }
```

Very open to thoughts on naming.  The PR isn't quite right, we'd need lots of thought and additional  regarding how this works with `endsParent`, `endsWithParent`, etc... `endWhen` probably needs to happen inside `endOfMode` so that if the first endWhen failed it would continue up the parent chain looking for a end that might match... vs right now it would find the first end that possibly matches, and then endWhen could cancel that match.

And already you have some good questions about nesting and when both end rules are the same (other than the callback)... if the first callback returns "false" is that a blocker, or does the chain of possible ends continue to be inspected moving up the parents until it finds one that might hit?

These same type of questions would likely apply to the `endFilter` also... whether a false is a HARD no, or a soft no, etc...

I MUCH prefer a generic solution here though because I very much dislike the idea of adding specific solutions to the parser each time we come across some weird language nuance.  For example that's what we did with `endSameAsBegin`.  A VERY specific grammar feature for only TWO languages (pgsql and Ruby)... when that exact problem could have been solved with the more generic solutions being proposed here.

@egor-rogov Any high-level thoughts?  Not looking for code critiques at this point but more of the idea conceptually... of adding callbacks to allow for this type of thing, etc...

yyyc514:
> that's what we did with `endSameAsBegin`.  A VERY specific grammar feature for only TWO languages (pgsql and Ruby)... when that exact problem could have been solved with the more generic solutions being proposed here.

Also note this is really **almost EXACTLY** the same problem as "endSameAsBegin" but because the solution was hard-coded it can't apply to this use case... just another point in why I dislike one off grammar "hack" rules vs more generic solutions that can be used more widely.

egor-rogov:
> I very much dislike the idea of adding specific solutions to the parser each time we come across some weird language nuance.

Absolutely. I'd gladly sacrifice `endSameAsBagin` in favour of something more general. I was thinking (back when it was invented) about a way to reference begin's matching groups from end regexp, but couldn't come to a good solution.

> Any high-level thoughts? Not looking for code critiques at this point but more of the idea conceptually... of adding callbacks to allow for this type of thing, etc...

I like the idea! Callbacks look promising.
For now I can't see what the use cases for `onBegin`/`onEnd` may be. Single `endFilter` is not enough?

yyyc514:
> For now I can't see what the use cases for onBegin/onEnd may be. Single endFilter is not enough?

- Not all matches result in an actual end match... the file could end before an actual ending is reached.
- It should probably be possible to reject/ignore a match completely at begin... (hence needing a begin callback of some sort)
- Perhaps the `onBegin` wants to simply change the end regex directly at the time the match starts... (not currently allowed, but I'm not opposed to the idea conceptually).
- Or perhaps you're counting tags because you only want to highlight the first occurrence of something or highlight additional occurrences as invalid?

The idea is to build out a flexible plugin system.  I feel like this comment you're making the same mistake as when `endSameAsBegin` was added... you're only imagining a single use case (filtering) instead of more generic system that people could build on top of.

egor-rogov:
>     * Or perhaps you're counting tags because you only want to highlight the first occurrence of something or highlight additional occurrences as invalid?

Yes, this makes sense.

> The idea is to build out a flexible plugin/event system. I feel like this comment you're making the same mistake as when `endSameAsBegin` was added... you're only imagining a single use case (filtering) instead of more generic system that people could build on top of.

I understand the intent, and I'm in no way against it. Just wanted to imagine a situation in which onBegin/onEnd can be useful.

yyyc514:
My latest thoughts on this include a response object:

```js
{
        begin: /(?:u8?|U|L)?R"([^()\\ ]{0,16})\(/,
        end: /\)([^()\\ ]{0,16})"/,
        onBegin: function(m, state, resp) { state.heredoc = m[1] },
        onEnd: function(m, state, resp) { if (state.heredoc === m[1]) resp.ignoreMatch()  },
      }
```

It's possible you could have `resp.data` and then you'd be back to two params... but then I come to a naming issue.  Technically when you have a potential match (remember our callback is going to decide what to do with the match) you have two main possibilities:

- "abort"... this is an indication of a failed match but you STILL advance the cursor, ie skipping the content in question
- "ignore"... failed match, but the cursor is not advanced.

I think I could use help with the naming.  Examples of both:

`beginKeywords` in the past used an "abort" style match... So if you have:

```
B.class = A.class
```

And `beginKeywords: 'class'`... the `.class` would have been matched, but the `.` would be detected and then the entire match ignored AND the cursor moved to right before the `=`... ie. any later modes you have that might have matched ".cl" or "las" would fail because the cursor had already moved past.

That's what I'm calling "abort"... meaning "ignore" would ignore class (because of the `.`) but your cursor would still be at the `.`... so `.cl` WOULD match... or `las` WOULD match.

I'm thinking perhaps we only should expose the ignore behavior, but still I could use some better names just for internal use and talking about these things.

yyyc514:
Example of usage:

```js
// in Ruby
          Object.assign({
             begin: /(\w+)/, end: /(\w+)/,
             contains: [hljs.BACKSLASH_ESCAPE, SUBST],
           }
           }, hljs.END_FIRST_MATCH_SAME_AS_BEGIN)

// in C++
      Object.assign({
         begin: /(?:u8?|U|L)?R"([^()\\ ]{0,16})\(/,
         end: /\)([^()\\ ]{0,16})"/,
       }, hljs.END_FIRST_MATCH_SAME_AS_BEGIN)
```

And the "logic" can then become a shared mode snippet:

```js
export const END_FIRST_MATCH_SAME_AS_BEGIN = {
   'after:begin': (m, resp) => { resp.data.heredoc = m[1]; },
   'before:end': (m, resp) => { if (resp.data.heredoc !== m[1]) resp.ignoreMatch(); }
 };
```

Not quite as pretty as `endSameAsBegin: true`, but completely modular and reusable.

src/lib/modes.js
```diff
@@ -112,3 +112,8 @@ export const METHOD_GUARD = {
   begin: '\\.\\s*' + UNDERSCORE_IDENT_RE,
   relevance: 0
 };
+
+export const END_FIRST_MATCH_SAME_AS_BEGIN = {
+  'after:begin': (m, resp) => { resp.data.heredoc = m[1]; },
+  'before:end': (m, resp) => { if (resp.data.heredoc !== m[1]) resp.ignoreMatch(); }
```
egor-rogov: Why `heredoc`?

docs/reference.rst
```diff
@@ -208,6 +208,26 @@ In this case you can't simply specify the same regexp for ``begin`` and
 ``end`` (say, ``"\\$[a-z]\\$"``), but you can use ``begin: "\\$[a-z]\\$"``
 and ``endSameAsBegin: true``.
 
+.. _endFilter:
+
+endFilter
```
egor-rogov: This was the original propose; do we still need it in presence of begin/end-plugins?

egor-rogov:
I really like it. Very flexible and concise.
`Response` sound good, but too general. Maybe `CallbackResponse` or something like that?
I consider this PR mergeable, as soon as these new callbacks are described in the docs.
Btw, what do you think about deprecating `endSameAsBegin`?

src/lib/modes.js
yyyc514: I guess now that I pulled it into a common module I should make the naming more specific.

docs/reference.rst
yyyc514: Yes, it's on the chopping block.

yyyc514:
> Btw, what do you think about deprecating endSameAsBegin?

That would be the next step. Though I'm not sure what that would mean... that we could rip it out in say v11?

yyyc514:
```js
export const END_SAME_AS_BEGIN = function(mode) {
  return Object.assign(mode,
    {
    'after:begin': (m, resp) => { resp.data._beginMatch = m[1]; },
    'before:end': (m, resp) => { if (resp.data._beginMatch !== m[1]) resp.ignoreMatch() }
    });
};
```

Now I'm wondering if defaulting to the first match group is assuming too much?  Hmmm...

yyyc514:
@egor-rogov Check out the docs.  Now that I had to document them I'm thinking perhaps `on:begin` and `on:end` might be better names...

I'm not sure we ever need such granularity as `before:begin` and `after:begin`... and if not I'm not sure the prior naming makes good sense.  Its kind of weird to remember one is before and one is after, etc... and begs the question why isn't is "before begin" (and maybe it should be, lol).

I guess I could see some crazy data transforms if we allowed that, but we could do the same thing by adding more flexibility to response... ie `response.addContent()` or some such...

Does the `on` naming seem clearer to you?

egor-rogov:
> > Btw, what do you think about deprecating endSameAsBegin?
> 
> That would be the next step. Though I'm not sure what that would mean... that we could rip it out in say v11?

Just so.

> Does the on naming seem clearer to you?

Yes, I have the same feeling that `on` is better.

Maybe that `END_SAME_AS_BEGIN` snippet worth mentioning in plugin-recipes?

yyyc514:
> Maybe that END_SAME_AS_BEGIN snippet worth mentioning in plugin-recipes?

Are all the other MODE helpers documented anywhere?  It'd probably be good if they were, but if you document one you have to document all...

egor-rogov:
> > Maybe that END_SAME_AS_BEGIN snippet worth mentioning in plugin-recipes?
> 
> Are all the other MODE helpers documented anywhere? It'd probably be good if they were, but if you document one you have to document all...

They are not, but END_SAME_AS_BEGIN is the first one that uses plugins and it is a good example. Anyway, that's a separate story.

yyyc514:
> They are not, but END_SAME_AS_BEGIN is the first one that uses plugins and it is a good example. Anyway, that's a separate story.

It uses callbacks, not plugins. :-)  But yes separate story.

egor-rogov:
> It uses callbacks, not plugins. :-)

Oh, right... :facepalm: There are no callback recipes yet (:

yyyc514:
@davidben Sorry for the delay on this one - and it's not exactly what you originally submitted - but I still gave you the byline for the bug fix. :-)  Merging now.

PR #2262, opened 2019-11-09
Update highlight.js

xp44mm:
provide with too many arguments to `startNewMode`


yyyc514:
@xp44mm thanks for your contribution. Merged.

PR #2265, opened 2019-11-10
(objectivec) Sync preprocessor highlighting with cpp

### LINKED ISSUE #1964: (c-likes) Macro including single < breaks because it thinks it's a string
See https://chromium-review.googlesource.com/c/chromium/src/+/669484/4/ios/chrome/browser/ui/ui_util.mm and https://bugs.chromium.org/p/gerrit/issues/detail?id=7248

davidben:
The objectivec highlighting had a few bugs and was missing a few
features:

- Expressions with < were misidentified as a string like <stdio.h>
  (fixes #1964).

- Keywords like if, else, endif, etc. were not highlighted.

- The escape sequences in string literals were not parsed.

- Line continuations were not parsed.

- Comments were not parsed.

Fix all these by adapting cpp's preprocessor definition. Pull in the
preprocessor test from cpp. (This is actually objectivec's first markup
test. It probably could do with tests beyond just this feature.)


src/languages/objectivec.js
```diff
@@ -69,16 +69,24 @@ function(hljs) {
...
+        begin: /#\s*[a-z]+\b/, end: /$/,
+        keywords: {
+          'meta-keyword':
+            'if else elif endif define undef warning error line ' +
+            'pragma ifdef ifndef include'
+        },
         contains: [
+          {
+            begin: /\\\n/, relevance: 0
+          },
```
yyyc514: Why do we need this?

yyyc514: Oh it eats line continuations?

src/languages/objectivec.js
```diff
@@ -69,16 +69,24 @@ function(hljs) {
       },
       {
         className: 'meta',
-        begin: '#',
-        end: '$',
+        begin: /#\s*[a-z]+\b/, end: /$/,
```
yyyc514: No macros with underscores?

src/languages/objectivec.js
davidben: Yup.

src/languages/objectivec.js
davidben: (I just grabbed this from the cpp language definition.)

But anyway the token immediately after the `#` isn't a macro but a preprocessing instruction like `if` or `endif` or `pragma`. I don't believe any of them have underscores. I'm not sure if this change is necessary, but since it's the same preprocessor that runs over C, C++, and Obj-C (and Obj-C++), I thought it'd be best to align the two.

src/languages/objectivec.js
```diff
@@ -69,16 +69,24 @@ function(hljs) {
...
+          },
+          hljs.inherit(hljs.QUOTE_STRING_MODE, {className: 'meta-string'}),
           {
             className: 'meta-string',
-            variants: [
-              { begin: '\"', end: '\"' },
-              { begin: '<', end: '>' }
-            ]
-          }
+            begin: /<[^\n>]*>/, end: /$/,
```
yyyc514: Would a non greedy ‘.*?’ work here instead?

src/languages/objectivec.js
davidben: git blame says it dates to 1bc269c8668786d5c3ea06bff14826543797fe71 and de1718e41a7cdf292898efd1adf03de113ae3556, but it's not clear why the stricter matching on #.

yyyc514: Right I know what it is. I’m just mislabeled it. Lol. Thanks for clarifying though.

src/languages/objectivec.js
davidben: Seems to work. This is taken from the cpp definition, so if we change one we should probably change the other one too. Do you have a preference?

yyyc514: Yeah we’d change both. I think the non greedy is a nicer and more elegant way to write it.

davidben: Done. (Probably worth noting that non-greedy `.*?` has more potential to make the regex go nuts with backtracking or otherwise misbehave. But, in this case, since there is only one byte after the pattern, it's fine.)

davidben:
Rebased and added.

src/languages/objectivec.js
yyyc514: > Probably worth noting that non-greedy .*? has more potential to make the regex go nuts with backtracking or otherwise misbehave

Is there a reference or any article you could point me to on this issue?

davidben: Not off-hand. Searching around I found this: http://blog.stevenlevithan.com/archives/greedy-lazy-performance

Though I was thinking a slightly different issue. Suppose you wanted to find empty divs, accepting random attributes. `<div .*?></div>` would misbehave a bit. Given `<div id=a>blahblahblah<div id=b></div>`, the regex would match the whole string rather than just the ending bit. Also if there are a lot of strings that almost match `></div>`, the regex might go down that route and backtracking a bunch. (Note sure whether JS implementations tend to use backtracking or DFA-based regex engines.)

yyyc514:
> <div .*?></div> would misbehave a bit. Given <div id=a>blahblahblah<div id=b></div>, the regex would match the whole string rather than just the ending bit. 

Well, then I presume it would be a poor use in that case... but that's not what we're doing here... :-)

PR #2266, opened 2019-11-10
(go) Fix escaped character literals

### LINKED ISSUE #1411: (go) Misunderstands Backslash Character Literals
In Go, when invoking the escape sequence for backslash in a character literal, HLJS gets stuck incorrectly interpreting the following lines as part of a string. Observe the following simple go program.

    package main
    import "fmt"
    func main() {
      s := "\\"
      c := '\\'
      fmt.Println("Hello, 世界", s, c)
    }

When highlighted with HLJS 9.5.0, the entire file following the `'\\'` is highlighted as a string. It doesn't have this issue with the string literal (`"\\"`).

davidben:
Fixes #1411

### DIFF
--- a/src/languages/go.js
+++ b/src/languages/go.js
@@ -30,7 +30,7 @@ function(hljs) {
         className: 'string',
         variants: [
           hljs.QUOTE_STRING_MODE,
-          {begin: '\'', end: '[^\\\\]\''},
+          hljs.APOS_STRING_MODE,
           {begin: '`', end: '`'},
         ]
       },
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,12 +19,14 @@ Improvements:
 - fix(parser): Fix merger HTML attribute quoting (#2235) [Josh Goebel][]
 - fix(parser): Look-ahead regex now work for end matches also (#2237) [Josh Goebel][]
 - fix(parser): Better errors when a language is missing (#2236) [Josh Goebel][]
+- fix(go): Fix escaped character literals (#2266) [David Benjamin][]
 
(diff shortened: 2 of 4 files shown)


src/languages/go.js
```diff
@@ -30,7 +30,7 @@ function(hljs) {
         className: 'string',
         variants: [
           hljs.QUOTE_STRING_MODE,
-          {begin: '\'', end: '[^\\\\]\''},
```
yyyc514: I’ve seen this a few places now. Does it just not work as intended at all?  Is it only escaping two slashes or?  Seems like it wouldn’t match super short strings either?

davidben: Yeah, it just doesn't work. It's looking for a `'` not preceded by `\`, but `'\\'` is perfectly legitimate and fails the regex.

yyyc514:
I’ll review all these PRs more carefully tomorrow.

yyyc514:
Any chance you have time/interest to look into the 6-7 other languages we're using this and see if they likewise suffer from the same issue?

One could imagine a language where \\ wasn't valid so this MIGHT be a reasonable regex...

src/languages/go.js
yyyc514: \Yeah I get confused by all the extra escaping with '' strings.  Looks like exactly what you say.

davidben:
> Any chance you have time/interest to look into the 6-7 other languages we're using this and see if they likewise suffer from the same issue?

I'm traveling for the next two weeks so probably not for a bit. But maybe if I get bored on the plane.

> Please add Changelog entry.

Done

PR #2268, opened 2019-11-10
(objectivec) Handle multibyte character literals

davidben:
See https://bugs.chromium.org/p/gerrit/issues/detail?id=9444


davidben:
Added.

davidben:
Rebased and added.

yyyc514:
For future, technically no need to rebase since we squash anyways...

PR #2274, opened 2019-11-12
(docs) clarify use of continuation

### LINKED ISSUE #2259: Incremental line-by-line highlighting breaks some languages

yyyc514:
Closes #2259.


yyyc514:
If it’d been markdown I probably would have. Don’t know how to do that in rst. Lol. 

I’ll look it up next time.

egor-rogov:
> If it’d been markdown I probably would have. Don’t know how to do that in rst. Lol.
 
AFAIK `*italics*` and `**bold**`.

yyyc514:
Know a good editor/previewer for rst files on Mac?

egor-rogov:
Nope.

PR #2281, opened 2019-11-13
(ebnf) add underscore to productions and dot to terminators

vancluever:
This adds the following:

* Underscores as a recognizable character in a production name
(meta-identifer). While I'm not necessarily sure if underscore is a
permitted character to the letter of ISO 14977, there are examples of
being used in practice, the Go spec being one of these.

* Adds "." as a production terminator. This is valid as per ISO 14977
(section 7.4 on alternate terminal characters).

### DIFF
--- a/src/languages/ebnf.js
+++ b/src/languages/ebnf.js
@@ -9,7 +9,7 @@ function(hljs) {
 
     var nonTerminalMode = {
         className: "attribute",
-        begin: /^[ ]*[a-zA-Z][a-zA-Z-]*([\s-]+[a-zA-Z][a-zA-Z]*)*/
+        begin: /^[ ]*[a-zA-Z][a-zA-Z-_]*([\s-_]+[a-zA-Z][a-zA-Z]*)*/
     };
 
     var specialSequenceMode = {
@@ -18,7 +18,7 @@ function(hljs) {
     };
 
     var ruleBodyMode = {
-        begin: /=/, end: /;/,
+        begin: /=/, end: /[.;]/,
         contains: [
             commentMode,
             specialSequenceMode,
--- a/test/markup/ebnf/terminators.txt
+++ b/test/markup/ebnf/terminators.txt
@@ -1,0 +1,3 @@
+full-stop = "." .
+semicolon = ";" ;
+end-test = "end" ; (* ending production to test semicolon *)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New styles:
   none.
 
 Improvements:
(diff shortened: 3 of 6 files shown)


yyyc514:
Please add markup test with example and changelog entry.

vancluever:
@yyyc514 done!

src/languages/ebnf.js
```diff
@@ -9,7 +9,7 @@ function(hljs) {
 
     var nonTerminalMode = {
         className: "attribute",
-        begin: /^[ ]*[a-zA-Z][a-zA-Z-]*([\s-]+[a-zA-Z][a-zA-Z]*)*/
+        begin: /^[ ]*[a-zA-Z][a-zA-Z-_]*([\s-_]+[a-zA-Z][a-zA-Z]*)*/
```
yyyc514: I know you didn't add \s but is a space truly allowed in an attribute name? How does that work?

Having to add the _ in two places is what got my attention here and caused me to notice this.

test/markup/ebnf/terminators.txt
```diff
@@ -0,0 +1,3 @@
+full-stop = "." .
+semicolon = ";" ;
+end-test = "end" ; (* ending production to test semicolon *)
```
yyyc514: Is this comment supposed to mean something if I know EBNF?  I find it indecipherable?

src/languages/ebnf.js
vancluever: It's not necessarily common practice, but spaces are consumed in meta identifiers and the parts concatenated. There's some discussion here about it: https://stackoverflow.com/questions/15268229/ebnf-whitespacing-in-meta-identifiers

test/markup/ebnf/terminators.txt
vancluever: The "bug" that gets exhibited when using an incorrect terminator (like a `.` before this patch) is that the name of the next rule (the _production_) doesn't get highlighted. So the final production is necessary to test that semicolons still work as a terminator.

src/languages/ebnf.js
vancluever: Also examples on the Wikipedia page. https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form

test/markup/ebnf/terminators.txt
yyyc514: Ah, makes perfect sense. :-)

src/languages/ebnf.js
yyyc514: Thanks, from what I'm reading this could probably be simplified, but this will work for now. ;-)

yyyc514:
@vancluever Thanks for your contribution!

vancluever:
Thanks @yyyc514!

PR #2285, opened 2019-11-14
enh(parser) post-highlightBlock callbacks via plugin

yyyc514:
- [x] add tests

---

First pass at simple plugin API to allow things like:

```js
    hljs.addPlugin( {
      "after:highlightBlock": ({block, result}) => {
        // move the language from the result into the dataset
        block.dataset.language = result.language }
    })
```

I realize I need to document the plugin API itself and the callbacks, but how does this look at a high level? We'd also support classes for more complex plugins:

```js
  class DataLanguagePlugin {
    constructor(options) {
      self.prefix = options.dataPrefix;
    }

    afterHighlightBlock({block, result}) {
      // ...
    }
  }

  hljs.addPlugin(DataLanguagePlugin, { dataPrefix: "hljs"})
```

Passing a class being the preferred way for more complex things, and the object based approach serves for one-liners to make them easy to whip up.


docs/index.rst
```diff
@@ -16,6 +16,8 @@ Contents:
    reference
    css-classes-reference
    style-guide
+   plugins
+   recipes
```
egor-rogov: Not sure if it's a good idea to separate recipes and plugins. I'd put them both on the same page unless there's some reason.

src/highlight.js
```diff
@@ -917,6 +922,24 @@ https://highlightjs.org/
...
+    var cb = eventToFuncName(event);
+    plugins.forEach(function (plugin) {
+      if (plugin[cb]) {
+        plugin[cb](args);
+      }
+    });
+  }
+
+
+  function eventToFuncName(event) {
```
egor-rogov: What is the purpose of having slightly different names for events and functions?

yyyc514: Easier to write, easier to read, and it's a convention i've seen used in other places.  It's just a nicety.

docs/index.rst
yyyc514: Maybe.  `recipes` is intended to be "recipes for small plugins" and really "plugins" is intended to be the list of hooks of plugin API.  That page is already going to get plenty long over time I think... and I think having the API docs separate from examples of usage makes sense to me.

I think the title even says "Plugin API".

yyyc514: I might rename the pages "plugin_api" and "recipes" does that make more sense to you?

egor-rogov: Maybe "plugin-api" and "plugin-recipes"?

src/highlight.js
egor-rogov: In "Callbacks" section you give the names of the events. And in the actual code (according to examples) the plugin author must use function names. How to match one with the other?

yyyc514: I changed the first example. :-)  I'm still thinking about the class based approach.  This would be much simpler in Ruby. :-)

Honestly it doesn't bother me to have/support both though... depending on how you write the plugin.

docs/index.rst
yyyc514: Sure.

src/highlight.js
egor-rogov: > I changed the first example

Ah, I see. What I can't see, is why it is better than simply `afterHighlightBlock` (:

>  we'd likely have a section explaining how the event names map to function names, which I think is pretty straight forward to understand

Perhaps, but still I can't see benefits of having this abstraction. I don't insist on 1:1 mapping, just trying to understand what it might be useful for.

egor-rogov: I was hasty about the first example. It makes sense, because in that case we don't bother with mapping, we just specify the event name, which is good.

But class-based approach doesn't fit. Of course you can instantiate an object of your class and register its methods as callbacks...

yyyc514:
```
  class DataLanguagePlugin {
    constructor(options) {
      this.prefix = options.dataPrefix;
    }

    'after:HighlightBlock'({block, result}) {
      // ...
    }
  }
```

?

Ha, I didn't even know you could do that.

egor-rogov: %)  How to invoke this thing?

yyyc514: You don't ever invoke it directly yourself, you just write it - the plugin system invokes it as necessary.  I also think it's nice because it differentiates the callbacks from regular internal functions your class might have.

But if you actually wanted to know you just use the `[""]` attribute accessor form:

```
object["weird:key:name"]()
```

yyyc514:
@egor-rogov How you feeling about this in general?  Can I polish it up for merging? (which requires building out the class functionality, writing some tests, better docs.)

I'll ping you again for a final review, but I wanted to make sure you're cool with the naming before I put more effort into this.

egor-rogov:
>How you feeling about this in general? Can I polish it up for merging? (which requires building out the class functionality, writing some tests, better docs.)

Definitely yes.
My only concern is events-functions mapping. I feel it's important to use only events names in plugin code.

yyyc514:
Don't we resolve that by just saying "always use the even names", as suggested above:

```
  class DataLanguagePlugin {
    constructor(options) {
      this.prefix = options.dataPrefix;
    }

    'after:highlightBlock'({block, result}) {
      // ...
    }
  }
```

egor-rogov:
> Don't we resolve that by just saying "always use the event names"

We do.

yyyc514:
I want to review the tests here and perhaps add a few more but is this otherwise good to merge?

yyyc514:
@egor-rogov Ping?

Issue #2264, opened 2019-11-09
(XML) Highlight !DOCTYPE tag components


w3suli:
Currently, the entire member is highlighted with the meta class.
```
<!DOCTYPE html>

<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN"  "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">

<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1 Basic//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11-basic.dtd">
```

The components could also be highlighted.
My question is, what classes would you use for this?

There are two possible classes between meta-classes. One is the **meta-keyword** for the keywords that mark the document and the other is the **meta-string** for the strings.

Look forward to your suggestions.

yyyc514:
Re XHTML I don’t think this really matters since so much of the web is just HTML 5 now anyways. So this is an edge case to begin with. 

Also not sure it’s a problem this is all meta. Ie, don’t see a lot of value in splitting it out. 

@egor-rogov thoughts?

w3suli:
Highlighting strings would be helpful.
This allows strings to be more separated, if any.
`<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1 Basic//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11-basic.dtd">`
```
<!DOCTYPE note [
<!ENTITY nbsp "&#xA0;">
<!ENTITY writer "Writer: Donald Duck.">
<!ENTITY copyright "Walt Disney Company.">
]>
```
The rest is less important.

egor-rogov:
We make use of `meta-keyword` and `meta-string` in other languages, so I have no objections for them in XML also.

yyyc514:
Well let’s do it then!   I guess I feel all languages are not the same but it hard to turn that into a persuasive argument.

w3suli:
@egor-rogov, @yyyc514 Is this the solution you like?
![doctype](https://user-images.githubusercontent.com/20889555/68788251-53febb80-0643-11ea-9c73-2de9025e5c79.png)

yyyc514:
Why the two different classes?  Tag vs meta?

w3suli:
If doctype is part, then meta, if not part, then not.

```
<!DOCTYPE note
[
<!ELEMENT note (to,from,heading,body)>
[...]

PR #2287, opened 2019-11-14
(xml) expand and improve document type highlighting

w3suli:
Solution to #2264 issue.


yyyc514:
Ping me when you think this is ready for review.  Look like you need some markups tests, and this has too many commits (looks like it's mixed with another of your PRs, has conflicts, etc, when it should probably start from a fresh master, etc)...

w3suli:
@yyyc514 Conflicts resolved.

src/languages/xml.js
```diff
@@ -6,10 +6,37 @@ Category: common
...
+	  begin: '\'', end: '\'',
+	  illegal: '\\n',
+	  contains: [hljs.BACKSLASH_ESCAPE]
+  };
+  var QUOTE_META_STRING_MODE = {
+	  className: 'meta-string',
+	  begin: '"', end: '"',
+	  illegal: '\\n',
+	  contains: [hljs.BACKSLASH_ESCAPE]
+  };
```
yyyc514: These should probably both inherit from QUOTE_STRING_MODE and APOS_STRING_MODE.  See other languages for examples.

w3suli: You're right. I did it.

w3suli:
Markup test added.

w3suli:
Example:
[https://jsfiddle.net/w3suli/axo0Lv8w/2/](https://jsfiddle.net/w3suli/axo0Lv8w/2/)

@yyyc514 Will this be good?

test/markup/xml/document-type-variations.expect.txt
```diff
@@ -0,0 +1,38 @@
...
+<span class="hljs-meta">&lt;!DOCTYPE<span class="hljs-meta-keyword"> svg</span><span class="hljs-meta-keyword"> PUBLIC</span><span class="hljs-meta-string">"-//W3C//DTD SVG 1.1 Basic//EN"</span><span class="hljs-meta-string">'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11-basic.dtd'</span>&gt;</span>
+
+<span class="hljs-meta">&lt;!DOCTYPE<span class="hljs-meta-keyword"> note</span> [
+<span class="hljs-meta">&lt;!ENTITY<span class="hljs-meta-keyword"> nbsp</span> <span class="hljs-meta-string">'&amp;#xA0;'</span>&gt;</span>
+<span class="hljs-meta">&lt;!ENTITY<span class="hljs-meta-keyword"> pound</span> <span class="hljs-meta-string">'&amp;#xA3;'</span>&gt;</span>
+<span class="hljs-meta">&lt;!ENTITY<span class="hljs-meta-keyword"> writer_person</span> <span class="hljs-meta-string">" Writer: Donald Duck. "</span>&gt;</span>
+<span class="hljs-meta">&lt;!ENTITY<span class="hljs-meta-keyword"> copyright</span> <span class="hljs-meta-string">"Copyright: Walt Disney Company."</span>&gt;</span>
```
yyyc514: Why aren't the `"` converted to `&quot;` here? (like in the prior expect) Am I missing something or have we found a bug?

src/languages/xml.js
```diff
@@ -6,10 +6,27 @@ Category: common
...
+	  className: 'meta-keyword',
+	  begin: '\\s#?[a-z_][a-z1-9_-]+',
+	  illegal: '\\n',
+  };
+  var XML_META_PAR_KEYWORDS = {
+	  begin: '\\(', end: '\\)',
+	  contains:[
+	    {
+	      className: 'meta-keyword',
+	      begin: '#?[a-z_][a-z1-9_-]+',
```
yyyc514: Looks like 15/23 share a common constant that you could factor out.

Why does Meta eat a space though but not meta_par?

test/markup/xml/document-type-variations.expect.txt
```diff
@@ -0,0 +1,38 @@
...
+]&gt;</span>
+
+<span class="hljs-meta">&lt;!ENTITY<span class="hljs-meta-keyword"> nbsp</span> <span class="hljs-meta-string">' &amp;#xA0; '</span>&gt;</span>
```
yyyc514: The space should not be a part of the keyword.

src/languages/xml.js
w3suli: Outside the parentheses, spaces define the keywords. No space is required within parentheses. I came up with another solution that also addresses the following criticism.

test/markup/xml/document-type-variations.expect.txt
w3suli: I wanted to save a step, but I am changing it so that the spaces are not in the markup.

test/markup/xml/document-type-variations.expect.txt
w3suli: I messed up the first solution. I inserted the code generated by the php htmlentities () function instead of the result of the hljs highlighting.

That is why the first solution code was bad and contained &quot; entities.

src/languages/xml.js
yyyc514: Generally I'd dislike something like line 14 because it's too broad (it matches ANY spaces anywhere). BUT since this is already scoped inside of ONLY `meta` modes then this is ok, since that limits the scope plenty.

src/languages/xml.js
```diff
@@ -6,10 +6,23 @@ Category: common
 
 function(hljs) {
   var XML_IDENT_RE = '[A-Za-z0-9\\._:-]+';
-    var XML_ENTITIES = {
+  var XML_ENTITIES = {
     className: 'symbol',
     begin: '&[a-z]+;|&#[0-9]+;|&#x[a-f0-9]+;'
   };
+  var XML_META_KEYWORDS = {
+	  begin: '\\s', end: '',
```
yyyc514: Is `end:` really needed?

w3suli: True! Corrected.

Issue #2173, opened 2019-10-09
Add all __ keywords to `cpp` in one fell swoop


yyyc514:
Stems from PR:
https://github.com/highlightjs/highlight.js/pull/2020

See the related discussion there.

>> I'd prefer a separate mode, something like

>So someone has to repeat that within every scope in the syntax file, yes? So if KEYWORDS is used 10 times, then they'd need to add that rule to all 50 place also, correct?

It'd be really nice to find a solution to this without a separate mode. This needs a little more though IMHO to see if that's possible before just breaking down and doing that.

yyyc514:
```js
// _Complex, __always_inline, etc.
var UNDERSCORE_KEYWORD_RE = ["_[A-Z][a-z\\d_]+", "__[a-z\\d_]+"]

var CPP_KEYWORDS = {
   keyword: 'int float while private char catch import module export virtual operator sizeof ' +
   ...
}

hljs.addKeywords(CPP_KEYWORDS["keyword"], UNDERSCORE_KEYWORD_RE)
// or 
hljs.addKeyword(CPP_KEYWORDS["keyword"], UNDESCORE_KEYWORD_RE, relevance: 1)
hljs.addKeyword(CPP_KEYWORDS["keyword"], DOUBLE_UNDESCORE_KEYWORD_RE, relevance: 2)
```

I think I'd prefer something like the above (a mini API to let someone add keywords to the actual keywords).  The would solve the problem of "not to depend on how keyword flattener works"... since we'd be free to change the keyword string format in the future (but how likely is that really?) just so long as we still supported `addKeyword`... right now all `addKeyword` would really do is just add the regex and a space... and it could blow up if (or do nothing) if someone included a space in their regex...

The problem with saying "just use another matcher" is (other than having to change the file perfectly in like 10 places, and then worry about how to TEST that those changes all work properly) that matchers actually have very different behavior from keywords... keywords are always the last to match, match the "gaps", etc...  lots of advantages to just piggybacking off the existing keyword system.

Thoughts?

egor-rogov:
Something like that would be OK, but I think it's a bit complicated than it looks.

> all addKeyword would really do is just add the regex and a space

So what about spaces in regex? And how about set some relevance to _every_ keyword introduced by `UNDERSCORE_KEYWORDS_RE`?
For me it seems that such API should somehow add keywords directly into `compiled_keywords` structure (bypassing flattener).

yyyc514:
[...]

PR #2289, opened 2019-11-16
(cpp) Add additional keywords

aostrowski:


src/languages/cpp.js
```diff
@@ -47,7 +47,7 @@ function(hljs) {
     keywords: {
       'meta-keyword':
         'if else elif endif define undef warning error line ' +
-        'pragma ifdef ifndef include'
+        'pragma ifdef ifndef include __has_include __has_cpp_attribute _Pragma'
```
yyyc514: Please remove the `__keywords` (double underscore).  If we're going to do this we need to do it completely, not partially:

https://github.com/highlightjs/highlight.js/issues/2173

src/languages/cpp.js
```diff
@@ -67,27 +67,27 @@ function(hljs) {
...
       'auto_ptr deque list queue stack vector map set bitset multiset multimap unordered_set ' +
-      'unordered_map unordered_multiset unordered_multimap array shared_ptr abort abs acos ' +
+      'unordered_map unordered_multiset unordered_multimap array shared_ptr abort terminate abs acos ' +
       'asin atan2 atan calloc ceil cosh cos exit exp fabs floor fmod fprintf fputs free frexp ' +
       'fscanf future isalnum isalpha iscntrl isdigit isgraph islower isprint ispunct isspace isupper ' +
       'isxdigit tolower toupper labs ldexp log10 log malloc realloc memchr memcmp memcpy memset modf pow ' +
       'printf putchar puts scanf sinh sin snprintf sprintf sqrt sscanf strcat strchr strcmp ' +
       'strcpy strcspn strlen strncat strncmp strncpy strpbrk strrchr strspn strstr tanh tan ' +
-      'vfprintf vprintf vsprintf endl initializer_list unique_ptr',
+      'vfprintf vprintf vsprintf endl initializer_list unique_ptr _Bool complex _Complex _Imaginary',
```
yyyc514: Shouldn't we also have `imaginary` as well?

yyyc514:
Also, please add a changelog entry.

src/languages/cpp.js
aostrowski: Surprisingly, no. Those are keywords coming from C99, not all of them have their equivalents in the C++ standard library.

src/languages/cpp.js
aostrowski: Done.

src/languages/cpp.js
yyyc514: I said that because of: https://en.cppreference.com/w/c/numeric/complex/imaginary

```
    double imaginary i = 2.0*I; // pure imaginary
```

> This macro expands to the keyword _Imaginary.

Defined in complex.h... if we don't count both how are we deciding that _Imaginary is a keyword but imaginary is not?

I'm welcome to be wrong here, but I'm trying to understand the logic we're using to call something a keyword.

yyyc514: You didn't respond?  Am I thinking about this wrong? ... ?

yyyc514:
```
double imaginary i = 2.0*I; // pure imaginary
double _Imaginary i = 2.0*I; // pure imaginary
```

Given this snippet, wouldn't both cases of imaginary there count as keywords? Yes/no?  Or are we drawing a distinction between build-in vs pulled in via some include?

Complex seems to be in the same boat:

https://en.cppreference.com/w/c/language/arithmetic_types#Complex_floating_types

So it seems to me if `complex` is a keyword, then `imaginary` also should be, and if not, then it should not?

yyyc514: I think if you just want to do the simplest thing to get this PR approved from where I'm setting it seems we need to add `imaginary` as a keyword also, and then we're consistent. But fell free to make a case the other way if you think that's not correct.

PR #2292, opened 2019-11-20
Add support for language NATURAL

MarkusAmshove:
This PR adds support for the language NATURAL by Software AG.

Sadly, it is a proprietary language, but you can see some example code [e.g. here](https://github.com/AlteOldenburger/NatUnit)


MarkusAmshove:
```
should be detected as cpp:

      AssertionError: expected 'natural' to be 'cpp'      

      actual expected

      

      naturalcpp
````

I'm not sure what happened here.

yyyc514:
Sadly, due to a lack of time and maintainers we are generally not accepting new languages into the core library.  Feel free to publish your library as a language extension though, and if you'd like to be this languages maintainer we can create even a repository here inside the `highlightjs` organization (for discoverability), if you would like.  It's reasonably easy for someone using highlightJS to load additional languages simply by including a JS tag for them and then using `hljs.registerLanguage()`.

Related discussion:
https://github.com/highlightjs/highlight.js/issues/2149

Feel free to join the discussion.

PR #2295, opened 2019-11-21
fix(kotlin): fix termination of """ string literals

### LINKED ISSUE #2280: (kotlin) Incorrect handling of strings with triple quotes
The following Kotlin code has incorrect highlighting. 
The triple quotes string ends insided myFun, but highlightjs highlight the string till the end of file.
```
fun myFun(f: Boolean): String {
  return if (f) """f="true"""" else """f="false""""
}

fun main(args: Array<String>) {
    println(myFun(true))
    println(myFun(false))
}
```
P.S. This is a correct Kotlin code, it prints
```
f="true"
f="false"
```

yyyc514:
- Properly handles cases: like """ends in a "quoted string""""
- Closes #2280.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@ New styles:
   none.
 
 Improvements:
-- fix(kotlin): fix termination of """ string literals () [Josh Goebel][]
+- fix(kotlin): fix termination of """ string literals (#2295) [Josh Goebel][]
 - fix(mercury): don't change global STRING modes (#2271) [Josh Goebel][]
 - fix: freeze built-in modes to prevent grammars altering them (#2271) [Josh Goebel][]
 - enh(xml) expand and improve document type highlighting (#2287) [w3suli][]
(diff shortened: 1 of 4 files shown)


egor-rogov:
I have had a look into Kotlin grammar (https://kotlinlang.org/docs/reference/grammar.html). The relevant parts are:

```
multiLineStringLiteral
  : '"""' (multiLineStringContent | multiLineStringExpression | '"')*
    TRIPLE_QUOTE_CLOSE
  ;

TRIPLE_QUOTE_CLOSE
  : ('"'? '"""') 
  ;
```

It seems that `end: '"""(?=[^"])'` is basically the same as `end: '"?"""'`, but the latter is a bit simpler and closer to the grammar. Thoughts?

yyyc514:
Screen Shot 2019-11-22 at 6 15 27 AM

I don' think those are the same... Is this not a problem?

egor-rogov:
Yeah, I played around a bit and looks like you're right. We need the rightmost match.

PR #2296, opened 2019-11-21
fix: packaging test is meaningless unless run in debug mode

yyyc514:


yyyc514:
Ha, now that I did this I actually don't even think it's true... I think the import part is the critical part here.

PR #2297, opened 2019-11-22
docs: update maintainers guide

### LINKED ISSUE #2252: Thoughts on release management
Since I've done 2 releases now I thought I'd type up my thoughts and see if we could find some kind of consensus.  I'd prefer to have somewhat stable releases that are separate from ongoing development work... IE, we can release 9.16.2 with a SINGLE hot fix without releasing a bunch of other random stuff that's slated for the 9.17 release.

This also gives us the power to slide bug fixes into patch releases easily if we choose to (cherry-pick, etc) while saving the enhancements for minor/major versions.  _Though of course I'd prefer not to release every time we fix a bug, unless it was critical._

To that end currently we have a `9_16_stable` branch that I'm releasing from.

Master tracks the ongoing 9.17 work...

Open to thoughts/discussion.

yyyc514:
Closes #2252.


yyyc514:
I tried to explain the philosophy a bit also, I think it makes sense.

egor-rogov:
It makes sense for me.
@marcoscaceres, please have a look at this also.

yyyc514:
@marcoscaceres Any thoughts?

marcoscaceres:
I'm ok with this - but personally I think it's a bit of overkill. However, if npm and other software know about "-stable", then it should be fine.

yyyc514:
> I'm ok with this - but personally I think it's a bit of overkill. 

Perhaps, but it has it's advantages. :-)  In the past I've found it very useful to isolate stable releases from ongoing development work.  That's all this does.

> However, if npm and other software know about "-stable", then it should be fine.

It's not anything NPM or other version management would see... the branches are purely for making life easier for maintainers and separating production releases from ongoing dev work (in master).

marcoscaceres:
> It's not anything NPM or other version management would see... the branches are purely for making life easier for maintainers and separating production releases from ongoing dev work (in master).

Ah, my apologies then. I had misunderstood the proposal.

yyyc514:
> Ah, my apologies then. I had misunderstood the proposal.

Did you find the docs unclear or were you only glancing quickly?  Since it doesn't affect NPM or anything should I assume you're OK with it then?  I think Egor was holding his approval until you weighed in pro or con.

PR #2300, opened 2019-11-23
fix(themes) Standardize the padding/spacing for all themes

yyyc514:
It's very weird that 2-3 of our themes completely change the padding/text-layout.  I understand why the "school paper" one does, but otherwise our themes should be consistent when it comes to whitespace.

This standardizes on:

```
  display: block;
  overflow-x: auto;
  padding: 0.5em;
```

Only 3 themes were changed to bring all our themes into alignment.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New styles:
   none.
 
 Improvements:
+- fix(themes): fix inconsistencies between some themes padding/spacing (#2300) [Josh Goebel][]
 - fix(mercury): don't change global STRING modes (#2271) [Josh Goebel][]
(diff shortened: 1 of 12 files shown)


yyyc514:
@egor-rogov We have some themes that specify no BG color... that seems very weird to me... if they are intended to be light themes would it hurt to add `white`?

If they are actually intended to be transparent then I think we should specify that explicitly. Thoughts?

yyyc514:
Screen Shot 2019-11-25 at 1 34 55 PM
 
Example of issue on our demo site... if it's intended to be transparent it's not obvious in demo so we'd have to think of a way to indicate that?

yyyc514:
> But I see no harm in adding backticks as rightful string terminators. 

Ha, convenient. :)  I hadn't seen that.

PR #2301, opened 2019-11-24
(sql) Backslash should not escape in string

### LINKED ISSUE #1748: (sql) single backslash inside a string literal seems to not be valid
I opened a ticket against VS Code for a colorization issue (https://github.com/Microsoft/vscode/issues/50059).  They closed it as upstream to this project.

It boils down to the following gets incorrect highlighting due to the backslash.

\```sql
select '\\' + foo from bar
\```

Same issue displays here on github:

```sql
select '\' + foo from bar
```
vs
```sql
select 'hello' + foo from bar
```

schallm:
Remove backslash_escape with in strings in standard SQL

Closes #1748.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New styles:
   none.
 
 Improvements:
+- fix(sql): backslash is not used to escape in strings in standard SQL (#1748) [Mike Schall][]
 - chore(javascript): add esm related extensions to aliases (#2298) [Rongjian Zhang][]
 - fix(kotlin): fix termination of """ string literals (#2295) [Josh Goebel][]
 - fix(parser/docs): disallow `self` mode at the top-level of a language (#2294) [Josh Goebel][]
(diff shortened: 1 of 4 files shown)


yyyc514:
Please add all 3 string types and examples of escapes to the expects test.  And we need a changeling entry also. :)

Thanks!

egor-rogov:
Hmm, strings in backticks are out of standard altogether. Anyone knows what database uses them?

yyyc514:
MySQL allows them for strings: https://stackoverflow.com/questions/11321491/when-to-use-single-quotes-double-quotes-and-backticks-in-mysql

yyyc514:
@egor-rogov Thoughts?  If there is still a separate issue here might we keep this focused on the original issue and perhaps create a new issue for the backtick issue?  

I think this solves the original issue, yes?

egor-rogov:
> @egor-rogov  Thoughts?  If there is still a separate issue here might we keep this  focused on the original issue and perhaps create a new issue for the  backtick issue? I think this solves the original issue, yes?

Absolutely.
As for backticks, looks like we cannot remove it until MySQL gets its own grammar, so let them be.

PR #2305, opened 2019-12-01
[pkg] Fix to be able to minify with closure compiler. 

saksmt:
Calling addEventListener on window instead of nothing/global scope

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@ Improvements:
 - fix(objectivec): Fix various preprocessor highlighting issues (#2265) [David Benjamin][]
 - fix(objectivec): Handle multibyte character literals (#2268) [David Benjamin][]
 - enh(cpp): Add additional keywords (#2289) [Adrian Ostrowski][]
+- fix(pkg): Prefix global addEventListener with window to be able to minify with closure compiler (#2305) [Kirill Saksin]()
 
 [Josh Goebel]: https://github.com/yyyc514
 [Liam Nobel]: https://github.com/liamnobel
@@ -44,6 +45,7 @@ Improvements:
 [Chris Marchesi]: https://github.com/vancluever
 [Adrian Ostrowski]: https://github.com/aostrowski
 [Rongjian Zhang]: https://github.com/pd4d10
+[Kirill Saksin]: https://github.com/saksmt
 
 
 ## Version 9.16.2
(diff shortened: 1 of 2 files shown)


yyyc514:
What does the closure compiler do without this change?

saksmt:
Complain about missing function (it reproduces only with ADVANCED compilation level)

yyyc514:
And is there not some option you can pass to tell it how to treat addEventListener?  This is very common JS code and convention going back forever.  Sounds like it's closure that's broken here, not us?

saksmt:
No, there is no such option. Also it's better not to rely on functions/variables from window scope to be present in global scope. Also if my memory not falling me such code will fail if `"use strict"` is enabled.

yyyc514:
> Also it's better not to rely on functions/variables from window scope to be present in global scope. 

@marcoscaceres Is this a thing?  I'm not super familiar with the many "best practices" in JS.  AFAIK it's always been this way and I didn't realize this was a problem.

> Also if my memory not falling me such code will fail if "use strict" is enabled.

Do you have a reference for that?  Pretty sure our existing library works in some "use strict" safe environments as it stands now, so not sure this is correct.

marcoscaceres:
> @marcoscaceres Is this a thing? I'm not super familiar with the many "best practices" in JS. AFAIK it's always been this way and I didn't realize this was a problem.

Yes, definitely a thing. ... I don't know about "use strict" failing - don't think that's the case... but being explicit about `window` is "a good thing"™️

saksmt:
> Do you have a reference for that?

Sorry, can't find anything about this, probably I was wrong (in my defense I'm not a JS dev at all, just use it when I need to do something on frontend, which is pretty rare case) or I just always used JS with both strict mode and closure compiler every time and it got in my head that this restriction is from strict mode.

Anyways this fix is really small and affect nothing except for best practices and stuff. So why don't we just merge this and be done with it? So I could happily go and write in scalajs without worrying too much about size of a bundle :)

yyyc514:
> Anyways this fix is really small and affect nothing except for best practices and stuff. So why don't we just merge this and be done with it? 

We will, I just wanted marco to sign off on the best practices stuff.  I'm worried if we made tiny [ultimately] unnecessary changes for every little build tool that's a rabbit hole I'd rather avoid.  But if it's best practice anyways... then sure, lets do it.

@saksmt Can you please add an entry to the changelog?

yyyc514:
@saksmt Thanks!

PR #2306, opened 2019-12-03
Add LaTeX

quantumsheep:
Add basic LaTeX language highlighting.

### DIFF
--- a/AUTHORS.en.txt
+++ b/AUTHORS.en.txt
@@ -282,3 +282,4 @@ Contributors:
 - Carl Baxter <carl@cbax.tech>
 - Thomas Reichel <tom.p.reichel@gmail.com>
 - G8t Guy <g8tguy@g8tguy.com>
+- Nathanael Demacon <nathanael.dmc@outlook.fr>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 ## Version 9.17.master
 
 New languages:
-  none.
+- *LaTeX* by [Nathanael Demacon][]
 
 New styles:
   none.
@@ -44,6 +44,7 @@ Improvements:
 [Chris Marchesi]: https://github.com/vancluever
 [Adrian Ostrowski]: https://github.com/aostrowski
 [Rongjian Zhang]: https://github.com/pd4d10
+[Nathanael Demacon]: https://github.com/quantumsheep
 
 
 ## Version 9.16.2
--- a/src/languages/latex.js
+++ b/src/languages/latex.js
@@ -1,0 +1,28 @@
+/*
+Language: LaTeX
+Author: Nathanael Demacon <nathanael.dmc@outlook.fr>
+Description: A document preparation system
+Website: https://www.latex-project.org/
+Category: common, markdown
+*/
+
+function(hljs) {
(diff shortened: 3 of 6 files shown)


yyyc514:
First question:

Is this different than:

https://github.com/highlightjs/highlight.js/blob/master/src/languages/tex.js

quantumsheep:
Oh, I didn't see it until now! Well, I think we can close this PR 😢

quantumsheep:
Reverting my choice, the actual implementation is not that great, I tried to fit the same highlighting as GitHub does for LaTeX files. This can be discussed tho.

yyyc514:
Our tex grammar looks a bit more complex than yours.  What does it NOT handle properly if you run the included sample tests against it?

quantumsheep:
It does not properly handle text inside `{}` and `[]`, it colors it as `string` but it shouldn't in certain cases, `\textit{}` for example. But I should make modifications to the existing code instead of creating a new one.

yyyc514:
> But I should make modifications to the existing code instead of creating a new one.

Exactly! Unless it's VERY VERY badly broken, indeed... but still the PR would need to be change against the existing source, not a whole new file...

And I think we probably could add "latex" as an alias here also (if you're going to submit a new PR).

Closing this PR.  Another one that modifies the existing grammar would be great!

quantumsheep:
I'll work on it!

yyyc514:
Should I assume you actually use Latex in a career or hobby?  Is Latex the more popular naming vs "Tex"?

quantumsheep:
I use it for both career and hobby! `TeX` is the programming language, `LaTeX` is a document maker using `TeX`, I badly named my PR too haha.

yyyc514:
Yes, I googled since then... though for practical purposes I'm still thinking perhaps Tex/Latex as a single language makes more sense here, what do you think?

Latex is really a super-set of Tex (or a library built on top of it, etc), yes?

quantumsheep:
LaTeX seems more like a package for the Tex language since it only introduce macros (`\texit`, `\section` and a lot more). But it could makes sense to use "LaTeX" while speaking about Tex since it's what's more commonly used. We can compare it to the JavaScript, we could say "ECMAScript" but we say "JavaScript".

yyyc514:
> LaTeX seems more like a package for the Tex language since it only introduce macros (\textit, \section and a lot more).

Right, so LaTex would cover both Tex and Latex... but Tex wouldn't include Latex. (if we're being technical). We have precedent already for using supersets... like using JSON for Json + Comments, and using INI for INI + TOML, etc...

quantumsheep:
Most LaTeX highlighters call the language "LaTeX" so yeah, it would include TeX and LaTeX coloration if we would make it perfect (italic text for `\textit`, specific coloration for `\documentclass`, etc...).

When I have time I'll work on an upgrade of the current version in hightlight.js.

egor-rogov:
I use LaTeX too. There are couple of other systems built on top of TeX, like XeTeX and LuaTeX, but looks like LaTeX is most popular. AFAIK no one uses bare TeX these days.
I totally agree that we should extend existing TeX grammar to cover LaTeX extensions. No need for two different languages.

PR #2309, opened 2019-12-04
(docs) More prominent list of supported languages

### LINKED ISSUE #2288: The supported languages list is hard to spot and read
[Documentation](https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html#language-names-and-aliases) points to [README](https://github.com/highlightjs/highlight.js) file, but spotting tine little '►' is hard and confusing task. More than that, even spotted it is hard to read and seem out of place. It might be a good idea to move it into a separate file for easy linking and easier searching.

codingthat:
I didn't see the list in the readme right away. It's a common thing to need to refer to, so it makes sense to make it a bit more visible, even if we want to collapse it by default to make the rest of the document more readable.

Resolves #2288


yyyc514:
What do you think about removing summary/details and just including the list inline?  I think that would help even further.

codingthat:
@yyyc514 I'm up for that, but since it used to be that way, I assumed it changed from that for a reason.  This PR was trying to stay in line with the collapsing of it while still helping its visibility a bit.

Samoxive:
This seems like a good middle way, keeping the list inline may be too annoying for people with small resolution screens or mobile readers.

yyyc514:
@codingthat If you want a mention in the changelog, please add one.

PR #2311, opened 2019-12-06
enh(parser): Better errors when one language requires another

### LINKED ISSUE #1899: error registering arduino language

yyyc514:
Closes #1899.

### DIFF
--- a/src/languages/arduino.js
+++ b/src/languages/arduino.js
@@ -94,7 +94,7 @@ function(hljs) {
         'DEFAULT OUTPUT INPUT HIGH LOW'
   };
 
-  var ARDUINO = hljs.getLanguage('cpp').rawDefinition();
+  var ARDUINO = hljs.requireLanguage('cpp').rawDefinition();
 
   var kws = ARDUINO.keywords;
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New styles:
   none.
 
 Improvements:
+- enh(parser): add better error when a language requirement is missing () [Josh Goebel][]
 - enh(ebnf) add backticks as additional string variant (#2290) [Chris Marchesi][]
 - fix(themes): fix inconsistencies between some themes padding/spacing (#2300) [Josh Goebel][]
 - chore(javascript): add esm related extensions to aliases (#2298) [Rongjian Zhang][]
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -126,6 +126,18 @@ Looks up a language by name or alias.
 Returns the language object if found, ``undefined`` otherwise.
 
 
+``requireLanguage(name)``
+---------------------
+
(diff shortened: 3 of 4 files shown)


yyyc514:
Screen Shot 2019-12-06 at 3 23 08 PM

yyyc514:
@egor-rogov Simply makes requires explicit rather than implicit so we can provide better errors when something goes wrong vs some generic error.

yyyc514:
Since this typically happens in registerLanguage it benefits from all the new safe mode stuff... ie you get the helpful errors in the console (shown above), but overall usage of Highlight.js continues to work (other than say the Arduino language, which would be broken).

src/languages/arduino.js
```diff
@@ -94,7 +94,7 @@ function(hljs) {
         'DEFAULT OUTPUT INPUT HIGH LOW'
   };
 
-  var ARDUINO = hljs.getLanguage('cpp').rawDefinition();
+  var ARDUINO = hljs.requireLanguage('cpp').rawDefinition();
```
egor-rogov: Shouldn't this go to v.10 as a breaking change?

yyyc514: Except it's not a breaking change.   Unless you're talking about using languages compiled for 9.17 with 9.16... that wouldn't work... but otherwise there is no breakage here... we're adding a new API, not renaming it.

egor-rogov: Ah, right. If you have no cpp, arduino won't work anyway (:

yyyc514: My new build system resolves dependencies (so asking for `arduino` will always also get you `cpp`)... but still people requiring languages one at a time with NPM or something can get bitten by this... not sure how to prevent people from shooting themselves in the foot.

Issue #1759, opened 2018-07-19
Build system for 2019


isagalaev:
Our build system has outlived its time and needs to be replaced.

## Current principles

- The library has to work in the browser as well as on the server under node.js. To achieve this the code is kept devoid of any module structure and the build system is responsible for wrapping and concatenating files into either a browser-ready script or a node.js compatible module.

- We don't simply host a library on npm and use a browserifying solution because our custom build system has custom compression code which is more efficient than generic solutions. It used to matter when we tried to keep the default build size under 30 KB.

- Browsers seriously suck at requesting many resources at once over HTTP/1.1, which is the reason to have a single packed file with the core and a subset of languages.

- There is a "common" subset of languages which is built for every release and hosted on a CDN. It is a subset because the full set weighs 500+ KB and most of the languages is a long tail that most users don't need.

## Pain points

- Contributors don't like to mess with weird custom systems. This discourages contributions.
- Users don't understand why their "uncommon" language doesn't work. Requiring to read README for basic usage is bad UX.
- Build system is implemented using an old unsupported library (if I remember correctly).

## Proposals

- Host the library in the most common and obvious way in today's JavaScript world (I'm not the one to offer specifics).
- Provide languages as separate files, so users can initialize the library with an explicit list of languages which will be downloaded dynamically. We had that in pre-historic time, but these days it should actually be feasible with HTTP/2. This should fix the second pain point, as users will know what they're requesting.
- Drop the custom compression scheme.
- (*optional*) It might make sense to severely limit the set of languages supported within the main repository, with others moving outside for community support, hosted somewhere else. For that the API between the library and languages needs to (finally) be stabilized.

marcoscaceres:
>  It might make sense to severely limit the set of languages supported within the main repository, with others moving outside for community support, hosted somewhere else. For that the API between the library and languages needs to (finally) be stabilized.

New languages should move to npm, and not maintained by us. We need to look at how Babel plugins work, for instance. People should not be blocked on us and should not depend on us for their own languages.

yyyc514:
I tested just now the custom build with and without compression.  I still minified it... I just skipped compress log, replace2,3,4, etc...

The file sizes (pre gzip) were 743kb vs 792kb... (that's a full build AFAIK, no limits specified)

And after gzip:
251kb vs 254kb

[...]

Issue #2291, opened 2019-11-20
highlightjs.org build seems to differ from `./tools/build.js -t browser :common`


yyyc514:
**Expected behavior: The build would be identical to building locally using the build process.**

Copying from https://github.com/highlightjs/highlight.js/issues/1245

> I've asked few times before, where is the code which is responsible for creating the download and/or concatenating the fragments into a single download?

You'd think it was the build process, but it seems the results are different. Only @isagalaev knows I think. The code generated on the website is putting hljs.registerLanguage OUTSIDE the original enclosure and the code generated by the build process places it inside. @isagalaev Is the website using a special build system instead of the one in the repo for some reason? hljs as a global only works in the web environment (where anything on window is treated as a global), so that is why it's failing in node.js.

yyyc514:
This will become even more important as we move forward and CHANGE the codebase and build system... the website needs to be using the same exactly build system we are using for development work and continuous integration or else this problem is going to get worse in the near future - with no guarantee that the builds the website produces are even usable.

yyyc514:
@isagalaev Ping.

isagalaev:
@yyyc514 I've been missing out on my bug mail lately… Life stuff, as usual. And I do remember I wanted to publish/open up the site's code, but haven't gotten to it yet.

Now, onto the build stuff. The site doesn't use `build.js -t browser` because it's supposed to build a custom package according to check boxes, and building it from scratch every time is too expensive. So instead I do this:

- upon a new release the site runs `build.js -t cdn` and stores all pre-build language files in a directory, and then `build.js -t cdn none` to create a pre-build core without any language files and stores it too
- upon a download request it concatenates them into a single piece and zips it on the fly, which is fast

So there is a (undocumented) assumption that a browser build is basically just a concatenation of files stored on a CDN.

Did `build.js -t browser` change recently which could break this assumption?

yyyc514:
> So there is a (undocumented) assumption that a browser build is basically just a concatenation of files stored on a CDN.
> Did build.js -t browser change recently which could break this assumption?

Obviously true, but I'd say long term that is a faulty assumption.  It prevents us from doing useful things (for a monolith build) like dead-code/constant elimination, allowing languages to share code (at compile time via ES6 modules), etc...

I've been playing with ES6 module based builds that import everything in one file and then compile/minify/terser that result, and it's also (not surprising) smaller than just concat-ing a bunch of individual files with all the repeated `registerLanguage` boilerplate.

[...]

PR #2312, opened 2019-12-06
New build system (including 3rd party languages)

### LINKED ISSUE #1425: replaceSkippingStrings does not properly understand /* style */ comments [auto-close new build system]
While working on a language definition file I stubmled upon the following error when building HLJS:

````
Starting build.
Copying documentation.
Writing documentation.
Generating demo.
Building highlight.js pack file.
Compressing highlight.js pack file.
{ task: 'tasks',
  error: { task: 'replaceSkippingStrings', error: 'Unmatched quote' } }
````

The error was risen by a multiline comments block I've added to the lang file using `/* ... */`. When I changed the comments to multiple single-line comments (`// ...`) the error was gone. So I worked to pin down the problem...

I've selectively removed comment lines until I isolated the culprit, but I can't make much sense from what I've found.

The following comment will raise the error:

````
/*
http://www.someurl.com/
*/
````

But if I remove the slash `/` at the end of the line the error dissapears:

````
/*
http://www.someurl.com
*/
````

The error also dissapears by adding another slash `/` at the end of the line:

````
/*
http://www.someurl.com//
*/
````

But a third slash bring the problem back again:

````
/*
http://www.someurl.com///
*/
````

And it doesn't matter is the slash is the last char on the line or if there are other characters following it — it's the presence of an ODD number of slashes on the line that raises the error!

It seems a problem relating to parsing multiline comments and the way slashes are interpreted — and the presence of two slashes in the URL is what sets into motion. A line with an odd number of slashes seems to raise the error, not so when their number is even. The  'Unmatched quote' error reported is a bit misleading, at the beginning I though it was due to the single quotes, but it wasn't.

The problem doesn't show up when using single line comments:

````
// http://www.someurl.com/
````

So we are not talking about an edge-case, similar URLs are common within comment blocks. The error blocks the building process, and doesn't provide much info on where the process failed, so users might need quite some time to isolate the specific line which is causing build failure.

### LINKED ISSUE #2061: Publish non-minified distributable on cdnjs
cdnjs provides only minified js files
I tied to use other providers,like npm, but that version doesn't work because it's intended to be used only on Node.js.
Can you publish non minified version on cdnjs?
Thank you.

### LINKED ISSUE #2101: building with recursive language dependencies should show a pretty error
I've been looking into improving the dart language definition; to get started, I built highlight.js locally for dart and received the following error (see below).

After some investigation, it seems that this is caused by the dependency on markdown which, in turn, depends on xml. This appears to be the only dependency chain in the language definition files. I haven't investigated in depth, but it seems that the build tools are currently unable to handle recursive dependencies.

```
  1 [I] username@username highlight.js> node tools/build.js dart
  2 Starting build.
  3 Copying documentation.
  4 Writing documentation.
  5 Generating demo.
  6 Building highlight.js pack file.
  7 /home/username/githubext/highlight.js/tools/tasks.js:35
  8     if(!object.processed) {
  9                ^
 10 
 11 TypeError: Cannot read property 'processed' of undefined
 12     at pushInBlob (/home/username/githubext/highlight.js/tools/tasks.js:35:16)
 13     at /home/username/githubext/highlight.js/tools/tasks.js:47:9
 14     at arrayEach (/home/username/githubext/highlight.js/node_modules/lodash/lodash.js:516:11)
 15     at Function.forEach (/home/username/githubext/highlight.js/node_modules/lodash/lodash.js:9342:14)
 16     at /home/username/githubext/highlight.js/tools/tasks.js:45:9
 17     at /home/username/githubext/highlight.js/node_modules/lodash/lodash.js:4905:15
 18     at baseForOwn (/home/username/githubext/highlight.js/node_modules/lodash/lodash.js:2990:24)
 19     at /home/username/githubext/highlight.js/node_modules/lodash/lodash.js:4874:18
 20     at Function.forEach (/home/username/githubext/highlight.js/node_modules/lodash/lodash.js:9342:14)
 21     at Queue.tasks.reorderDeps (/home/username/githubext/highlight.js/tools/tasks.js:41:5)
 22     at Queue._dispatch (/home/username/githubext/highlight.js/node_modules/gear/lib/queue.js:101:18)
 23     at Array.<anonymous> (/home/username/githubext/highlight.js/node_modules/gear/lib/tasks/tasks.js:34:18)
 24     at listener (/home/username/githubext/highlight.js/node_modules/async/lib/async.js:493:46)
 25     at /home/username/githubext/highlight.js/node_modules/async/lib/async.js:444:17
 26     at Array.forEach (<anonymous>)
 27     at _each (/home/username/githubext/highlight.js/node_modules/async/lib/async.js:46:24)
 28 

```

### LINKED ISSUE #2234: Current browser assets not strict-mode safe (langApiRestored)
In strict mode, running 'highlight.min.js' will throw an error.
Probably because the official compression destroys the properties of some objects
The file version obtained from the npm repository is different from the file obtained from other cdn repositories
```js
import('https://cdn.staticfile.org/highlight.js/9.15.10/highlight.min.js').catch(console.error)
```
```
TypeError: Cannot create property 'langApiRestored' on string 'self'
    at i (highlight.min.js:2)
    at Array.forEach (<anonymous>)
    at i (highlight.min.js:2)
    at Array.forEach (<anonymous>)
    at i (highlight.min.js:2)
    at Array.forEach (<anonymous>)
    at i (highlight.min.js:2)
    at Array.forEach (<anonymous>)
    at i (highlight.min.js:2)
    at Object.n.registerLanguage (highlight.min.js:2)
```
```js
    function i(e) {
        if (a && !e.langApiRestored) {
            for (var t in e.langApiRestored = !0,
            a)
                e[t] && (e[a[t]] = e[t]);
            (e.c || []).concat(e.v || []).forEach(i)
        }
    }
```

### LINKED ISSUE #2328: Discuss: 3rd party language packages spec
Starting a discussion for an official 3rd party grammar format/spec (directory structure, how modules are exported, etc).  I've been working on adding support to the build process for seamlessly building 3rd party grammars.  The idea being that you just check them out:

```
mkdir extra
cd extra
git clone git@github.com:highlightjs/highlightjs-robots-txt.git
git clone git@github.com:highlightjs/highlightjs-tsql.git
```

And then building "just works":

```
node ./tools/build.js -t browser :common tsql robots-txt
```

And high-level grammar testing just works (detect tests, markup tests, etc) in context of the whole test suite.  Just run tests like normal.  _The tests for extra languages are bundled with the extra languages, but magically "just work"._

In looking at the existing 3rd party stuff it seems we have two major types of layouts.  I think perhaps we could support both.

The slightly flatter npm package:
Screen Shot 2019-12-19 at 9 03 10 AM

A more "bundled"/traditional layout:
Screen Shot 2019-12-19 at 9 03 21 AM

**One alternative to supporting both is to always prefer/require the bundled layout.**  I prefer this layout slightly as it makes it easier to add multiple languages to a single "package/repo".  For example you might bundle Python, Python REPL... bundled layout gives you an easy way to do that (keeping the same directly structure we use internally).  Also if we ever again decided to pull a 3rd party into the core library it'd be trivial to do with the bundled layout as it'd just essentially be copying files from one repo to another.

As to how NPM would work with bundled layout I'd suggest an `index.js` that simply required the individual modules and exported a hash of them by name:

```js
# index.js
const python = require("./src/languages/python")
const pythonREPL = require("./src/languages/python_repl")
module.exports = { python, pythonREPL }
```

Thoughts?

### LINKED ISSUE #2334: Suggest: Change require API for just the core library
Suggestion:

```diff
// require the highlight.js library including all languages
const hljs = require('./highlight.js');

// require the highlight.js library without languages
- const hljs = require("highlight.js/lib/highlight.js");
+ const hljs = require("highlight.js/core.js"); 
```

We could slide this change in with v10.  I hate the redundancy and think this is a bit nicer.  Unless there is some JS convention here that says the other way is the normal way everyone expects libraries to be structured?

I think we could even preserve the old API and just deprecate it.

@egor-rogov Thoughts?

yyyc514:
The idea: Simplify the build process (that currently uses gear) to simple pure JS (with minimal dependencies) that is easy to read and understand.

Closes #2101.
Closes #2061.
Closes #2234 
Closes #1425 
Related #1759.
Should close #2328 also.
Closes #2334.

---

High level overview:

- `tools/rollup_browser.js` - New build script for Browser
- `tools/rollup_cdn.js` - New build script for CDN
- alters `tools/build.js` to hook the new CDN and Browser builds into the old Gear system
- `tools/lib` (some utils/tools that the new build scripts depend on)

This directly integrates Rollup, Terser, and CleanCSS for the minification, ES6 module support (all the languages are now ES6 modules), etc.

- The build code for `browser` (before the Gear stub) is < 126 lines.
- The build code for `cdn` (before the Gear stub) is < 64 lines.

The idea is clean and readable code that hopefully explains itself. 

**Building** the `browser` build:

```js
async function buildBrowser(options) {
  var languages = getLanguages()
  // filter languages for inclusion in the highlight.js bundle
  languages = filter(languages, options["languages"])

  await installDocs();
  await installDemo(languages);

  log("Preparing languages.")
  await Promise.all(
    languages.map(async (lang) => {
      await lang.compile({terser: config.terser});
      process.stdout.write(".");
     } )
  )
  log("")

  var size = await buildBrowserHighlightJS(languages)

  log("-----")
  log("Core                :", size.core ,"bytes")
  log("Languages           :", languages.map((el) => el.data.length).reduce((acc, curr) => acc + curr, 0), "bytes")
  log("Languages (min)     :", languages.map((el) => el.minified.length).reduce((acc, curr) => acc + curr, 0), "bytes")
  log("highlight.min       :", size.regular ,"bytes")
  log("highlight.min.js    :", size.minified ,"bytes")
  log("highlight.min.js.gz :", zlib.gzipSync(size.data).length ,"bytes")
  log("-----")
}
```

**Installing demo:**

```js
async function installDemo(languages) {
  log("Writing demo files.")
  mkdir("demo")
  installDemoStyles();

  const assets = await glob("./demo/*.{js,css}")
  assets.forEach((file) => install(file))

  const css = await glob("styles/*.css", {cwd:"./src"})
  const styles = css.map((el) => (
    { "name": _.startCase(path.basename(el,".css")), "path": el }
  ) )
  renderTemplate("./demo/index.html", "./demo/index.html", { styles , languages })
}

```

The build output looks something like (depending on build):

```
Starting build.
Writing docs files.
Writing demo files.
Writing style files.
Preparing languages.
..................................
Building highlight.js min file.
-----
Core                : 34596 bytes
Core (min)          : 12521 bytes
Languages           : 131641 bytes
Languages (min)     : 74562 bytes
highlight.js        : 167053 bytes
highlight.min.js    : 87116 bytes
highlight.min.js.gz : 28811 bytes
-----
Finished build.
```

The most complex high level code is `buildBrowserHighlightJS`, but I think that's understandable since this is the code building the full browser package really.

The dependency stuff is also a bit verbose (and probably could use a simpler algorithm)... but it works and is well isolated in `lib/dependencies.js`, so I think it's ok for now.


yyyc514:
My plan is to do this in two steps... Switch CDN/browser with one release, then finish up [NPM] with a second release... that second release is where we'd simplify `build.js` a bit further and remove all the old gear stuff completely.


There might also be a blocker with:
https://github.com/highlightjs/highlight.js/issues/2291

Not sure.

yyyc514:
I do use `child_process` to get the git revision, which I'd love to include in the built packages.  Open to thoughts on a better way to do this or if there are any concerns:

```
var git_sha = child_process.execSync("git rev-parse HEAD")
```

The new header:

```
/*
  Highlight.js 9.16.2 (a40e3dda)
  License: BSD-3-Clause
  Copyright (c) 2006-2019, Ivan Sagalaev
*/
```

egor-rogov:
This will require some time to get grasp of the changes. I can't promise to do it quickly, although I'll try to find time.

tools/lib/dependencies.js
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Reorders languages, moving dependencies above the languages
+ * that depend on them.
+ *
+ * @param {array<Language>} languages list of languages to be reordered
+ * @returns {array<Language>} ordered list of languages
+*/
+
+const reorderDependencies = (languages) => {
```
egor-rogov: This function is not evidently correct, so I'd like to have a test for it.
(I would trust in building a tree of dependencies and then pushing out languages in breadth-first order.)

tools/build.js
```diff
@@ -77,7 +77,14 @@ commander
...
 
-build     = require(`./${commander.target}`);
+var target = commander.target
+if (target=="cdn")
+  target = "./rollup_cdn"
+if (target=="browser")
+  target = "./rollup_browser"
+
+
+build     = require(`./${target}`);
```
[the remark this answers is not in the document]
yyyc514: I guess because I was still referencing them.  I don't mind renaming them though for the final version of this.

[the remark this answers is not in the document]
yyyc514: > And what about language definitions in separate repos, are they also supposed to use it?

I still need to review that.  It might make more sense for languages to use commons require format (making them usable with both node.js directly AND the build system - maximum flexibility).  I was going to play around and see what worked best.

But yes the idea is to come up with a common standard and then the blessed way to build a monolithic build would be if all the packages followed the preferred standard. 

This is still WIP. :-)  

> Is roll.js used at all? It looks like a first attempt that wasn't deleted.

Yeah, more WIP.  This hasn't been cleaned up yet.

tools/build_config.js
```diff
@@ -0,0 +1,66 @@
...
+          unsafe: true,
+          warnings: true,
+          dead_code: true,
+          toplevel: "funcs",
+          global_defs: {
+            LEGACY: false,
+            DEBUG: false
+          }
+        },
+        "mangle": {
```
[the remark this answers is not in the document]
yyyc514: Nope... or else I'll circle back around and end up uncommenting out a few things, we'll see.

tools/lib/dependencies.js
yyyc514: For building, I think the proof is in the results.  I'm not sure how I feel about writing tests for the build system (we don't have any presently).  Do you think the algorithm isn't self-evident?  It's really just a bubble sort.

If you insist I'd rather spend my time making the load order irrelevant and eliminate this as an issue rather than write tests... This is my least favorite code but I just wanted something simple that I knew worked.

tools/lib/bundling.js
```diff
@@ -0,0 +1,59 @@
+const rollup = require('rollup')
+
+async function buildOutput(inputOptions, outputOptions) {
+  const bundle = await rollup.rollup(inputOptions).catch(err => console.log(err));;
+  const { output } = await bundle.generate(outputOptions);
+  return output[0].code;
+}
+
+async function build(inputOptions, outputOptions) {
```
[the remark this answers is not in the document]
yyyc514: Yeah I might circle back around and try and make this a bit more modular.

yyyc514:
How does it feel overall though?  Does it seem clear and easy to follow at a big picture/high level?  I just wanted to make sure it's making sense before I throw more time at it.

tools/lib/dependencies.js
yyyc514: The high level (meta) test for the build system should be that our built deliverables actually work, and we already have tests for that (many for node, and a few for browser).  And CDN is really just browser + separate language modules.

egor-rogov: > Do you think the algorithm isn't self-evident? It's really just a dumb bubble sort.

It doesn't look self-evident to me. It looks like it can fail when depth of dependencies is more then MAX_REORDER_ATTEMPTS. I understand it's unlikely (probably even impossible), but I'd prefer more robust implementation.
E. g. first goes the languages which require nothing, then languages which require already processed languages, and so on. Repeat the process until the list of processed languages stabilizes. All languages that remain unprocessed are known to have recursive dependencies.

egor-rogov:
> How does it feel overall though? Does it seem clear and easy to follow at a big picture/high level? I just wanted to make sure it's making sense before I throw more time at it.

At a high level it looks pretty clear and certainly makes sense. Just requires some polishing.

yyyc514:
> At a high level it looks pretty clear and certainly makes sense. Just requires some polishing.

Perfect.  I'll circle back and get out my polish tools, and then ping you again later.

tools/lib/dependencies.js
yyyc514: You want to take a pass at it?  I just wanted something basic and functional. :-)

>  It looks like it can fail when depth of dependencies is more then MAX_REORDER_ATTEMPTS

Maybe, and then it fails with a nice error - that's intentional. :-)  But I think that very, very unlikely to happen.  It really would only happen in case of a circular dependency, which is what the error says.

yyyc514: I guess now I'll take a few minutes nd look into removing the whole dependency ordering issue in the first place.

yyyc514: npm to the rescue: 

```js
  let resolver = new DependencyResolver();
  for (let lang of languages) {
    resolver.add(lang.name);
    for (let required of lang.requires) {
      resolver.setDependency(lang.name, required);
    }
  }
```

yyyc514: I did first try to just remove the whole ordering thing but the tricky problem is aliases... we don't know which aliases map to what until AFTER we've truly registered the languages...

egor-rogov: I like it better this way.

> but the tricky problem is aliases
Do you mean specifying aliases in a build command (instead of "real" names)? AFAIK it doesn't work now, and I don't think we need to support it.

yyyc514: > Do you mean specifying aliases in a build command (instead of "real" names)? AFAIK it doesn't work now, and I don't think we need to support it.

No... I mean that if getLanguage only worked on non-aliases it'd be easy to defer the registration... but since you can pass aliases, now there is no way of knowing which language actually needs to be loaded (it might be an alias).

And you can't just load them all the first time it's called because getLanguage can be called recursively... so then it goes into an endless recursive loop.

Possibly it's fixable, but it wasn't the easy task I thought, so I just went and found a dependency resolver - and it works great. :)

yyyc514:
@egor-rogov You want to take another look at this now, it's a lot more polished.

I'm still waiting for more feedback on:

https://github.com/highlightjs/highlight.js/issues/2328

But really that only affects the actual detection of third party language packages (external_language.js)... and I've tried to keep all that bottled into a single place.

egor-rogov:
> @egor-rogov You want to take another look at this now, it's a lot more polished I think.

Yeah, I'm going to, but have to find a big enough time quantum...

tools/lib/makestuff.js
```diff
@@ -0,0 +1,34 @@
...
+  fs.copyFileSync(file, `${process.env.BUILD_DIR}/${dest}`);
+}
+
+function install_cleancss(file, dest) {
+  let content = fs.readFileSync(file, {encoding: "utf8"});
+  let out = new CleanCSS(config.clean_css).minify(content).styles;
+  fs.writeFileSync(`${process.env.BUILD_DIR}/${dest}`, out);
+}
+
+function mkdir(dirname) {
```
egor-rogov: I don't like the name very much, because it looks like a general-purpose function, but it's works for BUILD_DIR only. Something like `build_mkdir` would be better, although I'm not totally happy with this name either.

tools/lib/language.js
```diff
@@ -0,0 +1,112 @@
...
+
+  var original = data;
+  language.module = data;
+  data = Terser.minify(data, options["terser"]);
+  language.minified = data.code || original;
+}
+
+async function getLanguages() {
+  let languages = [];
+  fs.readdirSync("./src/languages/").forEach((file) => {
```
egor-rogov: Let's filter only `*.js` files? Git working directory can contain some other files (.orig/.rej from applying patches, backup copies etc).

tools/build_cdn.js
```diff
@@ -0,0 +1,90 @@
...
+  log("All Lang            :",
+    languages.map((el) => el.minified.length).reduce((acc, curr) => acc + curr, 0), "bytes");
+  log("highlight.js        :",
+    size.regular, "bytes");
+
+  if (options.minify) {
+    log("highlight.min.js    :", size.minified ,"bytes");
+    log("highlight.min.js.gz :", zlib.gzipSync(size.data).length ,"bytes");
+  } else {
+    log("highlight.js.gz     :", zlib.gzipSync(size.fullSrc).length ,"bytes");
```
egor-rogov: `fullSrc` -> `data`

tools/lib/dependencies.js
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+  }
+
+  // resolve requires
+  for (let lang of languages) {
+    lang.requires.forEach(needed => {
+      if (!languages.find((el) => el.name == needed)) {
+        console.info(`[INFO] Adding ${needed}... ${lang.name} requires ${needed}.`)
+        languages.push(allLanguages.find((el) => el.name == needed))
```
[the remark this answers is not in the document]
yyyc514: No, it's simply adding requires.  Duplicates would be resolved by the dependency resolver.

tools/lib/makestuff.js
yyyc514: Node already has a built in `mkdir` sync and unsync versions... I think these are build tool internals and if you want to know the specifics you read them.  i think contextual when you read the build scripts and see the build output it's pretty clean everything is working in terms of the build directory, so I'm not sure I see this as a real issue.

tools/lib/language.js
yyyc514: Agreed.

tools/build_cdn.js
yyyc514: Not they are different things, but the naming is poor.  I will improve naming of these items.

tools/lib/makestuff.js
yyyc514: So, willing to let this slide?  the one thing I can think of would be to namespace a little more:

```
const build = require("./lib/builddir_make");

...
build.mkdir()
```

Personally I think it's a bit verbose and doesn't really add that much clarity.

tools/lib/dependencies.js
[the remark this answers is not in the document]
egor-rogov: Then why do you need this check in line 75?
```js
 if (!languages.find((el) => el.name == needed)) {
```

tools/build_cdn.js
egor-rogov: Yeah, I see now. Previously I got an error during CDN build with --no-minify option in this line, so I thought that the name was incorrect. Can't reproduce now.

tools/lib/dependencies.js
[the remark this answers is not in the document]
yyyc514: There is no need to add a requirement twice... one task shouldn't NOT do something entirely obvious just because another later MIGHT take care of it... **when adding dependencies there is no need to add a dependent that already exists.** Even if it might get cleaned up later.

I see no issue here.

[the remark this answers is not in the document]
yyyc514: Perhaps if it was all a single function, but it's not... and also the logging would not be nearly so helpful if it just added things over and over even if they already existed.

yyyc514:
> (For the future, I'd prefer several smaller PRs, 'cause it's much easier to move in smaller steps...)

Agree, but so much of it was interconnected and the plan came together as it happened.  I see this as a pretty unusual type thing though in its complexity... should be simpler things in the future. :-)

tools/lib/makestuff.js
yyyc514: Yeah, I guess I see the slight issue with the name overlap now... something to think about next time we rewrite the build system. :-)

PR #2313, opened 2019-12-07
enh(docs) improve API docs

### LINKED ISSUE #1576: Documentation is confusing
http://highlightjs.readthedocs.io/en/latest/api.html#highlight-name-value-ignore-illegals-continuation

Listed arguments don't match with the signature. It's very confusing and I have to read the code to understand what the arguments are.

Edit: I just realized I confused input arguments with return value. It should be more clear. I suggest putting the "Returns an object..." sentence in a separate paragraph, and naming the first argument "language" or "language_name", the second "html_string", etc.

yyyc514:
Closes #1576.

This doesn't change any API, just renames a few arguments to be clearer.


src/highlight.js
```diff
@@ -513,16 +513,24 @@ https://highlightjs.org/
...
+   * @returns an object that represents the result
+   * @property {string} language - the language name
+   * @property {number} relevance - the relevance score
+   * @property {string} value - the highlighted HTML code
+   * @property {mode} top - top of the current mode stack
+   * @property {boolean} illegal - indicates whether any illegal matches were found
   */
-  function highlight(name, value, ignore_illegals, continuation) {
+  function highlight(languageName, code, ignore_illegals, continuation) {
+    var value = code;
```
egor-rogov: Is there any reason not to replace `value` with `code`?

yyyc514: Wasn’t in the mood for a search and replace battle and value is so generic and the scope was large. Lol.

yyyc514:
@egor-rogov I renamed value but left a longer name to make this easier to change in the future.  Right now our scope is too large IMHO... otherwise this wouldn't be an issue.

src/highlight.js
```diff
@@ -513,16 +513,24 @@ https://highlightjs.org/
...
+   * @returns an object that represents the result
+   * @property {string} language - the language name
+   * @property {number} relevance - the relevance score
+   * @property {string} value - the highlighted HTML code
+   * @property {mode} top - top of the current mode stack
+   * @property {boolean} illegal - indicates whether any illegal matches were found
   */
-  function highlight(name, value, ignore_illegals, continuation) {
+  function highlight(languageName, code, ignore_illegals, continuation) {
+    var codeToHighlight = code; // for search and replace
```
egor-rogov: `codeToHighlight` is fine, but again, why not simply
```js
function highlight(languageName, codeToHighlight, ignore_illegals, continuation) {
```
?
Doesn't renaming by assigning the value to another variable look strange?

yyyc514: I see what you're getting at, but doesn't bother me. In the method definition all that wording is extraneous... highlight(codeToHighlight) is silly and repetitive.  But in the huge scope this variable is used in the longer naming makes a lot more sense.

So this is the best of both worlds IMHO.

If you insist I'll change it, but I'd rather not. :-)

egor-rogov: Okay, agree. But let's remove `// for search and replace`?

yyyc514: LOL, I thought that would appease you... I'll remove.

egor-rogov: I think it's rather more confusing (:

PR #2319, opened 2019-12-11
fix: Issue with IE11 and Object.freeze()

### LINKED ISSUE #2318: Error with Object.Freeze() in IE 11
Hi, Thanks for the plugin!

When my site is run in IE 11 there is an error with the "object.freeze()". Do you know a workaround I can put in? Thanks!
image

yyyc514:
Closes #2318.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,13 @@ Language Improvements:
 - none yet.
 
 
+## Version 9.17.1
+
+Fixes:
+
+- fix(parser): resolve IE 11 issue with Object.freeze() (#2319) [Josh Goebel][]
+
+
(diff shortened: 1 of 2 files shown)


christopherpickering:
Thanks, I built and tested, but still get the same error. I think the function Object.freeze(o); just doesn't exist in IE 11.

image

If this section of the code is removed it still works well (not sure what else is affected!) but in my app its cool.. maybe performance is slightly affected? I'm using highlight.js for a handful of  sql code blocks.

If you have another idea I'd appreciate it! - otherwise I can use the code w/ this block commented out. (lines 1130 - 1151 in src/highlight.js)

image

PR #2322, opened 2019-12-13
(javascript) fix jsx self-closing tag issues

### LINKED ISSUE #1915: (jsx) multiple self-closing tags with attributes highlighting broken
Given jsx code:
```javascript
class App extends Component {
  render() {
    return (
      <BrowserRouter>
        <div>
          <Route path="/about" component={About} />
          <Route path="/contact" component={Contact} />
        </div>
      </BrowserRouter>
    );
  }
}
```

It seems that the XML code is not highlighted because of the second `Route` tag with attributes.
When I tried to modify `src/language/javascript.js`
```diff
          { // E4X / JSX
            begin: /</, end: /(\/\w+|\w+\/)>/,
            subLanguage: 'xml',
            contains: [
              {begin: /<\w+\s*\/>/, skip: true},
              {
-               begin: /<\w+/, end: /(\/\w+|\w+\/)>/, skip: true,
+               begin: /<\w+/, end: /(\/\w+[^>]*|\w+\S*\s*\/)>/, skip: true,

                contains: [
                  {begin: /<\w+\s*\/>/, skip: true},
                  'self'
                ]
              }
            ]
          }

```
The highlighting works as expected.

yyyc514:
Closes #1915.

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -188,17 +188,23 @@ function(hljs) {
             skip: true,
           },
           { // E4X / JSX
-            begin: /</, end: /(\/[A-Za-z0-9\\._:-]+|[A-Za-z0-9\\._:-]+\/)>/,
+            className: 'jsx',
+            begin: /<[A-Za-z0-9\\._:-]+/, end: /(\/[A-Za-z0-9\\._:-]+\/?)>/,
             subLanguage: 'xml',
             contains: [
+              // self-closing tag
               { begin: /<[A-Za-z0-9\\._:-]+\s*\/>/, skip: true },
+              // manually closed tag
               {
-                begin: /<[A-Za-z0-9\\._:-]+/, end: /(\/[A-Za-z0-9\\._:-]+|[A-Za-z0-9\\._:-]+\/)>/, skip: true,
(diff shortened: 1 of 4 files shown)


src/languages/javascript.js
```diff
@@ -6,6 +6,8 @@ Website: https://developer.mozilla.org/en-US/docs/Web/JavaScript
 */
 
 function(hljs) {
+  var TAG_START = /<[A-Za-z0-9\\._:-]+/;
+  var TAG_FINISH = /\/[A-Za-z0-9\\._:-]+>|\/>/;
```
luisherranz: Hey guys, React now support fragments which are usually written as `<>...</>`.

So maybe this could be
```js
var TAG_START = /<[A-Za-z0-9\\._:-]*/;
var TAG_FINISH = /\/[A-Za-z0-9\\._:-]*>|\/>/;
```

What do you think? Do you see any problem?

yyyc514: I think I need an actual real example of what you're talking about.

yyyc514: Well now if that ain't just ridiculous. :-)

https://reactjs.org/docs/fragments.html#short-syntax

Could you open an actual issue for this, please?

luisherranz: Sure: https://github.com/highlightjs/highlight.js/issues/2327

PR #2325, opened 2019-12-15
Version 10 misc chores and small breaking changes

### LINKED ISSUE #2273: Version 10 Breaking Changes
Creating an issue to keep track of all the "breaking changes" I'd like to do when we get around to major release time.

- [x] Expose parse tree, renderer, and new emitter API. #2404 
- [x]  Rename `cs.js` to `csharp.js`
- [x]  Rename `tex.js` to `latex.js`
- [x] No prohibition against using ES2015 features in the codebase (will cut out a few older browsers we currently technically might support)
- [x] Review ":common" group for potential removals
- [x] remove older onload hook, keeping newer DOMContentLoaded. 
- [x] remove `darkula.css` (mispelling)
- [x] remove `restoreLanguageAPI` (version 9 compressed languages are not compatible with version 10)
- [x] <del>Good time to split C/C++ with a common foundation if we have anyone to work on that</del> Create C/C++ dependency on 'c-like' to prevent future breaking changes https://github.com/highlightjs/highlight.js/issues/2146
- [x] consider XML/PHP entanglement and how to resolve https://github.com/highlightjs/highlight.js/issues/2330
- [ ] **[in progress]** Fix-up `htmlbars` and `handlebars` dependencies https://github.com/highlightjs/highlight.js/issues/2181
- [x] 3rd party language grammar official packaging format https://github.com/highlightjs/highlight.js/issues/2328
- [x] New build system https://github.com/highlightjs/highlight.js/issues/1759
- [ ] Resolve build inconsistency issues: https://github.com/highlightjs/highlight.js/issues/2291
- [x] Switch to createElement vs NS

Thing that probably aren't making it into 10.0

-  Consider deprecating `endSameAsBegin` and encourage callback use instead https://github.com/highlightjs/highlight.js/pull/2261
-  auto highlighting should return a list of possible matches, not just top two
-  Change `highlight(name, x, x, x)` API to `highlight(code, options={})`, we'd still support the legacy API until 11.0 I'd imagine. https://github.com/highlightjs/highlight.js/issues/2277
-  add `baseLanguage` to some language grammars and  https://github.com/highlightjs/highlight.js/issues/2275

yyyc514:
Should go a long ways towards closing #2273, but not completely.


src/highlight.js
```diff
@@ -274,18 +274,6 @@ https://highlightjs.org/
     return [mode];
   }
 
-  function restoreLanguageApi(obj) {
-    if(API_REPLACES && !obj.langApiRestored) {
```
egor-rogov: Do we still need `API_REPLACES` variable declared above?

yyyc514: The existing build system hooks on it I think - and this branch doesn't touch the build system.  I'll get it removed after the two branches are merged though.

yyyc514:
Anything else?

yyyc514:
I think I plan to merge this first, so now the other big PR is pending based on this one being approved. :-)

yyyc514:
> Just take care to resolve conflicts with the new build system PR, e. g. there will be no tools/tasks.js and probably something else.

You don't need to remind me. :-)  I'll make sure it all comes together properly. :-)  Then once this better foundation is in place I'll try to go back to some smaller changes you'll find easier to review. :-)

PR #2332, opened 2019-12-23
enh(cpp) Improve function declaration detection

### LINKED ISSUE #1222: (cpp) decltype(auto) preceded by specifiers
If `constexpr decltype(auto)` occurs in C++(14) code, `decltype` is not highlighted correctly, presumably since this looks like a function declaration with `decltype` being the name. This will also occur with other specifiers, e.g. `friend`, `typedef`, `inline` etc.

In fact, this can occur with keywords other than `decltype`. See e.g. 

```
#define X
X new (p) int;
```

I'd suggest dropping that heuristic entirely.

### LINKED ISSUE #1502: (cpp) template type in function declaration

### LINKED ISSUE #1730: (cpp) types with an underscore aren't recognized as the start of a function
Take for example the following code.
```cpp
    int8_t Get_Tile_Value()
    {
        return ( ( *Region & bitmask ) >> bit_position );
    }
```
If `int8_t` is a recognized type (a la `<script>hljs.getLanguage('cpp').keyword += '  int8_t';</script>`), highlight.js will recognize it and highlight it, but will not highlight the function name.

yyyc514:
- enh(cpp) Properly detect decltype(auto)
- enh(cpp) recognize primitive types (`int8_t`, etc.) as function types
- enh(cpp) Detect namespaced function types (`A::typeName func(...)`)
- enh(cpp) Detect namespaced functions also (`A::functionName`)
- enh(cpp) template type in function declaration (`vector<int> func(...)`)

Closes #1730.
Closes #1222.
Closes #1502.

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -7,6 +7,8 @@ Website: https://isocpp.org
 */
 
 function(hljs) {
+  var DECLTYPE_AUTO_RE = 'decltype\\(auto\\)'
+  var FUNCTION_TYPE_RE = '(' + DECLTYPE_AUTO_RE + '|[a-zA-Z_]\\w*)';
   var CPP_PRIMITIVE_TYPES = {
     className: 'keyword',
     begin: '\\b[a-z\\d_]*_t\\b'
@@ -64,7 +66,13 @@ function(hljs) {
     ]
   };
 
-  var FUNCTION_TITLE = hljs.IDENT_RE + '\\s*\\(';
+  var TITLE_MODE = {
+    className: 'title',
+    begin: '(' + hljs.IDENT_RE + '::)?' + hljs.IDENT_RE,
+    relevance: 0
+  };
+
+  var FUNCTION_TITLE = '(' + hljs.IDENT_RE + '::)?' + hljs.IDENT_RE + '\\s*\\(';
 
   var CPP_KEYWORDS = {
     keyword: 'int float while private char char8_t char16_t char32_t catch import module export virtual operator sizeof ' +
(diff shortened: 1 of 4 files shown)


src/languages/cpp.js
```diff
@@ -7,6 +7,16 @@ Website: https://isocpp.org
 */
 
 function(hljs) {
+  function optional(s) {
+    return '(' + s + ')?';
```
egor-rogov: This function looks useful, I like the clarity of the regexps below. Don't you want to put it into the library?
But probably non-capturing group `(?:`...`)` would be better.

src/languages/cpp.js
```diff
@@ -7,6 +7,16 @@ Website: https://isocpp.org
...
+  }
+  var DECLTYPE_AUTO_RE = 'decltype\\(auto\\)'
+  var NAMESPACE_RE = '[a-zA-Z_]\\w*::'
```
egor-rogov: `hljs.UNDERSCORE_IDENT_RE + '::'` ?
Btw, I'm not 100% sure but looks like C++ allows underscore as the first character in identifiers, so maybe all (most?) instances of `hljs.IDENT_RE` below should really be `hljs.UNDERSCORE_IDENT_RE`. (This is not showstopper of course.)

src/languages/cpp.js
yyyc514: I plan to add it more globally with  modules after the v10 build system hits and see how that 
looks.

> But probably non-capturing group (?:...) would be better.

Agree

yyyc514: > Looks like it will interfere with #2325 though.

Hopefully git will figure it out, time will tell. :-)  I did the rename in a sep commit.

src/languages/cpp.js
yyyc514: Pulling this thought out into it's own issue: https://github.com/highlightjs/highlight.js/issues/2353

Issue #2342, opened 2020-01-02
(fortran) does not highlight block statement


interkosmos:
The Fortran 2008 `block` statement is not recognised by highlightjs. It should be highlighted like any other intrinsic statement of the language. Example:

```
<pre><code class="fortran">program main
    implicit none

    block
        real :: r
    end block
end program main</code></pre>
```
(`block` is not formatted in bold.)

yyyc514:
Confirmed.  Block does not seem to be a keyword at present.

yyyc514:
@interkosmos Any chance you'd be willing to whip up a PR and quick test?

PR #2343, opened 2020-01-02
Added Fortran block statement.

interkosmos:
Added `block` keyword, as described in issue #2342.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,7 +14,7 @@ Core Changes:
 
 Language Improvements:
 
-- (fortran) added block keyword (#2343) [Philipp Engel][]
+- (fortran) added `block` and `endblock` keywords (#2343) [Philipp Engel][]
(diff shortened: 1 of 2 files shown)


yyyc514:
Please add to CHANGELOG.

If the `endblock` form without a space also allowed?  I see a lot of other keywords done like that.

interkosmos:
> Is the `endblock` form without a space also allowed? I see a lot of other keywords done like that.
Yes, `endblock` is accepted, even if `end block` should be preferred.

yyyc514:
You'll need to add `endblock` as a keyword also then.

interkosmos:
Can I somehow change the current PR?

yyyc514:
Just add a second commit.

yyyc514:
Yep. Nice.  Now just a `CHANGES` entry please.

yyyc514:
@interkosmos Thanks so much for this contribution!

PR #2344, opened 2020-01-02
feat(handlebars): add subexpressions, block-params and hash-parameters

### LINKED ISSUE #2181: (htmlbars) should inherit from Handlebars
I am currently working on refining the handlebars-language to support literal-segments:

* [example](https://github.com/nknapp/highlight.js/blob/expression-variants/test/markup/handlebars/block-expression-variants-in-param.txt)
* [description](https://handlebars-draft.knappi.org/guide/expressions.html#literal-segments)

The Handlebars-tests are working fine with my changes, but HTMLBars detection testcase no fails... because `highlight.js` detects Handlebars instead of HTMLBars.

Now, the file that should be detected as HTMLBars, might was well be a Handlebars file

```
<div class="entry">
  {{!-- only output this author names if an author exists --}}
  {{#if author}}
    {{#playwright-wrapper playwright=author action=(mut author) as |playwright|}}
    <h1>{{playwright.firstName}} {{playwright.lastName}}</h1>
    {{/playwright-wrapper}}
  {{/if}}
  {{yield}}
</div>
```
In my oppinion, there is no need for an HTMLBars language, because the syntax is actually the same as Handlebars. It even [uses the Handlebars parser](https://github.com/tildeio/htmlbars/blob/711264d565727ec86fbe9bdb7725706447f0c109/packages/htmlbars-syntax/lib/parser.js) to generate an AST. Couldn't htmlbars just be an alias of Handlebars.

I think the only difference really lies in the list of built-ins. HTMLBars/Glimmer/Ember seems to have many more builtin-helpers. But it should be possible to somehow handle that.

There are parts of the html-bars definition that should be moved into the handlebars-definition anyway (like sub-expressions).

I am willing to put some work into that. I just wanted to ask for your opinions first.

nknapp:
closes #2181 

* Support for sub-expressions: Helper names in sub-expressions are now highlighted properly. The mode-tree is built recursively for sub-expressions so that sub-elements (nested sub-expressions in params) are also highlighted properly (like `{{helper1 (helper2 (helper3 param) ) }}`).
* Support for hashes ("attr-assignments" in the "htmlbars"-language)  (like `{{helper key=value}}` or `{{helper key=(sub-helper value)}}`
* Path expressions are highlighted: `{{path.expression}}`. Up to know, only `path` would be highlighted.
* Block parameters are highlighted correctly: In `{{#with as as | as | }}` , only the middle `as` is a keyword.
* Literals in helper-parameters are highlighted as keyword, number or string.
* Merged built-ins with the list from 'htmlbars'
* Added 'htmlbars' as alias
* Deleted the 'htmlbars' language


nknapp:
@yyyc514 I think I am about ready except for some cleanup and the following questions:

* what should happen to 'htmlbars' now? 
  I have removed the whole language in my branch because the auto-detection tests were failing all the time.
* should the changes go to the wip_version_10 branch?
* The expression inside I basic mustache (`{{test}}`) is now highlighted as "hljs-name" just like in block-mustaches. I felt that they should because they essentially have the same semantics. Should I revert that to retain the previous behaviour?

yyyc514:
> what should happen to 'htmlbars' now?

I've changed my thinking on this and now think it should simply be an alias of handlbars.

> should the changes go to the wip_version_10 branch?

Sure, though I hope to merge it soon anyways, so that's not a necessity.

> The expression inside I basic mustache ({{test}}) is now highlighted as "hljs-name" just like in block-mustaches. I felt that they should because they essentially have the same semantics. Should I revert that to retain the previous behaviour?

What was the previous behavior again?

nknapp:
handlebars-testcases-before
handlebars-testcases-after

nknapp:
What I also changed (in my next version, is the attr-assignment-rule of htmlbars (in Handlebars-speak, this is calls "Hash"). The 'htmlbars'-language highlights the value (behind the equal-sign) as attr, but as I understand it, the part in front of the equal sign should be highlighted.

This change is not visible in the default theme, because the "attr"-class has no special highlighting, but it is there.

yyyc514:
> The expression inside I basic mustache ({{test}}) is now highlighted as "hljs-name" just like in block-mustaches.

I kind of dislike this for simple variable expressions, but I don't really know what to suggest as figuring out one from the other might not be worth the effort.

yyyc514:
>  but as I understand it, the part in front of the equal sign should be highlighted.

Makes sense to me, of course if the value was a number or string I'd still expect it to be highlighted as such.

nknapp:
> I kind of dislike this for simple variable expressions, but I don't really know what to suggest as figuring out one from the other might not be worth the effort.

The distinction here is not between blocks and basic mustaches (see https://jsfiddle.net/nL91aohe/1/):

* `{{test}}` can be a property-access or a function-call (if 'test' is a function) or a helper call (if a helper 'test' is registered)
* `{{#test}}...{{/test}}` can be a block-helper, or a property-access that dives into a object.

yyyc514:
> I just gave the first element in a a basic mustache the class "name", which makes it bold.

Don't presume this, how it looks all depends on the theme used...

nknapp:
This has gotten really complex now. I wonder if you'll be able to review...

It is really difficult to verify the tests manually. I tried to make them self-explanatory, but the HTML-tags in the expect-files really mess up with readability.

It might be a good idea to create use a visual regression-testing tool to create screenshots of the test-cases. Just an idea... Of course, we would need a theme that makes more classes visible...

I'll try to update the description of the PR with a summary of what has changed...

nknapp:
What I am still unsure about is the class for built-in helpers and keywords:

As far as I understand the [class reference](https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html#css-classes-reference) it should be like this:

* built-in helpers like 'each', 'with' (and probably all the other keywords in the built-ins list as well) should get the class `builtin-name` because they are a "name from the language standard library" .
* keywords like `{{else}}` or `as` in `{{#with x as | y |}}` should get the class "keyword" because they are part of the language and not the standard library

What is the 'built_in' class used for? Should any of those be `built_in` instead?

nknapp:
I am done here for now. I think you can start your review...

yyyc514:
I'd love a better way to write expects tests, but it seems a tough problem.  The way we do it now is surely because we never know what might be a validate language so there is no safe way to escape anything inside the test file.... hence we end up with raw text or using comments to describe individual test cases. (but those can get in the way of seeing the test cases themselves)

Something like JSON/YAML would probably be preferable but then optimally we'd have to build out a whole test case editor... which I don't think anyone has the time for.

nknapp:
I was just wondering if you are waiting for a answer to your last comment. But I didn't see a question, that's why I didn't write anything new.

I'm just doing it now, a friendly reminder that this is ready for review from my point of view.

yyyc514:
I asked a few questions in the sub-responses.

test/markup/handlebars/block-expression-variants-in-param.txt
```diff
@@ -1,7 +1,5 @@
-text {{#abc "lite]'ral}}segment" }}a{{/abc}}
+Block expression variants with literal segments in helper param:
```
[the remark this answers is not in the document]
yyyc514: I was hoping you would justify (perhaps I'm not getting the importance) or remove these to make it easier to review the actual meat of this commit and not just all the dressing...

nknapp:
Haven't seen then comment. Will look at it.

test/markup/handlebars/else-variants.expect.txt
```diff
@@ -0,0 +1,14 @@
+<span class="xml">Multiple possibilities of using "else":
+
+</span><span class="hljs-template-tag">{{#<span class="hljs-name">helper</span>}}</span><span class="hljs-keyword">{{else}}</span><span class="xml">else-block</span><span class="hljs-template-tag">{{/<span class="hljs-name">helper</span>}}</span><span class="xml">
```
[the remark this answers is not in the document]
nknapp: I change that. It is now 

```html
<span class="hljs-template-tag">{{<span class="hljs-keyword">else</span>}}</span>
```

test/markup/handlebars/block-expression-variants-in-param.txt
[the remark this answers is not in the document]
nknapp: Both titles and dots are there because I used `developer.html` to inspect the test results.

At first I started to insert one file after another into `developer.html`. Then I started concatenating files. The titles in the files make it easier to see, which file the tests actually belong to (as you can see in the screenshots above)

And sometimes, the output looked good, but on a second glance, I could see that a mode was spanning into the next textcase. The dots `...` are there to ensure that the mode ends where it should. The dots never be highlighted but always be displayed as "content" or "xml".

I think the dots are important. I can remove the titles, although if I start working on the language again, for whatever reason, I would like to have them in there.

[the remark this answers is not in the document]
yyyc514: > And sometimes, the output looked good, but on a second glance, I could see that a mode was spanning into the next textcase. The dots ... are there to ensure that the mode ends where it should. The dots never be highlighted but always be displayed as "content" or "xml".

This sounds like a real problem, but one that needs a better solution than adding a ton of ...s everywhere.  Have you have any lightbulb moment while working on tests on perhaps a better way of doing this?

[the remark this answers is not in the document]
yyyc514: I'm not completely sure I"m following how you're using developer.html... but it sounds like perhaps what your'e doing could/shoudl be automated rather than adding ... and titles everywhere you just run `test php in devtool` and it loads up the devtool with ALL the tests queued with separators, etc...

You can do that without adding all those things to the individual files.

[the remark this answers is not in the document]
nknapp: I think the dots are a valid method to ensure the transition from tag to text. Handlebars is primarily text, with some tags inserted, so we should put some text sections between tags in the test cases. The dots are the most unobtrusive chars I can imagine.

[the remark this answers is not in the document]
nknapp: As for the headers. I didn't want to implement something new. I just used used `ls`, `grep`, `xargs` and `cat` to generate a file with all test cases and copied that into the developer.html

I agree that a solution to view all test cases would be better, but currently there is none.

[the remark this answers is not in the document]
nknapp: I can remove the headers, but I think there should still be some text around the tags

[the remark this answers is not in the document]
yyyc514: > I think the dots are a valid method to ensure the transition from tag to text. Handlebars is primarily text, with some tags inserted, so we should put some text sections between tags in the test cases. 

You might have to try harder to explain the purpose of this to me again, but I still not sure I will be persuaded.  It feels like (for some reason) you're wanting to a way to SEPARATE individual test cases within a file.  That isn't a problem unique to HTMLBars... and I already dislike this solution and I'd hate to see it find it's way into all our other test cases (plus one can imagine a grammar where ... has real meaning, then UGH.  

And also dislike the idea that every grammar will just invent their own choice way of separating test cases...

There may indeed be a real problem here, but I don't think this is the right solution.

---

It might not be completely terrible if it was already there, but when you try to change the test AND add the ... at the same time the changes become impossible to review because the ... generates a ridiculous amount of noise in the diffs.

---

Lets start higher level.... can you explain clearly the problem you're trying to solve with "..." and why led you to add them in the first place?  I'm not sure I follow that part.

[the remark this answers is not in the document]
yyyc514: Currently in other cases we accomplish this with blank lines, or perhaps a comment (which then also gets highlighted, which adds noise)... I'd be open to coming up with a smarter encoding format for these test cases, but it needs to be clear, simple and something we can do globally in a single pass, not something that interferes with PRs and get added piecemeal.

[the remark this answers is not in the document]
yyyc514: >  Handlebars is primarily text, with some tags inserted, so we should put some text sections between tags in the test cases.

I think I may disagree with your premise here, at least in the general case.  I think of markup tests (at least THIS many and this specific) as unit tests.  They should be as specific as possible and test as narrowly as possible.  

Testing that the "overall" grammar works might be a markup test, but that could be a SINGLE test showing that tags and text mix well, it doesn't need to be proven over and over in a every single unit test.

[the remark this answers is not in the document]
yyyc514: Take another grammar for example that formats numbers inside of text... with 100 varieties of numbers... A good test might be:

```
# numbers mixed with content
This is a line with 32 numbers.
How do pigs fly if they weight 1,000 pounds?
# 100 different tests for individual number formats...
12323
123,320
123,233.232
0xFFFF
...
```
Adding  words to every line is noise and gets in the way of what the tests are trying to prove (that the number recognition regex works).

nknapp:
@yyyc514 
I have refactored the commits and some tests and rebased everything onto the master. The commits are now structured as follows:

1. Implementation of changes, adjustet test-cases expectations to reflect the new rules. This should allow you to see the changes that the new code does to the old tests
2. Add more tests for the new rules.
3. Change of the behaviour regarding `{{else}}` that you requested above.

I have not added any text parts to the new tests, except in the test "combinations-with-text.txt".

I have left the text parts in the old test-cases, because I didn't want to change the tests (make it easier for you to compare the results). I can clean-up those tests as well, if you want me to.

nknapp:
Somehow, GitHub displays the commits in the wrong order. The correct order (in my git repo) is:

c6fbcec6 fixup: make only "else" in {{else}} the keywords, not "{{else}}" as whole
9468f00e wip: adjustments for integrating handlebars rules
30a5e541 fixup: add more tests

yyyc514:
Sorry I've seen this just it hasn't bubbled to the top of my list yet with everything going on. Sorry.  I'll try to get to soon.

nknapp:
BTW. I have integrated the rules into the handlebars website. All code snippets 
are using them. Except in the editor of interactive examples, which uses codemiror

yyyc514:
Looks pretty good, about to take a more careful look.  We can't lose the htmlbars.js file though for compatibility issues, but I'll fix that myself.

src/languages/handlebars.js
```diff
@@ -8,37 +8,163 @@ Category: template
...
 
-  var EXPRESSION_OR_HELPER_CALL = hljs.inherit(IDENTIFIER_PLAIN_OR_QUOTED, {
-    keywords: BUILT_INS,
+  // as defined in https://handlebarsjs.com/guide/expressions.html#literal-segments
+  // this regex matches literal segments like ' abc ' or [ abc ] as well as helpers and paths
+  // like a/b, ./abc/cde, and abc.bcd
+  const IDENFIFIER_REGEX = /(".*?"|'.*?'|\[.*?\]|[^\s!"#%&'()*+,.\/;<=>@\[\\\]^`{|}~]+|\.|\/)+/;
+
+  // identifier followed by a equal-sign (without the equal sign)
+  const HASH_PARAM_REGEX = /(".*?"|'.*?'|\[.*?\]|[^\s!"#%&'()*+,.\/;<=>@\[\\\]^`{|}~]+)(?==)/;
```
yyyc514, line 60: A little duplication here, could we pull out the "IDENT" part and use it for both expressions?  Look at the new `regex.concat` helper if you want a nice abstraction.

src/languages/handlebars.js
```diff
@@ -8,37 +8,163 @@ Category: template
...
+    illegal: /\}\}/,
+    begin: /\(/,
+    end: /\)/
+    // the "contains" is added below when all necessary sub-modes are defined
+  };
+
+  const HASH = {
+    // fka "attribute-assignment", parameters of the form 'key=value'
+    className: 'attr',
+    illegal: /\}\}/,
```
yyyc514, line 81: Is a single `}` allowed (perhaps it is)?  If not then perhaps this should be just `\}`?

src/languages/handlebars.js
```diff
@@ -8,37 +8,163 @@ Category: template
...
+  const BLOCK_PARAMS = {
+    // parameters of the form '{{#with x as | y |}}...{{/with}}'
+    begin: /as\s+\|/,
+    keywords: { keyword: 'as' },
+    end: /\|/,
+    contains: [
+      {
+        // define sub-mode in order to prevent highlighting of block-parameter named "as"
+        begin: /\w+/,
+        keywords: ''
```
yyyc514, line 108: You don't need this, keywords don't inherit.

test/markup/handlebars/built-ins.expect.txt
```diff
@@ -6,7 +6,7 @@
 
 </span><span class="hljs-template-tag">{{#<span class="hljs-name"><span class="hljs-builtin-name">each</span></span> test}}</span><span class="xml">abc</span><span class="hljs-template-tag">{{/<span class="hljs-name"><span class="hljs-builtin-name">each</span></span>}}</span><span class="xml">
 
-</span><span class="hljs-template-variable">{{<span class="hljs-builtin-name">lookup</span> abc}}</span><span class="xml">
+</span><span class="hljs-template-variable">{{<span class="hljs-name"><span class="hljs-builtin-name">lookup</span></span> abc}}</span><span class="xml">
 
-</span><span class="hljs-template-variable">{{<span class="hljs-builtin-name">log</span> test}}</span><span class="xml">
+</span><span class="hljs-template-variable">{{<span class="hljs-name"><span class="hljs-builtin-name">log</span></span> test}}</span><span class="xml">
```
yyyc514, line 11: This is annoying but I'm not sure we should fix it, and perhaps it's technically accurate in any case?

yyyc514:
Actually I may have spoken too soon because of my last change before 10.0:

```
  var SHOULD_INHERIT_FROM_HANDLEBARS = hljs.requireLanguage('handlebars');
```

That would make a hard error if someone wasn't also loading handlebars, although I suppose someone could still be using the file in a manual `require` pipeline... so perhaps the `htmlbars.js` file is still required just to cover that edge case.

src/languages/handlebars.js
nknapp: After doing some tests, I'd say the `illegal`-statement does not make any sense, neither at this point nor in the sub-expressions-mode.

I have tried a few strings, but I couldn't find any where the `illegal`-regex really cause a match while parsing.

```handlebars
<p>{{helper }}=x}}</p>

and 

<p>{{helper a}}b=x}}</p>
```
are perfectly valid (although the `}}` is the end of the mustache here and everything after that is xml. It is parsed correctly with and without the `illegal`-regex.

In the sub-expressions only time the illegal `}}` really matches is in a string like 

```handlebars
before {{helper (}} 1 2)}} after
```

The effect is that the `1` is highlighted as `name ` and `2` as number. The same happens if I remove the `illegal`-regex.

I think it is best to remove both statements to avoid confusion.

src/languages/handlebars.js
nknapp: I actually found out the the `HASH_PARAM_REGEX` is wrong.

```
{{{helper [a}}a]=x}}}
```

is valid (`a}}a` is the key), but 

```
{{{helper 'a}}a'=x}}}
```
and 

```
{{{helper "a}}a"'=x}}}
```
are not. I don't know if we add this in the future.
I have tried a refactoring, but since the two regexes have more differences now, it has become a little more complex. Hopefully clearer, though.

src/languages/handlebars.js
nknapp: Added a commit to remove the `illegal`-regex statements

test/markup/handlebars/built-ins.expect.txt
nknapp: I am not sure how to interpret the definition in the [docs](https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html)

>  name | name of an XML tag, the first word in an s-expression
>  builtin-name | s-expression name from the language standard library

When I remove the `name` and try it in the `developer.html`, the tag name (also the `builtin-name` is not bold anymore. Of course, this depends on the theme.

But I am also not sure how to fix it. It is cause by the combination of the `name`-mode for mustache expressions and the `keywords`.

And it happens for block-helpers in the `highlight.js/master` as well:

```handlebars
{{#if test}}true{{/if}}
```

src/languages/handlebars.js
nknapp: I have removed it.

nknapp:
@yyyc514 I have added a few commits that hopefully address all of your points: Some commits, I have also noticed and fixed the following things:

* Typo in variable IDENFIFIER_REGEX (commit 74f0c87e6db62e96d564f21d04f5b122f4ffe29f was somehow attributed to you, not me. I made an interactive rebase and moved that commit to the beginning and something went wrong. I don't mind if you don't).
* The jsdoc-comment in `regex.concat` was not accurate. IntelliJ gave me a warning, a varargs-
  parameter is not an array in jsdoc.

I think it would generally be better to use 

```javascript
import {concat} from '../lib/regex'
```
then 
```javascript
import * as regex from '../lib/regex' 
```
because then, rollup can optimize using tree-shaking. I haven't done it because the wildcard import was used in other places. This might be an additional PR, for other languages as well.

src/languages/handlebars.js
yyyc514: Illegal is intended to say "this character should NEVER occur in the position in this type of grammar"... and if it is found then it will stop highlighting anything further (or consider the language a non-match for auto-detection purposes). (unless someone uses the ignoreIllegal flag of course)

_Yes, if your illegal is the same as your `end` expression that it will serve no purpose at all since end gets the match first._ Is that what was happening here?

test/markup/handlebars/built-ins.expect.txt
yyyc514: I think in a perfect world we'd see only one or the other, not both, but this isn't a big deal. You could even argue this is technically correct.

> When I remove the name and try it in the developer.html, the tag name (also the builtin-name is not bold anymore. Of course, this depends on the theme.

Ah, that makes a slight argument in favor of "it should be both because both styles would apply in cases where they can".

I will make this resolved.

yyyc514:
Thanks for the type fix. :-)  Your right.

---

I prefer seeing the `regex` namespace in the code. Though I suppose you could still important them one at a time then put them back in a the regex namespace, but ugh.

> because then, rollup can optimize using tree-shaking. I haven't done it because the wildcard import was used in other places. 

This is a point, but one I'm not 100% sure I'm in a hurry to address, though welcome to discuss it here further since you noticed it and brought it up. :-)

My thoughts:

- `regex` may eventually get hung off of `hljs` like everything else... I'm wait to see what "feels right"... but at that point it's just an API of the core library and can't be tree-shaken because we have no idea who might need it.
- Eventually we will build the "core" build (ie, common languages) as one huge module (rather than individual languages as it's built now) and in that case there will be effectively no tree-shaking because the code wouldn't existing in the codebase if it wasn't used SOMEWHERE. And when we build as one huge module the code will also only be included a single time, rather than duplicated.

At that point the only thing tree-shaking would do for us is make the CDN modules a bit "tighter"... and even that wouldn't be true if `regex` moved back into being hung off of the main library.  

Does that make sense?

src/languages/handlebars.js
yyyc514: I liked your refactoring, a bit more than I had in mind, but it's clean and clear.

nknapp:
Moving it into core makes sense.

I don't if making a huge module makes sense if people use their own bundler. If you just need 3 languages (like in the handlebars website) you probably don't want to have all other rules in the code as well. So it would be better to import distinct rules and let webpack do the, well, tree-shaking. From that point of view it also makes sense not to use the wildcard import, but

src/languages/handlebars.js
nknapp: ## In HASH

I am not sure: `HASH` does not have an `end`-regex, but only a `starts`. As far as I can see, the mode itself is only active for the token matching the `begin`-regex. And that can never be `}}`. Then it switches to different modes. 

## In SUB_EXPRESSION 

I made some tests  in the `developer.html`, with added `illegal: /}}/` in SUB_EXPRESSION.

When I set the language to `handlebars` and use 

```handlebars
some text {{log ( }} ) }} some text 
```

as input, every thing is still colored, even the part `) }}`. An added `console.log` shows me that the illegal-token was found. I would have expected everything to be black...

When I set `language` to `auto`, the language is detected as `AppleScript`. :2nd_place_medal: 

The question here is: Would you expect every snippet that causes a parse error in the Handlebars-parser, to match an `illegal`-regex somewhere? 

That would be a lot of work and we would mean rebuilding the Handlebars-compiler with highlight.js techniques. (It already feels like I have done so for a large part).

I think it is OK if some illegal constructs are still highlighted as good as possible. Maybe someone even wants to highlight a broken template...

I would like to leave it like this.

nknapp:
> Actually I may have spoken too soon because of my last change before 10.0:
> 
> ```
>   var SHOULD_INHERIT_FROM_HANDLEBARS = hljs.requireLanguage('handlebars');
> ```
> 
> That would make a hard error if someone wasn't also loading handlebars, although I suppose someone could still be using the file in a manual `require` pipeline... so perhaps the `htmlbars.js` file is still required just to cover that edge case.

Is there something to do for me in this respect?

yyyc514:
> Is there something to do for me in this respect?

Nope. I think the commit I added earlier adding the file back and making it point to handlebars resolves it until v11, when the file can be completely removed.

yyyc514:
> So it would be better to import distinct rules and let webpack do the, well, tree-shaking.

I’m not sure I follow.  Do you think the code should hung off of `hljs` or be imported?  If it hangs off of hljs it cannot be tree-shaken because it’s a public API and the dependencies are not known at compile-time... so all the code has to be included (not that it’s a huge amount of code).

So there are a few different decisions to make here.  Ex: If we hang all the until functions off of `hljs` then how we import them becomes irrelevant.

yyyc514:
If a size critical build was really an issue there are much better places to look for savings than whether all the functions in `regex.js` are tree-shaken or not.

nknapp:
You are right in that there is not much to gain and I wonder if this discussion is really worth it.

I understand your arguments and I don't want to tell you how to build `highlight.js`. I just wanted to point out that people may be using ESM-import directly. 

Since the core actually does an `import * as regex from './lib/regex'` the tree-shaking would not change much in this particular case. But there may be other cases in the future where it becomes relevant, who knows.

But I don't want to refactor higlightjs. I think you have done a great job moving to ESM in the first play. And to be honest, I am happy, when this PR is done and out of my mind :heavy_check_mark:

yyyc514:
> And to be honest, I am happy, when this PR is done and out of my mind ✔️

Sorry it got so complicated and I took so long to come back to it. :-)  It's awesome to finally get this finished.

yyyc514:
Would you mind writing up a changelog entry (or a  few)?  I was going to just do it myself but then wasn't sure I could do it all justice... since my big-picture take is that we finally killed htmlbars but I think this makes a lot of other improvements that perhaps you'd like to mention?

yyyc514:
@nknapp Take a deep breather and have a cold beverage today. :-) Thanks again!

nknapp:
I pushed an entry that contains that things that are important to me. But I haven't mentioned htmlbars, because I'm not sure what to write about it exactly.

Should I rebase and squash the commits into semantically relevant ones? 

Oh, you merged already :grin:

yyyc514:
>  But I haven't mentioned htmlbars 

I added an entry for the deprecation.

yyyc514:
Now we need someone as tenacious as you to tackle all the CSS stuff. :-)

PR #2345, opened 2020-01-06
feat(developer): add button to show parsed structure

nknapp:
This is a small extension to the developer.html

* It adds a button "Show structure" that adds a class to the "editor"-div
* Once this happens, additional styles come into effect, that render highlighted spans as blocks with borders and a small label containing the class name.

While this is not perfect (I would rather like to see all modes) this was very helpful when I made that latest changes to the Handlebars language and I would like to see this in the master branch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,11 +20,15 @@ Language Improvements:
 - (vbnet) add nameof operator to the keywords (#2329) [Youssef Victor][]
 - (stan) updated with improved coverage of language keywords and patterns. (#1829) [Jeffrey Arnold][]
 
+Developer Tools:
+
+- feat(developer): add button to show parsed structure (#2345) [Nils Knappmeier][]
+
 [Jeffrey Arnold]: https://github.com/jrnold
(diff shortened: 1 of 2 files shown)


yyyc514:
Snap of this in action? :)

nknapp:
Peek 2020-01-06 23-45

nknapp:
it's more helpful, if you define more classes on youre modes:

Peek 2020-01-06 23-57

yyyc514:
Did you try and inline version vs block?

nknapp:
No, I actually wanted to have a tree view. But it might look better inline, if that doesn't run into layouting problems 

I think the optimal solution for developing languages is still another one. Something displaying the test cases in a live reload server. I almost started building that, before I saw developer.html.

I didn't want to spend too much time on tooling, because the actual goal was to write the Handlebars definition.

yyyc514:
Add a changelog entry for this improvement?

nknapp:
Will do.

PR #2347, opened 2020-01-09
(clojure) Add support for global definitions name (class: title)

agrison:
Adds support for Clojure global definitions name with class name `title`, including:

* `def`
* `defonce`
* `defprotocol`
* `defstruct`
* `defmulti`
* `defmethod`
* `defn`
* `defn-`
* `defmacro`
* `deftype`
* `defrecord`

### DIFF
--- a/test/markup/armasm/default.expect.txt
+++ b/test/markup/armasm/default.expect.txt
@@ -1,24 +1,13 @@
-<span class="hljs-comment">//=================== YOUR CODE GOES IN THE SECTION BELOW ===================//</span>
+<span class="hljs-comment">//=================== YOUR CODE GOES HERE ===================//</span>
 
-    <span class="hljs-comment"># check for and handle zero</span>
-    <span class="hljs-keyword">orr</span> <span class="hljs-built_in">r1</span>, <span class="hljs-built_in">r0</span>, <span class="hljs-number">#128</span>            <span class="hljs-comment">// r1 = r0 | 128</span>
-    <span class="hljs-keyword">cmp</span> <span class="hljs-built_in">r1</span>, <span class="hljs-number">#128</span>                <span class="hljs-comment">// if r1 == 128 {</span>
-    <span class="hljs-keyword">bne</span> notZero                 <span class="hljs-comment">// .</span>
-    <span class="hljs-keyword">bl</span> zeroQuidFP2float         <span class="hljs-comment">//          zeroQuidFP2float() }</span>
-    <span class="hljs-keyword">b</span> <span class="hljs-meta">end</span>                       <span class="hljs-comment">// else {</span>
+<span class="hljs-comment"># logical operator demo (highlight.js arm comment test)</span>
+<span class="hljs-keyword">mov</span> <span class="hljs-built_in">r0</span>, <span class="hljs-number">#0x00F0</span>     <span class="hljs-comment">// r0 = 0x00F0</span>
+<span class="hljs-keyword">mov</span> <span class="hljs-built_in">r1</span>, <span class="hljs-number">#0x000F</span>     <span class="hljs-comment">// r1 = 0x000F</span>
+<span class="hljs-keyword">mov</span> <span class="hljs-built_in">r2</span>, <span class="hljs-number">#0xFF00</span>     <span class="hljs-comment">// r2 = 0xFF00</span>
 
-<span class="hljs-symbol">notZero:</span>                        <span class="hljs-comment">// .</span>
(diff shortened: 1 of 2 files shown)


src/languages/clojure.js
```diff
@@ -2,43 +2,45 @@
...
 Description: Clojure syntax (based on lisp.js)
 Author: mfornos
-Contributors: Martin Clausen <martin.clausene@gmail.com>
+Contributors: Martin Clausen <martin.clausene@gmail.com>; Alexandre Grison <a.grison@gmail.com>
 Website: https://clojure.org
 Category: lisp
 */
 
 function(hljs) {
+  var globals = 'def defonce defprotocol defstruct defmulti defmethod defn defn- defmacro deftype defrecord';
```
yyyc514: Missing a space here.

If these so different might one of the other keyword classes be more appropriate also?

yyyc514:
Wouldn't this patch remove the need for NAME and BODY?

yyyc514:
Please also add a changelog entry.

yyyc514:
Ping.

agrison:
Thanks for the feed-back it's my first time on highlightjs so didn't knew all the specifics I'll update with changes whenever I do have some free time :)

yyyc514:
Ping.

yyyc514:
Ping. Still out there?

yyyc514:
Ping.

agrison:
Hello.

Yes still out there, sorry for the delay, I've been busy and couldn't work on this.

Sadly, yesterday I tried but I don't know enough about the inners of the library to make it work like you need to, I tried moving some definitions, modify the lexeme but to no avail, it always break something else, and I don't have much time to investigate for this.

yyyc514:
Ok, Clojure actually is kind of a mess... and you might also be hitting an actual issue/bug with HLJS in that beginKeywords wants keywords to end on a word boundary but the "space" between `-` and ` ` (space) is NOT considered a word boundary evidentally... so it fails to match on: `defn-`... but that's only part of the problem...

The grammar is making a mess of things by trying to count almost everything as a NAME or a SYMBOL... which I'm pretty sure is wrong, and definitely now consistent with how we do other grammars...

But if you want to help, lets try and start on fixing some smaller things I see first... then maybe after we clean it up a bit we can circle back to this.

First question:

```
(def +x [(a 1) +2 -3.0 y-5])
```

What is `-5` there?  The grammar right now identifies `y-5` as a symbol because...

```
var SYMBOLSTART = 'a-zA-Z_\\-!.?+*=<>&#\'';
var SYMBOL_RE = '[' + SYMBOLSTART + '][' + SYMBOLSTART + '0-9/;:]*';
  
var SYMBOL = {
    begin: SYMBOL_RE,
    relevance: 0
  };
```

Is `y-5` truly an identifier/variable all on it's own or is that actually `y - 5` (y subtract 5)?

yyyc514:
I might be getting the naming confused... we also have `keys`, which get the class `symbol`... but that's not what I'm talking about here.  I'm talking about the literal rule that I show the source for above.

agrison:
Well Clojure is not really a mess, the syntax is simple and easy to pick-up, of course it can be a little disturbing the first times.

In your example `y-5` is a valid symbol identifier in Clojure, if you wanted y minus 5 it would have been written `(- y 5)`

Yes a symbol can start with a minus or a plus character.

```clojure
user=> (def y-5 7)
#'user/y-5
user=> (+ y-5 2)
9
user=> (def -foo "hello")
#'user/-foo
user=> (str -foo " world")
"hello world"
user=> (def -+* (+10 11))
#'user/-+*
user=> (def *-+ 21)
#'user/*-+
user=> (+ -+* *-+)
42
```

You're probably right, maybe the highlightjs parser is having some limitations that needs to be fixed prior to scan correctly clojure code. But it could also be that I don't know how to use it correctly to find the symbol boundaries effectively.

yyyc514:
Whoops, I accidentally pushed your master back to upstream/master instead of pushing my branch to yours - I forget the syntax cause I rarely do it - and now I can't push it because the PR is closed and the magic permissions go away.

I will re-open a new PR.

yyyc514:
https://github.com/highlightjs/highlight.js/pull/2419

yyyc514:
@agrison LOL.  No I didn't mean Clojure was a mess, I meant the grammar was... though it makes a little more sense with our explanation.  Anyways I went back over this and tried to just do the minimal amount to make this work for now.

Basically that meant making sure title only fires a single time and also removing some pieces of your rule that don't seem to have any real effect.

PR #2349, opened 2020-01-10
Added php7 to php classes list in readme

dmyers:


yyyc514:
Does anyone use all these variants anyways?  I've always wondered.  I think the idea was for it to be the file extension, but do people use `php*` as the file extension in real life anyways?

dmyers:
Sorry I forgot to reply on this but basically you are right I think, I haven't seen usage of the `.php$version` usage since php3 days, but I did wonder if it could be useful for some code syntax differences. I tried highlight.js with some newer PHP 7 syntax changes like newer typehints and they weren't highlighted yet.

yyyc514:
I can't see us maintaining separate grammars for different versions, so we'd always track the latest... if you see issues with parsing you could file an issue with specific examples.  I went ahead and merged this since we already have php7 as an alias anyways so we're down that rabbit hole whether we should be or not. :-) 

So at least now our docs are consistent.

PR #2360, opened 2020-01-22
Added `txt` as Alias

taufik-nurrohman:


yyyc514:
Can I ask the motivation before this or did it just seem missing?  

The problem is "plaintext" is more of a special/system level language... I've even though about moving it out of `languages` and simply into the source - so that people can't choose to not include it.  It's already used internally but a COPY of it is used rather than this version... not that that would make much difference for aliases, but it might for other things.

I'm not sure we want to start aliasing it like other things.  I need to give it some though.  "txt" naturally leads to "text" but "text" internally has a VERY different meaning to HLJS already - in that it will skip the block completely.  Although that's something I'd like to change in the future, perhaps with version 10.

taufik-nurrohman:
As `javascript` has an alias of `js` and `yaml` has `yml`, `markdown` has `md`, so `text` or `plaintext` would also has `txt`, logically.

Human Readable Name | File Extension
--- | ---
JavaScript | JS
YAML | YML, YAML
Markdown | MD, MKD
Text | TXT

I would also suggest to add common file extensions to every language into aliases as it will be easier to developers who want to create language detection without depending on the built-in language detection based on the file extension. For example, `apache` should have an alias `htaccess` too.

``` .js
let path = './path/to/file.js',
    container = document.querySelector('pre > code');

fetch(path).then(response => {
    let source = response.text(),
        x = path.split('.').pop(), // Take file extension as the language name
        out = hljs.highlight(x, source);
    container.className += ' hljs ' + out.language;
    container.innerHTML = out.value;
});
```

yyyc514:
> I would also suggest to add common file extensions to every language into aliases as it will be easier for developers who want to create language detection without depending on the built-in language detection based on the file extension. For example, apache should have an alias htaccess too.

Can you open an issue for this item/request?  I don't necessarily disagree but I wonder if we should add an "extensions" key to language definitions so we can be smarter about this in the future and split out actual extensions of a language from nicknames, etc.

yyyc514:
Could you please also add "text" as an alias and add an entry to the changelog and then I can approve this.

taufik-nurrohman:
I might also add a few new keywords for the PHP language in the future. Recent PHP version just provided some new keywords like `fn` , `iterable`, `void` and `scalar`.

yyyc514:
>  Recent PHP version just provided some new keywords like fn , iterable, void and scalar.

Will look forward to a PR.

Issue #2371, opened 2020-01-28
File Extension as Aliases


taufik-nurrohman:
I would suggest to add common file extensions to every language into aliases as it will be easier for developers who want to create language detection without depending on the built-in language detection based on the file extension. For example, `apache` should have an alias `htaccess` too.

Human Readable Name | File Extension
--- | ---
Apache | HTACCESS
JavaScript | JS
Markdown | MD, MKD
Text | TXT
YAML | YML, YAML

``` .js
let path = './path/to/file.js',
    container = document.querySelector('pre > code');

fetch(path).then(response => {
    let source = response.text(),
        x = path.split('.').pop(), // Take file extension as the language name
        out = hljs.highlight(x, source);
    container.className += ' hljs ' + out.language;
    container.innerHTML = out.value;
});
```

Imagine someone makes a [git repository viewer application](http://mecha-cms.com/git/tree/mecha-cms/x.highlight/highlight/engine/i/@scrivo/highlight.php/Highlight/Highlighter.php) and then uses _highlight.js_ to color the code syntax in their files. This will be more robust to automatically detecting language through file extensions than by reading the file contents over and over using the available language packages until it finds the most relevant match.

This also opens up various possibilities to load language packages asynchronously based on file extensions, so the amount of data transferred will be much smaller considering that JavaScript works on the client side.

yyyc514:
> This will be more robust to automatically detecting language through file extensions than by reading the file contents over and over using the available language packages until it finds the most relevant match.

Of course you wouldn't actually ever do it over and over.  If performance mattered at all you'd only do the auto-detect a single time and then cache that result for future use.  That said if you can trust the extensions, that could be much faster.

Of course historically we don't care about extensions since we aren't dealing with them, we're only dealing with source.

---

My suggestion here is that we make an actual `extensions` key so that extensions can be registered differently than aliases.  This would also help the stated use case here... as `file.javascript` would technically be a FALSE positive as `javascript` is not a valid extension for Javascript.  And god forbid the confusion that ensures when one languages extension happens to be the name (or common alias) of another language - though hopefully that's only a hypothetical problem.

So something like:

```
// plaintext.js
{
  aliases: ['text'],
  extensions: ['txt'],
  ...
}
```

yyyc514:
@isagalaev Do you have any historical context on what the actual original intention was here - that might guide us or add to the discussion?

[...]

Issue #803, opened 2015-04-05
Exclude parentheses from function parameter highlights


mariusschulz:
Take the following piece of JavaScript code as an example:

```js
var add = function(a, b) {
    return a + b;
};
```

Here's how the above snippet is highlighted:

```html
<code class=" hljs javascript"><span class="hljs-keyword">var</span> add = <span class="hljs-function"><span class="hljs-keyword">function</span><span class="hljs-params">(a, b)</span> </span>{
    <span class="hljs-keyword">return</span> a + b;
};
</code>
```

Notice how parameters are marked up:

```html
<span class="hljs-params">(a, b)</span>
```

Wouldn't it be nicer if the parentheses weren't included in the `<span>`? All text editors I use (Sublime Text, Brackets, Visual Studio, …) treat the highlighting this way, so it seems weird to me to see the parentheses colored as well (but not the curly braces) when I style `.hljs-params` with a different color.

isagalaev:
I personally wouldn't care either way but by your reasoning it does seem like a good idea. Would you be up to making a pull request?

mariusschulz:
Sure, I'll give it a spin.

mariusschulz:
Here's the commit which does the change for JavaScript:

https://github.com/mariusschulz/highlight.js/commit/10971e2b1096bf0eafd2a2ba4a5dbc233d1b9e1b

How do you want to deal with other languages, shouldn't they all behave similarly? In that case, would you like me to update them all, at least those that define params within parentheses?

isagalaev:
Oh, if you could go to the trouble of updating the other languages that would be awesome! Thanks!

mariusschulz:
For now, I've updated the following languages:

- JavaScript
- TypeScript
- C#

I've also added a simple markup test for TypeScript, which didn't exist before.

To avoid creating empty `<span class="hljs-params">` tags for parameterless functions, I've added a negative lookahead to the regular expression detecting parameters. The pattern only matches if the opening parenthesis isn't immediately followed by a closing one (with optional whitespace in between):

```js
{
  begin: /\((?!\s*\))/
}
```

I also wanted to update the rules for Swift, however that didn't work the same way as the other languages. I'm new to your codebase, so I'm not sure how to solve that. I'll leave it at this for now.

isagalaev:
Fixed in 18500454.

PR #2372, opened 2020-01-31
enh(php) Add additional keywords, built-in classes, and '<?=' syntax

taufik-nurrohman:
+ ~~Add `htaccess` as alias for `apache` ([#issuecomment-579757989](https://github.com/highlightjs/highlight.js/issues/2371#issuecomment-579757989))~~ ([#issuecomment-580984827](https://github.com/highlightjs/highlight.js/pull/2372#issuecomment-580984827))
 + ~~Highlight JSON and YAML attributes.~~

~~**PS:** For maintainers, you may need to unify `attr` class name to `attribute` or simply include `.hljs-attr` along with `.hljs-attribute` class declaration.~~

~~<https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html>~~ ([#issuecomment-580729326](https://github.com/highlightjs/highlight.js/pull/2372#issuecomment-580729326))


yyyc514:
Please don't mix a bunch of items in a single PR, that makes it much harder to review and will slow things down.

yyyc514:
I mean grouping the PHP things is ok, but the other stuff should preferably be split out...

yyyc514:
> PS: For maintainers, you may need to unify attr class name to attribute or simply include .hljs-attr along with .hljs-attribute class declaration.

These are very different things on purpose.  You can search issues in the past to find the write up, sorry I don't have it off the top of my head.  So you'd have to review other themes to see if attr aliasing to attribute is something we would accept. **Some themes purposely only highlight one and choose to ignore the other.**

yyyc514:
I will approve and flat-merge if you simply pull out 622f5cb.  We can discuss that elsewhere.

yyyc514:
OK but now you're going to have to look at the PHP/zephir collision.  It may be that you need to apply your PHP updates to Zephir as well.

Sorry just now noticed the failing test.

yyyc514:
Probably wouldn't hurt to add a comment to the PHP and Zephier files to tell maintainers to cross-reference them when making updates.

I also wouldn't mind if Zephir pulled in PHP as a requirement and built upon it (which would avoid these issues in the future) but that might be a bit more work than you were wanting to put it.  Let me know. :)

yyyc514:
```diff
--- src/languages/php.js	2020-01-31 08:28:57.000000000 -0500
+++ src/languages/zephir.js	2020-01-31 08:28:54.000000000 -0500
@@ -1,21 +1,14 @@
 /*
-Language: PHP
-Author: Victor Karamzin <Victor.Karamzin@enterra-inc.com>
-Contributors: Evgeny Stepanischev <imbolk@gmail.com>, Ivan Sagalaev <maniac@softwaremaniacs.org>
-Website: https://www.php.net
-Category: common
-*/
+ Language: Zephir
+ Description: Zephir, an open source, high-level language designed to ease the creation and maintainability of extensions for PHP with a focus on type and memory safety.
+ Author: Oleg Efimov <efimovov@gmail.com>
+ Website: https://zephir-lang.com/en
+ */
 
 function(hljs) {
-  var VARIABLE = {
-    begin: '\\$+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*'
-  };
-  var PREPROCESSOR = {
-    className: 'meta', begin: /<\?(php)?|\?>/
-  };
   var STRING = {
     className: 'string',
-    contains: [hljs.BACKSLASH_ESCAPE, PREPROCESSOR],
+    contains: [hljs.BACKSLASH_ESCAPE],
     variants: [
       {
         begin: 'b"', end: '"'
@@ -29,20 +22,21 @@
   };
   var NUMBER = {variants: [hljs.BINARY_NUMBER_MODE, hljs.C_NUMBER_MODE]};
   return {
-    aliases: ['php', 'php3', 'php4', 'php5', 'php6', 'php7'],
+    aliases: ['zep'],
     case_insensitive: true,
     keywords:
       'and include_once list abstract global private echo interface as static endswitch ' +
-      'array null if endwhile or const for endforeach self var while isset public ' +
+      'array null if endwhile or const for endforeach self var let while isset public ' +
       'protected exit foreach throw elseif include __FILE__ empty require_once do xor ' +
       'return parent clone use __CLASS__ __LINE__ else break print eval new ' +
       'catch __METHOD__ case exception default die require __FUNCTION__ ' +
       'enddeclare final try switch continue endfor endif declare unset true false ' +
       'trait goto instanceof insteadof __DIR__ __NAMESPACE__ ' +
-      'yield finally',
+      'yield finally int uint long ulong char uchar double float bool boolean string' +
+      'likely unlikely',
     contains: [
+      hljs.C_LINE_COMMENT_MODE,
       hljs.HASH_COMMENT_MODE,
-      hljs.COMMENT('//', '$', {contains: [PREPROCESSOR]}),
       hljs.COMMENT(
         '/\\*',
         '\\*/',
@@ -66,23 +60,9 @@
       ),
       {
         className: 'string',
-        begin: /<<<['"]?\w+['"]?$/, end: /^\w+;?$/,
-        contains: [
-          hljs.BACKSLASH_ESCAPE,
-          {
-            className: 'subst',
-            variants: [
-              {begin: /\$\w+/},
-              {begin: /\{\$/, end: /\}/}
-            ]
-          }
-        ]
-      },
-      PREPROCESSOR,
-      {
-        className: 'keyword', begin: /\$this\b/
+        begin: '<<<[\'"]?\\w+[\'"]?$', end: '^\\w+;',
+        contains: [hljs.BACKSLASH_ESCAPE]
       },
-      VARIABLE,
       {
         // swallow composed identifiers to avoid parsing them as keywords
         begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/
@@ -98,7 +78,6 @@
             begin: '\\(', end: '\\)',
             contains: [
               'self',
-              VARIABLE,
               hljs.C_BLOCK_COMMENT_MODE,
               STRING,
               NUMBER

```

Ok I never looked closely before.  I didn't realize they were so similar.  I can probably consolidate them myself, so I wonder if there is a simpler way just to get your tests passing so we could get the PHP portions of this merged?

taufik-nurrohman:
Hmm. I&rsquo;m confused. How can I revert specific item from a pull that has been requested? Can I just undo modifying the `agate.css` file with a new commit?

yyyc514:
The preferable way would be to rebase onto master and just snip out the bad commit... if you want to try:

```
# make sure master branch is up-to-date
git checkout [your branch]
git rebase -i master
# change the commit in question to removal
# save changes
# your commit history should be rewritten
# force push changes back to your branch/pull request
git push -f
```

yyyc514:
>  Can I just undo modifying the agate.css file with a new commit?

That would typically have been ok since we usually squash the commit history (although you'd do it with `revert` not manually).  But since your commits are so nice it'd be better here if you edited the history so we can preserve the commit history.

taufik-nurrohman:
Thanks for the instructions. It&rsquo;s almost midnight here. I will try it tomorrow.

yyyc514:
Once you see the power hopefully you'll go off and learn a bit more about rebasing, it's SUPER useful if you use git a lot and contribute to projects.

taufik-nurrohman:
Done with `drop` instead of `removal` :+1:

Just realized that HTML attributes looks ugly (too striking) when highlighted. I don&rsquo;t know, maybe this is very subjective.

### Highlighted as `.hljs-attribute` Color

Screenshot from 2020-02-01 09 47 41

### Not Highlighted

Screenshot from 2020-02-01 09 49 41

yyyc514:
> Just realized that HTML attributes looks ugly (too striking) when highlighted.

Well, that all depends on your theme.  Some are awful, some are good IME.

yyyc514:
This isn't truly rebased onto upstream master though or there wouldn't be any conflicts... you first need to make sure YOUR local master points to upstream/master.  Where upstream = the highlightjs core repository.

I usually have upstream as a git `remote` so I can just `fetch --all` and then `merge upstream/master`.

Oh you really should be using branches for development not master. :-)

taufik-nurrohman:
Most of the contributions I have made are because I use the application, not because I am proficient with Git. I will usually delete the fork repository when my pull requests has been merged, and will re-fork it when I want to make new pull requests :stuck_out_tongue:

yyyc514:
>  I will usually delete the fork repository when my pull requests has been merged, and will re-fork it when I want to make new pull requests 😛

That works, but makes it hard to make multiple PRs against a single repo at once.  :-)

yyyc514:
It seems you added:

```
Set(10) {
  '__TRAIT__',
  'bool',
  'callable',
  'float',
  'from',
  'int',
  'iterable',
  'object',
  'string',
  'void'
}
```

And removed:

```
Set(3) { 'exception', 'interface', 'use' }

Please explain the removals.
```

taufik-nurrohman:
### About `int|integer`, `double|float|real`, `bool|boolean`

At present, the PHP highlighter has no special way to distinguish between keywords for [type casting](https://www.php.net/manual/en/language.types.type-juggling.php) and [type hinting](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration):

~~~ .php
function foo(bool $a): bool {
  return $a;
}

$b = (bool) foo(false);
~~~

The problem comes with the alias of each of these keywords when used in function arguments or variable values.

This works:

~~~ .php
$b = (bool) $b;
$b = (boolean) $b;
~~~

But this doesn&rsquo;t:

~~~ .php
function foo(boolean $a, integer $b) {}

foo(false, 1);
~~~

Most of them will consider those keywords as class names:

~~~ .php
function foo(Boolean $a, Integer $b) {}
~~~

With _&ldquo;php difference A and B&rdquo;_ search key pattern, you will get some hints:

 - https://stackoverflow.com/a/48927247/1163000
 - https://stackoverflow.com/a/44009101/1163000
 - https://stackoverflow.com/a/44301450/1163000
 - https://stackoverflow.com/q/7839059/1163000

So I chose to take the safest way (to take keywords that works in both cases).

**PS:** `object` was accepted as type hint since PHP 7.2 ([link](https://www.php.net/releases/7_2_0.php))

### About `interface`, `use`

It&rsquo;s already there?

 - https://github.com/highlightjs/highlight.js/blob/e96ef4648b895d51bdef6e77cdb54dc2d691bdaf/src/languages/php.js#L117
 - https://github.com/highlightjs/highlight.js/blob/e96ef4648b895d51bdef6e77cdb54dc2d691bdaf/src/languages/php.js#L130

### About `exception`

There is [no such keyword](https://www.php.net/manual/en/reserved.php).

### About `from`

It is from the `yield from $foo` syntax &rarr; https://www.php.net/manual/en/language.generators.syntax.php#control-structures.yield.from

I wonder why this not detected as keyword.

yyyc514:
**Thanks for the detailed response!**

> The problem comes with the alias of each of these keywords when used in function arguments or variable values. This works:  But this doesn’t:

That's a user knowledge problem.  **We still should highlight all variants** valid in any context for the users that know how and when to use them correctly.

> It is from the yield from $foo syntax (link)

Looks good.

>  It misses the to keyword though.

Should we add `to` as a keyword then also?

> About interface, use..  It’s already there?

Well, LOL.  You have a point there.  I flagged them as missing because I was comparing just the `keywords` list from before/after.  Typically we "double list" keywords like this, so they'd be listed at the top also - it can make a difference in some grammars.  I'm not sure it matters in this specific case though - so this might be fine as is.

Now I'm really wondering why you're getting that zephir test failure though, grrr.

> exception There is no such keyword.

```php
try {
    // run your code here
}
catch (exception $e) {
    //code to handle the exception
}
```

https://stackify.com/php-try-catch-php-exception-tutorial/

Found it with a quick google.  Perhaps it's old syntax.  **We need to add it back.**

yyyc514:
@taufik-nurrohman Do you know anything about Zephir?  One problem is our Zephir sample code actually HAD `fn` but it wasn't detecting it... so suddenly PHP started claiming it (and a whole bunch of other function rules get scored, etc)... Once I add that to Zephir also now they are sitting equal... (because Zephir just had a bit better support already than PHP, and now you've caught up).

So I'm trying to figure out what the actual difference is to make the Zephir file look more like Zephir than PHP.  

Any thoughts/ideas?

yyyc514:
Pull my changes please, then you can build on top.  I think I got the Zephir tests passing now. :-)

taufik-nurrohman:
> Found it with a quick google. Perhaps it's old syntax.

`exception` is a class ([link](https://www.php.net/manual/en/class.exception.php)). In PHP, function and class name are case insensitive.

~~~ .php
try {
    // ...
} catch (Exception $e) {
    // ...
}
~~~

~~~ .php
class MyException extends Exception {}

try {
    // ...
} catch (MyException $e) {
    // ...
}
~~~

> Should we add `to` as a keyword then also?

No, it never exists :laughing:

taufik-nurrohman:
> Do you know anything about Zephir?

Unfortunately, no.

> So I'm trying to figure out what the actual difference is to make the Zephir file look more like Zephir than PHP.

Do Zephir has their own token? Something like Hack that uses `<?hh` instead of `<?php`?

yyyc514:
> exception is a class (link). In PHP, function and class name are case insensitive.

Ah.  Wouldn't matter since the grammar itsself case-insensitive (now that I checked).  When you add it back you can add it as `Exception` if you think that's more "correct".  **We've highlighted it in the past, so we should keep highlighting it I think.**

Perhaps it should be a `built_in` though, if you're taking issue with the keyword designation?

I'll let you decide if you want to spend time adding the other build in class names. :)

yyyc514:
> Do Zephir has their own token? Something like Hack that uses <?hh instead of <?php?

They have `unlikely`/`likely` which I don't believe PHP has, so I used those in order to differentiate the test cases.

taufik-nurrohman:
> https://github.com/highlightjs/highlight.js/blob/e96ef4648b895d51bdef6e77cdb54dc2d691bdaf/src/languages/php.js#L130

I believe this was for the class and function import (as it has `end: ';'`). Removing `use` from the keyword list may break the highlight for variable import. Example here:

~~~ .php
// Class import
use Author\Package\ClassName1;
use Author\Package\ClassName2;

// Variable import
$foo = 1;
$baz = function($qux) use($foo) {
    return $foo + $qux;
};
~~~

Maybe this was the reasons why we have keywords copy :thinking:

Didn&rsquo;t test it yet. I&rsquo;m on mobile now. If it fails, it means we need to put `class`, `interface` and `use` to the keyword list.

yyyc514:
>  Removing use from the keyword list may break the highlight for variable import. Example here:  

Doing my job for me! Thanks.  

> Maybe this was the reasons why we have keywords copy

Exactly.  And it's why we try to be very careful about removing things - I'm not an expert on every language and little nuance things like this are easy to miss.  **So we need use in the global keywords list also.**

I'd probably add them all back just to be sure unless your positively class and interface have no other use contexts.

taufik-nurrohman:
> Perhaps it should be a `built_in` though.

Agree. But PHP has a lot of built in classes. This change may increase the file size. But it&rsquo;s okay if you don&rsquo;t mind.

https://www.php.net/manual/en/book.spl.php

yyyc514:
If that's the full list that's not so terrible... what is that?  50-80 items maybe?  I'm fine with that IF you think they'd be useful for highlighting.  If not let jut add `Exception` and be done.  Your call. :)

taufik-nurrohman:
Trying to remove _Sync_ and _Update_ log but could&rsquo;t find it when trying to rebase.

taufik-nurrohman:
Before | After
--- | ---
before | after

``` .php
function foo(bool $a): bool {
  return $a;
}

$b = (bool) foo(false);


$b = (bool) $b;
$b = (boolean) $b;


function foo(boolean $a, integer $b = 1, string $c = 'wow') {}

foo(false, 1);


function foo(Boolean $a, Integer $b) {}

// TODO: Mark built-in interface/class name declaration as `built_in`
class Foo extends Bar implements ArrayAccess {
    public function bar(Boolean $a, Integer $b, Traversable $c) {
        $a = self::class;
        $b = static::class;
        $c = parent::class;
    }
}

$foo = new Foo;

$foo->bar(int $a = 1);


// Class import
use Author\Package\ClassName1;
use Author\Package\ClassName2;

// Variable import
$foo = 1;
$baz = function($qux) use($foo) { // TODO: Mark `use` as keyword here
    return $foo + $qux;
};


try {
    // ...
} catch (Exception $e) {
    // ...
}


class MyException extends Exception {}

try {
    // ...
} catch (MyException $e) {
    // ...
}
```

yyyc514:
> Trying to remove Sync and Update log but could’t find it when trying to rebase.

Does't matter, we'll squash this now... but what you probably want is to mark them as "fixup" then they'll be silently merged into the prior commit.  In order to rebase properly you really want your master to point to a clean state.  If you're rebasing onto yourself (which I think you're doing) I'm not sure how that works...  I always branch, so I'm rebasing my branches on top of master.

yyyc514:
Now your'e conflicting with Java... we may need to improve our Java sample.

yyyc514:
This may have to wait now until after https://github.com/highlightjs/highlight.js/pull/2387 gets merged.

egor-rogov:
Oh, I see you guys have done a great job here and in #2387. What do you think about adding some tests? We have not a single one for Zephir yet...

yyyc514:
I don't tend to require tests for contributors that are simply adding keywords.  That's really a feature of HLJS, not something we need a test fore (when the keywords are just global, like here).

> We have not a single one for Zephir yet...

I was using the detect file for all my testing... shall I just copy it wholesale over into the markup folder with it's expected result?

egor-rogov:
That's right, my concern was more about #2387.

> I was using the detect file for all my testing... shall I just copy it wholesale over into the markup folder with it's expected result?

It is better than nothing, and will make it easier to add other tests afterwards.

yyyc514:
Ok, the only failing test now seems to be regarding how you've changed params... I actually raised my eyebrows when I saw how you changed it, but I need to look at all our other grammars and see how we're doing it there.  Either the `()` are part of params or not, but we should be consistent about it.

Did you already look around before making this changes?

taufik-nurrohman:
> Did you already look around before making this changes?

I [did](https://github.com/highlightjs/highlight.js/issues/803).

> Either the `()` are part of params or not, but we should be consistent about it.

I could make you become very very confusing in seconds without magic: with ES6 arrow functions :laughing:

~~~ .js
const foo = (bar, baz) => {};
const bar = baz => {};
~~~

yyyc514:
> I did.

Ok, awesome.  Now can you just fix the one failing test and then I think this might be ready to go!

yyyc514:
I think you only fixed half of it.  If you can run the tests yourself locally that'd make checking a lot easier.

yyyc514:
@taufik-nurrohman Thanks for putting in the time on this one!

PR #2374, opened 2020-01-31
enh(parser) simplify nohighlight

yyyc514:


yyyc514:
@egor Any ideas on betting naming for `languagePrefixRe`? And should we perhaps remove the need to include the word at the end?  (hence it truly being prefix) or should we leave it super flexible?

I decided to make both configurable while I was in there since it was easy and it's come up in conversation before.

yyyc514:
I hadn't considered uses like:

```html
<pre><code class="text-content javascript">var x = 'foo';</code></pre>
 <pre><code class="clipboard-text javascript">var x = 'foo';</code></pre>
```

But think it's still safe to consider them deprecated for the brutal consistency of "nohighlight". And if anyone is seriously affected they can just use the configuration option.

docs/api.rst
```diff
@@ -74,6 +74,8 @@ Configures global options:
 * ``useBR``: a flag to generate ``<br>`` tags instead of new-line characters in the output, useful when code is marked up using a non-``<pre>`` container.
 * ``classPrefix``: a string prefix added before class names in the generated markup, used for backwards compatibility with stylesheets.
 * ``languages``: an array of language names and aliases restricting auto detection to only these languages.
+* ``languagePrefixRe``: allows you to configure the css language prefix  (to allow class names like, `color-as-php`, etc.)
+* ``noHighlightRe``: allows you to configure when CSS classes are considered "not not highlight" flags
```
egor-rogov: "not **to** highlight"?

docs/api.rst
```diff
@@ -74,6 +74,8 @@ Configures global options:
 * ``useBR``: a flag to generate ``<br>`` tags instead of new-line characters in the output, useful when code is marked up using a non-``<pre>`` container.
 * ``classPrefix``: a string prefix added before class names in the generated markup, used for backwards compatibility with stylesheets.
 * ``languages``: an array of language names and aliases restricting auto detection to only these languages.
+* ``languagePrefixRe``: allows you to configure the css language prefix  (to allow class names like, `color-as-php`, etc.)
```
egor-rogov: What about `languageNameRe` to make it clear that it doesn't actually have to be a prefix?

yyyc514: I'm open to a better name. :-)  Should we just drop the prefix part?  Or just remove the body from the regex and truly make it a prefix?  I'm not sure.  Do you have a preference or thoughts?

yyyc514: `languageDetectRe`?

allejo: Hmm what's the purpose of this variable? I don't quite understand the use case of `color-as-php`, like when would you use that?

yyyc514: In class on the client. 

code class="lang-php" 

Vs 

Code class="color-as-php"

yyyc514: In class on the client. 

code class="lang-php" 

Vs 

Code class="color-as-php"

allejo: So this configuration option will allow highlight.js to look for custom prefixes? So instead of looking for `lang-*` then it'll look for `color-as-*`?

yyyc514: Or both. However it's configured. It's regex.

egor-rogov: @allejo: Just so. I don't think someone will ever change it, but who knows...
@yyyc514: I actually proposed `languageNameRe` above, but `languageDetectRe` is fine.

yyyc514: How did I miss that? LOL.

yyyc514: I dislike languageName because it's not really the name, it's prefix + name... so I think `name` mistates it a bit.  I'll change it to Detect if you're cool with that.

egor-rogov: Agreed.

PR #2376, opened 2020-01-31
bug(coffeescript) fix freezing bug due to badly behaved regex

### LINKED ISSUE #2375: (coffeescript) Freezing, poorly behaved regex to match regex.
Tested in NodeJS 13 and Chrome 79 as well as v8 master.

Input text is around 5k.

In the attached file is a reproduction of a complete freeze when trying to highlight with coffeescript, this also affects auto language detection.
(Complete freeze in this case means, I killed the process after 10 minutes)

From debugging this a bit, it hangs in the exec of the coffeescript regex 
`var match = matcherRe.exec(s);`

I feel like it shouldn't hang and since this hang happens in v8 native code, I am not entirely sure this 
is actually an issue with highlightjs, so feel free to close this if appropriate.

Is it expected behavior for a browser to hang in such a case or abort after some amount of time? I am not entirely sure, would appreciate some info on this.

I was thinking about filing a but report on the v8 bug tracker but decided to hold off until I get some feedback here first.

[test.txt](https://github.com/highlightjs/highlight.js/files/4140031/test.txt)

yyyc514:
Closes #2375


yyyc514:
@marcoscaceres Or you could approve also if you have time to review.  Only pinging you since I consider freeze issues pretty serious but I'd still like to have more than just my eyeballs on this.

yyyc514:
Actually I just realized I can both release a new stable version and hold this open for review at the same time, so that's what I'm going to do. :-)  The power of Git.

yyyc514:
This has been live in 9.18.1 for 5 days now with no reported issue.  Going to merge and close this, but if anyone still seems an issue with it later we can always review.

egor-rogov:
Looks sane. Sorry I'm late.

yyyc514:
Better late than never, so I've heard. :-p

PR #2377, opened 2020-01-31
enh(java) improve Java annotation support

### LINKED ISSUE #1324: (java) parameter annotations that take parameters
Java parameter annotations that take parameters prematurely end the parameter list as delimited by the `span[class="hljs-params"]`.

As a simple example, consider the following code:

```
public class Example {
  public static void foo(@SuppressWarnings("unused") int bar) { }
}
```

In this case, when highlighted by HLJS, the `hljs-params` span ends before `int bar` -- apparently misinterpreting the close-parenthesis in the `@SuppressWarnings` annotation as the end of the function parameters.

### LINKED ISSUE #1478: (java) @annotations don’t support numbers
```java
@RunWith(SpringRunner.class)
@SpringBootTest
@AutoConfigureMockMvc
@Transactional     
@Slf4j
public class BoardControllerTest {
...

In hightlight js 

Slf``4j``

4j is not hightlighting correctly.

I use tomorrow skin

### PR DESCRIPTION
**Pull request by yyyc514**
Closes #1478 
Closes #1324.

### CODE REVIEW

File: src/languages/java.js [hunk 1]
```diff
@@ -15,6 +15,17 @@ function(hljs) {
     'package default double public try this switch continue throws protected public private ' +
     'module requires exports do';
 
+  var ANNOTATION = {
+    className: 'meta',
+    begin: '@[A-Za-z0-9]+',
```
egor-rogov: Shouldn't it be @ + JAVA_IDENT_RE?..

yyyc514: I don't know Java well enough to say for sure.  If you're confident I'll change it, if now lets leave it be.

egor-rogov: Well, let see... https://docs.oracle.com/javase/specs/jls/se7/html/jls-18.html
```
Annotation:
    @ QualifiedIdentifier [ ( [AnnotationElement] ) ]
```
So it's just a regular Java identifier.

PR #2378, opened 2020-01-31
enh(csharp) generic modifiers, reference paths, `where` keyword

### LINKED ISSUE #1714: (csharp) "where" keyword is not recognized as constraint
Here's an example:
```cs
public class TesterA<R, S>  where R : class where S : IComparable
```
Rendered  result:
image

### LINKED ISSUE #1947: (csharp) Reference path in inheritance breaks highlighting
The following snippet breaks the highlighting:

`public class AgentTryout : System.Windows.Forms.Form { }`

without reference path, it works as expected:

`public class AgentTryout : Form { }`

### LINKED ISSUE #2239: (csharp) highlight fail: `in` and `out` as generic variable modifier
this is the jsfiddle:
```html
<pre>
<code class="C#">
interface IObserver&lt;in T&gt;
{
  void OnNext(T item);
  void OnCompleted();
  void OnError(Exception error);
}
interface IObservable&lt;out T&gt;
{
  IDisposable Subscribe(IObserver&lt;T&gt; observer);
}
</code>
</pre>
```
`in` and `out` should be keywords.

yyyc514:
Closes #2239
Closes #1947.
Closes #1714.

- enh(csharp) add generic modifiers (`in`, `out`)
- enh(csharp) Allow reference path in class inheritance lists 
- enh(csharp) Support `where` keyword as class constraint


src/languages/csharp.js
```diff
@@ -18,10 +18,13 @@ function(hljs) {
       'uint ulong unchecked unsafe ushort using virtual void volatile while ' +
       // Contextual keywords.
       'add alias ascending async await by descending dynamic equals from get global group into join ' +
-      'let nameof on orderby partial remove select set value var when where yield',
+      'let nameof on orderby partial remove select set value var when where yield ' +
+      // generic modifiers
+      'in out',
```
egor-rogov: Why do you need them here, outside of `GENERIC_MODIFIER`?

yyyc514: I usually try to add keywords in both places (I think we have precedent for this in other grammars).  I don't 100% know for sure they are purely contextual keywords.

egor-rogov: Yeah, looking into https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure I see that these are rightful keywords. But we already have them above, in "normal keywords" section (:

yyyc514: Removing.

yyyc514:
Ah, ha you are totally right there!  So removing the extra set!

PR #2380, opened 2020-02-01
enh(parser) better regular expression detection

### LINKED ISSUE #2050: (javascript) division and regular expressions can get confused and break highlighting
**Version:** 9.15.8

**repo**
1. Run highlight.js on the the following javascript code:

```
console.log(`Bad ${foo / 1000}`)
const highlighting = 'broken'
```

**Bug**
Screen Shot 2019-06-03 at 6 05 07 PM

The division break highlighting

yyyc514:
- fewer false positives
- Uses a look-ahead to make sure we have a WHOLE regex and not
  just a partial (which is more likely a math expression)

Closes #2050.


yyyc514:
This paired with the "value" container support (and smart rule ordering) in JS actually catches a lot of common cases:

```
`Bad ${foo / 1000}`
foo = 2 / 2 + 2 / 2;
foo = / 2 + 2 /;
b = a++ / 2;
```

PR #2381, opened 2020-02-01
Update issue templates

yyyc514:
Add some issue templates to see if they help.


yyyc514:
Tests glitched, but are irrelevant here since this is just data.

PR #2385, opened 2020-02-01
(chore) automatic greeting for users first contrib

yyyc514:
@egor-rogov Any thoughts on verbiage?


.github/workflows/greetings.yml
```diff
@@ -0,0 +1,26 @@
...
+  greeting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/first-interaction@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+#        issue-message: 'Message that will be displayed on users'' first issue'
+        pr-message: >
+          Thank you so much for your first contribution!  It's appreciated.
+          One of our maintainers shouild review it in the next few days (if not sooner).
```
egor-rogov: I wouldn't make such promises... Maybe something like "we will do our best to review it as soon as possible"?

yyyc514: Agree.

egor-rogov: Otherwise it's seems good to me.

PR #2389, opened 2020-02-02
enh(apache) improve apache grammar

yyyc514:
- add `deny` and `allow` keywords
- highlight numbers as attribute values
- highlight IP addresses, ports, and strings in sections


yyyc514:
> but I'd certainly like to see some tests...

That's fair, I might hold this and go back and try to add a few.

PR #2395, opened 2020-02-06
(parser) plugin callbacks for `highlight`

yyyc514:


src/highlight.js
```diff
@@ -515,10 +515,34 @@ https://highlightjs.org/
    * @property {string} language - the language name
    * @property {number} relevance - the relevance score
    * @property {string} value - the highlighted HTML code
+   * @property {string} originalCade - the original raw code
```
egor-rogov: It seems that `originalCode` may be misleading, because a before plugin can change it and it will no longer be "original". Why not just `code` (or `codeToHighlight`)?

yyyc514: > because a before plugin can change it and it will no longer be "original".

Look again.  A plugin can't change it.

yyyc514: LOL, n/m. hmmmm

egor-rogov: // the plugin can change the desired language or the code to be highlighted
     // just be changing the object it was passed
     fire("before:highlight", context);

Am I missing something?

yyyc514: No, that's why I said "n/m".  I looked again. You're right.  I think we can come up with better naming as you suggest.

yyyc514: Resolved naming.

yyyc514:
Added docs.

docs/plugin-api.rst
```diff
@@ -61,6 +61,46 @@ This approach is best for simpler plugins.
...
+
+language
+  The language grammar that should be used for highlighting.
+
+Your plugin may modify either value and those new values will be used as input
+to the highlighting engine.  If you add a ``result`` key to the object that
+result will be returned as the overall result and the internal highlighting code
+will never even be called.
+
+If you're plugin plans to make its own recursive calls to ``highlight`` you'll
```
egor-rogov: "If **your** plugin ..."?

PR #2398, opened 2020-02-07
Updated .hljs-comment, .hljs-string color

Nikhar444:
Fixes the source file so that using the highlightjs, won't require to change the colors separately.

### DIFF
--- a/src/styles/atom-one-light.css
+++ b/src/styles/atom-one-light.css
@@ -28,7 +28,7 @@ hue-6-2: #c18401
 
 .hljs-comment,
 .hljs-quote {
-  color: #a0a1a7;
+  color: #717277;
   font-style: italic;
 }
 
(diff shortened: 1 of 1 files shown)


yyyc514:
Context?  Why are these colors changing?

yyyc514:
If the Atom theme recently change please provide us a link to confirm/cross-check, otherwise I'm not sure what is happening here.

yyyc514:
Closing this without additional context.  If context is provided this can always be reopened.

Nikhar444:
@yyyc514, check this [PR](https://github.com/w3c/respec/pull/2738). @marcoscaceres told me to make the same changes as I did in that PR.

yyyc514:
Ok, this context is a start... but I'm still a bit confused.  Is this theme intended to be a match for the Atom One Light theme (from the Atom editor) or not?  **If so then I'd expect it to be correct when it matches that theme.**  Many of our themes make questionable choices with colors/contrast IMHO, but that's why we have so many - there is something for everyone.

So has the official Atom One theme recently made changes to their color scheme that we should follow?  Otherwise I'm not sure we should just be changing the theme because not everyone agrees with the contrast. Am I missing something here?

@marcoscaceres

PR #2401, opened 2020-02-11
Remove .js extension in require

domoritz:


yyyc514:
When do we and when don't we?  We also have the extension on line 313.

PR #2402, opened 2020-02-15
split code into modules and simplify

yyyc514:


yyyc514:
Lots of little things here, but the individual PRs should be clear.  The commits that break things out into modules are just moving code around (and perhaps renaming the functions), not changing anything else.

src/highlight.js
```diff
@@ -3,34 +3,19 @@ Syntax highlighting with language autodetection.
...
-    globalObject.hljs = factory({});
-
-    // Finally register the global hljs with AMD.
-    if(typeof define === 'function' && define.amd) {
-      define([], function() {
-        return globalObject.hljs;
-      });
-    }
-  }
+import deepFreeze from './vendor/deep_freeze';
```
egor-rogov: What's the reason to put it into `vendor` (instead of `utils`)?

yyyc514: It's pretty much borrowed code from elsewhere, hence vendoring it.

egor-rogov: I don't insist, although for me a comment pointing to the original repo would be enough.

Are you done with the PR or you're going to add some more changes?

yyyc514: It's a Rails habit, but I like it.  Doesn't hurt anything to organize it nicely.

> Are you done with the PR or you're going to add some more changes?

Done.  I try to not assign things or label them WIP if they aren't finished.

yyyc514: This finally lays the groundwork for a parse tree, but that's another PR.

yakov116:
@yyyc514 your awesome! 
Quick question, some places your using var, some your using const. Is there a reason for that?

taufik-nurrohman:
@yakov116 migration requires adaptation and `var` is ~~back~~forward compatible. It should be removed in the future for module system but for day-to-day basis, `var` is perfectly okay for web browsers.

yyyc514:
@yakov116 It's a legacy codebase.  Over time we'll switch to let/const as I notice it.  It's not a high priority to do it all at once.

PR #2403, opened 2020-02-16
fix(javascript) comma is allowed in a "value container"

yyyc514:
- fixes case where a regex would not be detected if it was anything
  other than the first parameter of a function call
- in some cases this could actually cause the whole snippet to
  be flagged as illegal if the regex contained characters that
  were invalid at the top level (such as #)

This complexity is because we only detect regexs inside "value
containers" to prevent false positivies.

This issue was found when asking Highlight.js to highlight it's
own non-minified 1.2mb browser build.

Example of failing code (this was flagged illegal):

```
hljs.COMMENT(/\{#/, /#}/),
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Core Changes:
 
 Language Improvements:
 
+- fix(javascript) comma is allowed in a "value container" (#2403) [Josh Goebel][]
 - enh(apache) add `deny` and `allow` keywords [Josh Goebel][]
 - enh(apache) highlight numeric attributes values [Josh Goebel][]
 - enh(apache) highlight IP addresses, ports, and strings in sections [Josh Goebel][]
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -189,6 +189,9 @@ export default function(hljs) {
               }
             ]
           },
+          { // could be a comma delimited list of params to a function call
+            begin: /,/, relevance: 0,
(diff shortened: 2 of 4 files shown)


yyyc514:
Yeah, our regex regex should be decently good... the problem was just that it wasn't expecting the regex at all so then it would see the `#` and blow up...

I'm pretty impressed that it could highlight the whole 1.2mb source of itself without choking on anything else...

yyyc514:
> LGTM 👍 this works correctly even when there's a comma in the regex too.

If you're going to approve please actually click the Review/Approve button so it's counted as an official review so it'll trigger the ability to merge. :)

allejo:
> If you're going to approve please actually click the Review/Approve button so it's counted as an official review so it'll trigger the ability to merge. :)

I did approve, that's how it appears for me :sweat_smile:

yyyc514:
Now it's approved. :) Thanks.

yyyc514:
Hmmm, seems Github doesn't like it much when you remove reviewers, lol.

allejo:
I think it's because I approved as a triager, but GitHub wants an approval from someone with commit access. See how my approvals show up grey instead of green?

yyyc514:
> I think it's because I approved as a triager, but GitHub wants an approval from someone with commit access. See how my approvals show up grey instead of green?

Ah, hmmm...

PR #2404, opened 2020-02-16
enh(parser) Refactor parse tree and HTML render into components

### LINKED ISSUE #1086: Provide access to a parse tree as an alternative to HTML output

yyyc514:
This refactors the storage of the parse tree and the HTML rendering into their own components and separates them from the tokenizer itself.

Closes #1086

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -5,7 +5,6 @@ https://highlightjs.org/
 
 import deepFreeze from './vendor/deep_freeze';
 import TokenTreeEmitter from './lib/token_tree';
-import HTMLRenderer from './lib/html_renderer';
 import * as regex from './lib/regex';
 import * as utils from './lib/utils';
 import * as MODES from './lib/modes';
@@ -44,7 +43,10 @@ const HLJS = function(hljs) {
     classPrefix: 'hljs-',
     tabReplace: null,
     useBR: false,
-    languages: undefined
+    languages: undefined,
+    // beta configuration options, subject to change, welcome to discuss
+    // https://github.com/highlightjs/highlight.js/issues/1086
+    __emitter: TokenTreeEmitter
   };
 
   /* Utility functions */
@@ -322,7 +324,7 @@ const HLJS = function(hljs) {
     var top = continuation || language;
     var continuations = {}; // keep continuations for sub-languages
     var result;
-    var emitter = new TokenTreeEmitter();
+    var emitter = new options.__emitter(options);
     processContinuations();
     var mode_buffer = '';
     var relevance = 0;
@@ -341,7 +343,7 @@ const HLJS = function(hljs) {
       processLexeme(codeToHighlight.substr(index));
       emitter.closeAllNodes();
       emitter.finalize();
-      result = new HTMLRenderer(emitter, options).value();
+      result = emitter.toHTML();
 
       return {
         relevance: relevance,
@@ -395,7 +397,7 @@ const HLJS = function(hljs) {
     languageSubset = languageSubset || options.languages || Object.keys(languages);
     var result = {
       relevance: 0,
-      emitter: new TokenTreeEmitter(),
+      emitter: new options.__emitter(options),
       value: escape(code)
     };
     var second_best = result;
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -34,7 +34,6 @@ const HLJS = function(hljs) {
   // Regular expressions used throughout the highlight.js library.
   var fixMarkupRe      = /((^(<[^>]+>|\t|)+|(?:\n)))/gm;
 
-  var spanEndTag = '</span>';
   var LANGUAGE_NOT_FOUND = "Could not find the language '{}', did you forget to load/include a language module?";
 
   // Global options used when within external APIs. This is modified when
@@ -50,7 +49,7 @@ const HLJS = function(hljs) {
 
   /* Utility functions */
 
-  function isNotHighlighted(language) {
+  function shouldNotHighlight(language) {
     return options.noHighlightRe.test(language);
   }
 
@@ -73,7 +72,7 @@ const HLJS = function(hljs) {
 
     return classes
       .split(/\s+/)
-      .find((_class) => isNotHighlighted(_class) || getLanguage(_class))
+      .find((_class) => shouldNotHighlight(_class) || getLanguage(_class))
   }
 
   /**
@@ -462,7 +461,7 @@ const HLJS = function(hljs) {
     var node, originalStream, result, resultNode, text;
     var language = blockLanguage(block);
 
-    if (isNotHighlighted(language))
+    if (shouldNotHighlight(language))
         return;
 
     fire("before:highlightBlock",
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -4,6 +4,8 @@ https://highlightjs.org/
 */
 
 import deepFreeze from './vendor/deep_freeze';
+import TokenTree from './lib/token_tree';
+import HTMLRenderer from './lib/html_renderer';
 import * as regex from './lib/regex';
 import * as utils from './lib/utils';
 import * as MODES from './lib/modes';
@@ -11,10 +13,12 @@ import { compileLanguage } from './lib/mode_compiler';
 
 const escape = utils.escapeHTML;
 const inherit = utils.inherit;
+
 const { nodeStream, mergeStreams } = utils;
 
 
 const HLJS = function(hljs) {
+
   // Convenience variables for build-in objects
   var ArrayProto = [];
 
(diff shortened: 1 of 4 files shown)


allejo:
This looks a bit more of an invasive change, I can look this over in-depth later this week or weekend.

yyyc514:
Anyone?

src/lib/html_renderer.js
```diff
@@ -0,0 +1,42 @@
...
+
+  // renderer API
+
+  addText(text) {
+    this.buffer += escapeHTML(text)
+  }
+
+  openNode(node) {
+    let className = node.kind;
+    if (!node.kind) return;
```
egor-rogov: What's the meaning of nodes with no `kind`? It looks strange that the renderer has to check it all the time.

src/lib/token_tree.js
```diff
@@ -0,0 +1,92 @@
...
+      this.closeNode();
+  }
+
+  toJSON() {
+    return JSON.stringify(this.rootNode, null, 4);
+  }
+
+  finalize() {
+    return;
+    TokenTree._collapse(this.rootNode);
```
egor-rogov: `_collapse` after `return;`?

src/lib/token_tree.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  closeAllNodes() {
+    while (this.stack.length > 1)
+      this.closeNode();
+  }
+
+  toJSON() {
+    return JSON.stringify(this.rootNode, null, 4);
+  }
```
egor-rogov: Are you going to use it somehow?

src/lib/token_tree.js
```diff
@@ -0,0 +1,92 @@
...
+
+  addKeyword(text, kind) {
+    if (text === "") { return; }
+
+    this.top.children.push({
+      kind,
+      children: [ text ]
+    });
+  }
```
egor-rogov: It looks like it is essentially `openNode` + `addText` + `closeNode`. So do we need this method? Its name makes me think that a "keyword" is something special in the tree, but it is not.

src/lib/token_tree.js
```diff
@@ -0,0 +1,92 @@
+export default class TokenTree {
+  constructor() {
+    this.rootNode = { root: true, children: [] };
```
egor-rogov: I can't see how the `root` flag is used. Why do you need it?

src/lib/token_tree.js
```diff
@@ -0,0 +1,92 @@
...
+
+  closeNode() {
+    if (this.stack.length > 1)
+      this.stack.pop();
+  }
+
+  closeAllNodes() {
+    while (this.stack.length > 1)
+      this.closeNode();
+  }
```
egor-rogov: Just a matter of style, but for me `while (this.closeNode());` would look better (to not repeat the same check).

src/highlight.js
```diff
@@ -309,49 +311,48 @@ const HLJS = function(hljs) {
...
-    }
-
     function processKeywords() {
-      var keyword_match, last_index, match, result;
+      var keyword_match, last_index, match, result, buf;
 
-      if (!top.keywords)
+      if (!top.keywords) {
+        emitter.addText(mode_buffer);
         return escape(mode_buffer);
```
egor-rogov: Should be just `return;`?

src/lib/token_tree.js
yyyc514: It's very useful for debugging.

src/lib/html_renderer.js
yyyc514: It's for unnamed modes... ie, no `className`... there really isn't a reason to have a special node in the tree for them since they don't affect the output one bit.  You could argue this is a premature optimization except storing the tree itself has measurable memory implications relative to just using a string.  So I didn't see the need to track pointless nodes.

Technically if we were going to store `relevance` in the tree (which we don't now) then we'd have to revisit this as we'd actually need the node then.

---

You might have a point though that this should be a responsibility of the parser though, not the tree itself, that now the tree knows too much about the parser internals.  I'll give that some thought.

src/lib/token_tree.js
yyyc514: >  Its name makes me think that a "keyword" is something special in the tree, but it is not.

Here tree is just an implementation of an abstract interface.  Javascript doesn't have interfaces that I'm aware of... I'm defining the simplest API that is useful for the parser and then NodeTree (or whatever other object) is simply fulfilling that interface.

I found addKeyword much more useful than repeating the pattern you suggest over and over. 

I suppose you could argue that technically the emitter object should be separate from the implementation of the tree object... but I'm not sure it's worth it here since the objects are already so dead simple.

src/lib/token_tree.js
yyyc514: Removing.

src/lib/token_tree.js
yyyc514: This requires closeNode now return a value, which really it's now only doing to help out closeAllNodes, which has a certain smell for me, but I don't care that much, so I'll change.

src/highlight.js
yyyc514: Yep.

src/lib/token_tree.js
yyyc514: It seemed a helpful concept, does it bother you?

src/lib/html_renderer.js
yyyc514: **Oh, wait this is the HTML renderer.**  It's POSSIBLE that the tree might include notes that don't need require any output (such as future `relevance` nodes mentioned above) so hence this check is done on each node.  

It's not the tree's responsibility to know what should and should not be output, that's the renderers job, hence the check here.

src/lib/token_tree.js
egor-rogov: > I found addKeyword much more useful than repeating the pattern you suggest over and over.

It is called from exactly one place, actually. To my taste it introduces more complexity. At least, it could be implemented in terms of already defined methods (openNode, addText, closeNode). Otherwise it takes some time to understand that it's not something new but simply a shortcut.

src/lib/token_tree.js
egor-rogov: But you really ain't gonna need it (:
With this flag in place, there's no certainty about what is the root: a node that `rootNode` points to, or a node with `root` flag?
Moreover, you have to delete that root flag from sublanguage nodes.
Less code, less problems...

src/lib/html_renderer.js
egor-rogov: Hm. Maybe `kind` is not the best choice than... It's not that intuitive that `kind` reflects the need to output something when walking the tree.
Bwt, shouldn't relevance be an attribute of an existing node (not a separate node)?

src/lib/token_tree.js
yyyc514: > It is called from exactly one place, actually.

Ok that takes the wind out of my "over and over" sails, lol.  But I still think that doesn't mean it isn't an helpful abstraction.  It's my "separation of concerns" voice saying the parser doesn't need to know that much about the structure of the tree... hence treating it as an emitter object and just telling it "emit(keyword)", etc.

> At least, it could be implemented in terms of already defined methods

Now that's a useful observation for sure, although I worry it'd be slower - but since we have no real large benchmarks and premature optimization and all I'll refactor it to build on the existing blocks.

src/lib/token_tree.js
yyyc514: Removed.

src/lib/html_renderer.js
yyyc514: >  It's not that intuitive that kind reflects the need to output something when walking the tree.

**You have to think about them separately.**  Kind ONLY tells the tree what kind of node it is.  It has no other meaning INSIDE the class.

But I can make a helpful function to wrap this up conceptually...

> Bwt, shouldn't relevance be an attribute of an existing node (not a separate node)?

I'm not sure I follow... if we added relevance it would be an attribute of a node.  A node has a given relevance...

yyyc514: I think this help a lot:

https://github.com/highlightjs/highlight.js/pull/2404/commits/4d3078325c33155b8fba7495a9480127fe438422

egor-rogov: > if we added relevance it would be an attribute of a node. A node has a given relevance...

Right. Then I'm not sure I understand this:

> Technically if we were going to store relevance in the tree (which we don't now) then we'd have to revisit this as we'd actually need the node then.

(Not that it is important.)

yyyc514: I mean that right now the parser skips emitting nodes that don't have a className... because they don't affect output.  BUT if we wanted to store relevance in the tree then we'd need nodes to store it.

Right now if you had a bunch of rules but no classNames you'd end up with ONE large test node... with no where to store relevance...

I suppose another way would be to just accumulate relevance into whatever the current node is, but that would make it harder for 3rd parties to work with relevance data.

yyyc514: I may still rip out this whole thing before 10.0, LOL - leaving it only as a reference implementation.  It concerns me that there are increased memory requirements.  I don't think anyone using this on their blog is going to notice but someone using this on a backend system to process thousands of snippets or some sort of more "industrial" use this new version might have significantly different requirements than the old version.

So I weigh that against: is adding a parse tree that only a tiny portion of people have actually requested worth it?

Making it optional is another choice, we'd just have to alter the test suite to run both both emitters... that starts to sound like a lot of headache though.

So I think what might happen is the emitter API stays (that part of the work is a big improvement) but the parse tree implementation because reference and we allow people to configure their own emitters.  I really think this type of thing should be easy to do (emit to a tree instead of string), but that doesn't mean we  have to ship it by default.

Then by default we'd ship the "FastHTMLRender" emitter. :-)

yyyc514:
Please see the last commit, that's kind of the culmination of where I wanted to take this.

src/lib/token_tree.js
egor-rogov: Interestingly, `npm test` runs on my laptop 8s, and 7s with this PR applied. So in fact your refactoring speeds things up. Cool.

yyyc514: > Interestingly, npm test runs on my laptop 8s, and 7s with this PR applied. So in fact your refactoring speeds things up. Cool.

Oh - that's interesting?  Try measuring the memory usage though.  I only tried it once but it was like double or so.

yyyc514: With zsh `time`:

parse tree:

```
avg shared (code):         0 KB
avg unshared (data/stack): 0 KB
total (sum):               0 KB
max memory:                191216 MB
page faults from disk:     0
other page faults:         64816
```

master:

```
avg shared (code):         0 KB
avg unshared (data/stack): 0 KB
total (sum):               0 KB
max memory:                227424 MB
page faults from disk:     0
other page faults:         73349
```

So now it seems parse tree is the clear winner in all regards, hmmmmm...

yyyc514: I'm not sure WHY it's faster or more memory performant though... I'd have bet the opposite on both counts.

src/lib/html_renderer.js
yyyc514: So I take back everything I said here if the parse tree is actually FASTER and uses less memory then it's a clear win win, though I'm still puzzled.

egor-rogov: I got similar figures... but I'm not sure if `time` counts child processes?

src/lib/token_tree.js
egor-rogov: I got similar figures... but I'm not sure if `time` counts child processes?
I'm running `time -v npm test` and `ps` show two processes: `node` and `npm`.

yyyc514: I run the tests directly usually, so I'm not even invoking `npm` at all. :)

yyyc514: @marcoscaceres Any idea why we'd see these results?  I mean we're still building the huge output string, as before - just now we're ALSO first building a huge tree of nodes and then walking it to generate the output string.

yyyc514: Ok, I'm not completely crazy.  When running a single test of a 1.2mb JS file (our own built source) I get ~88mb with master and ~93mb with parse tree.  So "it depends"... but with such a small increase (and even decrease for smaller workloads) I don't think I have a problem with the parse tree being the new default.

egor-rogov: Agreed.
A profiling tool should give insights, but that's another story. I think we're done with it by now.

PR #2410, opened 2020-02-18
enh(armasm) support // and # style line comments

### LINKED ISSUE #2408: (armasm) Line comments do not seem to be handled correctly
I was referred to make this bug report by Discord support after seeing this behavior in Discord, which uses highlight.js.

When highlighting ARM assembly, `# comments` and `// comments` are not treated as comments but instead as live code. This leads to incorrect highlighting that can span the entirety of the code, see the included screenshot.

[Example (from Discord)](https://i.imgur.com/o2DKPtC.png)

yyyc514:
Fixes #2408.

### DIFF
--- a/src/languages/armasm.js
+++ b/src/languages/armasm.js
@@ -10,7 +10,7 @@ export default function(hljs) {
 
   const COMMENT = {
     variants: [
-      hljs.COMMENT('^[ ]*(?=#)', '$', {relevance: 0, excludeBegin: true }),
+      hljs.COMMENT('^[ \\t]*(?=#)', '$', {relevance: 0, excludeBegin: true }),
       hljs.COMMENT('[;@]', '$', {relevance: 0}),
       hljs.C_LINE_COMMENT_MODE,
       hljs.C_BLOCK_COMMENT_MODE,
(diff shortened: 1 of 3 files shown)


src/languages/armasm.js
```diff
@@ -7,6 +7,16 @@ Category: assembler
 
 export default function(hljs) {
     //local labels: %?[FB]?[AT]?\d{1,2}\w+
+
+  const COMMENT = {
+    variants: [
+      hljs.COMMENT('^[ ]*(?=#)', '$', {relevance: 0, excludeBegin: true }),
```
egor-rogov: Why just spaces, why not `\s`?

src/languages/armasm.js
```diff
@@ -87,8 +96,8 @@ export default function(hljs) {
       {
         className: 'symbol',
         variants: [
+            {begin: '^[ ]*[a-z_\\.\\$][a-z0-9_\\.\\$]+:'}, //GNU ARM syntax
```
egor-rogov: The same question about white-spaces here.

src/languages/armasm.js
yyyc514: `\s` eats new lines also and results in weirdness like the divs opening on the line before because the comment "includes" the newline.

src/languages/armasm.js
yyyc514: Same answer.

src/languages/armasm.js
egor-rogov: Ok. How about adding `\t` then?

yyyc514: We could do that.  Are we good to go if we add support for tabs?

egor-rogov: Yep.

PR #2412, opened 2020-02-19
WIP: try different approach to JSX

joshgoebel:
This is an attempt at a more modular approach to JSX.

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -5,15 +5,9 @@ Category: common, scripting
 Website: https://developer.mozilla.org/en-US/docs/Web/JavaScript
 */
 
+import { JSX_MODE, JSX_LANGUAGE } from "./lib/jsx";
+
 export default function(hljs) {
-  var FRAGMENT = {
-    begin: '<>',
-    end: '</>'
-  };
-  var XML_TAG = {
-    begin: /<[A-Za-z0-9\\._:-]+/,
-    end: /\/[A-Za-z0-9\\._:-]+>|\/>/
-  };
   var IDENT_RE = '[A-Za-z$_][0-9A-Za-z$_]*';
   var KEYWORDS = {
     keyword:
@@ -100,6 +94,15 @@ export default function(hljs) {
     excludeEnd: true,
     contains: PARAMS_CONTAINS
   };
+
+  const containers = [
+    hljs.APOS_STRING_MODE,
+    hljs.QUOTE_STRING_MODE,
+    TEMPLATE_STRING,
+    hljs.C_BLOCK_COMMENT_MODE,
+    hljs.C_LINE_COMMENT_MODE
+  ];
+  hljs.registerLanguage("_jsx", JSX_LANGUAGE({containers}));
 
   return {
     name: 'JavaScript',
@@ -206,19 +209,7 @@ export default function(hljs) {
             end: /\s*/,
             skip: true,
           },
-          { // JSX
-            variants: [
-              { begin: FRAGMENT.begin, end: FRAGMENT.end },
-              { begin: XML_TAG.begin, end: XML_TAG.end }
-            ],
-            subLanguage: 'xml',
-            contains: [
-              {
-                begin: XML_TAG.begin, end: XML_TAG.end, skip: true,
-                contains: ['self']
-              }
-            ]
-          },
+          JSX_MODE
         ],
         relevance: 0
       },
--- a/src/languages/lib/jsx.js
+++ b/src/languages/lib/jsx.js
@@ -1,0 +1,110 @@
+import { inherit } from "../../lib/utils";
+
+const FRAGMENT = {
+  begin: '<>',
+  end: '</>'
+};
+const XML_TAG = {
+  begin: /<[A-Za-z0-9\\._:-]+/,
+  end: /\/[A-Za-z0-9\\._:-]+>|\/>/
+};
+
+export const JSX_MODE = { // JSX
+  variants: [
+    { begin: FRAGMENT.begin, end: FRAGMENT.end },
+    { begin: XML_TAG.begin, end: XML_TAG.end }
+  ],
+  subLanguage: '_jsx',
+  contains: [
+    {
+      begin: XML_TAG.begin, end: XML_TAG.end, skip: true,
+      contains: ['self']
+    }
+  ]
+};
+
+// note: this is not a language grammar itself but when called
+// it returns a function that is the language grammar
+// This could be improved if our registerLanguage API was enhanced
+// to support passing options to grammars.
+export const JSX_LANGUAGE = function({containers}) {
+
+  // containers are any blocks which could prevent an inline JS mode
+  // from ending... inline javascript is denoted with `{ code }`
+  // so an example would be a string: ` { "}" }`  Become the } is
+  // inside the container it should be ignored and not counted as
+  // ending the inline JS snippet
+  //
+  // these are passed in because it makes sense that these would
+  // already be defined by the parent language
+  containers = containers.map((x) =>
+    // we need these to be quiet, and not output
+    inherit(x, {skip: true, className: null})
+  )
+
+  return function (hljs) {
+
+    const XML_IDENT_RE = '[A-Za-z0-9\\._:-]+';
+    const XML_ENTITIES = {
+      className: 'symbol',
+      begin: '&[a-z]+;|&#[0-9]+;|&#x[a-f0-9]+;'
+    };
+    const INLINE_JS = {
+      begin: /{/,
+      end: /}/,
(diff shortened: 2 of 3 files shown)


yyyc514:
I'm most curious for thoughts on desired output.  Since JSX can include embedded JS you get into this vicious cycle where now you really need to jump back into `javascript` sublanguage again... so you end up with code like:

```js
<span class="xml"><span class="hljs-tag">&lt;<span class="hljs-name">BrowserRouter</span>&gt;</span>
        <span class="hljs-tag">&lt;<span class="hljs-name">div</span>&gt;</span>
          <span class="javascript">{<span class="hljs-comment">/* comment */</span>}</span>
          <span class="javascript">{ test <span class="hljs-string">"}"</span> }</span>
```

Is it really critical/necessary to flag all of these as javascript, or should they really be something more like `subst`?

Then you also have comments inside tags (which seems to be valid and used):

```js
box = (<div
  box="yes" // comment
/>)
```

Which results in code like:

```html
<span class="hljs-attr">box</span>=<span class="hljs-string">"yes"</span> <span class="hljs-comment">// comment</span>
``

But technically the `//comment` is Javascript code, not HTML... so should it really also be in `javascript` or `subst` block?

I think the answers start to depend a lot on whether you think of this as a back and forth between JS and HTML or rather one larger meta-language such as "JSX".  In the latter case for example, I don't think the comment needs to be flagged as JS... though I do wonder in that case if `jsx` wouldn't be a better over-arching class than `html`.

@egor-rogov @allejo

**Comment by egor-rogov** [comment]
> But technically the `//comment` is Javascript code, not HTML... so should it really also be in `javascript` or `subst` block?
> 
> I think the answers start to depend a lot on whether you think of this as a back and forth between JS and HTML or rather one larger meta-language such as "JSX". In the latter case for example, I don't think the comment needs to be flagged as JS

I don't think it's our goal to mark different parts of code with appropriate languages. Our goal is to highlight comments as comments, no matter which language they belong to.

At the first glance the PR looks complicated. Registering of an internal language, a module in `lib`... I don't like the idea very much. 
Btw, please remind me, what's wrong with the current implementation?

**Comment by yyyc514** [comment]
> Btw, please remind me, what's wrong with the current implementation?

There is no good way to properly handle `{{ javascript.code() }}` or `// comments inside HTML` in the current implementation since neither of these things are valid HTML.

Hence needing an internal JSX variant of HTML.


>I don't think it's our goal to mark different parts of code with appropriate languages. 

`subLanguage` has always done that... are you suggesting that it should not?

> At the first glance the PR looks complicated. Registering of an internal language, a module in lib... I don't like the idea very much.

Would you prefer a run-time require vs the module?  In this case there is no point at all of the "_jsx" language without Javascript, so registering and linking at run-time seems wrong. And splitting the sublanguage out into it's own module is just good organization - so I'm not completely sure what you're objecting to WRT that.

> Registering of an internal language

Why? Does the idea of internal languages bother you in general?  Would it bother you less if they were hidden and not in the `listLanguages` namespace?

**Comment by yyyc514** [comment]
Also, the reason it's so dynamic is because eventually this can be used inside Javascript or Typescript, so you won't just have a single "_jsx" grammar. You'd have variants also: `_jsx_javascript`, `_jsx_typescript`, etc... since the JSX grammar needs a way to embed snippets of the parent grammar. So each variant would have a different internal `sublanguage` value with `{{ }}` blocks.

**Comment by yyyc514** [comment]
> Although I just had another thought that I need to look into.

Picking out HTML tags singly (one by one) inside of JS is another [much simpler] idea... But that is also a much "dumber" solution, in that _it can't identify blocks of HTML_ - and it's also broken in that it would incorrectly highlight JS code (or keywords) inside of HTML blocks. Ex:

```js
var x = (<div>class is not a keyword here</div>);
```

I'm not sure how people are using HLJS to highlight JSX code but if it was me I'd want a different background/style applied to `xml` blocks inside Javascript... they really should be rendered more like "string" or heretic something.

So it seems we need to detect actual **blocks** of HTML (which is how this has always worked in the past) not just individual HTML tags here and there.

allejo:
> You'd have variants also: `_jsx_javascript`, `_jsx_typescript`

What other variants are there? Also, wouldn't this be `jsx` and then `tsx`?

> I'm not sure how people are using HLJS to highlight JSX code but if it was me I'd want a different background/style applied to xml blocks inside Javascript... they really should be rendered more like "string" or heretic something.

While the React website uses Prism, can't we use the React website as the basis of how people [would] use highlight.js to highlight JSX? https://reactjs.org/tutorial/tutorial.html#what-is-react

> So it seems we absolutely need to detect actual **blocks** of HTML (which is how this has always worked in the past) not just individual HTML tags here and there.

Excuse the ignorance, when you say "blocks of HTML" do you mean matching the `<div>` with the `</div>` in your example and consider that a "block?"

yyyc514:
> What other variants are there? Also, wouldn't this be jsx and then tsx?

I'm not sure, maybe only those two.  The names don't matter they are just internal identifiers.  If the class/sublangauge names matter then we can encode pretty names and then we can discuss what those might be... but it's possible they might both actually be the same (xml).

> While the React website uses Prism, can't we use the React website as the basis of how people [would] use highlight.js to highlight JSX? https://reactjs.org/tutorial/tutorial.html#what-is-react

It's one point of data, true, but see my larger point about why we need to identify WHOLE blocks in any case - regardless of how we colorize the background.

> Excuse the ignorance, when you say "blocks of HTML" do you mean matching the <div> with the </div> in your example and consider that a "block?"

Yes.  We need to know "html starts here" and "html ends here" or else we have no idea whether we are highlighting JS or HTML.

yyyc514:
Actually, perhaps this could build on the work in https://github.com/highlightjs/highlight.js/pull/2261 and then we have a tiny HTML parser... so that we're actually MATCHING beginning and end tags... _I think JSX is always supposed to have a single wrapping tag in any case, is it not?_

You still have the case issue, but that can be solved for a small case like this with just double specifying the case in the HTML matching rules.

egor-rogov:
I had another look into this. Inasmuch I can't propose something better/simpler, I'm not against it.

> > Registering of an internal language
> 
> Why? Does the idea of internal languages bother you in general? Would it bother you less if they were hidden and not in the `listLanguages` namespace?
> 
> Would a single `internal: true` flag perhaps suffice?

My concern here is that an internal language can potentially cause some hard-to-troubleshot problem, like name collision or something. That's why I'd prefer either to make `_jsx` a rightful grammar or to encapsulate it in another namespace.

yyyc514:
> My concern here is that an internal language can potentially cause some hard-to-troubleshot problem, like name collision or something. That's why I'd prefer either to make _jsx a rightful grammar or to encapsulate it in another namespace.

Except JSX isn't a grammar, it's only a sub-grammar.  You don't think `_` is already a namespace? Would `__jsx` make you any more comfortable? :-)  The name will be auto generated for the language so conflicts isn't really a concern I have... Ie you'd have `__jsx_javascript` and `__jsx_typescript`, etc.

egor-rogov:
> The name will be auto generated for the language so conflicts isn't really a concern I have... Ie you'd have `__jsx_javascript` and `__jsx_typescript`, etc.

How's that?
```js
  hljs.registerLanguage("_jsx", JSX_LANGUAGE({containers}));
```

> A conflict isn't going to happen and if it does, it's the users problem - since these would be in the languages list they would be visible and avoidable.

It's not visible when you're looking into `src/languages` catalog...

But yes, underscore is like a namespace, so maybe we'll just need a word or two in developer's docs about this naming convention, and that's it.

yyyc514:
> How's that?

That's just how it would work internally - to prevent conflicts.

> It's not visible when you're looking into src/languages catalog...

No, because it's not a true grammar. :-)

> But yes, underscore is like a namespace, so maybe we'll just need a word or two in developer's docs about this naming convention, and that's it.

I agree only to the extent that we should make it easy for people to filter them out (if need be), say if they wanted to say build a dropdown or something - so to that end a small mention in the docs might make sense.  I still think I might scrap this whole approach in favor of the new callbacks though and just match the pair of HTML tags with callbacks... this was an experiment.  We'll see.

Right now I'm focused on other things and waiting for the website/server changes we need so we can do a proper actual v10 release.  So this stuff is likely pushed until 10.1 anyways.

jamiebuilds:
Hi @joshgoebel, can I ask why this was closed? I'd be happy to take this PR over and fix any remaining issues. Unless you want to take a different approach entirely, also happy to do that.

joshgoebel:
Wish I had left some notes, LOL.   Please see the newly created https://github.com/highlightjs/highlight.js/issues/2998

joshgoebel:
I'd suggest reading #2998 and the 3 other threads fully and then if you think you have some unique insights we'd be happy to hear them. :)

PR #2413, opened 2020-02-19
enh(python) Added `python-repl` for Python REPL sessions

### LINKED ISSUE #2025: (pycon) Add pycon (Python repl) support
For Python interactive sessions (with `>>>`), the output of commands is also syntax highlighted as Python. This is almost always not correct. It's generally quite distracting to have `in`, `for` and so on colored differently in output. 

For example

```py
>>> print("Ask not what your country can do for you")
Ask not what your country can do for you
```

The `not` and `for` in the output are colored (you can also see GitHub's highlighter has the same issue).

The exception might be tracebacks, but even these should be highlighted with care. Currently the line numbers and "in" from the file lines (like `  File "<stdin>", line 2, in test`) are colored. A special `pytb`/`python-traceback` language could be useful here. 

It would be better if the presence of `>>>` in Python disabled highlighting for any lines that don't start with `>>>` or `...`. Alternately, you could add a `pycon` language like Pygments.

yyyc514:
Closes #2025.


src/languages/python-repl.js
```diff
@@ -0,0 +1,29 @@
...
+          // a space separates the REPL prefix from the actual code
+          // this is purely for cleaner HTML output
+          end: / /,
+          starts: {
+            end: '$', subLanguage: 'python'
+          }
+        },
+        variants: [
+          { begin: /^>>>(?=[ ])/ },
+          { begin: /^\.\.\.(?=[ ])/ }
```
allejo: Python REPL accepts tabs (`\t`) as well in addition to spaces

yyyc514: But this is the character between the REPL lines in the output... are you saying that is a space and not a tab?  We're not talking about what the interpreter supports here.

yyyc514: I'm not even sure how I would test that... If I copy and paste it from terminal it'll be a space...

allejo: Ohhh oops! Sorry about that, I misunderstood the purpose of that space. You are correct, those are spaces coming in from terminal

yyyc514: I used to just throw \s everywhere but that can do very bad things sometimes when it starts matching line endings (CR/LF, etc) as white space.

allejo: Should the space be optional? See the copy/paste from my local terminal as an example

```
Python 3.7.3 (default, Dec 23 2019, 21:25:06)
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def hello():
...     print("hello")
...
>>> hello()
hello
>>>
>>>
```

yyyc514: I'm ok with making it space or $... that would cover that edge case.

src/languages/python-repl.js
```diff
@@ -0,0 +1,29 @@
...
+          // a space separates the REPL prefix from the actual code
+          // this is purely for cleaner HTML output
+          end: / |$/,
+          starts: {
+            end: '$', subLanguage: 'python'
+          }
+        },
+        variants: [
+          { begin: /^>>>(?=[ ]|$)/ },
+          { begin: /^\.\.\.(?=[ ])/ }
```
allejo:
```suggestion
          { begin: /^\.\.\.(?=[ ]|$)/ }
```

There can be an empty `...` line in REPL, no? (see my example)

yyyc514: Oh how did I miss that.  These are edge cases though really. :)

MartinThoma:
@yyyc514 Should `python-repl` be added to https://github.com/highlightjs/highlight.js/blob/master/SUPPORTED_LANGUAGES.md ?

yyyc514:
Indeed it should, want to make PR? :)

PR #2417, opened 2020-02-20
add(php-template) Explicit language to detect PHP templates

### LINKED ISSUE #2330: Discuss: How we handle the PHP templating language
Related:

- https://github.com/highlightjs/highlight.js/issues/725
- https://github.com/andreasjansson/language-detection.el

Technically PHP is a templating language that exists inside HTML, ie a full PHP template snippet would look like:

```php
<?php echo '<p>Hello World</p>'; ?> 
```

Although `echo '<p>Hello World</p>';` is still valid PHP code, per se... and it's possible to run this with `php -r`.   So it's possible to have PHP code without the PHP tags...

So keep these two very different context in mind...

---

Our current handling of PHP is problematic in several ways.  It's not very intuitive, and it gives people the wrong idea about how to handle templating languages in general.  PHP currently is handled by `xml` and `php`.  `php` is the actual php code itself, while the php templating functionality is buried inside `xml`.

That means a snippet such as `<?php echo '<p>Hello World</p>'; ?> ` would be auto-detected as XML, not PHP.  I believe that is one reason we score so poorly on some of the [language detection stats](https://github.com/andreasjansson/language-detection.el).  They are throwing PHP files at us and expecting us to return "php" but instead we return "xml".

This also isn't intuitive if you are manually trying to specify the highlighting language... if you have a FULL PHP template and specify "php" your highlighting will be pretty broken because HTML won't be highlighted at all... you'd instead need to specify "xml" to get proper highlighting.

But if you have only php *code*, without the `<?php ?>` start/end tags then `php` would work and `xml` would be badly broken.

---

Just including PHP in `xml` makes people start thinking about this wrongly (IMHO)... See issue #725, for example.  They see PHP handled by XML and get the idea that that's perhaps where we should dump other templating languages.

Should XML also handle Mojolicious? ERB?  Handlebars? etc... I think it's obvious this is not the correct path.  The correct way to specify a grammar for PHP would look something like:

```js
  return {
    subLanguage: 'xml',
    contains: [
      {
        begin: '<\?(php)?', end: '\?>',
        subLanguage: 'php-code'
      }
    ]
  };
}
```

If we just made this change today (moving the existing `php` to `php-code`) this should "just work" regarding auto-detect.  But now we'd have two different languages that PHP could be detected as, depending on whether it was enclosed in `<?php ?>` tags or not... `php` or `php-code`.  This seems suboptimal.

Also, this presents a new problem when someone wants to manually specify the language.  Should they use `php` or `php-code`?

I'm not 100% sure what the best course of action is here, but I think we need to figure this out and slide it into version 10, as it's likely going to be another breaking change.

yyyc514:
Closes #2330.


src/languages/php-template.js
```diff
@@ -0,0 +1,30 @@
...
+Requires: xml.js, php.js
+Author: Josh Goebel <hello@joshgoebel.com>
+Website: https://www.php.net
+Category: common
+*/
+
+export default function(hljs) {
+  return {
+    name: "PHP template",
+    subLanguage: 'xml',
```
egor-rogov: I can't say I like this two-grammars approach very much...
But looking at your PR I got a thought: what if we allow two sublanguages here? I. e.
```js
    subLanguage: ['xml','php'],
```
This way we got one grammar that resolves into either `xml` + contained `php`, or just `php`.
If only we had a way to disallow `php` to act as a top-level language...

yyyc514: > If only we had a way to disallow php to act as a top-level language...

But it IS a top-level language... someone could post a snippet of PHP without the tags... in which case it's a php snippet, ie `php` code... meanwhile a `php-template` is HTML+PHP...

> what if we allow two sublanguages here?

That makes me head hurt - and what is the advantage of doing it that way?  PHP code on it's own is NOT a php template.

yyyc514: > I can't say I like this two-grammars approach very much...

It's consistent with how we do VBScript-HTML and how I've suggested we do other templating languages.  For these type of things you have 3 things happening:

- The language (say Ruby)
- The templating language (say ERB)
- The actual output format (say HTML)

Those would be 3 different grammars.  This is true for PHP also... the language itself is technically distinct from the templating language.

egor-rogov: Just rename `php` to `_php`, and `php-template` to `php`.
So that you can use single `php` for both php snippets and php templates.

yyyc514: How would that work for ERB or VBscript?  I'm trying to also look at the BIG picture here on how we handling templating in general.

yyyc514: > So that you can use single php for both php snippets and php templates.

REMEMBER, no one is using `php` now for PHP... well they might be, but they're more likely using `xml` since that highlights HTML + PHP.

yyyc514:
>     subLanguage: ['xml','php'], 

**That doesn't work because that's EITHER OR.**  Either it will be colored HTML OR it will be colored PHP.  That's only 2 out of the 3, hence why we're adding php-template and why it has to be mixed.

src/languages/php-template.js
egor-rogov: > REMEMBER, no one is using php now for PHP... well they might be, but they're more likely using xml since that highlights HTML + PHP.

Right. And we consider it bad, because it makes xml grammar bloated.
On the other hand, adding xml support to the core language doesn't seem bad. Or it does?

> How would that work for ERB or VBscript?

The same way.
(leaving aside existing ERB and VBScript-html for the sake of discussion)

yyyc514:
I don't think we have a perfect way of doing this currently (see the longer issue thread).  I personally think `php` and `php-template` is better than PHP and XML.  And in the future if we add features to allow us to fix this then we could always just make `php-template` an alias of the new super powered `php`.

egor-rogov:
> That doesn't work because that's EITHER OR. Either it will be colored HTML OR it will be colored PHP. That's only 2 out of the 3, hence why we're adding php-template.

Why? This part takes care of it:
```
    contains: [
      {
        begin: /<\?(php|=)?/,
        end: /\?>/,
        subLanguage: 'php',
        ...
```

> I personally think `php` and `php-template` is better than PHP and XML.

Do you mean "is better than _existing_ PHP and XML"?
What do you think of (in principle, no matter how technically implemented) having `php` for both php code and php templates, and `xml` for XML/HTML? And likewise, `ruby` for both ruby and erb, etc.

yyyc514:
> Why? This part takes care of it:

I'm not sure I follow...  How? Where is that code snippet?  As previously mentioned you can't put every single templating language inside XML  - and it's inconsistent.  We already have many HTML templating languages (Twig, handlebars, HTMLBars, etc)... they are always their own grammar.

When you do something like sublanguage: `[php, html]` That means ONE will be picked. So if you're doing that at the top-level then you're really just picking php or html, but that does not get you php + html.  

So really what you're suggesting is `PHP` = `['_php', '_php-template']...


> Do you mean "is better than existing PHP and XML"?

Yes.

> What do you think of (in principle, no matter how technically implemented) having php for both php code and php templates, and xml for XML/HTML? 

As a general way of thinking about it I'm not opposed.  But I think this is specific to how PHP is thought about.  PHP has long been defined as "PHP/HTML"...

> And likewise, ruby for both ruby and erb, etc.

Very much disagree here.  **Ruby and ERB are not interchangeable.**  Ruby code is Ruby code and ERB is a templating language based on Ruby.

yyyc514:
> So really what you're suggesting is PHP = `['_php', '_php-template']...

I'd have to give this some more thought, but it produces really ugly HTML:

```
<div class="hljs php">
<div class="_php-template"><div class="xml">...</div>
<div class="_php>...</div>
<div class="xml">...</div>
<div class="_php>...</div>
<div class="xml">...</div>
</div>
...
```

yyyc514:
**This is going to a lot of trouble just to build a "nice php alias"...** Since that's the ONLY additional thing doing it this way gives you.  Having "php" and "php-template" for auto-detect works just fine, auto-detect will decide... but if you want the user to only ever have to think `php`... then you end up going down a more crazy path.

I'd like to suggest that `php` and `php-template` are clearer than `php` and `xml`. _And perhaps the aliasing problem needs to be solved at a different layer, not in the grammars themselves._

You could imagine something like:

```
hljs.addComplexAlias("php", to: ["php","php-template"])
```

yyyc514:
This touches on https://github.com/highlightjs/highlight.js/issues/2371 also.

Perhaps aliases should be more ambiguous.  As as soon as you add `asm` as an extension alias it could really refer to x86, ARM, AVR, etc... perhaps that's OK. and using `asm` should trigger auto-detect of all 3 languages.

egor-rogov:
> I'd like to suggest that php and php-template are clearer than php and xml

No doubt.

> PHP has historically been defined as "PHP/HTML"
> Ruby and ERB are not interchangeable. Ruby code is Ruby code and ERB is a templating language based on Ruby.

ERB is lucky to have its own name. But for PHP you're like "well, I'm using PHP inside HTML, so I'm going to use... what's-his-name... php-html? no... ah, php-template! some day I'll memorize..."

> So perhaps both PHP and PHP-template claim php as an alias and that "just works".
> I don't think that future feature should prevent merging this PR though - as I still feel this is a step in the right direction.

Yeah, this is a way to go.
I tried to do something like this on the level of grammar, but it doesn't look right.

yyyc514:
> But for PHP you're like "well, I'm using PHP inside HTML, so I'm going to use... what's-his-name... php-html? no... ah, php-template! some day I'll memorize..."

Yeah, I do understand the problem you're trying to solve. :-)  And I don't love `vbscript-html` either necessarily.  Technically it's "Active Server Pages" I think [that allows for Javascript also]... I used that stuff 20 years ago or so...

PR #2421, opened 2020-02-22
Add incorrect highlighting issue template

yyyc514:


.github/ISSUE_TEMPLATE/incorrect-syntax-highlighting.md
```diff
@@ -0,0 +1,26 @@
...
+A clear and concise description of what the issue seems to be.
+
+**Which language seems to have the issue?**
+Please specify exactly *which* language grammar you are using to highlight (`python`, `javascript`, etc.) or specify if you are using auto-detection.  If you are using auto-detection please first check to make sure the language you are expecting was properly detected.  If Highlight.js misidentifies the type of code - then the highlighting may be entirely incorrect - and there often isn't much we can do about that - auto-detection is only best-effort and can't guess right 100% of the time.
+
+**Sample Code to Reproduce**
+Please include text examples of the code that fails to highlight properly or can reproduce the bug.  You can attach a screenshot if you'd like, but the code is necessary because we may need to copy/paste it to help resolve the issue - or add it to a test.
+
+A jsfiddle can sometimes be even better.  You can fork an example test case:
+https://jsfiddle.net/ajoshguy/gfzujpyt/5/
```
allejo: Should we set a [base version](https://docs.jsfiddle.net/getting-started#setting-fiddle-as-base-version) here? This way we wouldn't have to update this template each time there's a new highlight.js release.

yyyc514: How would we do that exactly?

allejo: https://docs.jsfiddle.net/getting-started#setting-fiddle-as-base-version

yyyc514:
>  If users forget to delete the current text, it's a lot of fluff to sort through each time.

I prefer to try and avoid solving hypotheticals.  If that turns into an actual problem then we can circle back and review how to solve it.

.github/ISSUE_TEMPLATE/incorrect-syntax-highlighting.md
yyyc514: Ah yes, for sure.

allejo:
> I prefer to try and avoid solving hypotheticals. If that turns into an actual problem then we can circle back and review how to solve it.

I don't feel passionately about making these HTML comments, but I do think it'd make it easier for users to just come in, write a description and reference back to the instructions of each section to make sure they didn't forget anything and not have to worry about deleting it. Like if the description of each section is the first thing I delete and I start writing out a description but forget what the jsfiddle link is, I'd have to undo my changes and get the original template or open a new tab to reference the original prompt.

yyyc514:
I think HTML comments are weird here - and I didn't even really know you could do that and I've been using GitHub a LONG time... I've worked with template issues before (as a user) and it's not really that hard to keep things straight, copy/paste, etc...  So I think I still prefer "wait and see".

Can you point me to any big project using comments this way on GitHub or is that your own unique idea?

allejo:
> Can you point me to any big project using comments this way on GitHub or is that your own unique idea?

I got the idea from Jekyll, https://raw.githubusercontent.com/jekyll/jekyll/master/.github/ISSUE_TEMPLATE/feature_request.md

But I'm fine with waiting and seeing if this becomes a problem for us.

yyyc514:
Nice, they do use it to good effect.

yyyc514:
> But I'm fine with waiting and seeing if this becomes a problem for us.

Yep. That's the plan.

PR #2423, opened 2020-02-23
enh(powershell) Add PowerShell 5.1 default aliases as "built_in"s

### LINKED ISSUE #1338: (powershell) Add v5.0 aliases as builtins
Powershell v5.0 has, out of the box, about 150 different aliases for commonly used commands. It would be handy if these were recognized as builtins by HLJS.

The list:
`%, ?, ac, asnp, cat, cd, CFS, chdir, clc, clear, clhy, cli, clp, cls, clv, cnsn, compare, copy, cp, cpi, cpp, curl, cvpa, dbp, del, diff, dir, dnsn, ebp, echo, epal, epcsv, epsn, erase, etsn, exsn, fc, fhx, fl, foreach, ft, fw, gal, gbp, gc, gci, gcm, gcs, gdr, ghy, gi, gjb, gl, gm, gmo, gp, gps, gpv, group, gsn, gsnp, gsv, gu, gv, gwmi, h, history, icm, iex, ihy, ii, ipal, ipcsv, ipmo, ipsn, irm, ise, iwmi, iwr, kill, lp, ls, man, md, measure, mi, mount, move, mp, mv, nal, ndr, ni, nmo, npssc, nsn, nv, ogv, oh, popd, ps, pushd, pwd, r, rbp, rcjb, rcsn, rd, rdr, ren, ri, rjb, rm, rmdir, rmo, rni, rnp, rp, rsn, rsnp, rujb, rv, rvpa, rwmi, sajb, sal, saps, sasv, sbp, sc, select, set, shcm, si, sl, sleep, sls, sort, sp, spjb, spps, spsv, start, sujb, sv, swmi, tee, trcm, type, wget, where, wjb, write`

The only potential sticking point I could see is that `foreach` is both an alias to `ForEach-Object` and a language statement.

hmmwhatsthisdo:
This PR adds a list of default aliases pulled from two versions of PowerShell:
* Windows PowerShell 5.1, using the following expression:
  ```powershell
  (gal | ? Name -notin '?','%' | % name | % {"'$_'"}) -join ', '
  ```
* PowerShell Core, based on the list defined within [`[System.Management.Automation.Runspaces.InitialSessionState]::BuiltInAliases`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/InitialSessionState.cs#L4592) in the PowerShell Core source.

Resolves #1338.

### DIFF
--- a/src/languages/powershell.js
+++ b/src/languages/powershell.js
@@ -39,8 +39,9 @@ export default function(hljs){
     keyword: 'if else foreach return do while until elseif begin for trap data dynamicparam ' +
     'end break throw param continue finally in switch exit filter try process catch ' +
     'hidden static parameter',
+    // "echo" relevance has been set to 0 to avoid auto-detect conflicts with shell transcripts
     built_in: 'ac asnp cat cd CFS chdir clc clear clhy cli clp cls clv cnsn compare copy cp ' +
-    'cpi cpp curl cvpa dbp del diff dir dnsn ebp echo epal epcsv epsn erase etsn exsn fc fhx ' +
+    'cpi cpp curl cvpa dbp del diff dir dnsn ebp echo|0 epal epcsv epsn erase etsn exsn fc fhx ' +
     'fl ft fw gal gbp gc gcb gci gcm gcs gdr gerr ghy gi gin gjb gl gm gmo gp gps gpv group ' +
     'gsn gsnp gsv gtz gu gv gwmi h history icm iex ihy ii ipal ipcsv ipmo ipsn irm ise iwmi ' +
     'iwr kill lp ls man md measure mi mount move mp mv nal ndr ni nmo npssc nsn nv ogv oh ' +
--- a/src/languages/powershell.js
+++ b/src/languages/powershell.js
@@ -42,6 +42,23 @@ export default function(hljs){
     // TODO: 'validate[A-Z]+' can't work in keywords
   };
 
+  // Default aliases as defined in Windows PowerShell 5.1 (with exception of %/?)
+  var ALIASES = [
+    'ac', 'asnp', 'cat', 'cd', 'CFS', 'chdir', 'clc', 'clear', 'clhy', 'cli', 'clp', 'cls', 
+    'clv', 'cnsn', 'compare', 'copy', 'cp', 'cpi', 'cpp', 'curl', 'cvpa', 'dbp', 'del', 
+    'diff', 'dir', 'dnsn', 'ebp', 'echo', 'epal', 'epcsv', 'epsn', 'erase', 'etsn', 'exsn', 
+    'fc', 'fhx', 'fl', 'foreach', 'ft', 'fw', 'gal', 'gbp', 'gc', 'gci', 'gcm', 'gcs', 'gdr', 
+    'ghy', 'gi', 'gjb', 'gl', 'gm', 'gmo', 'gp', 'gps', 'gpv', 'group', 'gsn', 'gsnp', 'gsv', 
+    'gu', 'gv', 'gwmi', 'h', 'history', 'icm', 'iex', 'ihy', 'ii', 'ipal', 'ipcsv', 'ipmo', 
+    'ipsn', 'irm', 'ise', 'iwmi', 'iwr', 'kill', 'lp', 'ls', 'man', 'md', 'measure', 'mi', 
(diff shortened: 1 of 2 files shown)


yyyc514:
Nice!  Thanks for the contribution!  But is there some reason these couldn't simply be added to `keywords`?  If so that'd be better.

src/languages/powershell.js
```diff
@@ -116,6 +133,13 @@ export default function(hljs){
     ]
   };
 
+  var PS_ALIAS = {
+    className: 'built_in',
+    variants: [
```
yyyc514: I'm guessing the error is because you have no `begin` match here... but again I'm not sure why this is even needed and the keywords can't simply be added at the top level.

src/languages/powershell.js
```diff
@@ -42,6 +42,23 @@ export default function(hljs){
     // TODO: 'validate[A-Z]+' can't work in keywords
```
yyyc514: IE, you'd add a `built_in` key here...

hmmwhatsthisdo:
> is there some reason these couldn't simply be added directly to `keywords`? If so that'd be better.

@yyyc514 Within the language, aliases are analogous/interchangeable with cmdlet calls. There's a few spots where the control statements defined under `KEYWORDS` can't be used (the exact cases escape me, but nested parenthetical expressions might be one of them), but cmdlets/aliases can.

I could convert the list to a regex and add it as a variant of `CMDLETS` if you'd prefer that instead.

yyyc514:
Keywords are already scoped to the mode, they aren't global.  And CMDLETS is already part of GENTLEMANS whatever _at the top level_, so is PS_ALIAS as you tried in the original patch.

Adding them to `keywords` would be equivalent to your original approach and also equivalent to what you just suggested unless I'm missing something.  When `keywords` can be used instead of a rule its preferred as its' much simple to work with.

yyyc514:
> There's a few spots where the control statements defined under KEYWORDS can't be used, but cmdlets/aliases can.

Even if we wanted to handle this I don't see why it couldn't be done with `keywords`.

src/languages/powershell.js
hmmwhatsthisdo: So, something like this?
```js
  var KEYWORDS = {
   keyword: '...',
   built_in: 'ac asnp cat cd CFS chdir clc...'
  };

yyyc514: Yes.

https://highlightjs.readthedocs.io/en/latest/reference.html#keywords

yyyc514: I'm not exactly sure what PS_METHODS is trying to do.

hmmwhatsthisdo: > Yes.
> 
> https://highlightjs.readthedocs.io/en/latest/reference.html#keywords

I see! I'll make the appropriate changes.

yyyc514:
Any chances you could you add a markup test with some examples of using aliases?

Also a changelog entry, please?

yyyc514:
I might have to dig into the auto-detect conflicts unless you want to.  The problem is that now powershell looks too much like shell and the shell example is being mis-detected.

hmmwhatsthisdo:
> Any chances you could you add a markup test with some examples of using aliases?

I can certainly try to do so - I don't have a full dev env set up for this project so running test cases might be a bit difficult.

> Also a changelog entry, please?

Sure!

> I might have to dig into the auto-detect conflicts unless you want to.

I'm assuming this is because `echo` is now a `built_in`. I haven't really seen this one used that often in most scripts, and I feel like most PowerShell snippets that did in fact use it would be caught with everything else. I can try setting its relevance to 0 if that's the preferred path.

yyyc514:
>  I can try setting its relevance to 0 if that's the preferred path.

That would likely make the tests green and would be the simplest way forward - I'd suggest adding a comment to explain why.  Auto-detect tests are a mess when dealing with VERY similar languages/grammars.

>  I don't have a full dev env set up 

If you have `node` and `npm` it should be pretty trivial.

hmmwhatsthisdo:
> That would likely make the tests green and would be the simplest way forward - I'd suggest adding a comment to explain why.

Done. I've also added a changelog entry as requested.

hmmwhatsthisdo:
I'm looking at the latest Travis output and the test-case failure makes sense (I think) but I'm not sure what the appropriate method would be for fixing it. I might have to take another swing later.

yyyc514:
You may need to add a word boundary on `lexemes: /-?[A-z\.\-]+/,`

yyyc514:
@hmmwhatsthisdo Thanks so much for getting this done!

Issue #2269, opened 2019-11-10
CSS, SASS, LESS, STYLUS highlighting differences.


w3suli:
![img](https://user-images.githubusercontent.com/20889555/68548660-c7a88a80-03ef-11ea-9870-34e27afcbcec.png)

Sass bug list:
- attribute prefix -I like it but different from css and less - unfortunately bad (-webkit-animation-name:)
- for the selection attribute, the string is not highlighted (a[href*="example" I])
- lack of function highlighting (grayscale(0.5))
- at-rules with highlighting difference (@keyframes "from" "to")
- lack of selection of @media attributes ("(orientation: landscape)" or "(min-width: 600px)")
- other differences: ":lang(de)"

Less bug list:
- for the selection attribute, the string is not highlighted (a[href*="example" I])
- lack of pseudo selectors highlighting (q::after)
- lack of function highlighting (grayscale(0.5))
- lack of selection of @media attributes ("not", "only" and "and")
- other differences: ":lang(de)"

Stylus bug list:
- attribute prefix -I like it but different from css and less - unfortunately bad (-webkit-animation-name:)
- lack of attribute selectors highlighting (a[href*="example" I])
- lack of pseudo selectors highlighting (q::after)
- lack of function highlighting (grayscale(0.5))
- lack of at-rules highlighting ("@font-face" "@keyframes", "from", "to")
- lack of media at-rules highlighting ("@media", "not", "only" and "and")
- lack of selection of @media attributes ("(orientation: landscape)" or "(min-width: 600px)")
- other differences: ":lang(de)"

There is a lot of difference. If I have enough time, I'll make examples one by one. I welcome any help. Thanks in advance for all your help!

yyyc514:
>  I'll make examples one by one

Of each bug with each grammar?  I'm not sure that's a great use of time.  What really should happen here (if someone has a LOT of time) is to look at how to consolidate all of these, reuse some matchers, merge grammars where possible, etc.  That might start to look like more of a ground-level rewrite - and then this exact list of bugs wouldn't be useful at all.  It's be more useful to rebuild things and then see where we stood.

It's possible Less/CSS/SCSS could likely become a single grammar... (or very close)... and SASS and Stylus are very similar also.

Optimally someone (with a lot of time) needs to dig in here and think about this at a very high level.  If we can do something like build 5 on 2 core grammars... or 5 on 1... then keeping the features in sync with future improvements becomes much simpler.

[...]

PR #2425, opened 2020-02-24
(enh) Improve CSS consistency across CSS-like grammars

yyyc514:
Begins work on #2269


yyyc514:
Would `csslib` be better naming?

src/languages/css.js
```diff
@@ -21,7 +27,10 @@ export default function(hljs) {
         ]
       }
     ]
-  }
+  };
+  var VENDOR_PREFIX= {
+    begin: /-(webkit|moz|ms|o)-/
+  };
```
egor-rogov: A candidate for `css-shared`.

src/languages/scss.js
```diff
@@ -61,13 +66,18 @@ export default function(hljs) {
...
       {
         className: 'selector-pseudo',
-        begin: ':(visited|valid|root|right|required|read-write|read-only|out-range|optional|only-of-type|only-child|nth-of-type|nth-last-of-type|nth-last-child|nth-child|not|link|left|last-of-type|last-child|lang|invalid|indeterminate|in-range|hover|focus|first-of-type|first-line|first-letter|first-child|first|enabled|empty|disabled|default|checked|before|after|active)'
+        begin: ':(' + PSEUDO_CLASSES.join('|') + ')'
       },
       {
         className: 'selector-pseudo',
-        begin: '::(after|before|choices|first-letter|first-line|repeat-index|repeat-item|selection|value)'
+        begin: '::(' + PSEUDO_ELEMENTS.join('|') + ')'
       },
```
allejo: While not a CSS best practice, does Sass actually prevent using `:` for pseudo-elements? Looks to me like only the Sass grammar differentiates these.

```js
      {
        className: 'selector-pseudo',
        begin: ':(:)?(' + css_shared.PSEUDO_SELECTORS.join('|') + ')'
      },
```

Or would this not be in the scope of this PR?

src/languages/stylus.js
```diff
@@ -384,12 +375,14 @@ export default function(hljs) {
 
       // psuedo selectors
       {
+        className: "selector-pseudo",
```
allejo:
```suggestion
        className: 'selector-pseudo',
```

src/languages/less.js
```diff
@@ -124,17 +131,26 @@ export default function(hljs) {
       IDENT_MODE('selector-class', '\\.' + INTERP_IDENT_RE, 0),
       IDENT_MODE('selector-tag',  '&', 0),
       {className: 'selector-attr', begin: '\\[', end: '\\]'},
-      {className: 'selector-pseudo', begin: /:(:)?[a-zA-Z0-9\_\-\+\(\)"'.]+/},
+      {className: 'selector-pseudo', begin: /:(:)?[a-zA-Z0-9\_\-\+"'.]+/},
```
allejo: Can this regex also be shared? The CSS grammar is using this on [line 91](https://github.com/highlightjs/highlight.js/pull/2425/files#diff-335a561957078af90e8fbf6c8dfae8a7R91).

src/languages/css.js
```diff
@@ -77,7 +88,12 @@ export default function(hljs) {
       },
       {
         className: 'selector-pseudo',
-        begin: /:(:)?[a-zA-Z0-9\_\-\+\(\)"'.]+/
+        begin: /:(:)?[a-zA-Z0-9\_\-\+"'.]+/
```
allejo: Is there a reason the CSS grammar isn't using the new shared values?

src/languages/css.js
yyyc514: Yeah, that's why I created a branch for this.  It's going to take some time to work thru it all.

src/languages/scss.js
yyyc514: They are different things according to CSS, so since SCSS broke them out I went ahead and did the same and allowed it both ways.  The grammars that mix them are being slightly LAZY (my first thought).  Sidenote: keep in mind SASS is not SCSS.

_That said I don't necessarily mind sightly lazy._  Much of our parsing is based on what is expedient, not what is 100% correct.  Trying to parse these languages 100% correctly when parsing then 80% correctly will highlight them pretty  much exactly the same is often a battle with no real upsides (only resulting in increased complexity and bugs and maintenance time).

This first pass here is largely about consistency though not perfection in the code.  **First we need good tests, then we can refactor freely.**

src/languages/less.js
yyyc514: Agree, maybe I should have specified this is work in progress. :-)  Pulling these tiny things out is likely the last step since they are the most likely to have tiny differences hidden.

src/languages/css.js
yyyc514: I'm still not 100% convinced it should. Right now because CSS is a very specific grammar (more rules than the other variants) it can live with a ruleset based grammar instead of a list of attributes.  _That means that it gets updates "for free"._ 

**We'll never have an issue filed because CSS is mixing an attribute or missing a pseudo-selector.**  Less maintenance.  This is the same approach we've recommended other places (see the issue regarding __KEYWORDS in C++).  When the lists can get crazy long and be hard to maintain over time (or likely to age poorly) we prefer patterns.

**The BIG difference here though is that we already have to maintain the patterns for the other grammars though.**  So that might be enough to push it over the top and say just use the lists... welcome to thoughts.

yyyc514:
Take note here we're merging to `css_consistency`, not `master`... I'm ok the branch run on it's own for a bit since I expected this work to involve a lot of small PRs - plus that way they are easier to review than one much massive PR - which is where I was afraid we might be heading otherwise.

src/languages/css.js
allejo: I'm ok with either choice, I'd just push for consistency. For example, GitHub highlights only known pseudo selectors but it's consistent with highlighting `content` in CSS and Scss

**CSS**

```css
.hello::before {
  border: 1px solid;
  content: '';
  display: block;
}

.hello:toast {
  content: '';
  toast: '';
}
```

**Scss**

```scss
.hello::before {
  border: 1px solid;
  content: '';
  display: block;
}

.hello:toast {
  content: '';
  toast: '';
}
```

allejo:
What do you see being the end goal for the `css_consistency` branch? Like what are the goals for it so smaller PRs can be made against it.

src/languages/css.js
yyyc514: > GitHub highlights only known pseudo selectors

Perhaps cause that's a much shorter list than attribute names?

>I'm ok with either choice, I'd just push for consistency.

Well the most consistent way would be to use the same lists in ALL 4 of the grammars. At least when update happen they would happen to all 4 at the same time.

Perhaps first someone could see if there is ANYWAY to avoid an attribute list in some of the cleaner grammars, but I think it's going to be difficult.

>  but it's consistent with highlighting content in CSS and Scss

Highlighting content differently in interesting.  We could do that too, but we'd just have to decide which CSS class was appropriate.  I do see how you could argue it's technically different than other attributes.

yyyc514:
> What do you see being the end goal for the css_consistency branch? Like what are the goals for it so smaller PRs can be made against it.

Well the end goal is merging to master.  Preferably rebase and merge if we keep the commit history clean, which I'm going to try to do.  Having a separate branch and smaller PRs helps with that since the individual PRs can still be squashed and still the full history will remain in the parent branch.

If you're wanting to help you can look at the huge checklist in the related issue thread.

w3suli:
MISSED PSEUDO SELECTORS

:playing
:paused

:-moz-any
:-moz-any-link [Matches :link and :visited]
:-moz-broken
:-moz-drag-over
:-moz-first-node
:-moz-focusring
:-moz-full-screen
:-moz-full-screen-ancestor
:-moz-handler-blocked
:-moz-handler-crashed
:-moz-handler-disabled
:-moz-last-node
:-moz-list-bullet
:-moz-list-number
:-moz-loading
:-moz-locale-dir(ltr)
:-moz-locale-dir(rtl)
:-moz-lwtheme
:-moz-lwtheme-brighttext
:-moz-lwtheme-darktext
:-moz-native-anonymous
:-moz-only-whitespace
:-moz-placeholder
:-moz-read-only
:-moz-read-write
:-moz-submit-invalid
:-moz-suppressed
:-moz-system-metric(images-in-menus) 
:-moz-system-metric(mac-graphite-theme) 
:-moz-system-metric(scrollbar-end-backward) 
:-moz-system-metric(scrollbar-end-forward) 
:-moz-system-metric(scrollbar-start-backward) 
:-moz-system-metric(scrollbar-start-forward) 
:-moz-system-metric(scrollbar-thumb-proportional) 
:-moz-system-metric(touch-enabled) 
:-moz-system-metric(windows-default-theme)  
:-moz-tree-cell
:-moz-tree-cell-text
:-moz-tree-cell-text(hover)
:-moz-tree-checkbox
:-moz-tree-column
:-moz-tree-drop-feedback
:-moz-tree-image
:-moz-tree-indentation
:-moz-tree-line
:-moz-tree-progressmeter
:-moz-tree-row
:-moz-tree-row(hover)
:-moz-tree-separator
:-moz-tree-twisty
:-moz-ui-invalid
:-moz-ui-valid
:-moz-user-disabled
:-moz-window-inactive
:-moz-system-metric(images-in-menus) 
:-moz-system-metric(mac-graphite-theme) 
:-moz-system-metric(scrollbar-end-backward) 
:-moz-system-metric(scrollbar-end-forward) 
:-moz-system-metric(scrollbar-start-backward) 
:-moz-system-metric(scrollbar-start-forward) 
:-moz-system-metric(scrollbar-thumb-proportional) 
:-moz-system-metric(touch-enabled) 
:-moz-system-metric(windows-default-theme)

MISSED PSEUDO ELEMENTS
::-moz-anonymous-block
::-moz-anonymous-positioned-block
::-moz-canvas
::-moz-color-swatch
::-moz-cell-content
::-moz-focus-inner
::-moz-focus-outer
::-moz-inline-table
::-moz-page
::-moz-page-sequence
::-moz-pagebreak
::-moz-pagecontent
::-moz-placeholder
::-moz-progress-bar
::-moz-range-progress
::-moz-range-thumb
::-moz-range-track
::-moz-scrolled-canvas
::-moz-scrolled-content
::-moz-scrolled-page-sequence
::-moz-selection
::-moz-svg-foreign-content
::-moz-table
::-moz-table-cell
::-moz-table-column
::-moz-table-column-group
::-moz-table-outer
::-moz-table-row
::-moz-table-row-group
::-moz-viewport
::-moz-viewport-scroll
::-moz-xul-anonymous-block

::-ms-browse
::-ms-check
::-ms-clear
::-ms-expand
::-ms-fill
::-ms-fill-lower
::-ms-fill-upper
::-ms-reveal
::-ms-thumb
::-ms-ticks-after
::-ms-ticks-before
::-ms-tooltip
::-ms-track
::-ms-value

::-webkit-progress-bar
::-webkit-progress-value
::-webkit-slider-runnable-track
::-webkit-slider-thumb

::-webkit-scrollbar
::-webkit-scrollbar-button
::-webkit-scrollbar-thumb
::-webkit-scrollbar-track
::-webkit-scrollbar-track-piece
::-webkit-scrollbar-corner
::-webkit-resizer

PR #2426, opened 2020-02-25
Adding direct support for Cypher

obfischer:


yyyc514:
We are no longer merging languages into the core library because of lack of developer time to maintain and support them over the long-term.  We already have a repository for Cypher and it's easy for anyone to use 3rd party libraries with the main library.

PR #2427, opened 2020-02-25
(chore) update language contribution guidelines

### LINKED ISSUE #2055: Update the language contribution documentation and procedure

yyyc514:
Closes #2055

### DIFF
--- a/docs/language-contribution.rst
+++ b/docs/language-contribution.rst
@@ -1,17 +1,25 @@
 Language contributor checklist
 ==============================
 
-1. Put language definition into a .js file
+1. Read ``extra/3RD_PARTY_QUICK_START.md``
 ------------------------------------------
 
+It contains rough high-level steps for creating a 3rd party language grammar for Highlight.js.
+
+
+2. Create a language grammar definition file
+--------------------------------------------
+
 The file defines a function accepting a reference to the library and returning a language object.
-The library parameter is useful to access common modes and regexps. You should not immediately call this function,
-this is done during the build process and details differ for different build targets.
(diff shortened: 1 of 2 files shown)


yyyc514:
Further changes welcome.  This is all still rough and could be further improved but the key goal here was to minimal make it stop suggesting that we'll merge in anything and instead mention the new 3rd party process.

docs/language-contribution.rst
```diff
@@ -57,21 +66,27 @@ Take inspiration from other languages in ``test/detect/`` and read
...
 Describe shortly names of all meaningful modes used in your language definition.
 
+Note: If you use custom classes please be aware that all the build-in themes
+are not going to support your custom classes and you should likely also
+distribute your own custom theme.
+
 
-5. Add yourself to AUTHORS.*.txt and CHANGES.md
------------------------------------------------
+6. Request a repository at the ``highlightjs` organization
```
egor-rogov: One closing backtick is missing?

Issue #2429, opened 2020-02-25
Discuss: Keyword boundaries


yyyc514:
**Is your request related to a specific problem you're having?**

https://github.com/highlightjs/highlight.js/issues/2420
https://github.com/highlightjs/highlight.js/issues/2428

**The solution you'd prefer / feature you'd like to see added...**

I think perhaps languages need to be able to define their own keyword boundaries... we are considering adding `\b|[space]` to fix an issue with Clojure (because it allows `-` in keywords)... and then really (most of the time) keywords shouldn't be followed by `.` (then it's a method dispatch, not a keyword). Also, they really shouldn't be proceeded by `.` either (method name), but that would require negative look-behind [though eventually we'll get that.

So as I'm considering adding these things into the CORE highlighter as keyword boundaries it strikes me that I'm making some big assumptions and adding grammar specific things to the core, which seems quite bad.  If we support languages as diverse as [Brainfuck](https://en.wikipedia.org/wiki/Brainfuck) then we should have as little "generic" knowledge of languages as possible baked into the core highlighter.

Even `\b` as a separator for ALL languages is wrong as we're starting to see cracks with that already with Clojure, etc.

As languages can already configure `lexemes` I think they also should be able to configure boundaries.

yyyc514:
In the recent past the parser assumed any expression like `.keyword` or `keyword.` could not be a keyword... recent parser changes broken this (see #2428) but I'm not sure that was correct core behavior to begin with (though it might be a reasonable default).  Definitely there were issues with how it was implemented.

yyyc514:
Here is my issue with the prior behavior:

You write:

```js
contains: [
  {beginKeywords: "mud dirt gook"}
]
```

But what you really get is:

```js
contains: [
  { begin: /\.(mud|dirt|gook)\./ }, // first eat up any .keywords
  {beginKeywords: "mud dirt gook"
]
```

That solves the problem but it's implicit vs explicit and I don't think it's 100% behavior is predicable.  There is no *obvious* reason why this shouldn't work:

```js
contains: [
  { beginKeywords: "mud dirt gook" },
  { begin: ".dirt", className: meta" } // the .dirt macro has special meaning
]
```

Trial and error might teach you to put it first, but it's not clear at all WHY that would be so from reading the grammar itself.

yyyc514:
[...]

PR #2434, opened 2020-02-27
(bug) Fix `beginKeywords` to ignore . matches

### LINKED ISSUE #2428: (java) `.class` should not detect as keyword
**Describe the issue**
when have `class` key word and `=` in the arguments of a function, it will break the highlighting. A minimal code to reproduce this issue on 9.18.1 as follow:

```
public class T {
  public void a() {
    test( A.class, "a=b");
  }
  @Test
  public void b() {
    // ...
  }
}
```

while the code example works fine for 9.15.1.

9.18.1: http://jsfiddle.net/p9r1yd36/
9.15.1: http://jsfiddle.net/p9r1yd36/1/

**Expected behavior**
https://imgur.com/RTUfqny

**Actual behavior**
https://imgur.com/YU7ZDcj

yyyc514:
Related: https://github.com/highlightjs/highlight.js/issues/2429 Resolves #2428 

Attempts to restore as little as possible of the previous `beginKeywords` magic behavior as necessary just to prevent `.keyword` or `keyword.` from matching.  This is done via an internal `onBegin` callback that essentially ignores the match if a `.` is found.

**What behavior does this NOT restore?**  A failed match will NOT consume any content from the input stream, ie it will not advance the cursor.

Previously:

```
       v Parse cursor is here after failure to match.
A.class
```

Now:

```
  v Parse cursor is here (or earlier) after failure to match.
A.class
```

Ex:

```js
    contains: [
      { beginKeywords: "class" },
      // this rule will still match class if the above rules fails to (because of a `.`)
      { begin: "class", className: "found" }
    ]
```

Also previously the match would actually start at the `.` (when one was present), which would prevent any other rules from matching if there was a failed keyword match - that is no longer the case.  The match now starts with the keyword itself and then looks backwards to see if there was a preceding `.` or not.

---

**What else might still be an issue...**

This new behavior means that when `beginKeywords` fails to match `keywords` now has a chance to.  I think this is new behavior, and correct behavior.  **No other place in the parser does a rule silently consume content for a non-match**, and I don't think `beginKeywords` should do so either.

---

- `MultiRegex` is just a re-factoring of the behavior that was there before into a class, allowing us to work with a large group of regexes as a single search unit.
- `ResumableMultiRegex` wraps multiple MultiRegex objects to allow resuming a multi-regex search at the same cursor position, allowing for the possibility of multiple regexes matching, instead of just the first.

See comments in source for more details

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 New languages:
 
-- (php-template) Explicit language to detect PHP templates (vs xml) [Josh Goebel][]
+- add(php-template) Explicit language to detect PHP templates (vs xml) [Josh Goebel][]
 - enh(python) Added `python-repl` for Python REPL sessions
 
 New themes:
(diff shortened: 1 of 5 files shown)


yyyc514:
I think if you understand what's going on in `highlight.js` you understand the gist of it.  I wish I could think of better names though.  It's weird now because the MultiRegex has not only an `index` now but also a "which regex" to begin searching with index (which I called `startAt`), since now it's resumable at the same position to later regexes in the stack can be tried.

Wrapping a bunch of MultiRegexes was just the simplest way I could think of to accomplish the goal... but if someone knows a simpler way... the external API is simple enough:

- exec(string)
- matchAt
- index

And actually this now allows for lazy-compiling of multi regexes... so for some grammars with a ton of complex modes it might be possible that some of those regex never even have to be compiled if those modes are never actually hit - so that's a nice small win.

yyyc514:
Tried to improve naming.

egor-rogov:
Yes, it's a bit clearer this way.

PR #2439, opened 2020-02-28
Improve bash support [no squash]

### LINKED ISSUE #1504: (bash) $((Base#number)) variable syntax issue
In bash, you can specify the base of a number in an arithmetic expression with the construct B#n where the value of that construct is ``n`` is base ``B`` (B can be anything from 2 to 32, or can be specified with a ``$variable``).

The issue is that in highlight.js, in this configuration, # is considered as the start of a comment. Here are three examples use cases:
```bash
echo $(( 16#deadbeef / 1003 ))
```
```bash
yumi=deadbeef
echo $(( 16#$yumi / 1003 ))
```
```bash
B=20
yumi=deadbeef
echo $(( $B#$yumi / 1003 ))
```

What I see in hightlight js: everything after the # is a comment.

What I expect: ``16#`` and ``$B#`` should be considered as a numeric literal or something similar (at least when inside the ``$((...))`` arithmetic expression construct); And stuff after it should not be considered a comment. Typically, I expect normal syntax highlight to work after ``16#``.

Some tests gives me:

```node
> hljs = require("highlight.js")
> hljs.highlightAuto("echo $(( 16#deadbeef * 10))").value
'<span class="hljs-built_in">echo</span> $(( 16<span class="hljs-comment">#deadbeef * 10))</span>'
```
Note: this is the original snippet of code that prompted me to open this issue:
```bash
#!/bin/sh

#Sets the current system date to one hour before now.
#WARNING!! don't use after midnight.
curhour=$(date +'%H')
curdate=$((10#$curhour - 1)):$(date +'%M:%S')
date --set="$curdate"
```

### LINKED ISSUE #2099: (bash) Nested strings are not parsed correctly
Example:

```bash
SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
TLS_DIR="$SCRIPT_DIR/../src/main/resources/tls"
ROOT_DIR="$SCRIPT_DIR/.."
```

---

This is really two things.

- [x] String within string (the posted issue)
- [x] Variable within variable (issue below, from #1978)

yyyc514:
Closes #1504
Closes #2099


src/languages/bash.js
```diff
@@ -7,35 +7,66 @@ Category: common
 */
 
 export default function(hljs) {
-  var VAR = {
+  const VAR = {};
+  const BRACKET_VAR = {
```
egor-rogov: Looks like `man bash` tends to use "brackets" for `[ ]` and "braces" for `{ }`. So probably better use `BRACE_VAR` here.

PR #2444, opened 2020-03-01
enh(xml) deprecate ActionScript inside script tags

### LINKED ISSUE #1212: Discuss: Restricting language subset used for sublanguage detection
When limiting auto highlight to a language subset, it only works for top-level languages. So when parsing some HTML with JavaScript inside with:

``` js
hljs.highlightAuto(
  code,
  [ 'html', 'css', 'javascript' ]
).value
```

the JavaScript may still be interpreted as ActionScript, for example. 😕

yyyc514:
Resolves #1212.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ Core Changes:
 
 Language Improvements:
 
+- enh(xml) deprecate ActionScript inside script tags (#2444) [Josh Goebel][]
 - fix(javascript) prevent get/set variables conflicting with keywords (#2440) [Josh Goebel][]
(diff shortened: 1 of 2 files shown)


yyyc514:
> Shouldn't it be listed in VERSION_10_BREAKING?

I actually don't view this as a breaking change in the same way as the other things - we've done similar things (removing vbscript from the same list) in minor releases. I view this as more of a maintenance item.

> And btw it would be useful to have a link to that list in CHANGES.

I haven't decided how to do that yet.  Maybe - or I might end up moving the whole breaking changes content into the changelog before the actual release. If not, yes a link would make sense.

PR #2445, opened 2020-03-02
Corrected minor error in README

ohepworthbell:
Docs in README incorrectly advised users to import the core in `npm` as:

```javascript
import hljs from 'highlight.js/lib/core';
```

Rather than the correct location as:

```javascript
import hljs from 'highlight.js/lib/highlight';
```

As shown at:
https://highlightjs.org/usage/

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ const highlightedCode = hljs.highlightAuto('<span>Hello World!</span>').value
 
 ```js
 // require the highlight.js library without languages
-const hljs = require("highlight.js/lib/core");
+const hljs = require("highlight.js/lib/highlight");
 // separately require languages
 hljs.registerLanguage('html', require('highlight.js/lib/languages/html'));
 hljs.registerLanguage('sql', require('highlight.js/lib/languages/sql'));
@@ -184,7 +184,7 @@ import hljs from 'highlight.js';
 The default import imports all languages! Therefore it is likely to be more efficient to import only the library and the languages you need:
 
 ```js
-import hljs from 'highlight.js/lib/core';
+import hljs from 'highlight.js/lib/highlight';
 import javascript from 'highlight.js/lib/languages/javascript';
 hljs.registerLanguage('javascript', javascript);
 ```
@@ -192,7 +192,7 @@ hljs.registerLanguage('javascript', javascript);
 To set the syntax highlighting style, if your build tool processes CSS from your JavaScript entry point, you can import the stylesheet directly into your CommonJS-module:
 
 ```js
-import hljs from 'highlight.js/lib/core';
+import hljs from 'highlight.js/lib/highlight';
 import 'highlight.js/styles/github.css';
 ```
 


ohepworthbell:
Note to reviewer: I haven't updated the `CHANGES.md` file, though as the change is a typo correction to the `README` (and no functionality changes to actual code) I would assume this is probably unnecessary?

egor-rogov:
Hi @ohepworthbell,
this change was introduced in b64b0081be232715b1363b8710b8c78fc4bb340e, but not yet released. This is how you'll be supposed to import the library in version 10.0. So no error here.

ohepworthbell:
OK sorry - my bad.

Though that's still peculiar: I installed using:

```
npm install highlight.js --save-dev
```

And the installed version is `9.18.1` according to the `packages.json` (both in the root `package.json` and `node_modules/highlight.js/package.json`)... yet my local `/lib/` folder has `highlight.js` rather than `core.js`.

yyyc514:
> And the installed version is 9.18.1 according to the packages.json 

Correct, 9.18.1 is `highlight.js`.  10.0 (unreleased) is now `core.js`.

If you view the README for version 9.18.1 you'll see it is correct.

yyyc514:
@ohepworthbell What led you to this in the first place? Your not the first person to mention this.
Is it simply because so many people treat repository READMEs as "canonical" without taking into account that the repositories are always in flux and the readme might actually be for future releases?

We're probably in a better shape than many projects since we have good documentation in several other places (the website, readthedocs, etc)... but I'm hard pressed to see how we could solve this without purposely having held this change until closer to release. I'm interested though since this is the type of thing that one could easily imagine coming up again in the future.

CC @egor-rogov

ohepworthbell:
@yyyc514 sorry for the delay. Yes - I assumed that the README was contemporary and not a future release, and didn't pause for thought (entirely my bad - so apologies for wasting your time there).

I guess a simplest way around this in the future could be to specify a version number at the top of the README for clarification? It's easy to miss/forget to check the branch (especially as some projects don't branch their versions - and people tend to assume based on what is common, not what is correct), so many people will likely skip straight to the README and assume that it's for the current release, unless otherwise stated.

A lot of major projects typically header their README/tutorials with something like:

> ### Docs for v10.0.0
> For version 9.8.1 (current release) refer to [here](http://example.com)

Or perhaps even sub-notes throughout the documentation, like:

> The default import imports all languages! Therefore it is likely to be more efficient to import only the library and the languages you need:
> ```javascript
> // For version 10.^ (unstable)
> import hljs from 'highlight.js/lib/core';
>
> // OR for version 9.^ (current release)
> import hljs from 'highlight.js/lib/highlight';
> ```

I just assumed that - given no version was specified - the README was contemporary, not for a future release. And I guess there are perhaps many like me who didn't see this immediately spelled out to them, and therefore didn't immediately consider it.

I think you are right on your other points too. As you allude to, external/off-site documentation for other repos are often non-existent, forgotten, or not up-to-date. Many other projects - especially smaller ones - don't have dedicated websites at all, or certainly not with completely up-to-date information. As such, many developers resort to relying on a README file on GitHub as their only source of scant information, and assume that is canonical.

I guess you're also somewhat victims of your own success here... which you unfortunately can't do much about. It's rare to be as organised as this repo is. Again, people work from the assumptions they expect - i.e. what is common - not what is correct. We expect disorder, and so we operate like nothing is organised.

But yeah, I think just a version number at the top would probably make a big difference.

Hope that's all useful anyhow

yyyc514:
I added a one line mention at the top of the README.  Maybe it will help.

PR #2451, opened 2020-03-08
(docs) Correct the core lib path lib/core to lib/highlight

Menci:
There's no `lib/core` in highlight.js's NPM package. I think it's wrong. Maybe it should be `lib/highlight`.


yyyc514:
See the docs for 9.18.x, they are correct.  The repository reflects version 10, which is not released yet.  The new docs are correct if you build the source from master.

yyyc514:
Added a small note to top of README.

PR #2456, opened 2020-03-10
update docs to reflect new npm location

yyyc514:


yyyc514:
Actually now that I was able to simply move the existing npm package into the organization I'm not sure we need to do this... we'll just keep the global name... one less thing for people to have to change.

And we still get the admin benefits of it being in the organization.

PR #2466, opened 2020-03-22
Kusto language

kopytjuk:
Add kusto language support to `extra/`.

Fix path handling in `tools/lib/language.js` on windows machines.


yyyc514:
We don't merge languages into core anymore because of lack of developer time to maintain them in the long-run.  If you make your own repository though you can make a PR that adds a link to it so those who'd like to use your language as a 3rd party plugin can easily do see.

yyyc514:
> Fix path handling in tools/lib/language.js on windows machines.

If there is something here though we'd probably welcome a PR for a fix just for that.

tools/lib/language.js
```diff
@@ -63,8 +63,7 @@ class Language {
   }
 
   static fromFile(filename) {
-    if (filename.startsWith("/") || filename.startsWith("."))
-    {
+    if (filename.startsWith("/") || filename.startsWith(".") || process.platform === "win32") {
```
yyyc514: This doesn't look correct though, what exactly is the problem you were trying to solve here?

kopytjuk: Windows file paths start with `C:\Users\...`  - without this platform check the file name becomes: 

```
/src/languages/C:\Users\...
```

yyyc514: Ah, that's not a correct fix though, that will cause it to break elsewhere.

You'd need to do something like this:

https://stackoverflow.com/questions/21698906/how-to-check-if-a-path-is-absolute-or-relative

PR #2469, opened 2020-03-24
Fix windows absolute paths.

kopytjuk:
Fix windows absolute paths (building of extra languages breaks on Windows 10).

### DIFF
--- a/tools/lib/language.js
+++ b/tools/lib/language.js
@@ -63,7 +63,7 @@ class Language {
   }
 
   static fromFile(filename) {
-    if (filename.startsWith("/") || filename.startsWith("."))
+    if ( path.isAbsolute(filename) || filename.startsWith(".") )
     {
       var file = filename
     } else {
--- a/tools/lib/language.js
+++ b/tools/lib/language.js
@@ -63,7 +63,7 @@ class Language {
   }
 
   static fromFile(filename) {
-    if ( path.isAbsolute(filename) || filename.startsWith(".") )
+    if (path.isAbsolute(filename) || filename.startsWith("."))
     {
       var file = filename
     } else {


tools/lib/language.js
```diff
@@ -63,7 +63,7 @@ class Language {
   }
 
   static fromFile(filename) {
-    if (filename.startsWith("/") || filename.startsWith("."))
+    if ( path.isAbsolute(filename) || !path.isAbsolute(filename) )
```
yyyc514: I don't think that works. `ruby.js` is not an absolute path.  I think the "." check needs to remain.  Does that break on Windows?

yyyc514: All paths will be absolute OR not so then the bottom part of the condition will never fire, am I missing something?

kopytjuk: You used `filename.startsWith("/") || filename.startsWith(".")` which conceptually the same as `path.isAbsolute(filename) || !path.isAbsolute(filename)`. The first block both is used for relative `./xyz` and absolute `/xyz/...` paths.

yyyc514: Not at all.  There are THREE concepts here:

1. absolute paths
2. dot paths (relative by nature of the dot)
3. other (NOT relative, NOT absolute)

isAbsolute will return false for 2 and 3, which is bad.

Alway try to make the minimal invasive change.  Does this not resolve your problem:

```
    if ( path.isAbsolute(filename) || filename.startsWith(".") )
```

yyyc514: Your code might work if there was an `isRelative()` but that unfortunately `NOT isAbsolute` does not equivalent to isRelative (in this specific case).

kopytjuk: I agree.

yyyc514:
Thanks so much!

PR #2481, opened 2020-04-04
Never

smaludzi:
Please add Never language highlighter to the list of supported languages.


CHANGES.md
```diff
@@ -4,6 +4,7 @@ New languages:
 
 - add(php-template) Explicit language to detect PHP templates (vs xml) [Josh Goebel][]
 - enh(python) Added `python-repl` for Python REPL sessions
+- add(never) Added Never language support
```
yyyc514, line 7: Hmm.... I suppose we could mention it here but please say "Added 3rd party Never..." language support so it's clear this is not a core language.

PR #2487, opened 2020-04-09
(yaml) improve tag support; add verbatim tags

### LINKED ISSUE #2486: (yaml) Tags should allow non-word characters
**Describe the issue**
YAML tags can include some non-word characters (e.g. period, dollar sign) but current highlighting only matches letters and underscores

YAML

**Sample Code to Reproduce**
https://jsfiddle.net/7ukmfhsv/

**Expected behavior**
yaml-highlight

pplantinga:
Resolves #2486


src/languages/yaml.js
```diff
@@ -10,6 +10,9 @@ Category: common, config
 export default function(hljs) {
   var LITERALS = 'true false yes no null';
 
+  // YAML spec allows non-reserved characters in tags
+  var YAML_TAG_RE = '[a-zA-Z_.~@#$%():;+=-?/]\\w*'
```
yyyc514, line 14: So only the first character is the exception?  That doesn't seem right?

pplantinga: Ah, you're right. This might work better
```suggestion
  var YAML_TAG_RE = '[a-zA-Z_.~@#$%():;+=-?/]+'
```

yyyc514: Are there no rules about the first characters being different? Usually there is.

yyyc514: Are spaces always required after `:` in YAML? Trying to wonder how `:` in a name isn't somewhat problematic... I guess because they are all escaped with `!`?

pplantinga: Here's a link to the yaml spec for tags: https://yaml.org/spec/1.2/spec.html#id2782090
The intention is that they be able to encode URIs. Like you suspected, mappings do require a space after the colon.

yyyc514: I'd stick with what we're talking about here unless you really want to go broader. The discussion is about local tags... did you mean to ALSO change data types?

```

ns-uri-char | ::= | “%” ns-hex-digit ns-hex-digit \| ns-word-char \| “#”\| “;” \| “/” \| “?” \| “:” \| “@” \| “&” \| “=” \| “+” \| “$” \| “,”\| “_” \| “.” \| “!” \| “~” \| “*” \| “'” \| “(” \| “)” \| “[” \| “]”
-- | -- | --



```

pplantinga: In the spec, "data types" are just tags in a special namespace. The "secondary handle" !! is used to indicate this namespace, e.g. https://pyyaml.org/wiki/PyYAMLDocumentation#yaml-tags-and-python-types

yyyc514:
Can we add a markup file with a smattering of actual examples of tags in use and add the missing characters to the regex?

pplantinga:
All characters have been added, as well as a few tests. Ready for review.

yyyc514:
See my pending review on the removal of `!!`.

pplantinga:
I can't find the pending review. Where can I see it?

yyyc514:
Do you see it now?

src/languages/yaml.js
```diff
@@ -73,14 +73,10 @@ export default function(hljs) {
...
       },
-      { // local tags
+      { // tags
         className: 'type',
         begin: '!' + YAML_TAG_RE,
       },
-      { // data type
-        className: 'type',
-        begin: '!!' + YAML_TAG_RE,
-      },
```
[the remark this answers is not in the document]
pplantinga: I took it out because the spec allows for '!' in tags, and I thought it would be confusing to leave in a pattern that never matches. I can see that leaving in the pattern could be useful, for example, if eventually different highlights were wanted for the two patterns. What if we allow '!' everywhere except the first character (after the !), would that be an appropriate solution?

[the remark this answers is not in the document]
yyyc514: You might be right. I think it's original intent was to note that technically `!!` tags are a different type of tag, but whether that's useful or not I'm not sure.   Generally when writing new grammars I try and keep them as vague as possible why still doing the job well but when maintaining old ones I tend to not remove complexity without completely understanding WHY it was added in the first place.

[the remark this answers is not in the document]
yyyc514: Plus technically I think a lot of what we're talking about are tag prefixes... The tag portion can only contain ASCII, but I don't think we're differentiating.

src/languages/yaml.js
```diff
@@ -10,6 +10,10 @@ Category: common, config
 export default function(hljs) {
   var LITERALS = 'true false yes no null';
 
+  // YAML spec allows non-reserved characters in tags
+  var NON_EX_CHARS = '\\w#;/?:@&=+$,.~*\\\'()[\\]'
```
yyyc514, line 14: What is "EX"?

src/languages/yaml.js
```diff
@@ -72,11 +76,11 @@ export default function(hljs) {
       },
       { // local tags
         className: 'type',
-        begin: '!' + hljs.UNDERSCORE_IDENT_RE,
+        begin: '!' + YAML_TAG_RE,
```
yyyc514, line 79: Let's change it to YAML_TAG_PREFIX - unless you tell me I'm mistaken? And then I think this looks ok.

src/languages/yaml.js
pplantinga: It was originally `NON_EXCLAMATION_CHARS` but that was too long. Maybe `NON_PREFIX_CHARS` is better?

src/languages/yaml.js
pplantinga: Tag prefixes are available in YAML but I'm not sure they work exactly the way you think they do. Basically, at the top of a YAML document you can define a tag prefix which is basically just a shorthand for a longer tag, e.g.:
```
%TAG !e! tag:example.com,2000:app/
---
!e!foo "bar"
```
Which converts `!e!foo` to `!<tag:example.com,2000:app/foo>` during parsing, according to the spec.

pplantinga: I'm open to other names for this variable, but I'm not sure `YAML_TAG_PREFIX` is appropriate, since we're matching the whole tag.

src/languages/yaml.js
yyyc514: It wasn't too long.

src/languages/yaml.js
yyyc514: https://yaml.org/spec/1.2/spec.html#id2783273

I'm thinking "local tag prefix" which start with `!` from my read but globals don't have `!` evidentaly.  Is your intention here that this is matching tags as well as tag prefixes, yes?

yyyc514: Ie, we're matching `!tagname!` with this also, yes?

src/languages/yaml.js
yyyc514: Wait, are they non prefix character or prefix characters?

pplantinga: You're right, prefix is confusing (here I meant '!' as prefix). Going back to exclamation.

src/languages/yaml.js
pplantinga: After reading more carefully, I don't think it works the way I thought it did either. Most of what I've read so far merely relates to the `%TAG` directive, which allows three tag prefixes, !, !!, and ! \w+ !, then it allows a replacement that includes an URI characters.

However, this doesn't define what is allowed in full tags. I actually can't find what is allowed in full tags in this document, but given how PyYAML uses the tags, it seems they can include at least `/` and `:` and `.`

pplantinga: I propose allowing all URI characters in tags since this is how other syntax highlighters are doing it (e.g. vim, although I don't think vim allows square brackets).

yyyc514: > However, this doesn't define what is allowed in full tags.

What are you calling a "full tag" here?

A tag is only a handle and a prefix unless I'm missing something.

Screen Shot 2020-04-17 at 5 18 14 AM

Then there are primary handles `!`, secondary `!!` and named handles:

```
[92] | c-named-tag-handle | ::= | “!” ns-word-char+ “!”
```

Isn't that it?  I feel like this rule (as currently written) is also going to capture named tag handles - and that might be ok... if not we should perhaps add matchers for the 3 tag handle variants.

yyyc514: If all this extra complexity is to deal with tag handles vs prefixes would it be simpler to have a sep rule for the handles? entirely? ` !, !!, and ! \w+ !` as you already said. Would that simplify the prefix rule?

pplantinga: To clarify I by "full tag" I meant `!yaml!str` where the tag prefix (or tag handle) refers to `!yaml!`. What is allowed to come after the tag prefix (in this example: `str`) is not very clearly defined in the spec. The only thing I can find is in section `3.2.1.2 Tags` and it goes like this:

`local tags are restricted to the URI character set and use URI character escaping.`

And it goes on to describe how different languages will use the tags differently (e.g. Java tags use `.`). So full URI character set is what I recommend for tags (the part after the prefix). If you want to do it entirely properly, probably what would be required is three rules for the three valid prefixes, plus excluding exclamation points from the character set.

yyyc514:
One of your tests is off.

pplantinga:
Are the regexes applied in order? The failing test splits `!foo!bar` into two separate tags, which perhaps can only be fixed if the named tag regex is applied before the local tag regex. Otherwise I can just go back to including `!` in tags which would capture named tags too

yyyc514:
Let me know when you're happy and I will pull this local for one last review and maybe a tweak or two and then try to get this merged.

yyyc514:
> Are the regexes applied in order?

They are.

pplantinga:
Okay, I reverted adding named tag handles, since I'm not quite familiar enough with how the regexes are applied. The fact that they are captured by the local tag regex should be enough in practice, its just not quite as explicit as it could be. I'm happy with this request, feel free to pull and test and make changes.

yyyc514:
All I did was add verbatim tags since I came across them while doing some final reading on this.  Waiting for a newer changelog in another PR then I'll merge this.  Right now its gonna wait until 10.1 anyways.

PR #2488, opened 2020-04-10
(dev) Developer tool should use un-minified library

mondeja:
- Following the docs, I've found that in *Basic Testing* section says: ["You need to build highlight.js with only the language you’re working on (**without compression**, to have readable code in browser error messages)"](https://highlightjs.readthedocs.io/en/latest/building-testing.html#basic-testing), but in `tools/developers.html` file the library is included minified. Documentation or file was wrong, I've supposed that the file.
- Removed a import in `test/browser/worker.js` that is not doing nothing.

### DIFF
--- a/test/browser/worker.js
+++ b/test/browser/worker.js
@@ -2,7 +2,7 @@
 
 const Worker   = require('tiny-worker');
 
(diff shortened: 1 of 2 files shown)


yyyc514:
@mondeja Thanks again.

PR #2489, opened 2020-04-11
Fix Typescript markup for comment after enclosed lambda.

### LINKED ISSUE #2189: (typescript) snippet containing a lambda which parameters are enclosed in parens fails
**Editor:** Adding a list of samples we should add tests for:

```js
const array = [1, 2, 3].reduce<number[]>((acc, next) => [...acc, next], []);
return this.sides.every((length,width=(3+2+(4/5))) => length > 0 );
const bad = ((a, b) => [...a, b]);
const bad = (a => [...a, b]);
const bad = (_ => doSomething());
const good = () => 0;
const bad = (() => 0);
```

- [x] Nested parens in expressions
- [x] () surrounded arguments
- [x] flag argument with no ()
- [x] `_` argument

Possibly relevant:
https://stackoverflow.com/questions/18703187/count-parentheses-with-regular-expression

Since to do this I think we really need to count parens in the regex that's matching => functions.

---

This is an issue I noticed downstream in VS Code itself and in the output of Asciidoctor which uses highlightjs. The snippet causing issues is this one:

````typescript
const array = [1, 2, 3].reduce<number[]>((acc, next) => [...acc, next], []);
//                             ~~~~~~~~
// Type of the initial value, in our case making it an array of numbers.
````

And gets rendered in VS Code and Asciidoctor generated HTML as follows:

image

image

When removing the arrowhead of the lambda (even though syntactically incorrect) the formatting is correct once again (since the comment at the ends gets picked up as a comment):

image

image

### PR DESCRIPTION
**Pull request by mondeja**
This resolves #2189. Function params can't contains `self`.

### DIFF
--- a/test/markup/typescript/comment-after-enclosed-lambda.expect.txt
+++ b/test/markup/typescript/comment-after-enclosed-lambda.expect.txt
@@ -1,0 +1,2 @@
+<span class="hljs-keyword">const</span> bad = <span class="hljs-function">(<span class="hljs-params">(a, b</span>) =&gt;</span> [...a, b]);
+<span class="hljs-comment">// https://github.com/highlightjs/highlight.js/issues/2189</span>
--- a/src/languages/typescript.js
+++ b/src/languages/typescript.js
@@ -159,7 +159,6 @@ export default function(hljs) {
                     excludeBegin: true, excludeEnd: true,
                     keywords: KEYWORDS,
                     contains: [
-                      'self',
                       hljs.C_LINE_COMMENT_MODE,
                       hljs.C_BLOCK_COMMENT_MODE
                     ]
--- a/test/markup/typescript/comment-after-enclosed-lambda.txt
+++ b/test/markup/typescript/comment-after-enclosed-lambda.txt
@@ -1,0 +1,2 @@
+const bad = ((a, b) => [...a, b]);
+// https://github.com/highlightjs/highlight.js/issues/2189

### CODE REVIEW

File: test/markup/typescript/comment-after-enclosed-lambda.expect.txt [hunk 1]
```diff
@@ -0,0 +1,2 @@
+<span class="hljs-keyword">const</span> bad = <span class="hljs-function">(<span class="hljs-params">(a, b</span>) =&gt;</span> [...a, b]);
```
yyyc514, line 1: This is not correct output, please compare to https://github.com/highlightjs/highlight.js/blob/master/test/markup/typescript/functions.expect.txt

The () aren't being wrapped or paired properly.  The proper fix here is likely more complex.

yyyc514: You would also need some tests (if we don't have them) for nested parenthesis inside arguments, which is likely what the self was trying to catch in the first place.

```js
return this.sides.every((length,width=(3+2+(4/5))) => length > 0 );
```

mondeja: Yes, but Javascript has the same problem. At least this change produces the same output that JS language produces. What is the purpose of 'self' in Typescript params?

yyyc514: In a perfect world something like I think:

```
const bad = ((a, b) => [...a, b]);
const bad = (a => [...a, b]);
```

```
bad = (<span class="hljs-function"><span class="hljs-params">(a, b)</span> =&gt;</span> [...a, b]);
bad = (<span class="hljs-function"><span class="hljs-params">a</span> =&gt;</span> [...a, b]);
```

mondeja: Ok, I see.

yyyc514: > Yes, but Javascript has the same problem. At least this change produces the same output that JS language produces.

Then it would seem they are both broken, just in different ways.  Fixing some of these requires really digging in deep and understanding what is happening, why the rules are there, the different possible issues (like in this case nested parens), etc. My bet is this is one of the harder ones, but I"m not certain.

yyyc514: I see you were trying to follow my last comment (thanks) but now that I've looked a little closer I don't think the fix here is just "do what JS does".

yyyc514:
```
------------v
const bad = ((a, b) => [...a, b]);
```

The problem here is that first ( is matched as the start of a function when it is not, it's just an expression.  That's what someone needs to tackle here - this misdetection.

yyyc514:
If you're curious I believe this actually resolves the issue:

https://github.com/highlightjs/highlight.js/pull/2502

PR #2490, opened 2020-04-11
(python) don't include `(` `)` in functions `params` match

mondeja:
Python functions parameters before this pull are including `(` and `)` characters. This not is the desired behaviour as commented in pull #2489.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -71,6 +71,7 @@ Developer Tools:
 [Josh Goebel]: https://github.com/yyyc514
 [Sean Williams]: https://github.com/hmmwhatsthisdo
 [Adnan Yaqoob]: https://github.com/adnanyaqoobvirk
+[Álvaro Mondéjar]: https://github.com/mondeja
 
 
 ## Version 9.18.1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Parser Engine Changes:
 
 Language Improvements:
 
+- enh(python) Exclude parens from functions params (#2490) [Álvaro Mondéjar][]
 - enh(swift) Add `compactMap` to keywords as built_in (#2478) [Omid Golparvar][]
 - enh(nim) adds `func` keyword (#2468) [Adnan Yaqoob][]
 - enh(xml) deprecate ActionScript inside script tags (#2444) [Josh Goebel][]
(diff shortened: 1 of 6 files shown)


yyyc514:
Please add changelog entry.

test/markup/python-repl/sample.expect.txt
```diff
@@ -11,7 +11,7 @@ foo = 42"
 <span class="hljs-meta">&gt;&gt;&gt;</span> <span class="python"><span class="hljs-string">"""</span></span>
 <span class="hljs-meta">...</span> <span class="python"><span class="hljs-string">abc</span></span>
 <span class="hljs-meta">...</span> <span class="python"><span class="hljs-string">"""</span></span>
-<span class="hljs-meta">&gt;&gt;&gt;</span> <span class="python"><span class="hljs-function"><span class="hljs-keyword">def</span> <span class="hljs-title">test</span><span class="hljs-params">()</span>:</span></span>
+<span class="hljs-meta">&gt;&gt;&gt;</span> <span class="python"><span class="hljs-function"><span class="hljs-keyword">def</span> <span class="hljs-title">test</span>(<span class="hljs-params"></span>):</span></span>
```
yyyc514, line 14: I'll accept as is, but would it be too much trouble to handle the edge case of () so it doesn't produce noise?

It'd prolly be something like:

```
  var PARAMS = {
    className: 'params',
    variants: [
     {begin: /\(/, end: /\)/, excludeBegin: true, excludeEnd: true, contains: ['self', PROMPT, NUMBER, STRING, hljs.HASH_COMMENT_MODE]},
     {begin: /\(\s*\)/, skip: true, className: null }
    ]
    
  };
```

Or some such.

CHANGES.md
```diff
@@ -22,6 +22,7 @@ Parser Engine Changes:
 
 Language Improvements:
 
+- enh(python) Exclude parens from functions params (#2490) [Álvaro Mondéjar][]
```
yyyc514, line 25: You also have to add your url in the footnotes around line 60 or so.

test/markup/python-repl/sample.expect.txt
mondeja: Ok, it's included.

yyyc514:
@mondeja Thanks for your contribution!

PR #2496, opened 2020-04-17
(yaml) add support for timestamps

### LINKED ISSUE #2475: (yaml) add support for dates/timestamps
**Describe the issue**
Highlighting of YAML date/datetime objects is inconsistent.

Screen Shot 2020-03-30 at 12 31 59

Some things that are parsed by YAML as dates get highlighted partly as strings, some things that are parsed by YAML as strings get highlighted as dates (or maybe numbers).

yaml

`initHighlightingOnLoad` with a language classname (see codepen below).

**Sample Code to Reproduce**
https://codepen.io/tamlyn/pen/mdJvMYv

**Expected behavior**
I expect the highlighting to be consistent with YAML's complex datatype parsing. But given the highlighting rules are based on regexes, maybe I expect too much.

pplantinga:
Add support for timestamps to YAML. Closes #2475.


src/languages/yaml.js
```diff
@@ -44,6 +44,15 @@ export default function(hljs) {
     ]
   };
 
+  var DATE_RE = '[0-9]{4}(-[0-9][0-9]){0,2}';
+  var TIME_RE = '([Tt \\t][0-9][0-9]?(:[0-9][0-9]){2})?';
+  var FRAC_RE = '(\\.[0-9]*)?';
```
yyyc514, line 49: Please no abbreviations.

src/languages/yaml.js
```diff
@@ -44,6 +44,15 @@ export default function(hljs) {
...
+  var ZONE_RE = '([ \\t])*(Z|[-+][0-9][0-9]?(:[0-9][0-9])?)?';
+  var TIMESTAMP = {
+    className: 'number',
+    begin: DATE_RE + TIME_RE + FRAC_RE + ZONE_RE + '\\b',
```
yyyc514, line 53: Why only a boundary at the end and not the beginning?

yyyc514:
Only those two notes, otherwise this looks great.  Told ya it'd be easier. :)

PR #2502, opened 2020-04-21
fix(javascript/typescript): lambda with parens in parameters fails

### LINKED ISSUE #2189: (typescript) snippet containing a lambda which parameters are enclosed in parens fails

yyyc514:
- Fixes both JavaScript and TypeScript grammars

Fixes samples like:

    const bad = ((a, b) => [...a, b]);
    sides.every((length,width=(3+2+(4/5))) => length > 0 );

This is done by counting parens in the regex that finds arrows
functions. Currently we can only handle 2 levels of nesting as
shown in the second example above.

Resolves #2189.


yyyc514:
Note this isn't a full parser.... this will also fall apart if say you had parenthesis inside strings inside the params... since the regex can't tell a paren inside a string from a regular one.

Always open to better suggestions.

src/languages/javascript.js
```diff
@@ -90,6 +90,10 @@ export default function(hljs) {
     hljs.REGEXP_MODE
   ];
   var PARAMS_CONTAINS = SUBST.contains.concat([
+    // eat recursive parens in sub expressions
+    { begin: /\(/, end: /\)/,
+      contains: ["self", hljs.C_LINE_COMMENT_MODE,hljs.C_BLOCK_COMMENT_MODE]
+    },
```
egor-rogov, line 96: What is the purpose of this "parens eater"? Due to this, some highlighting is lost inside params.
E. g. for this snippet:
```
sides.every((length,width=(3+2+(4/5))) => length > 0 );
```
we had before:
```
sides.every(<span class="hljs-function">(<span class="hljs-params">length,width=(
<span class="hljs-number">3</span>+<span class="hljs-number">2</span>+(<span class="hljs-number">4</span>/<span class="hljs-number">5</span></span>))
) =&gt;</span> length &gt; <span class="hljs-number">0</span> );
```
and after the patch we have:
```
sides.every(<span class="hljs-function">(<span class="hljs-params">length,width=(
3+2+(4/5))</span>
) =&gt;</span> length &gt; <span class="hljs-number">0</span> );
```
I. e. numbers are no more highlighted.
(In typescript, numbers are not highlighted even now)

yyyc514: We have to count parens to know which one is the proper one to stop at... but this issue you pointed out can be resolved by just adding more things to the params ruleset. I'll look into that.

taufik-nurrohman: https://github.com/highlightjs/highlight.js/pull/2372#issuecomment-583271131

yyyc514: ?

taufik-nurrohman: @yyyc514 It is okay. It&rsquo;s just that I know this issue will come up as I thought about the function argument pattern problem, like in the comment I have made there.

yyyc514: This fix handles both those cases without issue.

yyyc514: @egor-rogov Have another look please now.

PR #2506, opened 2020-04-24
fix(parser) fixes sublanguage with no rule matches

### LINKED ISSUE #2504: (html) Code inside <script> tag disappeared
**Describe the issue**

When applying highlight.js to the following code block,

```
<pre><code class="html">&lt;script&gt;foo();&lt;/script&gt;</code></pre>
```

 I am expecting `<script>foo();</script>`, but it prints `<script></script>` without `foo();` inside. I understand if the javascript cannot be highlighted since the language is set as html. However, the disappearance of the contents does not look good.

html

I used `initHighlightingOnLoad`.

...

**Sample Code to Reproduce**

Here is jsfiddle: <https://jsfiddle.net/6ey310qL/1/>

```
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/10.0.0/styles/default.min.css">
    <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.0.0/highlight.min.js"></script>
  </head>
  <body>
    <pre><code class="html">&lt;script&gt;foo();&lt;/script&gt;</code></pre>
  </body>
</html>
```

I changed:

* version 10.0.0
* language as html
* script source to cloudfare: `https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.0.0/highlight.min.js`

**Expected behavior**

```
<script>foo();</script>
```

is printed in the code block.

**Additional context**

Is this a bug? or a feature?

yyyc514:
Resolves #2504.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+## Version 10.0.1
+
+Parser Engine Changes:
+
+- (bug) Fix sublanguage with no relevance score (#) [Josh Goebel][]
+
+[Josh Goebel]: https://github.com/yyyc514
+
+
 ## Version 10.0.0
 
 New languages:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 Parser Engine Changes:
 
-- (bug) Fix sublanguage with no relevance score (#) [Josh Goebel][]
+- (bug) Fix sublanguage with no relevance score (#2506) [Josh Goebel][]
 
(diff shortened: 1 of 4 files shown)


yyyc514:
In the future (if we pull plaintext into the core library so it's impossible to build without) it'd probably make sense to use that here, but this works for now.

Issue #1245, opened 2016-07-14
Discuss: Website/browser build is not intended for use with Node.js.


avih:
`var <any-name-hljs-included> = require('./highlight.pack');` fails with `ReferenceError: hljs is not defined.` when executed in node.js.

This happens when downloading and extracting highlight.zip from https://highlightjs.org/download/ (tested both the default selected languages or with just one - javascript)

But it doesn't fail (and I can later use it nicely) if I build it on my own with `node tools/build.js javascript` or any other languages which I tried.

This happens with 9.50 but also happened with 9.40 (worked when I build it, didn't from the download).

Looking at `highlight.pack.js` - both the downloaded and the one which I built, it seems that the code is supposed to work also in a module environment (tests for `exports` availability at the beginning - identical first few LOC at both files), but it doesn't with the download.

Is it a bug? expected? IMO it should be possible to require it also with the downloaded package.

isagalaev:
We have different build targets for node.js and for the browser/AMD environment. The `highlight.pack.js` you're getting from the site is the browser variant. Have a look here: http://highlightjs.readthedocs.io/en/latest/building-testing.html

avih:
I know. I'm saying that if you build for the browser, like so: `node tools/build.js -t browser :common` then the resulting `highlight.pack.js` can still be used with node, but if you download it from the site then it cannot be used with node.

isagalaev:
Aha… Then it might be due to the bug #1220 that I fixed just yesterday. Could you please check again?

avih:
I just check. It still cannot be used from node when downloading from the site, but can be used from node when building locally for the browser (with 9.6.0 - same as it was on 9.5.0 and 9.4.0).

At the site download, I selected a single language: javascript and clicked download.

Inside the downloaded zip I opened `highlight.pack.js` in an editor, and it starts with `/*! highlight.js v9.6.0 | BSD3 License | git.io/hljslicense */` so this is the latest version.

I created a test file `test.js` which contains a single line: `var x = require('./highlight.pack');`

It fails with the downloaded `highlight.pack.js`, but works with `highlight.pack.js` which I built myself for the browser.

I then unminified (or at least beautified) `highlight.pack.js` which I downloaded - using http://jsbeautifier.org/

This code is at the global scope (towards the end of the file):
``` javascript
hljs.registerLanguage("javascript", function(e) {
    return {
    // ... the rest of the javascript support is here
```

However, there's no global object `hljs`, so it fails.

[...]

PR #2511, opened 2020-04-25
(build) browser build is CommonJS and IIFE (global)

### LINKED ISSUE #2291: highlightjs.org build seems to differ from `./tools/build.js -t browser :common`

### LINKED ISSUE #2505: Require using AMD broken in 10.0.0
Loading via SystemJS in AMD used to work with v9.18.1. Broken in v10:

```
system.js:4 Uncaught (in promise) ReferenceError: hljs is not defined
    at eval (highlight.js:1566)
    at eval (highlight.js:6322)
    at eval (<anonymous>)
    at Qe (system.js:4)
    at system.js:4
    at async Promise.all (:63342/Kidoju.Widgets/src/js/widgets/index 10)
```

hljs is undefined when calling  L1566

```
hljs.registerLanguage('apache', function () {
   ...
}());
```

In System.config.js, I have:

```
packages: {
    'src/js': {
        defaultExtension: 'js',
        meta: {
           '*.js': {
                babelOptions: {
                    es2015: false,
                    stage2: false,
                    stage3: false,
                },
                format: 'amd',
            },
            '*.es6': {
                format: 'esm',
            },
            '*.jsx': {
                format: 'esm',
            },
            '*.mjs': {
                format: 'esm',
            },
         },
      },
 },
```

A workaround for now is to add an exception in meta:

```
 'vendor/highlight/highlight.js': {
      // Note: AMD broken as of v10
      format: 'global',
},
```

yyyc514:
- dropping support for AMD, which we never truly supported
  in the first place

  Resolves #2505. Also think this will close #2291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,5 +29,7 @@ script:
       npm run test
     else
       npm run test-browser
+      # our browser build should also work fine as a Node.js CommonJS module
+      node test/builds/browser_build_as_commonjs.js
     fi
 sudo: false  # Use container-based architecture
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,7 +40,7 @@ Brower build:
 
 Parser Engine Changes:
 
-- [Issue](https://github.com/highlightjs/highlight.js/issues/2504) (bug) Fix sublanguage with no relevance score (#2506) [Josh Goebel][]
+- ...
 
 
 [Josh Goebel]: https://github.com/yyyc514
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,16 @@
 ## Version 10.0.1
 
+Brower build:
+
+- [Issue](https://github.com/highlightjs/highlight.js/issues/2505) (bug) Fix: Version 10 fails to load as CommonJS module. (#2511) [Josh Goebel][]
+- [Issue](https://github.com/highlightjs/highlight.js/issues/2505) (removal) AMD module loading support has been removed. (#2511) [Josh Goebel][]
+
+
 Parser Engine Changes:
 
-- (bug) Fix sublanguage with no relevance score (#2506) [Josh Goebel][]
+- [Issue](https://github.com/highlightjs/highlight.js/issues/2504) (bug) Fix sublanguage with no relevance score (#2506) [Josh Goebel][]
+
+
 
 [Josh Goebel]: https://github.com/yyyc514
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,7 +30,7 @@ Language Improvements:
 [Vania Kucher]: https://github.com/qWici
 
 
-## Version 10.0.1
+## Version 10.0.2 (pending)
 
 Brower build:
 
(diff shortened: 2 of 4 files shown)


yyyc514:
Essentially this is just IIFE (global) with a small footer to support CommonJS (since I didn't see the harm):

```
var hljs = (function () {
  'use strict';

  ...

}());
if (typeof module !== 'undefined') { module.exports = hljs; }
```


Attaching a file for testing:

[highlight.js.zip](https://github.com/highlightjs/highlight.js/files/4534086/highlight.js.zip)

avih:
I find it a bit hard to understand from the description, does this merge mean that the concatenated web-download can now work in node.js? does it have an effect on #1245

Thanks in advance.

yyyc514:
> does this merge mean that the concatenated web-download can now work in node.js?

@avih YES, AFAIK it SHOULD now work when used like that. **I'm still not sure that's optimal**, but it was an easy thing to enable in hindsight. There is a chance in the future we might compile the web build ES5 which would make it larger/slower than the Node.js build... but for now...

If you test it and it works, please let me know and I'll bump that thread with an update.

avih:
Thanks a lot! I'll bump this thread once I have tested it (not right away, but I intend to)

avih:
> @avih **YES, AFAIK it should now work when used like that.**

Indeed it seems to be working :)

I went to https://highlightjs.org/download/ , created some custom package (the default languages and few more), and it seems to be loading fine and working in node.js as is.

Thank you! (even it it won't last forever, but then again, what does last forever? ;) )

yyyc514:
> As a side note, any chance you can add a button at the download page to select all/default checkboxes?

That'd be a task for @isagalaev ... but the page should default to the defaults by default and remember custom choices in a  cookie.

avih:
> the page should default to the defaults by default and remember custom choices in a cookie.

Correct, but if someone wants to download everything then it's quite tedious.

yyyc514:
Literally everything isn’t even advised. It will be huge and slow to autodetect anything.

avih:
You could make exactly the same argument that anything beyond what you need exactly would be a waste of cpu cycles and storage space or network bandwidth, and yet your default includes 38 languages.

yyyc514:
I would say that’s entirely correct, yet we need some defaults so as to be immediately useful to people as something you can easily load via CDN and just quickly get started easily.

I think we’re still pretty conservative with qhat we include by default with regard to how large some JS libraries are there days.  Of course if someone knows they are NEVER going to need to highlight a given language it’d be better to not use the library with that language compiled in. That’s just common sense.

Anyways, as I said any changes to the website are fully up to @isagalaev and his time.

_It’s already super easy with the main build tool to build a tiny or HUGE package, whatever one desires._

avih:
You provide a useful thing, but you limit your users to what you think would be reasonable usage. Let users go wild, maybe someone uses it locally with enough space and cpu cycles to spare? maybe they don't do autodetection? etc.

yyyc514:
No limits.  People can do whatever they want.  :-). The defaults work well for “most people”.

Issue #2498, opened 2020-04-19
(yaml) Add support for inline sequences and mappings


pplantinga:
**Describe the issue**
Current highlighting does not support inline sequences and mappings.

**Which language seems to have the issue?**
YAML

**Sample Code to Reproduce**
https://jsfiddle.net/sda50gr8/

**Expected behavior**
![sequence-mapping](https://user-images.githubusercontent.com/4369106/79674618-dc652780-81b2-11ea-8d73-a284b4ce5485.png)

yyyc514:
Can these be recursive also?

pplantinga:
Yes, these can be nested. I've started working on this, but the simple implementation (just add a rule for all brackets and commas) auto-detected a "clean" file as yaml.

yyyc514:
Wait. What is the goal here again?  I’m not certain it’s clear from the snap.

pplantinga:
I think top priority should be given to making sure numbers are not highlighted as strings (e.g. `{1` or `2.0,` or `3]` would all be highlighted as strings). A secondary concern might be highlighting braces, brackets, and commas as a separate type than strings.

yyyc514:
> A secondary concern might be highlighting braces, brackets, and commas as a separate type than strings.

Traditionally we don't highlight symbols.

You might look into the JSON grammar for inspiration.

PR #2513, opened 2020-04-27
(yaml) Add support for inline sequences and mappings

pplantinga:
This merge request closes issue #2498

### DIFF
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -39,7 +39,7 @@ export default function(hljs) {
     variants: [
       {begin: /'/, end: /'/},
       {begin: /"/, end: /"/},
-      {begin: /\S+/}
+      {begin: /[^\s,{}[\]]+/}
     ],
     contains: [
       hljs.BACKSLASH_ESCAPE,
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -47,20 +47,15 @@ export default function(hljs) {
     ]
   };
 
-  // Strings inside objects can't contain braces, brackets, or commas
-  var INSIDE_OBJECT_STRING = {
-    className: 'string',
-    relevance: 0,
+  // Strings inside of value containers (objects) can't contain braces,
+  // brackets, or commas
+  var CONTAINER_STRING = hljs.inherit(STRING, {
     variants: [
       {begin: /'/, end: /'/},
       {begin: /"/, end: /"/},
       {begin: /[^\s,{}[\]]+/}
-    ],
-    contains: [
-      hljs.BACKSLASH_ESCAPE,
-      TEMPLATE_VARIABLES
     ]
-  };
+  });
 
   var DATE_RE = '[0-9]{4}(-[0-9][0-9]){0,2}';
   var TIME_RE = '([Tt \\t][0-9][0-9]?(:[0-9][0-9]){2})?';
@@ -69,9 +64,30 @@ export default function(hljs) {
   var TIMESTAMP = {
     className: 'number',
     begin: '\\b' + DATE_RE + TIME_RE + FRACTION_RE + ZONE_RE + '\\b',
-  }
+  };
 
-  var INSIDE_OBJECT_TYPES = [
+  var VALUE_CONTAINER = {
+    end: ',',
+    endsWithParent: true,
+    excludeEnd: true,
+    contains: [],
+    keywords: LITERALS,
+    relevance: 0
+  };
+  var OBJECT = {
+    begin: '{', end: '}',
+    contains: [VALUE_CONTAINER],
+    illegal: '\\n',
+    relevance: 0
+  };
+  var ARRAY = {
+    begin: '\\[', end: '\\]',
+    contains: [VALUE_CONTAINER],
+    illegal: '\\n',
+    relevance: 0
+  };
+
+  var MODES = [
     KEY,
     {
       className: 'meta',
@@ -120,8 +136,8 @@ export default function(hljs) {
     },
     { // array listing
       className: 'bullet',
-    // TODO: remove |$ hack when we have proper look-ahead support
-    begin: '\\-(?=[ ]|$)',
+      // TODO: remove |$ hack when we have proper look-ahead support
+      begin: '\\-(?=[ ]|$)',
       relevance: 0
     },
     hljs.HASH_COMMENT_MODE,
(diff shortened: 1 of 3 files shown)


src/languages/yaml.js
```diff
@@ -39,7 +39,7 @@ export default function(hljs) {
     variants: [
       {begin: /'/, end: /'/},
       {begin: /"/, end: /"/},
-      {begin: /\S+/}
+      {begin: /[^\s,{}[\]]+/}
```
yyyc514, line 42: I'm not sure about this.  What is the correct behavior for:

```
foo: I am a string { boo: hoo } 
```

Also I'm not sure this is the desired fix in any case.  What you'd really want to do is match on `[` or `{` and then go into an alternative mode (to allow matching embedded strings, numbers, etc)... not overload string in this fashion.

pplantinga: This change simply makes it so strings do *not* match braces or brackets or commas. I don't think your example is valid yaml since it mixes two different types of nodes (they must be one of scalar, sequence, or mapping).

pplantinga: I don't think that what is valid inside of braces or brackets is any different from outside, except  that the brackets and braces must of course be matching, and commas are only valid inside.

yyyc514: It seems these are certainly characters that are valid inside strings:

```rb
> YAML.load("---\n:test: boo[]ger{mode:[32]}").to_h
=> {:test=>"boo[]ger{mode:[32]}"}
```

Again negative matching isn't the way to go about this.  You need to detect `[` or `{` and then switch modes.

yyyc514: Again, looking at the JSON stuff might be a big help.

yyyc514: > I don't think that what is valid inside of braces or brackets is any different from outside, except that the brackets and braces must of course be matching, and commas are only valid inside.

Right but if you don't want them to be counted as strings then you need to explain to the ruleset what they are. :-)  Since the general case is "anything following a key is a string"... these are exception cases and must be accounted as such.

And surely what is valid inside is a little different... at least regarding line breaks, yes?

pplantinga:
I'm a little stuck to be honest. How can I prevent strings from matching commas and end-brackets or braces without adjusting the rule for strings?

yyyc514:
> I'm a little stuck to be honest. How can I prevent strings from matching commas and end-brackets or braces without adjusting the rule for strings?

Are naked strings actually allowed inside arrays and objects?  

If you have some failing tests cases please submit them so i can see what you're talking about exactly.

yyyc514:
I would imagine though that **while inside the array/object** you'll need a different type of string, one that doesn't allow the array or brace characters...

pplantinga:
Yes naked strings are allowed inside arrays and objects.

Failing test is the one that I added:
```
AssertionError: expected 

'<span class="hljs-attr">foo:</span> [<span class="hljs-string">bar,</span> <span class="hljs-string">bar2,</span> <span class="hljs-number">1</span><span class="hljs-string">,</span> <span class="hljs-number">2</span><span class="hljs-string">]</span>\n<span class="hljs-attr">foo:</span> {<span class="hljs-attr">bar:</span> <span class="hljs-number">1</span><span class="hljs-string">,</span> <span class="hljs-attr">bar:</span> <span class="hljs-number">2</span><span class="hljs-string">}</span>'

to be
'<span class="hljs-attr">foo:</span> [<span class="hljs-string">bar</span>, <span class="hljs-string">bar2</span>, <span class="hljs-number">1</span>, <span class="hljs-number">2</span>]\n<span class="hljs-attr">foo:</span> {<span class="hljs-attr">bar:</span> <span class="hljs-number">1</span>, <span class="hljs-attr">bar:</span> <span class="hljs-number">2</span>}'
```

yyyc514:
Can tags really appear anywhere inside the YAML, including nested in a hash or array?

yyyc514:
Do you love YAML? :) You surely pick some of the harder problems here to solve when there are a lot of easier ones in the issues. :-)

pplantinga:
> Do you love YAML? :) You surely pick some of the harder problems here to solve when there are a lot of easier ones in the issues. :-)

Hah, no kidding ;-) I had been planning to use highlight.js for a project that heavily uses YAML (perhaps in some non-standard ways), but I was bothered by some of the incorrect highlighting.

yyyc514:
> I had been planning to use highlight.js for a project that heavily uses YAML (perhaps in some non-standard ways), but I was bothered by some of the incorrect highlighting.

Not to dissuade you but I think you'd likely do much better be parsing it with a library that fully understands it and they just iterating over the raw object data and rendering that - although that wouldn't guarantee the preservation of the exactly structure of the encoding itself.

src/languages/yaml.js
```diff
@@ -46,6 +46,19 @@ export default function(hljs) {
       TEMPLATE_VARIABLES
     ]
   };
+  var INSIDE_OBJECT_STRING = {
+    className: 'string',
+    relevance: 0,
+    variants: [
+      {begin: /'/, end: /'/},
+      {begin: /"/, end: /"/},
+      {begin: /\S*[^\s,\]}]/}
```
yyyc514, line 55: `{[` and are off limits too IIRC.

src/languages/yaml.js
```diff
@@ -56,76 +69,95 @@ export default function(hljs) {
...
+  ];
+  var VALUE_CONTAINER = {
+    end: ',', endsWithParent: true, excludeEnd: true,
+    contains: TYPES.concat([INSIDE_OBJECT_STRING]),
+    keywords: LITERALS,
+    relevance: 0
+  };
+  var OBJECT = {
+    begin: '{', end: '}',
+    contains: [hljs.inherit(VALUE_CONTAINER)],
```
yyyc514, line 146: Why inherit?

src/languages/yaml.js
yyyc514: Perhaps this should be \S+ with a look-ahead instead? Would that more accurately express the intent?

src/languages/yaml.js
pplantinga: This was copied from the json, where it had the note: "inherit is a workaround for a bug that makes shared modes with endsWithParent compile only the ending of one of the parents"

src/languages/yaml.js
pplantinga: I don't think a look-ahead is what we want here, since it doesn't matter what is immediately before or after the string, it just can't start/end with a comma or brace or bracket.

yyyc514: >  it just can't start/end with a comma or brace or bracket.

That's exactly what is "after the string"...

```
\S+(?=[\s,\]}])
```

Plus now the negation is removed which is easier to follow.

src/languages/yaml.js
yyyc514: That goes back to 2015... does it break if you remove the inherit?

pplantinga: This has been removed in most recent commit

src/languages/yaml.js
```diff
@@ -47,6 +47,22 @@ export default function(hljs) {
...
 
+  // Strings inside objects can't start with { or } or [ or ] or ,
+  // and can't end with } or ] or ,
+  var INSIDE_OBJECT_STRING = {
+    className: 'string',
+    relevance: 0,
+    variants: [
+      {begin: /'/, end: /'/},
+      {begin: /"/, end: /"/},
+      {begin: /[^\s,{}[\]](\S*[^\s,\]}])?/}
```
yyyc514, line 58: Why would a simple repeated negation not work:

```
[^\s,{}[\]]+
```

That's the simplest thing I came up with playing.

pplantinga: I was attempting to still allow these characters in strings, just not as the first or last character.

yyyc514: In my testing in Ruby at least that is not allowed.  The character means the end of the string.

pplantinga: Ah, you're right, its the same way in Python. I guess I shouldn't have assumed. I'll use the repeated negation.

yyyc514: Never assume with YAML. :-)  Although it's a good rule in other context also. :-)

yyyc514:
I don't see any tests for nested, nesting is allowed is it not?

pplantinga:
> I don't see any tests for nested, nesting is allowed is it not?

Tests with nested elements are included in most recent commit

Issue #2514, opened 2020-04-27
(javascript) (typescript?) setTimeout incorrectly detected as a setter


qWici:
<details>
  <summary>Screenshot of highlighted code</summary>

![image](https://user-images.githubusercontent.com/11472929/80415196-da9a1300-88da-11ea-8d2d-fd10cff05b7d.png)

</details>

<details>
  <summary>Screenshot of html output</summary>

![image](https://user-images.githubusercontent.com/11472929/80415307-0c12de80-88db-11ea-9207-4ec33022afdb.png)

</details>

It looks like this regexp wrong. 
https://github.com/highlightjs/highlight.js/blob/d4b047b4d89bcc302f50f079ba189b0ad0ccb91b/src/languages/javascript.js#L266

yyyc514:
> It looks like this regexp wrong.

I concur with your conclusion.  Would you be willing to whip up a PR?  If so:

- Don't forget `CHANGES.md`
- We'll need to add a quick markup test for this.

It should likely be `\s+` me thinks.

yyyc514:
This quite likely is also broken in typescript I'd guess.

qWici:
@yyyc514 yeah, I tested it on [regex101](https://regex101.com/r/dgIXfJ/1). And I want to try to fix it.

yyyc514:
Oh wow that site is super nice!

qWici:
@yyyc514 ok, I fixed it. But I have 2 questions before PR.
1. Should these functions be highlighted? Because it's native JS functions (setTimeout, setInterval)
2. I don't see such a rule for typescript. I can add. Do I need to create a separate PR for this?

yyyc514:
> Should these functions be highlighted? Because it's native JS functions (setTimeout, setInterval)

You could argue that either way.  We barely highlight any built in functions right now. Since those are more like window globals I wouldn't mind if we added them as built-ins.

> I don't see such a rule for typescript. I can add. Do I need to create a separate PR for this?

Oh, it doesn't have it.  I wouldn't.  Technically the rule would need to be a lot more complex in TypeScript with all it's type details... so unless you want to potentially tackle that I'd probably just focus on JS.

Long-term I'd like to see if we can find a way to build TS and JS on the same foundation since they are so similar (other than typing).

qWici:
[...]

PR #2515, opened 2020-04-27
fix(javascript) fix set/get issue, add `set/clear`...`Timeout/Interval` keywords  (#2514)

qWici:


src/languages/javascript.js
```diff
@@ -263,7 +263,7 @@ export default function(hljs) {
         beginKeywords: 'constructor', end: /\{/, excludeEnd: true
       },
       {
-        begin:'(get|set)\\s*(?=' + IDENT_RE+ '\\()',
+        begin: "(get|set)\\s+(?=" + IDENT_RE + "\\()",
```
yyyc514, line 266: Please don't change the string type for no reason.

src/languages/javascript.js
```diff
@@ -272,6 +272,12 @@ export default function(hljs) {
           PARAMS
         ]
 
+      },
+      {
+        beginKeywords: "(setInterval|setTimeout)",
+        keywords: "setInterval setTimeout",
+        end: /\(/,
+        excludeEnd: true,
```
yyyc514, line 280: Is there some reason these can't simply be added to built-ins in `keywords`?

qWici: Oh yes I missed that

CHANGES.md
```diff
@@ -13,6 +13,8 @@ Deprecations:
 
 Language Improvements:
 
+- enh(javascript) setTimeout|setInterval now highlighted (#2514) [Vania Kucher][]
+- fix(javascript) prevent setter keyword `set` conflicting with setTimeout|setInterval (#2514) [Vania Kucher][]
```
yyyc514, line 17: You need to add yourself around line 23/24 also.... so it knows where to link to.

src/languages/javascript.js
yyyc514: N/m, I'll take this back.  This kind of feedback would drive me nuts since we don't have clear guidelines.   As long as you aren't changing the whole file I'm good.

src/languages/javascript.js
```diff
@@ -263,7 +263,7 @@ export default function(hljs) {
         beginKeywords: 'constructor', end: /\{/, excludeEnd: true
       },
       {
-        begin:'(get|set)\\s*(?=' + IDENT_RE+ '\\()',
+        begin: '(get|set)\\s+(?=" + IDENT_RE + "\\()',
```
yyyc514, line 266: pretty sure you can't mix them like this. :-)

src/languages/javascript.js
```diff
@@ -33,7 +33,7 @@ export default function(hljs) {
       'Float64Array Int16Array Int32Array Int8Array Uint16Array Uint32Array ' +
       'Uint8Array Uint8ClampedArray ArrayBuffer DataView JSON Intl arguments require ' +
       'module console window document Symbol Set Map WeakSet WeakMap Proxy Reflect ' +
-      'Promise'
+      'Promise setInterval setTimeout'
```
yyyc514, line 36: Don't we also have clearInterval() and clearTimeout()?

test/markup/javascript/keyword-as-part-of-method-name.expect.txt
```diff
@@ -0,0 +1,2 @@
+<span class="hljs-keyword">setTimeout</span>();
+<span class="hljs-keyword">setInterval</span>();
```
yyyc514, line 2: Is this correct? it should be built-in now, yes? Are you running the tests still?

yyyc514:
I'll swing back for a final review later, but this is looking pretty good!

qWici:
@yyyc514 okay, now it works. All tests were successful.

image

PR #2517, opened 2020-04-28
Add jsdoc annotations and TypeScript type file

yyyc514:
Adds JSDoc annotations and a `.tsconfig` that allows TypeScript to be run in it's "allowJS" mode and apply type and sanity checking to JavaScript code also. See [Type Checking JavaScript Files](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html).

I've been using TypeScript a lot lately and finding it very beneficial and wanted to get  those same benefits here but without converting the whole project to TypeScript.  It was rough at the beginning but now that this is finished I think it's about 80%-90% of the benefits without any of the TS compilation pipeline. The big difference in being JSDoc for adding typing information vs inline types with TypeScript.

Should be super helpful for maintainers using an editor with tight TypeScript integration and the improved docs/comments should help everyone else.

- [x] Adds `types/index.d.ts` to NPM build (should be useful for TypeScript peeps)
- [x] Improves documentation of many functions
- [x] Adds JSDoc annotations to almost all functions
- [x] Adds JSDoc type annotations to variables that can't be inferred
- [x] Refactors a few smaller things to allow the TypeScript compiler to better infer what is happening (and usually also made the code clearer)

Other than the small bits of refactoring this is mostly a docs/comments update.


allejo:
Apologies for coming in late on this one, but yes! :heart: thank you for this PR! This will make my life so much easier when reading through the code ☺️

yyyc514:
@allejo Glad to help.  It's even more amazing when writing it in a TS enabled editor. :)

julien-c:
Just a note that because the types refer to `HTMLElement` they don't compile out of the box in Node.

The community-maintained typings over at [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/highlightjs) work around this by defining an empty `interface Node { }`

yyyc514:
I'm not sure exactly where we'd add that in our type file or how we'd export it... I'm still confused on the whole ambient module stuff at the moment.

eytienne:
why not "converting the whole project to TypeScript" ?

joshgoebel:
> why not "converting the whole project to TypeScript" ?

JS is more popular (for those who sometimes contribute) and simply converting the whole project is not necessary.

PR #2519, opened 2020-04-29
fix(ts/js) use identifier to match potential keywords

### LINKED ISSUE #2516: (typescript) $class variable is incorrectly highlighted as a keyword
**Describe the issue**

**Are you using `highlight` or `highlightAuto`?** `highlight` 

...

**Sample Code to Reproduce**

https://jsfiddle.net/fsLxdrpm/1/
```ts
const ElementPrototype = Element.prototype;
const isBool = (v: any) => typeof v === 'boolean';
const $class = (v: any) => typeof v === 'boolean';
const whether = isBool(true);
const test = $class(true);
```

**Expected behavior**

1. `$class` prefix with `$` went wrong
2.  method with `class` name went wrong
3. method isn't be highlighted
4. Element isn't be highlighted 
5. isBool isn't be highlighted

yyyc514:
- fix(typescript) use identifier regex to match potential keywords
- fix(javascript) use identifier regex to match potential keywords

The treats expressions like `$class` as a whole identifier and therefore
avoids the class portion of that identifier incorretly being highlighted
as a keyword.

Resolve #2516.


yyyc514:
``js`
  var KEYWORDS = {
     $lexemes: ECMAScript.IDENT_RE,
     keyword: ECMAScript.KEYWORDS.join(" "),
     literal: ECMAScript.LITERALS.join(" "),
     built_in: ECMAScript.BUILT_INS.join(" ")
```

I can document this syntax, but I think this makes a lot of sense as the whole purpose of lexemes is to process keywords. And many syntaxes embed the keyword object over and over in many places...

**Comment by yyyc514** [comment]
Also the name "lexemes" drives me nuts and we could use this as an opportunity to start renaming it... `$pattern`?

**Comment by allejo** [comment]
> * mode level `lexemes` (if present) would supersede `keywords/$lexemes`

Mode level `lexemes` currently exist and are potentially in use right now. Wouldn't we want `keywords/$lexemes` to supersede that? i.e. we're introducing new behavior and the old behavior will potentially be deprecated, so shouldn't that new behavior be more important?

> * having both is a hard error and the language does not compile (ie, grammar designers would not be allowed to do this when designing grammars)

Do you know if there's a use case of when anyone would need to use both? If not, then I would think that discouraging this behavior would be ideal so a hard error.

Unless there's a use case, this has my vote :+1:

> * Although perhaps no special behavior is needed here... we could let `$lexemes` always win and it can always be destroyed with `inherit` if they REALLY need an exception case for some deep mode - the same as you would do for any other key.

If supporting both methods at the same time is important, then I think this idea is the same as my thoughts on the first option.

---

> Long term I think perhaps we should deprecate `MODE.lexemes` in favor or `MODE.keywords.$lexemes`.

If my understanding is correct, lexemes are currently defined per mode (as a sibling to `keywords`). However, a common practice in writing grammars is that `keywords` is a single object/string and the lexemes for said keywords are outside that object, correct? If so, I'm in favor of deprecating `MODE.lexemes`.

---

> Also the name "lexemes" drives me nuts and we could use this as an opportunity to start renaming it... `$pattern`?

:+1: for `$pattern`. What this feature is defining is the _pattern_ for a lexeme or token, not the lexemes themselves.

**Comment by egor-rogov** [comment]
> This adds $lexemes

It looks reasonable, but given we already have one method (`MODE.lexemes)), do we really need another one? I'm not strongly against it, just in a doubt.

> Long term I think perhaps we should deprecate MODE.lexemes in favor or MODE.keywords.$lexemes.

We'll have troubles doing so, because a lot of grammars use it.

> Also the name "lexemes" drives me nuts and we could use this as an opportunity to start renaming it... $pattern?

I agree, "lexemes" always seemed strange to me. `keywords.$pattern` looks better. But why do you want to use this dollar?

**Comment by yyyc514** [comment]
> keywords.$pattern looks better. But why do you want to use this dollar?

Because it's a reserved namespace.  Someone might be writing a grammar for a pattern matching library and want to have a "pattern" type of keyword and have "pattern" as the className... so we have to have some type of "reserved" namespace if we are going to allow configuration options right next to class names.

> but given we already have one method (`MODE.lexemes)), do we really need another one? I'm not strongly against it, just in a doubt.

I was cutting and pasting the code 4-5 times for Javascript and then about to have to do it gain for TypeScript - and this is literally why we program, to eliminate this type of manual cutting and pasting - and I'm about ready for grammars to get a bit easier to write.  Plus cutting and pasting is error prone, and it's too easy to miss a spot, etc...

> Do you know if there's a use case of when anyone would need to use both? If not, then I would think that discouraging this behavior would be ideal so a hard error.

We could always discourage it UNTIL someone finds a use case. :-)

---

The other idea here would be some type of inheritance setting for keywords/lexemes... but I'm not sure what that would look like.  You'd want it at a single level (top) to prevent hunt and find and cut and paste over and over...

```
keywords: { ... },
lexemes: /[a-z]+/,
inheritKeywords: true,
inheritLexemes: true,
// or
keywords: { 
  $inherit: true,
  $lexemes: /[a-z_]+/,
  keywords: " ... ",
  built_ins: " ... "
}
```

I'm already liking this idea of keeping the configuration with the keywords themselves though, as honestly it seems the most natural place to me.

So once inherit was turned on the keywords would be inherited down... and then (say inside a string or comment) you'd have to say `keywords: null` to stop the inheritance chain.

My worry is there we might start getting people accidentally highlighting keywords inside strings/comments just because they fail to turn off inheritance... but I'm not sure that's much worse than the current behavior of not highlighting keywords at all because people forgot to do the keyword inheritance manually?

**Comment by yyyc514** [comment]
> We'll have troubles doing so, because a lot of grammars use it.

We'd deprecate it like you do anything... with a major version, first to deprecate it - add warnings... and then in the next major version to turn it off completely. But I'm fine with it living for a long while if need be... 

If the new location becomes the blessed one it's one line of code to just copy `lexemes` into `keywords.pattern` when compiling modes... and I think we can maintain that one line of code many years into the future if we MUST. :-)

**Comment by egor-rogov** [comment]
Okay, I'm fine with a hard error.
So we agreed on `$pattern`, I think? And it should be documented, of course.

**Comment by yyyc514** [comment]
No worries I will document it. :)

**Comment by yyyc514** [comment]
How is that?

File: src/languages/php.js [hunk 1]
```diff
@@ -84,8 +84,7 @@ export default function(hljs) {
         false,
         {
           endsWithParent: true,
-          keywords: '__halt_compiler',
-          lexemes: hljs.UNDERSCORE_IDENT_RE
```
yyyc514, line 88: Not needed.

yyyc514:
Better?

docs/language-guide.rst
```diff
@@ -85,19 +87,25 @@ To define such keyword groups the attribute ``keywords`` becomes an object each
...
+``\w+``, and that works well for many languages. Different lexing rules can be
+defined by the magic ``$pattern`` attribute:
 
 ::
 
   {
-    lexemes: '-[a-z]+',
-    keywords: '-import -export'
+    keywords: {
+      $pattern: /-[a-z]+/,        // allow keywords with dash in them
```
egor-rogov, line 102: with dash in them -> that start with a dash?

src/lib/mode_compiler.js
```diff
@@ -207,11 +207,25 @@ export function compileLanguage(language) {
     mode.__beforeBegin = null;
 
     mode.keywords = mode.keywords || mode.beginKeywords;
+
+    let kw_pattern = null;
+    if (typeof mode.keywords === "object") {
+      kw_pattern = mode.keywords.$pattern;
+      delete mode.keywords.$pattern;
```
egor-rogov, line 214: Can't figure out... why delete?

docs/language-guide.rst
yyyc514: I was upgrading not bug fixing. :-)

yyyc514: And I don't really know for sure so I left it alone.

src/lib/mode_compiler.js
yyyc514: So it doesn't get passed into the keyword compiler engine along with other groups.

docs/language-guide.rst
egor-rogov: That's right, but I'm talking about the comment you've added (:  It is misleading.

yyyc514: Oh yes you're right! I will fix that.

src/lib/mode_compiler.js
egor-rogov: Ok, now I see.

docs/language-guide.rst
yyyc514: Oh because it's the docs and I was trying to be helpful, I'l lcorect it to be more specific.

PR #2524, opened 2020-05-02
fix(parser) Fix freezing issue with illegal 0 width matches

### LINKED ISSUE #2522: Infinite loop for illegal tokens when `ignore_illegals` or `highlightAuto`
On 10.0.1, the following XML:

```xml
<?xml version="1.0"?>
<response value="ok" xml:lang="en">
  <text>Ok</text>
  <comment html_allowed="true"/>
  <ns1:description><![CDATA[
  CDATA is <not> magical.
  ]]></ns1:description>
  <a></a> <a/>
</response>
```

When passed through `highlightAuto` or `highlight` with `ignore_illegals` turned on, will cause an infinite loop, specifically, because of `'$'` as an illegal in CSS selector attributes:

https://github.com/highlightjs/highlight.js/blob/e8624ccb65cfd859a0273b58ec3897afd16a5498/src/languages/css.js#L72

yyyc514:
Closes #2522

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -373,6 +373,15 @@ const HLJS = function(hljs) {
         return 1;
       }
 
+      // infinite loops are BAD, this is a last ditch catch all. if we have a
+      // decent number of iterations yet our index (cursor position in our
+      // parsing) still 3x behind our index then something is very wrong
+      // so we bail
+      if (iterations > 100000 && iterations > match.index * 3) {
+        const err = new Error('potential infinite loop, way more iterations than matches');
+        throw err;
+      }
+
       /*
       Why might be find ourselves here?  Only one occasion now.  An end match that was
       triggered but could not be completed.  When might this happen?  When an `endSameasBegin`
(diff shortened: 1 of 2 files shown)


yyyc514:
@allejo @egor-rogov Will publish this tomorrow if no one reports an issue until then.

wooorm:
Thanks Josh for the quick fix! ✨

yyyc514:
10.0.2 being released now.

src/highlight.js
```diff
@@ -365,6 +365,23 @@ const HLJS = function(hljs) {
...
+      }
+
+      // infinite loops are BAD, this is a last ditch catch all. if we have a
+      // decent number of iterations yet our index (cursor position in our
+      // parsing) still 3x behind our index then something is very wrong
+      // so we bail
+      if (iterations > 100000 && iterations > match.index * 3) {
+        const err = new Error('potential infinite loop, way more iterations than matches');
+        throw err;
+      }
```
egor-rogov, line 383: Sorry I'm late to the party. Quick fix was the right thing to do.
I'd move this check out to the `for (;;)` loop, in order not to deal with `iterations` inside `processLexeme` function.

yyyc514: Except then we'd literally have to check it ALL the time... where-as right now it's more of a fallback... but I'll play around with it.

yyyc514: But technically your right in that if there were a bug in the above code that this wouldn't be well suited to catch it... :-/

PR #2526, opened 2020-05-03
added rails-log as an 3rd-party language

sweetppro:
modified the ruby.js language to provide better syntax highlighting for Rails Unicorn logging.


yyyc514:
I didn't look closely before.  How is this different from Ruby?

Did you just add

```
  'Rendering Rendered Completed Started POST Processing by Redirected to Delivered mail Performed GET'
```

?

yyyc514:
Why would there be raw Ruby code in the output of the Rails log? Like an exception?

sweetppro:
> Did you just add
> 
>   'Rendering Rendered Completed Started POST Processing by Redirected to Delivered mail Performed GET'
> ?
> 


Theres a number of small regex tweaks which give a much better result in Unicorn logging
If the changes could be merged into the ruby language, that would be maybe cleaner than a whole new third party language

> Why would there be raw Ruby code in the output of the Rails log? Like an exception?
> 
Yup

yyyc514:
> Theres a number of small regex tweaks which give a much better result in Unicorn logging

I'd be curious to see a diff against `ruby.js`... but this sounds like a pretty specific use case...

I was about to say with only the keywords you could do this with a few lines of code... (see `arduino.js`...)... but now I'm curious what these other tweaks might have been.

sweetppro:
here's an example:

```
I, [2020-05-03 13:57 #21962]  INFO -- : [c1bbbae9-2bcb] [0.0.0.0]   Rendered vendor/bundle/ruby/2.6.0/gems/exception_notification-4.4.0/lib/exception_notifier/views/exception_notifier/exception_notification.text.erb (Duration: 27.4ms | Allocations: 6416)
I, [2020-05-03 13:57 #21962]  INFO -- : [c1bbbae9-2bcb] [0.0.0.0] Delivered mail 5eaeb18edfdc_55ca2ada442f56546042e@lvps92-51-148-79.dedicated.hosteurope.de.mail (1729.9ms)
F, [2020-05-03 13:57 #21962] FATAL -- : [c1bbbae9-2bcb] [0.0.0.0]   
[c1bbbae9-2bcb] [0.0.0.0] ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken):
[c1bbbae9-2bcb] [0.0.0.0]   
[c1bbbae9-2bcb] [0.0.0.0] actionpack (6.0.2.2) lib/action_controller/metal/request_forgery_protection.rb:217:in `handle_unverified_request'
[c1bbbae9-2bcb] [0.0.0.0] actionpack (6.0.2.2) lib/action_controller/metal/request_forgery_protection.rb:249:in `handle_unverified_request'
[c1bbbae9-2bcb] [0.0.0.0] actionpack (6.0.2.2) lib/action_controller/metal/request_forgery_protection.rb:244:in `verify_authenticity_token'
[c1bbbae9-2bcb] [0.0.0.0] activesupport (6.0.2.2) lib/active_support/callbacks.rb:429:in `block in make_lambda'
```

yyyc514:
So far it looks like the better/simpler choice would be to concentrate on parsing the log file itself and use the real `Ruby` as a sublanguage just for the relevant part of the stack trace.

yyyc514:
If this is mostly/solely for unicorn I'm wondering if unicorn-rails-log might be more appropriate?

sweetppro:
here's a diff:

[rails-log.txt](https://github.com/highlightjs/highlight.js/files/4572058/rails-log.txt)

sweetppro:
oops, closed accidentally

yyyc514:
What kind of diff is that? My editor doesn't like it one bit and won't color it.

sweetppro:
I created it like so:
`git diff > rails-log.patch`
then renamed it rails-log.txt to upload

yyyc514:
Ok, wow most themes have terrible support for diffs in VS code.  Annoying. :)

yyyc514:
Yeah all these changes to work around line endings would probably have been better handled by Ruby as a sublanguage. And then only turn on the Ruby parsing for those parts of the line that are actually Ruby code.  And the parent rule could always force the Ruby to end at the end of each line and then you wouldn't need to hack Ruby itself.

Any thoughts on the naming?

sweetppro:
I can change the name no problem.
I'll have another look in the morning, and see if I can figure out how to use Ruby as a sub language.

yyyc514:
We have reasonable documentation and you can look at some of our grammars for samples... like XML can includes CSS and JS sublanguages, etc...

sweetppro:
nice one thx.
ill take a look

sweetppro:
new pull request here:
https://github.com/highlightjs/highlight.js/pull/2528

yyyc514:
This PR would have still worked, but whatever. :)  Thanks for contributing!

PR #2531, opened 2020-05-05
fix(javascript) fix regex inside parens after a non-regex

### LINKED ISSUE #2530: (javascript) Incorrect highlight after '/' symbol
**Describe the issue**
In some cases, the highlight is not correct after the '/' symbol.
Bellow is 2 examples.

*Which language seems to have the issue?**
`javascript`

`highlight`

**Sample Code to Reproduce**
Example 1:
```
const x = `${duration / 1000}`;
func();
```

Example 2:
```
a(b, /.*c.*/);
func();
```

**Expected behavior**
In the example 1, the text after the '/' symbol doesn't change highlight color (including the symbol itself)

In the example 2 highlight after regexp must be correct

**Additional context**
It seems, problem appears in 9.18.1. I can't reproduce the problem with 9.18.2

yyyc514:
Resolves #2530.


src/languages/javascript.js
```diff
@@ -148,13 +149,30 @@ export default function(hljs) {
...
+          // <comments up next>` and then immediately end after the , because it
+          // fails to find any actual attrs. But this still does the job because
+          // it prevents the value contain rule from grabbing this instead and
+          // prevening this rule from firing when we actually DO have keys.
+          regex.lookahead(regex.concat(
+            // we also need to allow for multiple possible comments inbetween
+            // the first key:value pairing
+            /(\/\/.*\s*)*/,
+            /(\/\*.*\*\/\s*)*/,
+            IDENT_RE + '\\s*:'))),
```
egor-rogov, line 168: This trick doesn't work in case like this:
```
  key5: value, /* with a multiline comment */ // another comment
  key6: value,
```
or like this:
```
  key5: value, /* with a REAL multiline
comment */
  key6: value,
```

yyyc514: Good catch, pretty sure I can fix those cases.

yyyc514: Do you understand why the last one doesn't work?

yyyc514: Better?

egor-rogov: Yes, it seems fine.
(Still it looks somewhat ugly, although I can't see how to handle it more elegantly. We can't express it by saying "first match all comments and then just ignore them altogether")

yyyc514: Our syntax isn't expressive enough. :-)  You could have something like:

```js
alwaysIgnore: COMMENTS,
match: ["{", followedBy: "[ident]: "]
```

Where match is pretty much any sequence and then the always ignore item would turn into a huge regex inserted in the middle that was always ignored. Although I guess here I'm talking about look-ahead where it's easy... as soon as you want to emit them while you're ignoring them then it's a whole other thing. :-)

Although I'm not sure that'd be much different actually... just a way of inserting random modes in-between any other modes... seems most useful for comments... but for another day...

PR #2532, opened 2020-05-05
(fix) `fixMarkup` would rarely destroy markup when `useBR` was enabled

### LINKED ISSUE #2529: fixMarkup `useBR: true` removes HTML tag if it starts at index 0 of string
`fixMarkup` and `highlightBlock` methods with `{useBR: true}`  break markup.

I noticed, that `hljs.highlight(...)` generates the following code:

`"<span class="hljs-symbol">[Created By]</span>"`

Then `fixMarkup` for this string generates the following:

`"[Created By]</span>"`

Just comment `configure` method call and you'll see the difference (console output provided as well)
https://jsfiddle.net/t1maqeb7/6/

I believe it should not break markup this way.

yyyc514:
Will resolve #2529.


test/api/fixmarkup.js
```diff
@@ -5,11 +5,21 @@ const hljs   = require('../../build');
...
 describe('.fixmarkup()', () => {
   after(() => {
-    hljs.configure({ useBR: false })
-  })
+    hljs.configure({ useBR: false });
+  });
+
+  it('should not strip HTML from beginning of strings', () => {
+    hljs.configure({ useBR: true });
+    const value = '<span class="hljs-attr">"some"</span>: \n <span class="hljs-string">"json"</span>';
```
andrewQwer, line 13: the issue occurs when there is no any `\n` in the original string.
So test config looks like this: 
```
 hljs.configure({ useBR: true });
    const value = '<span class="hljs-attr">"some"</span>';
```

yyyc514: Yes, because the regex matches against ^... so any content (just not a newline) would prevent you from seeing this bug.  What I need to know is WHY we have this behavior in the first place - because it looks very intentional.

src/highlight.js
```diff
@@ -32,7 +32,7 @@ const HLJS = function(hljs) {
   var SAFE_MODE = true;
 
   // Regular expressions used throughout the highlight.js library.
-  var fixMarkupRe = /((^(<[^>]+>|\t|)+|(?:\n)))/gm;
+  var fixMarkupRe = /((^(<[^>]+>|\t|)+|\n))/gm;
```
allejo, line 35:
```suggestion
  var fixMarkupRe = /(^(<[^>]+>|\t|)+|\n)/gm;
```

yyyc514:
> I have not looked but do we have a unit test that will show fixMarkupRe's purpose? 

Yes, `customMarkup.js`. I just had no idea until you pointed me to it.

allejo:
> > I have not looked but do we have a unit test that will show fixMarkupRe's purpose?
> 
> Yes, `customMarkup.js`. I just had no idea until you pointed me to it.

I know we have this test case for making sure leading tabs and tabs in tags are replaced.

https://github.com/highlightjs/highlight.js/blob/04e4991d66657bd3c23d5356e069591f281910c6/test/fixtures/index.html#L46-L50

But I don't see if there are's a test confirming tabs in the "content" are _not_ replaced. Similar to how I couldn't replicate it yesterday until you explained the difference of tabs in the "content"

yyyc514:
> But I don't see if there are's a test confirming tabs in the "content" are not replaced. Similar to how I couldn't replicate it yesterday until you explained the difference of tabs in the "content"

Oh no, I don't think we have tests for that... I'd rather forget about this entirely - are you going to make me go and write more tests? :-)

allejo:
> are you going to make me go and write more tests? :-)

I think it'd be a good idea to have if this is going to be revisited in the future and the implementation is going to be changed; that way we can ensure consistent behavior. Even though this shouldn't be part of the public API it's still used by `highlightBlock`, which is part of the public API so we could inadvertently change behavior.

But, since it took a whole 5 major versions for this to be discovered, it's such an edge case that I won't block it.

yyyc514:
I wish we had some tally of how many people used these more obscure features.

sa-orlov (commented):
Hi

Issue #2523, opened 2020-05-02
Add support for `registerAlias` similar to `registerLanguage`


wooorm:
**Is your request related to a specific problem you're having?**

Aliases are useful for example to support [GitHub’s names](https://github.com/github/linguist/blob/40992ba7f86889f80dfed3ba95e11e1082200bad/lib/linguist/languages.yml#L4987-L4991), and map them to highlightjs names.

**The solution you'd prefer / feature you'd like to see added...**

Such as like so: https://github.com/wooorm/lowlight#lowregisteraliasname-alias, but anything that allows mapping a new label to an (existing) language name.

**Any alternative solutions you considered...**

There is no access to `languages` or `aliases`, so it’s not possible to do this in userland.

**Additional context...**

n/a

yyyc514:
FYI: You can do this now by just getting a languages raw interface and re-registering the language (which will update it's aliases).

I'm not opposed to a simple registerAlias method though:

```
registerAlias(name, alias|list)
```

I don't think the second form that you support is necessary at all since it seems an edge case and easy enough to write a loop.  And if we added this we'd of course need to add docs for it.

Want to make a PR?

wooorm:
The reason for `registerAlias` as opposed to something else is that it aligns with `registerLanguage`, but I don’t feel strongly about that.

The reason for the other option is so that a module could expose that mapping between highlightjs names and GH aliases as an object, which could be required and then directly passed.

yyyc514:
I don't love `register*` but as you say at least it's consistent. :-)

Sure, but you can write 2 lines of code to use `forEach` to do the same thing... :-)  I just think the core library should provide low-level APIs and integrators build on top of those.  Your library can support the more complex version if you think that's best, and then loop and hook into our lower-level method.

yyyc514:
@wooorm You willing to whip up a PR for this one?

wooorm:
I’m a bit swamped at the moment; if @taufik-nurrohman is interested, I would be grateful!

yyyc514:
Closing as resolved by #2540

PR #2540, opened 2020-05-07
Add hljs.registerAlias Method

taufik-nurrohman:
**Related:**

 - https://github.com/highlightjs/highlight.js/issues/2523
 - https://github.com/highlightjs/highlight.js/issues/2371

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5,7 +5,7 @@ Highlight.js exports a few functions as methods of the ``hljs`` object.
 
 
 ``highlight(languageName, code, ignore_illegals, continuation)``
----------------------------------------------------------
+----------------------------------------------------------------
 
 Core highlighting function.
 Accepts a language name, or an alias, and a string with the code to highlight.
@@ -32,7 +32,7 @@ Returns an object with the following properties:
 
 
 ``highlightAuto(code, languageSubset)``
-----------------------------------------
+---------------------------------------
 
 Highlighting with language detection.
 Accepts a string with the code to highlight and an optional array of language names and aliases restricting detection to only those languages. The subset can also be set with ``configure``, but the local parameter overrides the option if set.
@@ -76,7 +76,7 @@ Configures global options:
 * ``classPrefix``: a string prefix added before class names in the generated markup, used for backwards compatibility with stylesheets.
 * ``languages``: an array of language names and aliases restricting auto detection to only these languages.
 * ``languageDetectRe``: a regex to configure how CSS class names map to language (allows class names like say `color-as-php` vs the default of `language-php`, etc.)
-* ``noHighlightRe``: a regex to configure which CSS classes are to be skipped completely
+* ``noHighlightRe``: a regex to configure which CSS classes are to be skipped completely.
 
 Accepts an object representing options with the values to updated. Other options don't change
 ::
@@ -85,15 +85,14 @@ Accepts an object representing options with the values to updated. Other options
     tabReplace: '    ', // 4 spaces
     classPrefix: ''     // don't append class prefix
                         // … other options aren't changed
-  })
+  });
   hljs.initHighlighting();
 
 
 ``initHighlighting()``
 ----------------------
 
-Applies highlighting to all ``<pre><code>..</code></pre>`` blocks on a page.
-
+Applies highlighting to all ``<pre><code>…</code></pre>`` blocks on a page.
 
 
 ``initHighlightingOnLoad()``
@@ -113,11 +112,19 @@ Adds new language to the library under the specified name. Used mostly internall
   to use common regular expressions defined within it.
 
 
+``registerAlias(alias|aliases, {languageName})``
+------------------------------------------------
+
+Adds new language alias or aliases to the library for the specified language name defined under ``languageName`` key.
+
+* ``alias|aliases``: a string or array with the name of alias being registered
+* ``languageName``: the language name as specified by ``registerLanguage``.
+
+
 ``listLanguages()``
-----------------------------
+-------------------
 
 Returns the languages names list.
-
 
 
 .. _getLanguage:
@@ -132,7 +139,7 @@ Returns the language object if found, ``undefined`` otherwise.
 
 
 ``requireLanguage(name)``
----------------------
+-------------------------
 
 Looks up a language by name or alias.
 
@@ -150,5 +157,5 @@ Enables *debug/development* mode.  **This mode purposely makes Highlight.js more
 
 For example, if a new version suddenly had a serious bug (or breaking change) that affected only a single language:
 
-* **In Safe Mode**: All other languages would continue to highlight just fine.  The broken language would appear as a code block, but without any highlighting (as if it were plaintext).
+* **In Safe Mode**: All other languages would continue to highlight just fine. The broken language would appear as a code block, but without any highlighting (as if it were plaintext).
 * **In Debug Mode**: All highlighting would stop when an error was encountered and a JavaScript error would be thrown.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -92,7 +92,7 @@ Accepts an object representing options with the values to updated. Other options
 ``initHighlighting()``
 ----------------------
 
-Applies highlighting to all ``<pre><code>…</code></pre>`` blocks on a page.
+Applies highlighting to all ``<pre><code>...</code></pre>`` blocks on a page.
 
 
 ``initHighlightingOnLoad()``
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -685,6 +685,14 @@ const HLJS = function(hljs) {
     return languages[name] || languages[aliases[name]];
   }
 
+  function registerAlias(name, names) {
+    if (typeof names === 'string') {
+      aliases[names] = name;
+    } else {
+      names.forEach(function(alias) { aliases[alias] = name; });
+    }
+  }
+
   function autoDetection(name) {
     var lang = getLanguage(name);
     return lang && !lang.disableAutodetect;
@@ -716,6 +724,7 @@ const HLJS = function(hljs) {
     registerLanguage,
     listLanguages,
     getLanguage,
+    registerAlias,
     requireLanguage,
     autoDetection,
     inherit,
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -685,11 +685,12 @@ const HLJS = function(hljs) {
     return languages[name] || languages[aliases[name]];
   }
 
-  function registerAlias(name, names) {
-    if (typeof names === 'string') {
-      aliases[names] = name;
-    } else {
-      names.forEach(function(alias) { aliases[alias] = name; });
+  function registerAlias(alias, {languageName}) {
+    let list = alias;
+    if (typeof list === 'string') {
+      list = [alias]
+    }
+    list.forEach(alias => aliases[alias] = languageName);
     }
   }
 
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -691,7 +691,6 @@ const HLJS = function(hljs) {
       list = [alias]
     }
     list.forEach(alias => aliases[alias] = languageName);
-    }
   }
 
   function autoDetection(name) {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Parser Engine:
 
 - (parser) Adds `keywords.$pattern` key to grammar definitions (#2519) [Josh Goebel][]
 - (parser) Adds SHEBANG utility mode [Josh Goebel][]
+- (parser) Adds `registerAlias` method (#2540) [Taufik Nurrohman][]
 - (enh) Added `on:begin` callback for modes (#2261) [Josh Goebel][]
 - (enh) Added `on:end` callback for modes (#2261) [Josh Goebel][]
 - (enh) Added ability to programatically ignore begin and end matches (#2261) [Josh Goebel][]
--- a/test/api/registerAlias.js
+++ b/test/api/registerAlias.js
@@ -1,0 +1,24 @@
+'use strict';
+
+const hljs   = require('../../build');
+const should = require('should');
+
+describe('.registerAlias()', () => {
+  it('should get an existing language by alias', () => {
+    hljs.registerAlias('jquery', {
+      languageName: 'javascript'
+    });
+    const result = hljs.getLanguage('jquery');
+
+    result.should.be.instanceOf(Object);
+  });
+
+  it('should get an existing language by aliases', () => {
+    hljs.registerAlias(['jquery', 'jqueryui'], {
+      languageName: 'javascript'
+    });
+    const result = hljs.getLanguage('jquery');
+
+    result.should.be.instanceOf(Object);
+  });
+});


src/highlight.js
```diff
@@ -685,6 +685,14 @@ const HLJS = function(hljs) {
     return languages[name] || languages[aliases[name]];
   }
 
+  function registerAlias(name, names) {
```
yyyc514, line 688:
```suggestion
  function registerAlias(name, langugeNames) {
```

src/highlight.js
```diff
@@ -685,6 +685,14 @@ const HLJS = function(hljs) {
...
+    if (typeof names === 'string') {
+      aliases[names] = name;
+    } else {
+      names.forEach(function(alias) { aliases[alias] = name; });
```
yyyc514, line 692:
```suggestion
  function registerAlias(aliases, {languageName}) {
    if (typeof aliases === 'string') {
      aliases = [aliases]
    }
    aliases.forEach(alias => aliases[alias] = languageName);
```

yyyc514, line 692:
```suggestion
  function registerAlias(alias, {languageName}) {
    let list = alias;
    if (typeof list === 'string') {
      list = [alias]
    }
    list.forEach(alias => aliases[alias] = languageName);
```

yyyc514:
Nice. Changelog and perhaps a simple test or two?

src/highlight.js
yyyc514: Still don't love the naming, our global here is annoying when it competes for namespace with the arguments.

yyyc514:
Oh and you're going to need to update some docs also to add this since it's now public API.

src/highlight.js
taufik-nurrohman: Would you think it be better to pass a function to it too so that the API would have the same usage as `registerLanguage`, to prevent confusion?

~~~ .js
hljs.registerAlias(['jquery', 'jqueryui', function(hljs) {
    return {
        languageName: 'javascript'
    };
});
~~~

yyyc514: No, that's because languages are executable code... no need for that complexity here.

taufik-nurrohman: Just curious. What makes you decided to use `languageName` key and not `language`?

yyyc514: They are two different concepts - and it causes all sorts of headaches elsewhere int he codebase when they are used interchangably.

- `language` <Language> the compiled language object returned by a grammar function
- `languageName` <string> "javascript", etc.

yyyc514: Actually it's messier than that since "cpp" isn't a language name at all (the name is "C++"), it's an identifier... really just another alias!... but that's a problem for another day.

taufik-nurrohman: Okay.

docs/api.rst
```diff
@@ -85,15 +85,14 @@ Accepts an object representing options with the values to updated. Other options
...
+  });
   hljs.initHighlighting();
 
 
 ``initHighlighting()``
 ----------------------
 
-Applies highlighting to all ``<pre><code>..</code></pre>`` blocks on a page.
-
+Applies highlighting to all ``<pre><code>…</code></pre>`` blocks on a page.
```
yyyc514, line 95: I don't know about this change...

docs/api.rst
```diff
@@ -150,5 +157,5 @@ Enables *debug/development* mode.  **This mode purposely makes Highlight.js more
 
 For example, if a new version suddenly had a serious bug (or breaking change) that affected only a single language:
 
-* **In Safe Mode**: All other languages would continue to highlight just fine.  The broken language would appear as a code block, but without any highlighting (as if it were plaintext).
+* **In Safe Mode**: All other languages would continue to highlight just fine. The broken language would appear as a code block, but without any highlighting (as if it were plaintext).
```
yyyc514, line 160: LOL.

docs/api.rst
```diff
@@ -113,13 +112,21 @@ Adds new language to the library under the specified name. Used mostly internall
...
+
+Adds new language alias or aliases to the library for the specified language name defined under ``languageName`` key.
+
+* ``alias|aliases``: a string or array with the name of alias being registered
+* ``languageName``: the language name as specified by ``registerLanguage``.
+
+
 ``listLanguages()``
-----------------------------
+-------------------
```
yyyc514, line 125: Thanks for cleaning all these up, or is that a linter of some sort?

docs/api.rst
taufik-nurrohman: Forgive me about my perfections. Do I need to revert it back?

docs/api.rst
taufik-nurrohman: No. it&rsquo;s just me.

docs/api.rst
yyyc514: Maybe I'm just old school for preferring ASCII... I'd have turned it into three periods personally.  If you don't mind, but I don't care enough to force the issue.  Is it common to use the ... thingy in mostly ASCII text documents?

taufik-nurrohman: > Is it common to use the ... thingy in mostly ASCII text documents?

I don&rsquo;t think so. But some users prefer post-parsing those characters into some typographical characters like `…`, `‘`, `’`, `“`, `”`, `–` and `—` in the HTML output.

Will convert it to `...` ASAP!

yyyc514: >  But some users prefer post-parsing those characters into some typographical characters like

Yeah I've always thought of it as a post-parsing concern of the renderer...

src/highlight.js
```diff
@@ -685,6 +685,15 @@ const HLJS = function(hljs) {
     return languages[name] || languages[aliases[name]];
   }
 
+  function registerAlias(alias, {languageName}) {
+    let list = alias;
+    if (typeof list === 'string') {
+      list = [alias]
+    }
+    list.forEach(alias => aliases[alias] = languageName);
+    }
```
yyyc514, line 694: Extra }?

src/highlight.js
```diff
@@ -685,6 +685,15 @@ const HLJS = function(hljs) {
...
+  }
+
```
yyyc514, line 696:
```suggestion
  function registerAlias(alias, {languageName}) {
    let list = alias;
    if (typeof list === 'string') {
      list = [alias]
    }
    list.forEach(alias => aliases[alias] = languageName);
  }

```

wooorm:
Any reason why `languageName` is in an object, and not directly a parameter? Are there other values expected? Isn’t `languageName` the only really essential part of this call, whereas the aliases are optional?

yyyc514:
The older and wiser I get the any function that takes more one or two unnamed parameters is a code smell - Objective C got this right long ago with named arguments.  But the short form of the idea is that often functions are verbs that act on nouns... so the first param is often clear from context.

```
startEngine(a,...)
registerAlias(b,...)
```

But the second, third, forth... all start to be completely meaningless... the closest thing we have to named arguments in JS are objects, and the new destructuring stuff names it quite easy to work with them in this way. (it wasn't hard before, but it's nicer now)

This way it's 100% clear what parameter is which.

```js
registerAlias("js", { languageName: 'javascript' });
```

yyyc514:
> Isn’t languageName the only really essential part of this call, whereas the aliases are optional?

I don't know what this means.  You can't register an alias without telling it what language to point to... Both arguments are "essential".

wooorm:
As one could pass an empty list of aliases, it is not far off to see no alias the same as a list of no aliases.

yyyc514:
> As one could pass an empty list of aliases

Could, but that would serve no purpose.

yyyc514:
If I were being fancier I'd go with more English language:

> register the alias "js" with the language: "javascript"

Translates to:

```
registerAlias("js", { withLanguage: 'javascript' });
```

wooorm:
> Could, but that would serve no purpose

Indeed, it should be a noop. Whereas no language should be an error.

---

I’m very glad this isn’t Java with its extremely long names. Or natural language.

For ordering: `key, value|values` is more in line with how other APIs, at least in JavaScript, work.

yyyc514:
> For ordering: key, value|values is more in line with how other APIs, at least in JavaScript, work.

I don't think it's very useful to say "other APIs" as a general term as if all APIs were to be somehow treated the same - although if you had one particular article that made a general case fairly well I wouldn't mind reading it...  This type of thinking makes big assumptions about the internal data storage. _It's not the callers concern to know how this data is represented inside the library._

From the callers perspective they are "registering an alias" (the name of the function you recommended) and attaching it to a given language. There isn't necessarily any key or value here.

The alias IS actually the literal key here and the language the value, FYI... but again I only use that to make the point of how silly it is to base the ordering purely on the internal data structures. I wouldn't use that to support the ordering I've chosen - because again how it's stored internally is irrelevant. :-)

yyyc514:
> Indeed, it should be a noop. Whereas no language should be an error.

An empty array would be a noop right now, yes.  No alias at all should be an error, just like no language should be.  And I could imagine someone making a good argument for an empty array being an error condition too...

I get where you're going _sort of_, but I just don't agree.** The whole idea makes no sense without both arguments.** 

```
registerAlias("javascript") // passing a language
```

Makes no sense... I can't do anything, I have no aliases, I have no subject for the verb.  I could error or do literally nothing.

```
registerAlias("js") // passing an alias
```
I could actually do something here... I could keep a list of aliases waiting to be registered with languages, etc... `pendingAliases()`.  Doesn't seem super useful for us... **but since I have both the verb and subject I have no options.**

yyyc514:
Anyways, if you have an article I'm happy to read it - I do try to keep an open mind... but right now verb/noun seems crystal clear here to me (as I've explained above) - and I find this verb/noun pattern to be extremely common as a pattern for naming functions.

My only thought is that it may have been better named `registerAliases`... revealing the full power of the function - that it takes an array - which probably could be an array of 1.  Where as `registerAlias` doesn't reveal the use case of passing multiple items.

wooorm:
> An empty array would be a noop right now, yes. No alias at all should be an error, just like no language should be.

The reason for my thinking is to allow for the current code:

https://github.com/highlightjs/highlight.js/blob/8c7422958b8cca585e1866cde0d9f9ea6c7a8e43/src/highlight.js#L660-L662

To be written as:

```js
if (lang.aliases) {
  registerAlias(name, lang.aliases)
}
```

Or even:

```js
registerAlias(name, lang.aliases)
```

...instead of:

```js
if (lang.aliases) {
  registerAlias(lang.aliases, {languageName: name})
}
```

My proposal is based on my mental model of alternative names: there is a primary name, and there are zero or more alternative names. That it is stored in reverse, is indeed an implementation detail that I don’t need to know about.
I think of this as a setter, hence my comparison to `map.set(key, value)`, `element.setAttribute(key, value)`, etc.
One could think up other examples such as getters (where I prefer `getter(name)` instead of `getter({languageName: name})`) or a way to clear aliases (`delete(name)`)?  All probably not very useful, but this hopefully explains why I think what I think.

yyyc514:
This seems completely clear to me:

```js
if (lang.aliases) {
  registerAliases(lang.aliases, {languageName})
}
```

Again there are no keys or values here, this isn't a Map.  **It's not a setter either, by it's very name.**  Your argument would hold up better if the language was the noun, but it's not.

```js
if (lang.aliases) {
  alterLanguage(languageName, addAliases: lang.aliases)
}
```

I would still have the conditional for clarity.   **I don't think passing null to the function and treating that the same as an empty array is a great habit to get into.** (Writing TypeScript has made me think a lot harder about input and output APIs).  So I think your stated goal here (`registerAlias(name, lang.aliases)`) is NOT desirable. Explicit almost always tops implicit for me.

> One could think up other examples such as getters (where I prefer getter(name) instead of getter({languageName: name}))

Not sure I follow... this isn't [as large a] problem for single argument functions.

```js
getLanguage(a)
```
Although this isn't immediately clear either... We can be sure what is returned by not was is passed, there could be multiple ways to find a language... by name, by unique ID, etc...

```js
getLanguage("C++")
getLanguage("cpp")
getLanguage(32)
// vs
getLanguageBy({prettyName: "C++"})
getLanguageBy({name: "cpp"})
getLanguageBy({id: 32})
```

**I think we're just going to have to agree to disagree.**

yyyc514:
**It's also not fully correct to think of aliases as purely an attribute of a single language** - even more so now that we can add them dynamically - opening them up to all sorts of creative uses.   I've already seriously considered things like the following:

```js
registerAlias("php", languageNames: ["php_snippet", "php_template"])
```

IE, PHP could refer to either language... perhaps decided based on relevancy, etc... and we actually have another language where there are two languages with the same name... the kind of thing you could solve with more powerful aliases, also...

```js
registerAlias("basic", languageNames: ["trs100_basic", "microsoft_basic","c64_basic"])
```

yyyc514:
I do think I finally understand you, but for me **it's all about the naming**. (that and my last point that I don't think of aliases are purely attributes of a language)

For me though as soon as the function name was `registerAlias(es)` (which is a reasonable enough name, and congruent with `registerLanguage` as you pointed out) there is no discussion to be had, the subject/first argument is going to be alias/aliases. _Again, I think this verb/noun pattern is super common and used everywhere._ 

IF we had a different name then a different API might make sense (as I mentioned above somewhere). But we started with the name and we ended here.

yyyc514:
The swift guidelines make a lot of sense to me:

https://swift.org/documentation/api-design-guidelines/#argument-labels

yyyc514:
The explicit naming also makes it clear that we expect a _language name_, not a language object... 

```js
var javascript = hljs.getLanguage("javascript")
// which is correct?  Now way to know.
hljs.registerAlias("js", javascript)
hljs.registerAlias("js", "javascript")
```

It's completely unclear which would be the correct form (**perhaps neither!**), since you have no clue what that second argument is supposed to be.

yyyc514:
I just had an interesting discussion where a colleague pointed out to me that native language could have an effect on what feel naturals regarding naming. 

> I think native language may have an effect though on what feels natural to someone. Word order being rather different, as I recall, in German vs. English.

I never even considered that aspect and I definitely write code from a purely English grammatical perspective.  It's possible that a language that sequences things entirely different could lend itself to a different naming approach. Hard to say for certain, but it's easy to imagine. But English is all I have to work with.

wooorm:
Thanks for all your comments! Yes, I do believe natural language is important. What I’m mostly worried about is that you point to Swift guidelines, which has named parameters, whereas JavaScript does not. By accepting an object, you can indeed label them, but for me objects imply optionality and that multiple fields are (or will be) supported.

I still feel I prefer `addIngredient(salad, 'tomato')` or `addIngredient(pasta, 'capers')` over the inverse. And would think `addIngredient(soup)` a noop as there is no ingredient. But indeed, we do disagree about this style. And that’s fine!

yyyc514:
>  What I’m mostly worried about is that you point to Swift guidelines, which has named parameters, whereas JavaScript does not.

Objects are the tool we have in JS to get named arguments.  This was also a SUPER common pattern in Ruby also until we got real named arguments.  I'd imagine it's a pattern in other languages, also. This is what type definitions and documentation is for. :-) I pointed to that because it was an easy find vs spending more time looking for a JS example of someone preaching the same thing.

> but for me objects imply optionality and that multiple fields are (or will be) supported.

You know what they say about implications, no wait that's assumptions, but still. :-)  Consider letting go of that bias and I think you'll find some amazing new options in JS land.

> And that’s fine!

Indeed. :-)

PR #2541, opened 2020-05-08
Recognize `priority_queue` `pair` as cpp containers

Linhk1606:


yyyc514:
What is the C++ terminology for this? Are they actually called containers?

Linhk1606:
Yes. They are called "containers" formally. They are special templates that ability to manipulate multiple data structures and algorithms

src/languages/c-like.js
```diff
@@ -213,7 +213,7 @@ export default function(hljs) {
       [
       PREPROCESSOR,
       {
```
yyyc514, line 215:
```suggestion
      { // containers: ie, `vector <int> rooms (9);`
```

PR #2543, opened 2020-05-09
(swift) @objcMembers not completely highlighted

nicked:
Would match `@objc` first, and the `Members` part would be unhighlighted

### DIFF
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -125,11 +125,11 @@ export default function(hljs) {
       {
         className: 'meta', // @attributes
         begin: '(@discardableResult|@warn_unused_result|@exported|@lazy|@noescape|' +
-                  '@NSCopying|@NSManaged|@objcMembers|@objc|@convention|@required|' +
+                  '@NSCopying|@NSManaged|@objc|@objcMembers|@convention|@required|' +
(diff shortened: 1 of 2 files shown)


src/languages/swift.js
```diff
@@ -125,7 +125,7 @@ export default function(hljs) {
       {
         className: 'meta', // @attributes
         begin: '(@discardableResult|@warn_unused_result|@exported|@lazy|@noescape|' +
-                  '@NSCopying|@NSManaged|@objc|@objcMembers|@convention|@required|' +
+                  '@NSCopying|@NSManaged|@objcMembers|@objc|@convention|@required|' +
```
yyyc514, line 128: How are these used in code?  Should we perhaps just add a \b before and after?

yyyc514: Or at least after? IE, this looks like the kind of thing that would keep coming up every time they added a new keyword that started with @objc, or same for any of the other keywords...

nicked: I don't know enough about how this matching works, but if it's a standard regex, then looking for a non-word char would be better. I'd suggest `\W` though, because we're only looking for anything that isn't a continuation of the word

nicked: Ah you're right, a trailing `\b` works best, don't want to include the extra matched char. Thanks

PR #2544, opened 2020-05-11
Add basic Vue.js component and examples.

### LINKED ISSUE #2535: Propose: Better support Vue simple use cases
## Vue

The two examples I found fast:

- https://github.com/metachris/vue-highlightjs/blob/master/index.js
- https://github.com/ItsJimi/v-highlightjs/blob/master/index.js

Both use directives... wouldn't you really want a component for this though? What I have so far for Vue:

```html
<highlightjs language="javascript" code="var x = 5;" />
<highlightjs language="javascript" v-bind:code="sourceCode" ignore_illegals />
<highlightjs v-bind:code="sourceCode" />
<!-- or to be explicit -->
<highlightjs auto :code="sourceCode" />
<highlightjs autodetect :code="sourceCode" />
```
Using a slot for passing in code has all the same problems of using a non-PRE tag in HTML and many additional problems so that doesn't seem like a good route, so I've rule that out.

### Vue questions:

- Component vs directive.
- Dash or no dash or is that something that you should configure when you setup the plugin, ie using a custom tag name?
- Would people be expecting to configure the whole library via HTML "attributes" though?  I suppose you could have an options attribute thought that took an object...

joshgoebel:
Resolves #2535.


yyyc514:
Still need to go back and clean up the developer stuff to properly integrate it.  Figured developer tool is an easy please to eat our own dogwood on this one.

egor-rogov:
Do I understand correctly that this is still WIP?

yyyc514:
Only the developer tool tweaks. I think the plugin itself is good.

yyyc514:
@egor-rogov Done.

yyyc514:
Tests are green: https://travis-ci.org/github/highlightjs/highlight.js/pull_requests

src/plugins/vue.js
```diff
@@ -0,0 +1,64 @@
+import { escapeHTML } from "../lib/utils";
+
+function hasValueOrEmptyAttribute(value) {
+  return Boolean(value || value === "");
+}
+
+const NO_HIGHLIGHT = "no-highlight";
```
egor-rogov, line 7: Hmm, looks like this won't work in case of `hljs.configure({noHighlightRe: 'something'})`...
(and the same is true for `blockLanguage` function in `highligh.js`)

yyyc514: Damn. :-) Grrr.

yyyc514:
> But what are the advantages of using Vue in developer.html, I wonder? 

**To eat our own dog food.**  If we support Vue, we might as well use it in developer - even if only as a quick way to get bugs reported if the integration breaks for some reason.

> Why do we need vue.js and vue.runtime.js in the repository?

I vendored them for the same reason we vendor `jquery` - to have a pinned version.  Or are you asking why we need both?  Technically we could just include the one...

src/plugins/vue.js
egor-rogov: Probably it's safe to always check for `/^(no-?highlight)$/i` in `shouldNotHighlight`, and additionally check for the provided regex?

egor-rogov:
> To eat our own dog food

It makes sense, but what if we decide to provide another plugin for some another framework?

src/plugins/vue.js
yyyc514: HLJS doesn't expose it's internal settings though. :-)

yyyc514:
> It makes sense, but what if we decide to provide another plugin for some other framework?

*shrugs* Can't solve all the worlds problems in one day. :-)

egor-rogov:
What helps Vue is the plugin, not our developer tool. Keeping it simple also makes sense for me, as well as "to eat our own dog food".

yyyc514:
I meant it helps Vue indirectly as in gives us another reason to keep the plugin maintained. :-)

src/plugins/vue.js
yyyc514: Actually we still hard-code it for when an explicit language "fails" to load.  So this has the exact same [fixed] behavior as the main library. So I'm calling this good enough for now. :-)

```js
      if (!language) {
        console.warn(LANGUAGE_NOT_FOUND.replace("{}", match[1]));
        console.warn("Falling back to no-highlight mode for this block.", block);
      }
      return language ? match[1] : 'no-highlight';
    }
```

src/highlight.js
```diff
@@ -11,6 +11,7 @@ import * as utils from './lib/utils';
 import * as MODES from './lib/modes';
 import { compileLanguage } from './lib/mode_compiler';
 import * as packageJSON from '../package.json';
+import { VuePlugin } from "./plugins/vue";
```
yyyc514, line 14:
```suggestion
import { VuePlugin } from "./plugins/vue.js";
```

src/plugins/vue.js
egor-rogov: > Actually we still hard-code it for when an explicit language "fails" to load.

Yes, that's exactly what I mentioned in my first message (: 
So what do you think about changing `shouldNotHighlight` to always check for this hardcoded value? (in addition to whatever may be passed in `options`)

yyyc514: I see two choices:

- Drop the configurability completely and just have a static value/regex.
- Improve the configurability to make this "no-highlight" value something you could provide via options

This configurability was only added preemptively when we changed the meaning of "plain" and "text" - to allow people to easily get the "old" behavior.

Really this behavior is just a way to "alert" the user visually of broken highlighting without making a big fuss about it (break the styles a little bit).  We could simply deprecate that behavior and when there is an error apply no css classes at all, simply let the user read the console.log if they want to know what went wrong?

yyyc514: I think I'd rather just remove the hard-coded string... so the only way any type of "no-highlight" gets on there is if the user added it themself... the library would never do it.

tools/developer.html
```diff
@@ -70,7 +70,7 @@ <h3>Code</h3>
     </div>
     <div>
       <h3>Rendered in <span class="rendering_time">...</span> ms [<span class="rendering_stats">...</span>]</h3>
-      <pre><code class="hljs">...</code></pre>
+      <highlightjs auto :code="code" ignore_illegals />
```
allejo, line 73: I see `ignore_illegals` was commented out in the component, should it be removed from here too?

tools/developer.html
```diff
@@ -97,10 +98,14 @@ <h3>Markup</h3>
...
-        $(".hljs span").each((_,el) => {
-          $(el).attr("data-klass", el.className.replace("hljs-",""))
-        })
+        vue.code = source;
+        // give Vue a moment to update
+        setTimeout(() => {
+          $(".hljs span").each((_,el) => {
+            $(el).attr("data-klass", el.className.replace("hljs-",""))
+          })
+        }, 10);
```
allejo, line 107: Instead of using a timeout, can't this be done in the `Vue()` as part of the lifecycle? (i.e. `mounted`)

This is some very weird jQuery + Vue butchery lol

src/plugins/vue.js
```diff
@@ -0,0 +1,64 @@
...
+        this.detectectLanguage = this.language;
+      }
+      return result.value;
+    },
+    autoDetect() {
+      return !this.language || hasValueOrEmptyAttribute(this.autodetect);
+    },
+    ignoreIllegals() {
+      return true;
+      // return !hasValueOrEmptyAttribute(this.ignore_illegals);
```
allejo, line 45: Is this gonna be removed or why is it commented out?

src/plugins/vue.js
```diff
@@ -0,0 +1,64 @@
...
+        return escapeHTML(this.code);
+      }
+
+      let result;
+      if (this.autoDetect) {
+        result = hljs.highlightAuto(this.code);
+        this.detectedLanguage = result.language;
+      } else {
+        result = hljs.highlight(this.language, this.code, this.ignoreIllegals);
+        this.detectectLanguage = this.language;
```
allejo, line 36:
```suggestion
        this.detectedLanguage = this.language;
```

src/highlight.js
```diff
@@ -846,7 +847,9 @@ const HLJS = function(hljs) {
     requireLanguage,
     autoDetection,
     inherit,
-    addPlugin
+    addPlugin,
+    // plugins for frameworks
+    vuePlugin: VuePlugin
```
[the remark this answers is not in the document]
yyyc514: Nope, this breaks old versions of Node.js and someone filed an issue just the other day to NOT do this, which we merged... if you think this needs a real discussion, please file a separate issue.  Here isn't the place to discuss.

tools/developer.html
yyyc514: I'll take another look... this wasn't meant to be a perfect example of usage. :)  It would need to be an "onchange" hook on the code attribute of our app, not on mount I don't think... this needs to fire anytime the code is changed.

src/plugins/vue.js
egor-rogov: > I think I'd rather just remove the hard-coded string... so the only way any type of "no-highlight" gets on there is if the user added it themself... the library would never do it.

Probably it's the right way... but you have to do something in the language-not-found case.
Just wanted to say that this issue shouldn't stop the PR. Looks like It should be addressed separately.

tools/developer.html
yyyc514: It's not when it's mounted though it's every single time it's updated... I'm not aware that Vue has a "afterRerender" callback, does it?  `$watch` is too early from my understanding of it.

yyyc514: Ok, using $nextTick now. :)

yyyc514:
@allejo All good?

PR #2549, opened 2020-05-12
Add Svelte to list of supported languages

ryanditjia:


yyyc514:
Could you include the standard `dist` folder with ready to go CDN module?

Issue #2550, opened 2020-05-12
(Dart) Support Dart's new Null Safety type system


srawlins:
**Describe the issue**

Dart's new [Null Safety language feature](https://github.com/dart-lang/language/blob/master/accepted/future-releases/nnbd/feature-specification.md) introduces some new keywords and a new way to specify types:

* `late` and `required` are new keywords
* `Never` is a new type found in `dart:core`
* (almost) all types can be suffixed with a `?` to indicate a nullable type. For example, in:

  ```dart
  f(int x) { ... }
  g(int? x) { ... }
  ```

  `f` is a function which accepts a non-null `int` parameter, and `g` accepts a nullable `int` parameter.

**Which language seems to have the issue?**

Dart.

**Are you using `highlight` or `highlightAuto`?**

Either.

**Sample Code to Reproduce**

```dart
late int x;
f({required int? y}) {}
const List<Never> z = [];
```

**Expected behavior**

`late` and `required` should be highlighted as keywords. `int?` should highlight the same as `int`. `Never` should be highlighted like the other built-in types.

**Additional context**

yyyc514:
Please add a usage of Never to your sample.

A PR would be welcome.  Tagging this as beginner friendly.

srawlins:
Should the long list of built-in types just be duplicated to include, e.g. `int?`, `String?`, ... or is there a better way?

yyyc514:
Duplication is probably best, but you can do it dynamically with JS... break out the types (or do all the Built-Ins count?) split them, then build a second set with `?` appended, add that to the first...

yyyc514:
You'll need to add `?` to the keyboard $pattern... and is that going to cause any false positives with ternary?

yyyc514:
Does this apply to user types as well? Is so might want to think about making it a mode instead that matches all `[identifier]?`, but again would need to consider false positives.

srawlins:
> Please add a usage of Never to your sample.

Done.

> You'll need to add ? to the keyword $pattern... and is that going to cause any false positives with ternary?

It shouldn't; most Dart code is auto-formatted so that a ternary `?` has whitespace before it.

> Does this apply to user types as well? Is so might want to think about making it a mode instead that matches all [identifier]?, but again would need to consider false positives.

It does, but I think it would be more confusing to match `[identifier]?`, since that would match differently from `[identifier]`.

[...]

PR #2551, opened 2020-05-12
enh(dart) Add `late` and `required` keywords, and `Never` built-in type (Dart 2.9)

srawlins:
Addresses some of #2550 

I think there are open questions about supporting `?` on every type, so that will be addressed later.


yyyc514:
Please add a changelog entry in CHANGES.md.

srawlins:
Done

PR #2554, opened 2020-05-13
(erlang) add support for underscore separators in numeric literals

seriyps:
Erlang/OTP v23.0 was released today and it includes this PR https://github.com/erlang/otp/pull/2324 that adds support for underscore group separators in numeric literals:

```
1_000_000
16#DEAD_BEEF
1_000.234_567
```

Reference regexp is the one from Emacs erlang-mode: https://github.com/erlang/otp/pull/2324/files#diff-196d091d94ca59ba10083cf66a7e8cbf extended to support floats as well.


yyyc514:
You'd really want to add the tests to a `markup` test, not put them in `detect` where they could have an effect on our very brittle auto-detect stuff.

And could you also add an entry to a CHANGES.md?

seriyps:
> You'd really want to add the tests to a markup test
The problem is that there is no `markup` tests for Erlang at all. I can try to add some though...

> they could have an effect on our very brittle auto-detect stuff
Not sure I understand, should I revert my changes in auto-detect?

> add an entry to a CHANGES.md
sure!

yyyc514:
>  should I revert my changes in auto-detect?

If you don't mind.

> The problem is that there is no markup tests for Erlang at all.

seriyps:
I added `markup` (mostly copied from Elixir) and reverted chages for `autodetect`. CHANGES was also updated.

yyyc514:
@seriyps Thank you for your contribution!

PR #2556, opened 2020-05-14
(docs) add Jolie to Supported Languages

xiroV:
Adds syntax highlighting support for the [Jolie language](https://www.jolie-lang.org/).


yyyc514:
Your instructions could perhaps also include a Node.js `require` example.

PR #2562, opened 2020-05-17
fix(typescript) Add missing read-only keyword

Lhoerion:
I've fixed detection of keywords in property name, along with moving the basic types to keywords.

### DIFF
--- a/src/languages/typescript.js
+++ b/src/languages/typescript.js
@@ -31,7 +31,8 @@ export default function(hljs) {
     "protected",
     "implements",
     "declare",
-    "abstract"
+    "abstract",
+    "readonly"
   ];
   var KEYWORDS = {
     $pattern: ECMAScript.IDENT_RE,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,7 +50,7 @@ Language Improvements:
 [Sam Rawlins]: https://github.com/srawlins
 [Sergey Prokhorov]: https://github.com/seriyps
 [Nils Knappmeier]: https://github.com/nknapp
-[Martin]: https://github.com/Lhoerion
+[Martin (Lhoerion)]: https://github.com/Lhoerion
 
 
 ## Version 10.0.2
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,7 @@ Deprecations:
 
 Language Improvements:
 
+- fix(typescript) add `readonly` keyword (#2562) [Martin (Lhoerion)][]
 - fix(javascript) fix regex inside parens after a non-regex (#2530) [Josh Goebel][]
 - enh(typescript) use identifier to match potential keywords, preventing false positivites (#2519) [Josh Goebel][]
 - enh(javascript) use identifier to match potential keywords, preventing false positivites (#2519) [Josh Goebel][]
@@ -49,6 +50,7 @@ Language Improvements:
 [Sam Rawlins]: https://github.com/srawlins
 [Sergey Prokhorov]: https://github.com/seriyps
 [Nils Knappmeier]: https://github.com/nknapp
+[Martin]: https://github.com/Lhoerion
 
 
 ## Version 10.0.2


src/languages/typescript.js
```diff
@@ -222,6 +222,9 @@ export default function(hljs) {
       {
         begin: '\\.' + hljs.IDENT_RE, relevance: 0 // hack: prevents detection of keywords after dots
       },
+      {
+        begin: '\\b[A-Za-z$_][0-9A-Za-z$_]*:', relevance: 0 // prevents detection of keywords in property name
```
yyyc514, line 226: Please show an example or two of exactly what you’re trying to avoid by adding this.

Lhoerion: Pretty straightforward example:
`public readonly type: number;`

yyyc514: I’m thinking this should probably reuse `ECMAScript.IDENT_RE`, isn’t that what that is?... you can use `regex.join` to glue the pieces together.

src/languages/typescript.js
```diff
@@ -35,9 +35,9 @@ export default function(hljs) {
   ];
   var KEYWORDS = {
     $pattern: ECMAScript.IDENT_RE,
-    keyword: ECMAScript.KEYWORDS.concat(TS_SPECIFIC_KEYWORDS).join(" "),
+    keyword: ECMAScript.KEYWORDS.concat(TYPES).concat(TS_SPECIFIC_KEYWORDS).join(" "),
     literal: ECMAScript.LITERALS.join(" "),
-    built_in: ECMAScript.BUILT_INS.concat(TYPES).join(" ")
+    built_in: ECMAScript.BUILT_INS.join(" ")
```
yyyc514, line 40: This was intentional - to classify the types differently than keywords.  This is in line with how all the built-in JS classes are classified (which are also technically types): as built-ins.

Lhoerion: Yeah, my misunderstanding of the TS context, removed it because it's indeed nonsense

src/languages/typescript.js
Lhoerion: Replaced it with simple concatenation

Lhoerion:
Added also missing `readonly` TS specific keyword.

src/languages/typescript.js
```diff
@@ -222,6 +223,14 @@ export default function(hljs) {
       {
         begin: '\\.' + hljs.IDENT_RE, relevance: 0 // hack: prevents detection of keywords after dots
       },
+      {
+        begin: '\\b' + IDENT_RE + ':', relevance: 0 // prevents detection of keywords in property name
+      },
+      {
```
yyyc514, line 229: I don't think we can add this.  Many of our built-ins ARE function names (all BUILT_IN_GLOBALS for example)... what specific issues are you having that caused you to add this rule?

Lhoerion:
```
yyyc514 21 hours ago Member
Please show an example or two of exactly what you’re trying to avoid by adding this.

Lhoerion 21 hours ago Author
Pretty straightforward example:
public readonly type: number;

yyyc514 21 hours ago Member
I’m thinking this should probably reuse ECMAScript.IDENT_RE, isn’t that what that is?... you can use regex.concat to glue the pieces together.

Lhoerion 21 hours ago Author
Replaced it with simple concatenation
```

yyyc514: I'm referring to the block STARTING on line 229...  not the proceeding.

yyyc514: How annoying how GitHub did this. Let me try again.

src/languages/typescript.js
```diff
@@ -222,6 +223,14 @@ export default function(hljs) {
...
+        className: 'function',
+        begin: IDENT_RE + '\\s*\\(', returnBegin: true,
+        relevance: 0
+      },
```
yyyc514, line 233: I'm not sure this is the correct solution here. Many of our built-ins ARE function names (all BUILT_IN_GLOBALS for example)... what specific issues are you having that caused you to add this rule?

yyyc514: Better. Sorry for confusion.

Lhoerion: It turned out that same problem affects methods name too, so for ex.
`delete(): void`
will have `delete` keyword highlighted even when it shouldn't

yyyc514: This may be starting to get a little bigger than you were hoping for, but here is what I see off the top of my head:

- Lose the `className`, traditionally we don't highlight function invocation.  I'm open-minded to changing that, but this PR isn't the place to start that journey.
- `returnBegin` looks wrong and probably only works because you have `ARGUMENTS` later to save you from an infinite loop.  Or it could also be our infinite loop detection logic saving the day... there is no reason to return to the beginning here. (unless I'm missing something)
- You probably want to remove that and instead use a look ahead for just the `(` part... that way your rule doesn't eat the `(` and it can be matched by ARGUMENTS later  or VALUE_CONTAINER...
- We need to add a rule BEFORE this one to match the BUILT_IN_GLOBALS functions and still highlight them as keywords properly.
- Then this rule would come after, prevent highlighting any other keywords as function names.
- And then technically this same fix would need to be applied to Javascript also (or at the very least an issue opened).

That's off the top of my head.

yyyc514: You may also need to run the tests and see why they are currently red.  I'm guessing you either actually fixed something that was actually incorrectly highlighted in a test case or else you've accidentally broken some of the tests with these rule additions.

yyyc514: Hang in there, try not to get discouraged.  Often contributing a small fix isn't nearly as simple as you first though... particularly once you get into the grammar rules, and inclusion, exclusion, etc... there are often unintended side effects, etc... so it takes time to work thru all those, cover all the cases, etc.

That's why we have the `markup` tests, to help prevent unintended regressions when changes like this are made, etc... so if you aren't running the tests you probably really should be.

Lhoerion: I'm not too familiar with `highlight.js`, so it's sometimes trial and error.
I use Windows OS, it fails for me and I don't have time to fix another test task.

yyyc514: If you wanted to pull out this last part and just go with the first two fixes (assuming they pass tests on their own) then we could get those merged and add an issue to track the other stuff...

You'd still want to add a quick CHANGES.md mention.

Lhoerion: @yyyc514
I've come up to something like this
```js
begin: '\\b(?<!\\.)' + IDENT_RE + '(?=[\\(])', relevance: 0
```
but still ``(`` get's eaten somehow, so until next try to fix it, I don't know how to prevent this.

yyyc514: That looks right regarding look ahead. Look behind is not well supported yet so we cannot use that.

Lhoerion: Could you suggest in what direction I should push?

Lhoerion: I don't know what can be done if positive lookahead doesn't work

yyyc514: > I don't know what can be done if positive lookahead doesn't work

It most definitely works in general, we use it in many places.  What might help here is if you wrote a failing markup test and then pushed your current code AND the failing markup tests and then I could take a look at it locally.

And also please be explicit in why you say it's "not working".

yyyc514: `test/markup/typescript/identifiers_that_include_keywords.txt` might be a good place to start.

yyyc514: > I use Windows OS, it fails for me and I don't have time to fix another test task.

If you can't run the tests though I'm really not sure what to tell you - this is going to be a major pain to check all the results by hand for all cases.  Is there a particular reason you can run the test suite?  It should be possible on Windows if you have Node.js, etc installed.

Lhoerion: I fixed it in different way and tests have passed, but take a look again if it's ok in current shape.
After that I'll focus on making other necessary preparations.

yyyc514:
Please provide me with the code samples you were using locally to test these changes.

src/languages/typescript.js
```diff
@@ -222,8 +223,14 @@ export default function(hljs) {
...
         begin: '\\.' + hljs.IDENT_RE, relevance: 0 // hack: prevents detection of keywords after dots
       },
+      {
+        begin: '\\b' + IDENT_RE + ':', relevance: 0 // prevents detection of keywords in property name
+      },
       DECORATOR,
-      ARGUMENTS
+      ARGUMENTS,
+      {
+        begin: '(^(?!\\.$).)' + IDENT_RE + '\\(', relevance: 0 // prevents detection of keywords in method name
```
yyyc514, line 232: Not sure what the intention is here since we already have line 223 which already does this?  Unless I'm misunderstanding your regex... I'm not sure what you're intending the beginning part there to be doing exactly.

yyyc514:
> Please provide me with the code samples you were using locally to test these changes.

And can you please attach a code snippet to the start of this message to so anyone else looking at it can see what you're trying to accomplish without digging thru the whole chat log?

yyyc514:
This might be more ambitious than you were even thinking:

```
delete(a)
```

Without more context there is no way to know if that is a delete operator followed by the expression `(a)` or a function definition.

src/languages/typescript.js
```diff
@@ -222,8 +223,14 @@ export default function(hljs) {
       {
         begin: '\\.' + hljs.IDENT_RE, relevance: 0 // hack: prevents detection of keywords after dots
       },
+      {
+        begin: '\\b' + IDENT_RE + ':', relevance: 0 // prevents detection of keywords in property name
```
yyyc514, line 227: The better way to do this might be the very complex way we do it in JS now, and also to highlight these as `attr`.  That's probably something we should move into a `js-like` library so that both JS and TS should share it.

I think the best thing at this point might be to scale this back to JUST adding the `readonly` keyword and to open an issue (with examples, please) for the other issues so they can best be discussed how to solve them keeping the bigger picture in mind (like JS grammar, etc)

yyyc514:
@Lhoerion Thanks! I opened https://github.com/highlightjs/highlight.js/issues/2568 as a result of the discussion here.

PR #2565, opened 2020-05-18
fix(groovy) strings are not allowed inside ternary clauses

### LINKED ISSUE #2217: (groovy) strings are not allowed inside ternary clauses
If you do for example:

```groovy
def createCodeStyleMsgBody(build, label) {
    def codeStyleFiles = findCodestyleFilesInLog(build)
    def formattingMsg = label < 0 ? ('The following files need formatting:\n    ' +
        codeStyleFiles.join('\n    ')) : 'All files are correctly formatted'
    def url = build.url + "consoleText"

    return "${Globals.resTicks[build.result]} $formattingMsg\n    (${url})"
}
```

'The following files need formatting:\n    ' is not highlighted correctly as a string, secondly everything after ' + is coloured in blue (so if you create another def after, it colours all that in blue).

yyyc514:
Closes #2217

### DIFF
--- a/src/languages/groovy.js
+++ b/src/languages/groovy.js
@@ -97,7 +97,7 @@ export default function(hljs) {
             },
             {
               // highlight map keys and named parameters as attrs
-              className: 'attr', begin: IDENT_RE + '[ ]*:'
+              className: 'attr', begin: IDENT_RE + '[ \t]*:'
             },
             {
               // catch middle element of the ternary operator
@@ -115,7 +115,7 @@ export default function(hljs) {
             {
                 // highlight labeled statements
                 className: 'symbol',
-                begin: '^[ ]*' + regex.lookahead(IDENT_RE + ':'),
+                begin: '^[ \t]*' + regex.lookahead(IDENT_RE + ':'),
                 excludeBegin: true,
                 end: IDENT_RE + ':',
                 relevance: 0
--- a/src/languages/groovy.js
+++ b/src/languages/groovy.js
@@ -5,69 +5,84 @@
  Website: https://groovy-lang.org
  */
 
+import * as regex from "../lib/regex";
+
+function variants(variants, obj = {}) {
+  obj.variants = variants;
+  return obj;
+}
+
 export default function(hljs) {
+  const IDENT_RE = '[A-Za-z0-9_$]+';
+  const COMMENT = variants([
+    hljs.C_LINE_COMMENT_MODE,
+    hljs.C_BLOCK_COMMENT_MODE,
+    hljs.COMMENT(
+      '/\\*\\*',
+      '\\*/',
+      {
+        relevance : 0,
+        contains : [
+          {
+            // eat up @'s in emails to prevent them to be recognized as doctags
+            begin: /\w+@/, relevance: 0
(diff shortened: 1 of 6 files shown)


yyyc514:
Sorry, I should have split the commit in two.  Mostly it's just a reorg of the modes into constants to make them easier to reuse inside TERNARY.

src/languages/groovy.js
```diff
@@ -77,25 +92,35 @@ export default function(hljs) {
...
+                REGEXP,
+                NUMBER,
+                'self'
+              ]
             },
             {
                 // highlight labeled statements
-                className: 'symbol', begin: '^\\s*[A-Za-z0-9_$]+:',
+                className: 'symbol',
+                begin: '^[ ]*' + regex.lookahead(IDENT_RE + ':'),
```
egor-rogov, line 118: You've changed \s to space, is it ok?

src/languages/groovy.js
```diff
@@ -5,69 +5,84 @@
...
+    hljs.QUOTE_STRING_MODE,
+    ],
+    { className: "string" }
+  );
+
     return {
         name: 'Groovy',
         keywords: {
-            literal : 'true false null',
+            built_in: 'this super',
```
egor-rogov, line 69: What is the reason to make these built-ins? I looked into https://docs.groovy-lang.org/latest/html/documentation/ and found them among keywords.

src/languages/groovy.js
```diff
@@ -77,25 +92,35 @@ export default function(hljs) {
...
             },
-            hljs.C_NUMBER_MODE,
             {
                 className: 'meta', begin: '@[A-Za-z]+'
             },
             {
-                // highlight map keys and named parameters as strings
-                className: 'string', begin: /[^\?]{0}[A-Za-z0-9_$]+ *:/
+              // highlight map keys and named parameters as attrs
+              className: 'attr', begin: IDENT_RE + '[ ]*:'
```
egor-rogov, line 100: Should here be \s in place of a space?..

src/languages/groovy.js
yyyc514: This is so the match starts on the current line, not many lines previous.

src/languages/groovy.js
yyyc514: They are still keywords, everything in `keywords` is keywords, the question is what kind of keyword on they.  These are very clearly not language keywords in the same vein as `for`, `if`, etc... they are built in variables, same as `global` or `window`, etc...

src/languages/groovy.js
yyyc514: See prior note, again do not want to eat new lines.

yyyc514:
>  I mean that \t is also a possible character.

Ugh, I guess... if I have to worry about that everywhere I'm going to go and write some kind of whiteSpaceButNotNewline helper. :-)

src/languages/groovy.js
yyyc514: Adding support for tabs.

src/languages/groovy.js
yyyc514: Adding support for tabs.

PR #2569, opened 2020-05-19
enh(typescript,javascript) Build both JS and TS on a common foundation

### LINKED ISSUE #1155: (typescript) Add JSX support to Typescript

### LINKED ISSUE #2568: Proposal: TypeScript should be built on the solid foundation of JS
Since TypeScript is purely a superset of Javascript (and all valid JS is also valid TS) and that  we should merge our JS and TS grammars into a single grammar for ease of maintenance - this probably should have been done from the beginning, but water under the bridge. Right now many things require double the effort and there are strange differences between the grammars (like JS supports JSX but TS does not support TSX).

It also makes things harder for contributors... they need to do more work (perhaps double) be more aware of the bigger picture, etc...

Most recent example: https://github.com/highlightjs/highlight.js/pull/2562

We could literally make `js` and `javascript` as an alias of TypeScript or JS could be more of a "dummy" grammar that really just re-uses the TypeScript grammar under the hood. I'm not sure what the value there is though other than that `lang.name` would be "Javascript".

@egor-rogov Any objections?

joshgoebel:
Resolves #1155.
Resolves #2568.


yyyc514:
I'm stuck... it seems there would be no easy way to distinguish JSX from an old style type cast?

```ts
// 1
return getModuleInstanceState((<ModuleDeclaration>node).body);
// 2
return getModuleInstanceState(
  <ModuleDeclaration>
     node.body this is text inside html
  </ModuleDeclaration>
);
return getModuleInstanceState(<ModuleDeclaration/>);
```

I'm not sure how to easily tell the difference between 1 and 2.  Since #1 just looks like the beginning of a pattern like #2.

yyyc514:
@egor-rogov  @allejo Any ideas here?  I was really making progress until I hit this.

allejo:
> I'm stuck... it seems there would be no easy way to distinguish JSX from an old style type cast?

I don't believe that casting in manner 1 is legal in the TypeScript + JSX (aka TSX) world. In TSX, you need to use the [`as` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-6.html#new-tsx-file-extension-and-as-operator) for casting; so if anything, we'd need to differentiate between `ts` and `tsx` and have them as separate languages (from an end-user perspective).

i.e.

~~~markdown
```tsx
```
~~~

vs.

~~~markdown
```ts
```
~~~

yyyc514:
In that case I'm not sure I'm interested in supporting the "old" TS syntax than and I'd just go straight for TSX support.  I don't have time to figure out both and not sure it's worth the time.

allejo:
I wouldn't consider it being the "old" TS syntax, the `as` operator is just an alternate syntax when you're in the scope of TSX. So if you have `<Whatever>`, it can only be either TS or TSX (and I'd lean towards having the user specify which).

If you're interested in differences, one difference would be that TS casting can only ever be `<DataType>`. Whereas TSX can take two syntaxes: `<Element>...</Element>` (open and close) or `<Element />` (needs to have a `/`).

allejo:
Sorry, let me clarify something I think I could have been misleading about. Generics are still valid in TSX like so:

```tsx
class MyComponent extends Component<Props> {
  render() {
    return (
      <div />
    );
  }
}
```

It's just when it comes to casting, that `<Whatever>` is not legal.

yyyc514:
> I wouldn't consider it being the "old" TS syntax, the as operator is just an alternate syntax when you're in the scope of TSX.

My understanding from mentoring TS is that `as` is prefer now, except for templates like `Set<Person>`, etc.

yyyc514:
Please have a look at my `on:begin` filter for beginning JSX.  It does two things:

- Detects "tags" with `<` inside them such as "<Array<Array<number>>", and ignore them - since this would be invalid HTML.
- Detects tags that immediately close `<booger>` and searches the remaining text for a matching close tag... if none is found it will assume this is a type cast NOT JSX since JSX should include closing tags.

This will not highlight incomplete JSX but that doesn't bother me lot.  If someone can find a better way to solve this problem I'm all ears.

yyyc514:
@allejo @egor-rogov Ping.

egor-rogov:
The idea look good to me. I had a look at the filter and the code seems reasonable.

joshgoebel:
Don't forget #2702.

joshgoebel:
Thoughts on reviewing:

This one is big and a bit messy to review, sorry.  Essentially what I did here at a high-level:

- Moved all unique rules from JS into the TS grammar (making TS a super-set that should then handle both JS and TS)
- Made JS point to the new TS and fix all broken tests (often new things being highlighted that were missed before)
- Move the new super-TS grammar back into the JS grammar and strip out it's "TS only bits"
- Add those "TS only bits" back in the form of a TS grammar that now builds on top of the JS grammar
- Look at the old grammars add some small nuances back (different shebangs, `strict asm`, etc)

I'm assuming I got this mostly right based on being careful and our test coverage. :-)

joshgoebel:
@allejo @egor-rogov Actually I guess you all had a glance at this before... but any final thoughts?  I think it's cleaned up and pretty much ready to merge now.

egor-rogov (approved):
Cool! I just went through the patch and it looks good to me.

PR #2580, opened 2020-05-27
Update the core library import from /lib/core to /lib/highlight. /lib/core no longer exists.

gruns:


yyyc514:
What version are you using? core exists from v10 on forward.

yyyc514:
If you want to use an older version for some reason you could find the older README here:

https://github.com/highlightjs/highlight.js/blob/9-18-stable/README.md

gruns:
:+1:

Issue #2500, opened 2020-04-21
(csharp) Extending highlighting support for more specific language constructs


Turnerj:
The short version is I want to get this (the C# snippet from the highlight.js website):
![image](https://user-images.githubusercontent.com/904226/79879670-27d73a00-842e-11ea-9ff6-ca69f311ef0c.png)

Closer to this (same snippet but in the latest version of Visual Studio 2019):
![image](https://user-images.githubusercontent.com/904226/79880250-ebf0a480-842e-11ea-8498-540c794bb605.png)

Actual snippet of code:
```
using System.IO.Compression;

#pragma warning disable 414, 3021

namespace MyApplication
{
    [Obsolete("...")]
    class Program : IInterface
    {
        public static List<int> JustDoIt(int count)
        {
            Console.WriteLine($"Hello {Name}!");
            return new List<int>(new int[] { 1, 2, 3 })
        }
    }
}
```

My question isn't about writing the CSS to do that for a theme, its about having classes to target those styles in the first place. Where does the use of a specific highlight.js class start and end (eg. "hljs-function" for declarations but not calls?). Like, is there a spec somewhere for when something should be used?

I've read through some of the related issues like #1731 and #1378 so I do understand reluctance for adding classes that can target these types of cases as core functionality especially when not all languages do this. I guess I'm querying, more as a guide, if I were to implement (some of) this for C#, I don't want to step on toes about class name consistency.

yyyc514:
Also if you attached plain-text versions for anyone who wanted to play with this that'd be greatly appreciated.

Turnerj:
> That said the VS Code is all OSS. If you want syntax highlighting "that good" you might want to look at just using the same code Visual Studio uses to do the highlighting. But their syntax file for some languages is like 1mb while ours is like 10kb.... so that level of detail comes at a price.

Yeah, there are definitely trade-offs if I wanted to match my example images 100%. If I can bring them a little closer though that would be great. It was just funny seeing a code snippet of mine, with highlighting enabled, seem like nothing really was highlighted except for like 1 if-statement.

[...]

PR #2593, opened 2020-06-07
Rust unsafe keyword also gets `emphasis strong`.

### LINKED ISSUE #2585: [Request] Highlight Rust `unsafe` keyword differently from others
I'd appreciate a more distinct visual indicator of unsafe blocks and functions when reviewing highlighted code.

Currently it appears that [`unsafe` is a keyword](https://github.com/highlightjs/highlight.js/blob/debfdf7cc9822a49e7f033b2c286ef025ce8d3c8/src/languages/rust.js#L15). I'm not very familiar with hljs and so I'm not sure if the best option is to give it a custom mode(?) or if there's a built-in mode (a warning?) or if there's a way to e.g. apply a bold to the keyword but treat it as a keyword otherwise.

anp:
This differentiates the `unsafe` keyword from others with additional styling.

Closes #2585.

### DIFF
--- a/src/languages/rust.js
+++ b/src/languages/rust.js
@@ -12,7 +12,7 @@ export default function(hljs) {
     'abstract as async await become box break const continue crate do dyn ' +
     'else enum extern false final fn for if impl in let loop macro match mod ' +
     'move mut override priv pub ref return self Self static struct super ' +
-    'trait true try type typeof unsafe unsized use virtual where while yield';
+    'trait true try type typeof unsized use virtual where while yield';
   var BUILTINS =
     // functions
     'drop ' +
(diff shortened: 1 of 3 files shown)


yyyc514:
This is how you might do it for personal use. But it’s also a total hack. Something we can’t merge at this time. 

Right now we don’t have an official way to apply multiple classes.

yyyc514:
https://github.com/highlightjs/highlight.js/issues/2500

Keep an eye on this for how we MIGHT want to handle this in the future.

yyyc514:
Sorry if there was some confusion before.  I wasn’t trying to suggest that anyone open a PR for this. I was only trying to explain how you might go about actually doing it for personal use or for a specific service.

Issue #2594, opened 2020-06-08
(TOML) Incorrect syntax highlighting


aduh95:
**Describe the issue**

The library fail to parse some valid TOML lines.

**Which language seems to have the issue?**

TOML, also INI.

**Are you using `highlight` or `highlightAuto`?**

Both.

**Sample Code to Reproduce**

Using this code sample as test case (copied from [TOML readme](https://github.com/toml-lang/toml#user-content-keys)):

```toml
# Bare keys:
key = "value"
bare_key = "value"
bare-key = "value"
1234 = "value"

# Quoted keys:
"127.0.0.1" = "value"
"character encoding" = "value"
"ʎǝʞ" = "value"
'key2' = "value"
'quoted "value"' = "value"

"key \"containing\" backslash" = 6
'key \"containing" backslash\' = 6

# empty quoted key is allowed
"" = "blank"     # VALID but discouraged
'' = 'blank'     # VALID but discouraged

# Dotted keys:
name = "Orange"
physical.color = "orange"
physical.shape = "round"
site."google.com" = true
3.14159 = "pi"

# Whitespace around dot-separated parts is ignored:
hello  .  world  =  "!"

# Whitespace is ignored around key names and values
  hello = "World!"
```

JS code to render the above using `highlight.js`:

```js
import hljs from "highlight.js/lib/core.js";
import toml from "highlight.js/lib/languages/ini.js";

const TOML = "toml";
hljs.registerLanguage(TOML, toml);

console.log(hljs.highlight(TOML,`
  …above toml code here…
`, true));
```
**Actual behavior**

Only bare keys and some dotted keys are highlighted, while the other ones are parsed as `illegal`.

<details>

```html
<span class="hljs-comment"># Bare keys:</span>
<span class="hljs-attr">key</span> = <span class="hljs-string">"value"</span>
<span class="hljs-attr">bare_key</span> = <span class="hljs-string">"value"</span>
<span class="hljs-attr">bare-key</span> = <span class="hljs-string">"value"</span>
<span class="hljs-attr">1234</span> = <span class="hljs-string">"value"</span>

<span class="hljs-comment"># Quoted keys:</span>
"127.0.0.1" = "value"
"character encoding" = "value"
"ʎǝʞ" = "value"
'key2' = "value"
'quoted "value"' = "value"

"key "containing" backslash" = 6
'key "containing" backslash' = 6

<span class="hljs-comment"># empty quoted key is allowed</span>
[...]

PR #2595, opened 2020-06-08
fix(toml) Improve key parsing

aduh95:
Fixes: https://github.com/highlightjs/highlight.js/issues/2594


src/languages/ini.js
```diff
@@ -66,7 +66,7 @@ export default function(hljs) {
         begin: /\[+/, end: /\]+/
       },
       {
-        begin: /^[a-z0-9\[\]_\.-]+(?=\s*=\s*)/,
+        begin: /([A-Za-z0-9_-]+|(\s*\.\s*)|"(\\"|[^"])*"|'[^']*')+(?=\s*=\s*)/,
```
yyyc514, line 69: Can we please use `variants` for this to make it more obvious what is happening here, what the different flavors of tags are supported, etc.

yyyc514: Is A-Z correct? I don't see any uppercase keys in the test?

yyyc514: Can you please explain `(\s*\.\s*)`?

yyyc514:
If you can get to this in the next day or two it'll likely land in 10.1 which I hope to push out this week.

src/languages/ini.js
aduh95: > Can you please explain `(\s*\.\s*)`?

According to the TOML spec, "whitespace around dot-separated parts is ignored".

> Is A-Z correct? I don't see any uppercase keys in the test?

It is in TOML, let me add a test for that.

yyyc514: You removed the `^` so I'm worried about this matching who knows what now... Why is the ^ not correct?

aduh95:
There are three variants of keys in TOML:

```js
        variants: [
          {begin: /[A-Za-z0-9_-]+(?=\s*=\s*)/}, // bare keys
          {begin: /"(\\"|[^"])*"(?=\s*=\s*)/}, // quoted keys with double quotes
          {begin: /'[^']*'(?=\s*=\s*)/}, // quoted keys with single quotes
        ]
```

On top of this, there are the dotted keys, which are "a sequence of bare or quoted keys joined with a dot" ("whitespace around dot-separated parts is ignored"). It seems hard to find an efficient regex for that case…

```js
/([A-Za-z0-9_-]+|"(\\"|[^"])*"|'[^']*')(\s*\.\s*([A-Za-z0-9_-]+|"(\\"|[^"])*"|'[^']*'))*(?=\s*=\s*)/
```

Should I go for the above or is there a more readable alternative?

yyyc514:
Why wouldn't matching a key drop into a different mode that looks for more [.][key] matches? And when none are found ends the match?

Or you you'd split them into variables and then a super match would look some like:

```
([key1]|[key2]|[key3]) + optional(([.]([key1]|[key2]|[key3]))*)
```

Or something.  Look at the `regex` helpers.

src/languages/ini.js
aduh95: The `^` was preventing indented keys to be selected:

```toml
key = "value"
  indented_key = "value"
```

I originally thought I could add a `\s*` to fix that, but the whitespace is not actually part of the key so I'm not sure. What do you think?

yyyc514: What are the full rules for values... what would this be/do?

```tom
indented_key = amiakey = "value"
```

aduh95: > what would this be/do?
> 
> ```
> indented_key = amiakey = "value"
> ```

That's not valid TOML, though without the lookbehind, `amiakey` would be highlighted as a key..

src/languages/ini.js
```diff
@@ -54,6 +56,17 @@ export default function(hljs) {
...
   };
 
+  var BARE_KEY = /[A-Za-z0-9_-]+/;
+  var QUOTED_KEY_DOUBLE_QUOTE = /"(\\"|[^"])*"/;
+  var QUOTED_KEY_SINGLE_QUOTE = /'[^']*'/;
+  var ANY_KEY = regex.either(
+    BARE_KEY, QUOTED_KEY_DOUBLE_QUOTE, QUOTED_KEY_SINGLE_QUOTE
+  );
+  var DOTTED_KEY = regex.concat(
+    '(?<=^\s*)', ANY_KEY, '(\s*\.\s*', ANY_KEY, ')+',
```
yyyc514, line 66: Is this look behind? We can't do this because it's not supported everywhere... I guess for now we can just drop this and forget the pinning to the beginning of the line.  I'm not sure it matters.

yyyc514: Rest of it looks pretty nice.

yyyc514:
@aduh95 Thanks for working on this, looking pretty nice!

src/languages/ini.js
aduh95: Yeah right, not worth the breakage for sure. Thanks a lot for your help, I removed the lookbehind and the tests are now passing fine 🙏

yyyc514:
Please add an entry in CHANGES.md.

PR #2598, opened 2020-06-09
(dart) highlight built-in nullable types

### LINKED ISSUE #2550: (Dart) Support Dart's new Null Safety type system

srawlins:
Fixes #2550 

This change separates the "built-ins" into those which can be marked nullable (like `int` and `String`) and those which cannot (like `dynamic` and `window`). Nullable built-in types like `int?` are highlighted as such.

### DIFF
--- a/src/languages/dart.js
+++ b/src/languages/dart.js
@@ -8,23 +8,23 @@ Category: scripting
 */
 
 export default function(hljs) {
-  var SUBST = {
+  const SUBST = {
     className: 'subst',
     variants: [{
       begin: '\\$[A-Za-z0-9_]+'
     }],
   };
 
-  var BRACED_SUBST = {
+  const BRACED_SUBST = {
     className: 'subst',
     variants: [{
       begin: '\\${',
       end: '}'
-    }, ],
+    }],
     keywords: 'true false null this is new super',
   };
 
-  var STRING = {
+  const STRING = {
     className: 'string',
     variants: [{
         begin: 'r\'\'\'',
@@ -72,7 +72,7 @@ export default function(hljs) {
     hljs.C_NUMBER_MODE, STRING
   ];
 
-  var BUILT_IN_TYPES = [
+  const BUILT_IN_TYPES = [
     // dart:core
     'Comparable',
     'DateTime',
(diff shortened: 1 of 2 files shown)


src/languages/dart.js
```diff
@@ -72,17 +72,28 @@ export default function(hljs) {
     hljs.C_NUMBER_MODE, STRING
   ];
 
+  var BUILT_IN_TYPES =
```
yyyc514, line 75: Please just go ahead and explode this into an array with one item per line.  We need to start doing this anyways and it will clean up the logic below.

src/languages/dart.js
```diff
@@ -72,17 +72,28 @@ export default function(hljs) {
...
   var KEYWORDS = {
     keyword: 'abstract as assert async await break case catch class const continue covariant default deferred do ' +
       'dynamic else enum export extends extension external factory false final finally for Function get hide if ' +
       'implements import in inferface is late library mixin new null on operator part required rethrow return set ' +
       'show static super switch sync this throw true try typedef var void while with yield',
     built_in:
+      // non-nullable built-in types
+      BUILT_IN_TYPES + ' ' +
+      // nullable built-in types
+      BUILT_IN_TYPES.split(' ').map((e) => { e + '?' }).join(' ') + ' ' +
```
yyyc514, line 91: Use template literal?

src/languages/dart.js
srawlins: Done

src/languages/dart.js
srawlins: Done

yyyc514:
@srawlins Does this still look good to you?

srawlins:
Looks great, thanks!

yyyc514:
Can you add a changes.md entry, please?

srawlins:
I just edited the existing line, is that good?

yyyc514:
Good enough.

src/languages/dart.js
```diff
@@ -72,17 +72,59 @@ export default function(hljs) {
...
+    'ElementList',
+  ];
+  const NULLABLE_BUILT_IN_TYPES = BUILT_IN_TYPES.map((e) => `${e}?`);
+
+  const KEYWORDS = {
     keyword: 'abstract as assert async await break case catch class const continue covariant default deferred do ' +
       'dynamic else enum export extends extension external factory false final finally for Function get hide if ' +
       'implements import in inferface is late library mixin new null on operator part required rethrow return set ' +
-      'show static super switch sync this throw true try typedef var void while with yield',
+      'show static super switch sync this throw true try typedef const void while with yield',
```
yyyc514, line 111:
```suggestion
      'show static super switch sync this throw true try typedef var void while with yield',
```

yyyc514: Fixing this back.

yyyc514:
@srawlins Thanks for putting in the time!

Issue #2596, opened 2020-06-08
(csharp) Declaration for generic functions does not highlight


djoyman:
**Describe the issue**
Declaration for generic methods does not highlight

**Which language seems to have the issue?**
C#

**Are you using `highlight` or `highlightAuto`?**
highlight

**Sample Code to Reproduce**
<pre><code class="csharp">
  ObjectType FindObject&lt;ObjectType&gt;(
    Criteria criteria
  )
</code></pre>

**Expected behavior**
highlight like visual studio
![image](https://user-images.githubusercontent.com/28087745/84023309-8467c780-a990-11ea-8697-a80f8921b794.png)

yyyc514:
And how are we currently highlighting the same code?

djoyman:
This code is not highlighted at all:

![image](https://user-images.githubusercontent.com/28087745/84039016-4cb94980-a9a9-11ea-965c-7cef783f3132.png)

yyyc514:
Yeah, I don't think templates are supported in function declarations right now at all.

yyyc514:
@djoyman Any chance you want to take a try at making a PR?

djoyman:
Not now, maybe later. If I am going to fix, I will write

yyyc514:
Closing per #2599.  @djoyman

PR #2599, opened 2020-06-10
enh(csharp) highlight generics in more cases

TupikovVladimir:
https://github.com/highlightjs/highlight.js/issues/2596

### DIFF
--- a/src/languages/csharp.js
+++ b/src/languages/csharp.js
@@ -178,12 +178,12 @@ export default function(hljs) {
       },
       {
         className: 'function',
-        begin: '(' + TYPE_IDENT_RE + '\\s+)+' + hljs.IDENT_RE + '\\s*(\\<.+\\>\\s*)?\\(', returnBegin: true,
+        begin: '(' + TYPE_IDENT_RE + '\\s+)+' + hljs.IDENT_RE + '\\s*(\\<.+\\>)?\\s*\\(', returnBegin: true,
         end: /\s*[{;=]/, excludeEnd: true,
         keywords: KEYWORDS,
         contains: [
           {
-            begin: hljs.IDENT_RE + '\\s*(\\<.+\\>\\s*)?\\(', returnBegin: true,
+            begin: hljs.IDENT_RE + '\\s*(\\<.+\\>)?\\s*\\(', returnBegin: true,
             contains: [
               hljs.TITLE_MODE,
               GENERIC_MODIFIER
--- a/src/languages/csharp.js
+++ b/src/languages/csharp.js
@@ -92,8 +92,7 @@ export default function(hljs) {
     begin: "<",
     end: ">",
     contains: [ 
-      { className: "keyword", begin: "in\\s" },
-      { className: "keyword", begin: "out\\s" },
+      { beginKeywords: "in out"},
       TITLE_MODE 
     ]
   };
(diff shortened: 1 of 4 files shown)


yyyc514:
Tests will also need to be updated:

test/markup/csharp/generic_modifiers*.txt

src/languages/csharp.js
```diff
@@ -177,13 +178,16 @@ export default function(hljs) {
       },
       {
         className: 'function',
-        begin: '(' + TYPE_IDENT_RE + '\\s+)+' + hljs.IDENT_RE + '\\s*\\(', returnBegin: true,
+        begin: '(' + TYPE_IDENT_RE + '\\s+)+' + hljs.IDENT_RE + '\\s*(\\<.+\\>\\s*)?\\(', returnBegin: true,
```
yyyc514, line 181: What is the trailing `\\s*` doing?

src/languages/csharp.js
```diff
@@ -177,13 +178,16 @@ export default function(hljs) {
...
         end: /\s*[{;=]/, excludeEnd: true,
         keywords: KEYWORDS,
         contains: [
           {
-            begin: hljs.IDENT_RE + '\\s*\\(', returnBegin: true,
-            contains: [hljs.TITLE_MODE],
+            begin: hljs.IDENT_RE + '\\s*(\\<.+\\>\\s*)?\\(', returnBegin: true,
```
yyyc514, line 186: What is the trailing `\\s*` doing?

src/languages/csharp.js
yyyc514, line 181: I think the trailing `\\s*` should be outside the optional match, yes?

src/languages/csharp.js
yyyc514, line 186: I think the trailing `\\s*` should be outside the optional match, yes?

src/languages/csharp.js
TupikovVladimir: There is no difference in result, but it is better for grouping. Ok, I'll modify it.

TupikovVladimir:
I fixed keywords `in` and `out` for generic parameters, but following white space is included to this keyword. Also I expanded test for generic parameters.

src/languages/csharp.js
```diff
@@ -91,8 +91,11 @@ export default function(hljs) {
   var GENERIC_MODIFIER = {
     begin: "<",
     end: ">",
-    keywords: "in out",
```
yyyc514: Why was this a problem?

src/languages/csharp.js
```diff
@@ -91,8 +91,11 @@ export default function(hljs) {
...
-    contains: [ TITLE_MODE ]
+    contains: [ 
+      { className: "keyword", begin: "in\\s" },
+      { className: "keyword", begin: "out\\s" },
```
yyyc514, line 96: Spaces should NOT be including in keyword tags.

src/languages/csharp.js
TupikovVladimir: Because it does not work, I've made it by another way using keywords, and now space is not  including in keyword tags

yyyc514: Ah yes, TITLE would eat them...  but the original solution wasn't optimal.  `beginKeywords` is the best quick way to do this!   As you're doing now.

yyyc514:
And while I have your attention should we consider tagging these differently in the future?

title.class vs title.template?  What do you call those items? Do most editors syntax highlight them differently?

We're considering such things for future improvements.

Issue #2600, opened 2020-06-10
Discuss: Should Highlight.js version 11 npm release be ESM only?


aduh95:
**Is your request related to a specific problem you're having?**

I'm always frustrated when starting a project, I don't want to set up a whole build chain and just want to get started:

```js
import highlight from './node_modules/highlight.js/lib/core.js'
```

And I get an error `Error: 'default' is not exported by /node_modules/highlight.js/lib/core.js`...

**The solution you'd prefer / feature you'd like to see added...**

I'd much rather prefer having standard ES6 modules in the npm package, and being able to use them as advertised in the `README.md`.

**Any alternative solutions you considered...**

`s/module.exports =/export default/` on the `node_modules/highlight.js/` files, but that's not very convenient.

**Additional context...**

Support for ESM has landed in Node.js current and LTS lines recently, with features such as [Conditional Exports](https://nodejs.org/docs/latest/api/esm.html#esm_conditional_exports) (useful in case we want to keep the CJS version around, still required for Node.js v10 support).

Having ESM files would also make the library compatible with Deno for free, which would be nice.

yyyc514:
Is anyone using `.jsm`?  Or a special extension for modules?

> s/module.exports =/export default/ on the node_modules/highlight.js/ files, but that's not very convenient.

Perhaps I should have looked at this a bit more carefully.  Is this really ALL you were looking for here?  If we shipped those 2-3 entry files (or variants) with ONLY this change?

aduh95:
> Is anyone using .jsm? Or a special extension for modules vs normal CJS?

`.mjs` is the one recognized by Node.js.

> would that resolve this?

Yes, that would be totally fine! I think Rollup can handle that, right?

https://github.com/highlightjs/highlight.js/blob/b1bce6e3ada485b89696f554878b2ef44a73d94a/tools/build_config.js#L10

Haven't tried it, but I think you can replace it with:

```js
output: [
  { format: "cjs", strict: false },
  { format: "esm" },
]
```

Would you interested in a PR?

yyyc514:
Actually that will mess with all the individual languages we require also, no?  Those would then also need to also be modules, yes?  That's what I was fearful of...  I'm not sure we want to ship +180 additional/duplicate JS files just because of ESM... is that the only way to do this - or am I thinking about it wrong?  I wonder if there are any other large modular libraries and how they are dealing with this problem?

[...]

PR #2601, opened 2020-06-10
(chore) add `.js` extension to import statements

aduh95:
Add file extension to all `import` specifiers in `./src/` files.
This is useful to run the files straight from source with a web browser, Node.js ESM or Deno.

Refs: https://github.com/highlightjs/highlight.js/issues/2600#issuecomment-642282239

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,9 +38,6 @@ module.exports = {
       "no-undefined": "error",
       // ensure import specifier contains file extension
       "import/extensions": ["error", "always"],
-      // make sure there is no Node.js specific API slipping into the source files
-      "import/no-nodejs-modules": "error",
-      "import/no-commonjs": "error",
 
       // TODO maybe
       "camelcase": "off", // TODO: turn on later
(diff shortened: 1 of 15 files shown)


yyyc514:
What about the requires inside src/languages?   Are all those already good?

yyyc514:
Is this some way to make rollup REQUIRE this?  Any idea?  This isn't going to magically keep itself maintained...

.eslintrc.js
```diff
@@ -36,6 +36,11 @@ module.exports = {
       "indent": ["error", 2, {"VariableDeclarator":2}],
       // seems like a good idea not to use explicit undefined
       "no-undefined": "error",
+      // ensure import specifier contains file extension
+      "import/extensions": ["error", "always"],
+      // make sure there is no Node.js specific API slipping into the source files
+      "import/no-nodejs-modules": "error",
+      "import/no-commonjs": "error",
```
yyyc514, line 43: Do we need to add a npm library for these?  Where can I find the docs?  I don't see these on the main ESLIst list of rules?

aduh95: The library is already there, we just don't use it yet for some reason:

https://github.com/highlightjs/highlight.js/blob/1936dbd59dcd4de7391a9ea0413a68906fb04c14/package.json#L50

yyyc514: I think these (or at least the nodejsone) should be moved to `overrides` and scope to `src/*`, no?  We definitely DO use nodejss modules in tests.  Or the rule could just be removed; I don't see a big risk of Node.js modules accidentally slipping into our client libraries.

PR #2602, opened 2020-06-11
(chore) use eslint on src files

aduh95:
This is an attempt to answer @yyyc514's request on https://github.com/highlightjs/highlight.js/pull/2601#issuecomment-642361317.

This PR is adding eslint to the test command, to enforce rules on source files. I have also run `eslint --fix` on the source files.

I have changed a few rules from the `eslintrc` file:

- Remove TypeScript parser and plugins. This repo doesn't use TS at all, it was probably an oversight.
- Disable `no-useless-escape` on the `./src/languages/…` files. There are just too many of them.


yyyc514:
**Sorry but I don't think we can merge this one.**

> I have also run eslint --fix on the source files.

This destroys/pollutes Git history. I'm usually against large style changes like this because as a maintainer I find clean history to be MUCH more valuable than perfect style.  (and the style issues here aren't terrible)  We'll improve them with time.

> This PR is adding eslint to the test command, to enforce rules on source files. 

Maybe one day, not today.  I don't think adding more friction for maintainers/contributors right now isn't optimal.

Out of curiosity: Is there perhaps some easy way to only lint MODIFIED files?

>Remove TypeScript parser and plugins. This repo doesn't use TS at all, it was probably an oversight.

It's used in "jsonly" mode and for all the type inference it can do plus the JSDoc types really helps with catching errors while coding.  It supplements the linter nicely in this regard.

> Disable no-useless-escape on the ./src/languages/… files. There are just too many of them.

I have definitely consider doing this. :-)  I'm still debating if they add some type of documentation value (better a useless escape than missing a REAL escape) or if they are truly just clutter and should all be cleaned up.

yyyc514:
I might be more open to adding this (lint on test) for JUST the core parser source (excluding languages for now)... I might experiment with that locally and see how that works out.

aduh95:
> better a useless escape than missing a REAL escape

Actually there are some that outlines a real issue:

https://github.com/highlightjs/highlight.js/blob/1936dbd59dcd4de7391a9ea0413a68906fb04c14/src/languages/yaml.js#L96

Here eslint repports `Unnecessary escape character: \s`, and indeed the backslash should be escaped instead of the s:
 
```js
  begin: '^---\\s*$',
```

> Is there perhaps some easy way to only lint MODIFIED files?

I think solutions exist involving commit hooks. I've never use it myself though.

aduh95:
> I might be more open to adding this (lint on test) for JUST the core parser source (excluding languages for now)... I might experiment with that locally and see how that works out.

FWIW I think that would be the right approach. Maybe let the linter check the languages modules but disable all the rules that do not currently pass?

yyyc514:
>  Maybe let the linter check the languages modules but disable all the rules that do not currently pass?

That's kind of what I do now, but I don't feel any compulsion to "update them all at once".  I keep an eye on the linter warnings (they are toned way done for languages) when I'm updating a language... but if I'm making larger changes I might quickly comment out all the language rules to get a much stricter ruleset in place and update some styles while I'm in there.

yyyc514:
Added task for now: https://github.com/highlightjs/highlight.js/commit/dfeb3a131fbb59f91f20894f471866667ca276c7

PR #2605, opened 2020-06-12
(feat): added `record` as keyword to Java

### LINKED ISSUE #2604: [Request] Adding `record` as keyword to Java
According to [JEP 359](https://openjdk.java.net/jeps/359) since Java 14 `record` is a new kind of type declaration, so it should be a keyword like `enum`?
Currently `record` is not in the list of keywords in java.js

See https://github.com/highlightjs/highlight.js/blob/b1bce6e3ada485b89696f554878b2ef44a73d94a/src/languages/java.js#L13-L17

Adding `record` to list of keywords

**Example Code**

<em>Added by Josh Goebel</em>

```java
record Range(int lo, int hi) {
  public Range {
    if (lo > hi)  /* referring here to the implicit constructor parameters */
      throw new IllegalArgumentException(String.format("(%d,%d)", lo, hi));
  }
}
```

0xflotus:
should fix #2604


0xflotus:
this means that?

yyyc514:
Changes.md

0xflotus:
From where comes this `span class="hljs-function"` in the expected string?

yyyc514:
Because this looks exactly like a function definition does it not?

0xflotus:
I don't think so. Functions in Java are methods and look like 

```java
public foo(int x) {}
private foo() {}
abstract protected foo();
```

yyyc514:
Can I declare a method without one of those keywords in front of it?

0xflotus:
Yes this means it is implicitly `package private`

yyyc514:
Yeah that’s not right. I’ll look at on my computer later.

yyyc514:
So what's a REAL example?

```
record Point(int x, int y) { }
```

What goes inside the {}? Can there be code there? data?

0xflotus:
According to [JEP 359](https://openjdk.java.net/jeps/359) there can be 
Bildschirmfoto 2020-06-13 um 18 47 09

0xflotus:
More examples

Bildschirmfoto 2020-06-13 um 18 49 36
Bildschirmfoto 2020-06-13 um 18 49 55

yyyc514:
Ok, so this is a kind of class.

> Records are a new kind of type declaration in the Java language. Like an enum , a record is a restricted form of class.

What we need to do here is have a new variant rule that's a mix of the existing function/class/interface rules... We'd probably still want `class` for the className... And we possibly could pull out the params into a PARAMS const and then re-use that mode for `record`.

Does that make sense?

yyyc514:
@0xflotus Are you wanting to continue to do work on this to get it done along the lines of what I mentioned in my last comment?

0xflotus:
@yyyc514 sorry I'm not sure what to do next. I would be fine, when we close this PR.

Issue #2603, opened 2020-06-12
typescript build problem: node_modules/highlight.js/types/index.d.ts is not a module


limbowang:
ts file content: 
```
// index.ts
import * as hljs from 'highlight.js';
```
after I run command
```
tsc index.ts
```
I got error
```
index.ts:1:23 - error TS2306: File '/Users/limbo/tmp/ssss/node_modules/highlight.js/types/index.d.ts' is not a module.

1 import * as hljs from 'highlight.js';
                        ~~~~~~~~~~~~~~


Found 1 error.
```
The version I use is 10.1.0.

Try to use 10.0.2 and it build successfully.

yyyc514:
Not sure what you did exactly to get that error?

yyyc514:
May need some TypeScript people to weigh in here...  we only have the `index.ts.d` so we can check our own code in TS's `jsonly` mode - and so that perhaps it can be helpful to people using TS enabled editors for the code competition, etc...

It's possible we may need to change something slightly?

yyyc514:
Is this a regression since 10.0?  I don't think this has changed from 10.0 to 10.1

limbowang:
> Is this a regression since 10.0? I don't think this has changed from 10.0 to 10.1

Have tried v10.0.x and it built successfully. There is no types/index.d.ts under v10.0.x

yyyc514:
> There is no types/index.d.ts under v10.0.x

Ah, so it was more recent... but I still don't know why your TS is trying to load that file?

yyyc514:
Maybe we need `export = HLJSApi` or something?

limbowang:
Suggest to use index.d.ts in @types/highlight.js. Or you can remove `types: 'types/index.d.ts'` in package.json. The ts compiler will read it by default.

yyyc514:
>  The ts compiler will read it by default.

Only if it's in the root is my understanding from reading the docs.

limbowang:
> A types package is a folder with a file called index.d.ts or a folder with a package.json that has a types field.

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

plantain-00:
Same problems here.
Suggest to refer to `export` from index.d.ts in @types/highlight.js: 

```ts
declare namespace hljs {
 ...
}
export = hljs;
export as namespace hljs;
```
[...]

PR #2606, opened 2020-06-13
fix "is not a module" #2603

rodrigojcmello:


nikeee:
You can introduce support for `highlight.js/lib/core.js` like this:
https://github.com/highlightjs/highlight.js/compare/master...nikeee:ts?expand=1

Basically, you could do it for `highlight.js/lib/highlight.js` the same way.

However, I think it is not possible to do something for `highlight.js/lib/languages/basic.js` via `./languages/*.js`.

Also, the types are not exported and available in TS (only the hljs object). It would require something different to also export the types.

IMHO, the easiest solution would be to migrate to TS and generate the declaration files via tsc. However, this would require some extra work to get the CJS version right.

yyyc514:
I don’t want type files everywhere. Does the ambient module stuff simply not work at all?  Why is it in the documentation? It looks like the perfect solution to this and I see lots of packages on @types that seem to be using it.

nikeee:
You could try something like this (omitting the extension):
```typescript
declare module 'lib/language/basic' {
    export = fn;
}
```
(When importing, the .ja extension is optional in TS). fn being a const of the languageFn.

Not tested.

yyyc514:
I've been trying that, but nothing works so far.  I guess today I'll try to find an NPM package and @types that actually seems to work correctly and then try to extrapolate from there.

rodrigojcmello:
I created a typing file for each submodule, and changed the types in the package.json to get the typing files out of the "types" folder.

I realized that there are two HLJSApi interfaces, I commented one because I didn't really understand the reason for having two. I need help with that.

It was also not very clear to me, what are the differences in the "highlight.js", "index.js" and "core.js" interface, I left them all the same. It would be good for someone to review this.

yyyc514:
Can you please provide thoughts on https://github.com/highlightjs/highlight.js/pull/2608

This seems to work in my testing.  And it supports not just pulling in the library itself, but also the variations people do most often.

yyyc514:
Oh I'm just now seeing this, let me have a look.

yyyc514:
I think ambient modules are much cleaner, no?

yyyc514:
> I realized that there are two HLJSApi interfaces, I commented one because I didn't really understand the reason for having two. I need help with that.

It's just extending the first, this is a feature of TS.  I was trying to separate the helpers form the main API... though perhaps I should instead do this with a second interface and then extend it... or a `&` type...

> It was also not very clear to me, what are the differences in the "highlight.js", "index.js" and "core.js" interface, I left them all the same. It would be good for someone to review this.

Well `highlight.js` and `highlight.js/lib/index`, are the same, so I removed the latter in my PR.  `core` loads just the core library (no languages) where by default `index` pulls in EVERYTHING (core + load all languages, then export core).

rodrigojcmello:
> I think ambient modules are much cleaner, no?

I did several tests, this was the only way I managed to make it work lol

I tried the way here # 2608 but was unsuccessful.

yyyc514:
> I tried the way here #2608 but was unsuccessful.

Can you show me how you were unsuccessful or show me your test cases?

yyyc514:
@rodrigojcmello How to structure the type file is a whole other discussion. :-)  I'm sure it could be improved greatly and perhaps it should be more "type definition like".  I'm not sure yet.

Right now I'm just first trying to solve the basic issue of importing our library in a TS project without blowing up. :)

rodrigojcmello:
> Can you please provide thoughts on #2608
> 
> This seems to work in my testing. And it supports not just pulling in the library itself, but also the variations people do most often.

Here it is not clear whether you tested #2606 or #2608

I didn't get to test the #2608, yesterday I did something like that and it didn't work out. I'm going to test #2606 now.

yyyc514:
I didn’t test. I’ve very not interested in multiple files - if we can handle this instead with a single file and ambient modules. Which seems like they were designed for exactly this purpose

tools/build_node.js
```diff
@@ -75,11 +76,14 @@ async function buildNode(options) {
...
-  mkdir("types");
 
   install("./LICENSE", "LICENSE");
   install("./README.md","README.md");
-  install("./types/index.d.ts","types/index.d.ts");
+  install("./types/index.d.ts","lib/index.d.ts");
+  install("./types/types.ts","lib/types.ts");
+  install("./types/core.d.ts","lib/core.d.ts");
+  install("./types/highlight.d.ts","lib/highlight.d.ts");
+  install("./types/languages/basic.d.ts","lib/languages/basic.d.ts");
```
yyyc514, line 86: There are 180+ languages... this is why I'm not a fan of the individual file approach here.

rodrigojcmello:
Apparently the #2608 works very well.

My test yesterday probably failed because I declared the modules at the beginning of the file. Anyway, this solution is even better. 🥇

PR #2612, opened 2020-06-16
Removed offending comma

ehoogerbeets:
This causes a syntax error on node 6 and earlier:

```
edwin@Lisbon:~/ht/loctool/test$ node -v
v6.17.1
edwin@Lisbon:~/ht/loctool/test$ node
> var hl = require("highlight.js");
/data/root/home/edwin/ht/loctool/node_modules/highlight.js/lib/languages/java.js:89
          ),
          ^

SyntaxError: Unexpected token )
```

Without the comma:

```
edwin@Lisbon:~/ht/loctool/test$ node -v
v6.17.1
edwin@Lisbon:~/ht/loctool/test$ node
> var hl = require("highlight.js");
>
```
Much better! That comma is a trailing comma after the last parameter to the function call. Node 6 and earlier do not like it. Node 7 and beyond just ignores it.

### DIFF
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -42,7 +42,7 @@ export default function(hljs) {
           regex.either(
             regex.concat(SEQUENCE_ALLOWING_UNDERSCORES('a-fA-F0-9'), /\./, SEQUENCE_ALLOWING_UNDERSCORES('a-fA-F0-9')),
             regex.concat(SEQUENCE_ALLOWING_UNDERSCORES('a-fA-F0-9'), /\.?/),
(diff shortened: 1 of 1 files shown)


yyyc514:
Is this really the only one?  I feel like I see this all the time.

ehoogerbeets:
It's possibly not the only one there, but it is the only one that prevented highlight.js from even loading in node 6 and earlier! Thanks for merging that. What does the release schedule look like? I have a tool (ilib-js/loctool) that is broken because it depends on this and folks are asking me when the loctool will be fixed...

yyyc514:
I released 10.1.1 to fix.

PR #2617, opened 2020-06-20
fix(js) Implement Numeric Separator proposal

aduh95:
This PR implements the Numeric Separator proposal, currently stage 3 of TC39 process, and supported by all major browsers.

Limitations:
- the proposal doesn't allow multiple adjacent separator. Because it makes the regex overcomplicated, I've ignored it.
- Number literals with leading zero (which represents either decimal or octal literal) are not supported by this PR. Worth noting they are illegal in strict mode (incl. ES modules), so I believe it makes sense to not highlight them.

Refs: https://github.com/tc39/proposal-numeric-separator
Refs: https://tc39.es/ecma262/#sec-literals-numeric-literals

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -75,7 +75,7 @@ export default function(hljs) {
       { begin: nonDecimalLiterals('oO', '0-7') }, // Octal literals
       { begin: nonDecimalLiterals('xX', '0-9a-fA-F') }, // Hexadecimal literals
       { begin: regex.concat(/\b/, noLeadingZeroDecimalDigits, 'n') }, // Non-zero BigInt literals
-      { begin: regex.concat(/\b0?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
+      { begin: regex.concat(/(\b0)?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
       { begin: regex.concat(
         /\b/,
         noLeadingZeroDecimalDigits,
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -64,7 +64,7 @@ export default function(hljs) {
     built_in: ECMAScript.BUILT_INS.join(" ")
   };
   const nonDecimalLiterals = (prefixLetters, validChars) => 
-    `(-?)\\b0[${prefixLetters}][${validChars}]([${validChars}_]*[${validChars}])?n?`;
+    `\\b0[${prefixLetters}][${validChars}]([${validChars}_]*[${validChars}])?n?`;
   const noLeadingZeroDecimalDigits = /[1-9]([0-9_]*\d)?/;
   const decimalDigits = /\d([0-9_]*\d)?/;
   const exponentPart = regex.concat(/[eE][+-]?/, decimalDigits);
@@ -74,15 +74,15 @@ export default function(hljs) {
       { begin: nonDecimalLiterals('bB', '01') }, // Binary literals
       { begin: nonDecimalLiterals('oO', '0-7') }, // Octal literals
       { begin: nonDecimalLiterals('xX', '0-9a-fA-F') }, // Hexadecimal literals
-      { begin: regex.concat(/(-?)\b/, noLeadingZeroDecimalDigits, 'n') }, // Non-zero BigInt literals
-      { begin: regex.concat(/(-?)0?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
+      { begin: regex.concat(/\b/, noLeadingZeroDecimalDigits, 'n') }, // Non-zero BigInt literals
+      { begin: regex.concat(/0?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
       { begin: regex.concat(
-        /(-?)\b/,
+        /\b/,
         noLeadingZeroDecimalDigits,
         regex.optional(regex.concat(/\./, regex.optional(decimalDigits))), // fractional part
         regex.optional(exponentPart)
         ) }, // Decimal literals >= 1
-      { begin: /(-?)\b0[\.n]?/ }, // Zero literals (`0`, `0.`, `0n`, `-0`, ...)
+      { begin: /\b0[\.n]?/ }, // Zero literals (`0`, `0.`, `0n`)
     ],
     relevance: 0
   };
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -63,12 +63,26 @@ export default function(hljs) {
     literal: ECMAScript.LITERALS.join(" "),
     built_in: ECMAScript.BUILT_INS.join(" ")
   };
+  const nonDecimalLiterals = (prefixLetters, validChars) => 
+    `(-?)\\b0[${prefixLetters}][${validChars}]([${validChars}_]*[${validChars}])?n?`;
+  const noLeadingZeroDecimalDigits = /[1-9]([0-9_]*\d)?/;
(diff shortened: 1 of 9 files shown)


yyyc514:
I only took a glance.  This may have to wait until after the JS/TS merge I've been working on.  But even though I don't think it'd be a big deal to reintegrate it after.

I'll try to take a deeper look next week.

src/languages/javascript.js
```diff
@@ -24,12 +24,26 @@ export default function(hljs) {
...
+      { begin: nonDecimalLiterals('xX', '0-9a-fA-F') }, // Hexadecimal literals
+      { begin: regex.concat(/(-?)\b/, noLeadingZeroDecimalDigits, 'n') }, // Non-zero BigInt literals
+      { begin: regex.concat(/(-?)0?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
+      { begin: regex.concat(
+        /(-?)\b/,
+        noLeadingZeroDecimalDigits,
+        regex.optional(regex.concat(/\./, regex.optional(decimalDigits))), // fractional part
+        regex.optional(exponentPart)
+        ) }, // Decimal literals >= 1
+      { begin: /(-?)\b0[\.n]?/ }, // Zero literals (`0`, `0.`, `0n`, `-0`, ...)
```
joshgoebel, line 46: Is there a strong reason you have included `-`?  Typically we do not do this because it will break math expressions like `3-2`... there is no simple way to detect that it's truly a unary expression or not...

Sorry for the long delay on this... finally polished up my TS/JS work and hope to have it merged soon so I was taking another look at this.

aduh95: > Is there a strong reason you have included `-`?

TBH I have no recollection what I was trying to achieve here. Once the other PR gets merge, I'll try to come back at this one to see if that's OK to just remove that dash.

joshgoebel: > Once the other PR gets merge,

Merged. :)

aduh95:
I've rebased and included a second commit to remove the `-` from the regex. I guess I should include a few more test cases to make explicit what is the policy of `highlight.js` when it comes to dash and numbers:

```js
let num = -3; // `-` _should_ be included, but is not anymore
let num = - 3; // Who writes JS like that??
let diff = 5 -3; // Who writes JS like that??
let diff = 5- 3; // Who writes JS like that??
let diff = 5 - 3; // yep, pretty classic
let diff = 5-3; // minified version

**Comment by joshgoebel** [comment]
> EDIT: Let's try if my original commit is able to handle these cases before dismissing it.

Nope.

File: test/markup/javascript/keywords.expect.txt [hunk 2]
```diff
@@ -1,6 +1,6 @@
 <span class="hljs-function"><span class="hljs-keyword">function</span> <span class="hljs-title">$initHighlight</span>(<span class="hljs-params">block, cls</span>) </span>{
   <span class="hljs-keyword">try</span> {
-    <span class="hljs-keyword">if</span> (cls.search(<span class="hljs-regexp">/\bno\-highlight\b/</span>) != <span class="hljs-number">-1</span>)
+    <span class="hljs-keyword">if</span> (cls.search(<span class="hljs-regexp">/\bno\-highlight\b/</span>) != -<span class="hljs-number">1</span>)
```
aduh95, line 3: That looks like a regression, but maybe it's fine.

joshgoebel:
```
<span class="hljs-keyword">let</span> diff = <span class="hljs-number">5</span><span class="hljs-number">-3</span>;
```

Screen Shot 2020-10-03 at 10 44 05 AM

I'd say actually it's a FIX - just we didn't have good numeric tests in the past. :-)  C_NUMBER includes `-?` in the regex, which as I said is problematic... it's probably something we should consider changing (for exactly this reason), but that's a separate issue.

When look behind is added we could add `-` back when it immediately follows certain things like `(` or `=`, but that's also another day.

src/languages/javascript.js
```diff
@@ -63,12 +63,26 @@ export default function(hljs) {
...
     className: 'number',
     variants: [
-      { begin: '\\b(0[bB][01]+)n?' },
-      { begin: '\\b(0[oO][0-7]+)n?' },
-      { begin: hljs.C_NUMBER_RE + 'n?' }
+      { begin: nonDecimalLiterals('bB', '01') }, // Binary literals
+      { begin: nonDecimalLiterals('oO', '0-7') }, // Octal literals
+      { begin: nonDecimalLiterals('xX', '0-9a-fA-F') }, // Hexadecimal literals
+      { begin: regex.concat(/\b/, noLeadingZeroDecimalDigits, 'n') }, // Non-zero BigInt literals
+      { begin: regex.concat(/0?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
```
joshgoebel, line 78: Is there a reason we have no `\b` here?

src/languages/javascript.js
```diff
@@ -63,12 +63,26 @@ export default function(hljs) {
...
+      { begin: regex.concat(
+        /\b/,
+        noLeadingZeroDecimalDigits,
+        regex.optional(regex.concat(/\./, regex.optional(decimalDigits))), // fractional part
+        regex.optional(exponentPart)
+        ) }, // Decimal literals >= 1
+      { begin: /\b0[\.n]?/ }, // Zero literals (`0`, `0.`, `0n`)
```
joshgoebel, line 85: Is sequencing off here?  Isn't it `[.]?n?`?  `0n.` would not be correct, would it?

joshgoebel: Oh, it's either... :-) So then I think this is ok since `.0n` does not seem valid.

joshgoebel:
Looking good.  Left one comment.  Also please add a CHANGES.md entry. :-)

Almost there. :)

src/languages/javascript.js
aduh95: Probably not, let's try with it

```suggestion
      { begin: regex.concat(/\b0?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
```

aduh95: Hum so now the tests don't pass…

joshgoebel (approved):

src/languages/javascript.js
joshgoebel: Does the \b need to be grouped with just the 0?  I don't think there is a boundary between say a " " and a "."?

src/languages/javascript.js
```diff
@@ -63,12 +63,26 @@ export default function(hljs) {
...
     className: 'number',
     variants: [
-      { begin: '\\b(0[bB][01]+)n?' },
-      { begin: '\\b(0[oO][0-7]+)n?' },
-      { begin: hljs.C_NUMBER_RE + 'n?' }
+      { begin: nonDecimalLiterals('bB', '01') }, // Binary literals
+      { begin: nonDecimalLiterals('oO', '0-7') }, // Octal literals
+      { begin: nonDecimalLiterals('xX', '0-9a-fA-F') }, // Hexadecimal literals
+      { begin: regex.concat(/\b/, noLeadingZeroDecimalDigits, 'n') }, // Non-zero BigInt literals
+      { begin: regex.concat(/\b0?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
```
joshgoebel, line 78:
```suggestion
      { begin: regex.concat(/(\b0)?\./, decimalDigits, regex.optional(exponentPart)) }, // Decimal literals between 0 and 1
```

joshgoebel (approved):

joshgoebel:
@aduh95 Thanks for all the work on this, and sorry for the lag.

PR #2620, opened 2020-06-29
(octave) Add Octave language

apjanke:
Would you accept a separate Octave language definition? Octave is a clone of Matlab, but is syntactically different enough that I think it deserves its own entry. In particular:

* Octave has several additional keywords not present in Matlab
* Its string literal semantics are somewhat different
* It allows "#" comments

Alternately, I could do this as a 3rd party definition – https://github.com/highlightjs/highlight.js/issues/2621 – in which case, just close this PR without merging.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -288,3 +288,4 @@ Contributors:
 - Alexandre Grison <a.grison@gmail.com>
 - Jim Mason <jmason@ibinx.com>
 - lioshi <lioshi@lioshi.com>
+- Andrew Janke <floss@apjanke.net>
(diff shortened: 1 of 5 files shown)


apjanke:
Or should I be doing this as a "third party" language definition in a separate repo?

apjanke:
I've read through the "3rd party quickstart" and "language-contribution.rst" docs and am unclear on how adding a third-party language to the `extra/` directory works. Do I make the `extra/<language>` directory a Git submodule/subrepo?

apjanke:
Actually, now that I've read your docs again, I'm pretty sure this would work better as a 3rd party definition. Closing this in favor of #2621.

PR #2623, opened 2020-07-01
(kotlin) remove very old keywords and update example code

kageru:
These aren’t valid Kotlin keywords and just cause issues with autodetection
because Scala, which is syntactically quite similar to kotlin, actually has a `trait` keyword.
It even says `// to be deleted soon`, whenever soon is, so I decided to do it now.

Also replace `trait` in the example
(which would no longer be highlighted)
and format the code according to the Kotlin coding conventions.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,9 +7,11 @@ New themes:
 Language Improvements:
 
 - enh(matlab) Add new R2019b `arguments` keyword and fix `enumeration` keyword (#2619) [Andrew Janke][]
+- fix(kotlin) Remove very old keywords and update example code (#2623) [kageru][]
 
 [Andrew Janke]: https://github.com/apjanke
 [Samia Ali]: https://github.com/samiaab1990
+[kageru]: https://github.com/kageru
 
 
 ## Version 10.1.1
--- a/src/languages/kotlin.js
+++ b/src/languages/kotlin.js
@@ -14,9 +14,7 @@ export default function(hljs) {
       'crossinline dynamic final enum if else do while for when throw try catch finally ' +
       'import package is in fun override companion reified inline lateinit init ' +
       'interface annotation data sealed internal infix operator out by constructor super ' +
-      'tailrec where const inner suspend typealias external expect actual ' +
-      // to be deleted soon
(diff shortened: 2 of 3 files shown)


yyyc514:
Your removing 5 different keywords here, not just trait. Do you know when they were they removed from the Kotlin language?  We don't have a specific policy on handling deprecations like this but in general if code still exists using these things we'd like to still be highlighting them - so we've never been in a hurry to remove older keywords.

Auto-detect is always problematic and not a justification in and of itself for removing older keywords.

kageru:
As far as I know, none of these were ever valid keywords. I don’t know why they’re in the syntax definition to begin with.
`trait` is a keyword from Scala, and the other four are Java keywords.

They’re nowhere to be found in the [list of keywords in the Kotlin language documentation](https://kotlinlang.org/docs/reference/keyword-reference.html).

yyyc514:
> https://stackoverflow.com/questions/46757825/does-kotlin-have-a-trait-type

Trait most definitely was.  Though I'm happy to say 5 years was "some time ago though"... I wonder about the other keywords.    Would be nice to have a Kotlin person weigh in on this.

If the problem that let you here is truly just relevancy though how about updating the comment to "old syntax" and simply making them all 0 relevancy?  Are they harming current highlighting of Kotlin code?

yyyc514:
Ok let me review this again.  I think we can put some weigh into Sanni's comment 4 years ago: "delete soon". https://github.com/highlightjs/highlight.js/commit/af601cb4ca53fdbf6cd8f33f9ea9603581691113

kageru:
I don’t think they harm anything. The reason I removed them was because I saw the example on the hljs website using `trait`. We can keep the wrong keywords, but then I’d at least change the example to avoid future confusion.

yyyc514:
Can you please also add an entry to CHANGES.md?

kageru:
I’ll do that.

yyyc514:
> The reason I removed them was because I saw the example on the hljs website using trait. 

Ah I forget we use the detect samples as sample code for the website.

yyyc514:
@kageru Thanks for spotting!

PR #2636, opened 2020-07-20
use null prototype objects for languages/aliases

night:
`getLanguage` currently returns values for object prototype method names. By swapping to objects with a null prototype, we can fix this weird behavior.

### DIFF
--- a/test/api/getLanguage.js
+++ b/test/api/getLanguage.js
@@ -41,4 +41,34 @@ describe('.getLanguage()', () => {
     result.should.have.property('aliases').with.containEql('cs');
     should.strictEqual(result, hljs.getLanguage('csharp'))
   });
+
+  it('should not succeed for hasOwnProperty', () => {
+    const result = hljs.getLanguage('hasOwnProperty');
+
+    should.strictEqual(result, undefined);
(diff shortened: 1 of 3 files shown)


test/api/getLanguage.js
```diff
@@ -41,4 +41,34 @@ describe('.getLanguage()', () => {
...
+    const result = hljs.getLanguage('constructor');
+
+    should.strictEqual(result, undefined);
+  });
+
+  it('should not succeed for __proto__', () => {
+    const result = hljs.getLanguage('__proto__');
+
+    should.strictEqual(result, undefined);
+  });
```
yyyc514, line 73: I don't one or two tests here are sufficient to prove the point, we don't need to test for EVERY built-in key.

night:
Discord uses `getLanguage` to validate that a language specified exists in highlightJS and retrieve metadata about the language for code block highlighting in chat. Because highlightJS returns prototype values instead of the highlight languages themselves, the result is a few different bugs in our clients which expect the return type to be only `Language | undefined`.

yyyc514:
@night Thanks so much!

PR #2640, opened 2020-07-27
Update README.md

ElectricRCAircraftGuy:
Add missing `</link>` closing tag. Add clarity for people to help them more easily and clearly know how to use this.

Side note: the missing `</link>` closing tag bug is found on both examples at the top of this webpage too: https://highlightjs.org/download/. Please fix that page as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,20 @@ Please read [VERSION_10_UPGRADE.md](https://github.com/highlightjs/highlight.js/
 
 The bare minimum for using highlight.js on a web page is linking to the
 library along with one of the styles and calling
-[`initHighlightingOnLoad`][1]:
+[`initHighlightingOnLoad`][1]. 
 
 ```html
-<link rel="stylesheet" href="/path/to/styles/default.css">
+<link rel="stylesheet" href="/path/to/styles/default.css"></link>
 <script src="/path/to/highlight.min.js"></script>
+<script>hljs.initHighlightingOnLoad();</script>
+```
+
+True example: place the following code between the `<head> </head>` tags on your website to use the hosted **cdnjs** CDN (Content Delivery Network).  
+_Note: you may find the latest **cdnjs** or **jsdelivr** Content Delivery Network (CDN) URLs here: https://highlightjs.org/download/._
+```html
+<link rel="stylesheet"
+      href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.2/styles/default.min.css"></link>
+<script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.2/highlight.min.js"></script>
 <script>hljs.initHighlightingOnLoad();</script>
 ```
 


taufik-nurrohman:
That&rsquo;s not the way HTML works bro! :wink:

ElectricRCAircraftGuy:
@taufik-nurrohman, Google Blogger disagrees: 
image

It accepts either `</link>` at the end or just the shortened `/>` form at the end, but it throws errors and refuses to work without one of those explicitly closed tag forms.

taufik-nurrohman:
The second one is because Blogger uses XHTML. The third one is HTML5. The first one is invalid. For Blogger, use the second one.

ElectricRCAircraftGuy:
I used the first one and Blogger did accept it, but it converted it to the 2nd one. Can we update the readme here to use the 2nd one then? As far as I understand it, it works in cases where the first one would work too, so it is the most universally-accepted version which works in the most places.

yyyc514:
The README already contains all the same information (CDN hosting info is towards the bottom), so the added chunk of examples is unnecessary.

If someone is using XHTML it goes without saying that every tag needs to be closed, that's something someone using XHTML *needs* to know and it has nothing to do with our library.  As XHTML is an older standard now, our documentation is for HTML5, which works just fine without a closing tag.  We don't need to update our docs because of an old standard - and we definitely don't want to encourage wasted  markup for 99% of users who don't need it.

ElectricRCAircraftGuy:
Alright. I can't convince you. For what it's worth, I lost about 3 hrs on this today and about 2 years of anxiety leading up to it since I didn't know how to get syntax highlighting on my blog for about that long, so I guarantee you you'd have more users and non-html-experts would have an easier time with the process with my additions.

ElectricRCAircraftGuy:
I also didn't know what a CDN was. Your readme is not written for a beginner (the well-above-average blogger) to understand. My pretty minor changes made it better for that person.

yyyc514:
> For what it's worth, I lost about 3 hrs on this today

I'm truly sorry to hear that but we also don't want to teach 99% of the people using more modern systems the WRONG way to write HTML, so it's complicated.   Most new things are built in HTML5.  If you'd posted an issue after spending only 30 minutes we could have helped you out and you might have saved 2.5 hours.  It's not obvious to me that the improvements you're proposing here would help others in the future.

> I also didn't know what a CDN was.

Hmmmm, I think I'd either consider that knowledge someone needed to learn before using the library via CDN.  Also the only other change you made was smash two code blocks together... one part of the readme explains how to use up the library, another explains how to fetch the library via CDN.  _The docs already cover both items - and no one else has raised an issue on this matter._

A "Using with Blogger" section MIGHT be more helpful, but I'm not sure that belongs in the README either vs some sort of extra documentation "using Highlight.js with different systems".

Right now it seems the README is sufficient for most people.

taufik-nurrohman:
I still remember that time when I thought jQuery was a new programming language specialized for adding animation effect to the website.

And so, every article about jQuery I have found on the internet told me to copy and paste that damn jQuery library from Google CDN into my template, just above the `</head>` _code_.

Ended up having a lot of jQuery library with various versions loaded everywhere into my website haha 😄

Someday you will be able to escape from that level if you continue to learn 👍

yyyc514:
So be clear I'm not saying our docs can't be improved, just that this PR is not an improvement.  Some ideas that might be useful improvements:

- Adding a sentence in the CDN section explaining what a CDN is.
- Adding a note to the top example explaining it's using the library locally on your website and that there are lots of ways to get the library (with maybe a link to "how to get the library" below).

taufik-nurrohman:
Prism has [a section](https://prismjs.com/#tutorials) contains links pointed to external tutorial, for beginners.

yyyc514:
Yes, I think we'd be happy to link to some resources like that... although I'm not sure anything short of a Blogger/XHTML example would have solved the exactly issue here - which seems to mostly be about HTML 5 vs XHTML.

PR #2643, opened 2020-08-02
class definitions could start with keyword enum

ezksd:
Enum is a special type in java, which doesn't like enum in other programming languages. It's not a primitive type, enum has the same construct of class.  We also consider it a syntax sugar to define a special class. And enum definition located at subchapter of classes in [java language specification](https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.9) too. 
```java
enum Coin {
    PENNY(1), NICKEL(5), DIME(10), QUARTER(25);
    Coin(int value) { this.value = value; }

    private final int value;
    public int value() { return value; }
}
```


yyyc514:
Looks good.  Please also update CHANGES.md

PR #2644, opened 2020-08-04
Remove console warnings

Himura2la:
The deleted outro adds to every language after the return statement, which generates the:
`Warning: unreachable code after return statement (Firefox)`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Stmt_after_return

### DIFF
--- a/tools/build_config.js
+++ b/tools/build_config.js
@@ -43,7 +43,6 @@ module.exports = {
         },
         output: {
           format: "iife",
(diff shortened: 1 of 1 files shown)


joshgoebel:
This is a remnant of the build system.  Some 3rd party languages need this line because they don't export a single function they actually export a module.  So a PR couldn't be accepted here unless you wanted to dig into the complexities of the build process, track down such a 3rd party module, etc...

PR #2646, opened 2020-08-05
[cherry-pick] enh(csharp) highlight generics in more cases (#2599)

Himura2la:
Port the #2599 pull-request into version 9, because we need this in IE11

### DIFF
--- a/src/languages/cs.js
+++ b/src/languages/cs.js
@@ -86,6 +86,14 @@ function(hljs) {
     ]
   };
 
+  var GENERIC_MODIFIER = {
+    begin: "<",
+    end: ">",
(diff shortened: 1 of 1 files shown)


joshgoebel:
@Himura2la Thank you for your contribution, but the `9-*-stable` branch is closed for all but critical security updates and we have no plans to maintain it otherwise.  You may want to just maintain these back ports in  your own private repository (which you're welcome to share with others, or not).

I'll be updating the README/docs shortly to make our policy on this clear.

joshgoebel:
https://github.com/highlightjs/highlight.js/blob/master/OLD_VERSIONS.md

Issue #2651, opened 2020-08-10
(php) Support indented heredoc closing marker


oadam:
Indented closing markers are [valid in heredoc strings since PHP version 7.3
](https://laravel-news.com/flexible-heredoc-and-nowdoc-coming-to-php-7-3)

Here is a fiddle that reproduces the issue : https://jsfiddle.net/ywhrtvzo/
The string mode should end at the second `EOTEXT` marker

I've checked the code to see if I could fix it myself, and the [END_SAME_AS_BEGIN](https://github.com/highlightjs/highlight.js/blob/a010c151328b8944a7a03e5335956de8baf63295/src/lib/modes.js#L157) used in the ruby language for heredoc seems very promising, but I was not able to figure out how to adapt the php mode.
If I understand it well, a capturing group in the `begin` regex could be used for the `end` matching, which would be perfect !

joshgoebel:
> If I understand it well, a capturing group in the begin regex could be used for the end matching, which would be perfect !

Yes, the first capture group in both the being and end is what is used by default with END_SAME_AS_BEGIN... looks like it'd be a fairly straight forward PR.

You'd just need to add the match groups and remove the `^` from the end regex.

We've a love a PR!

eytienne:
I am working on a PR but sadly, it is not as simple as you said, look at the this for the details : https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes
It will need a bit of custom code, but maybe I am confusing an highlighter with a parser.

joshgoebel:
What seems to be the problem?  We just have to match the beginning and ending.  We already do it for "normal" heredocs (although poorly since we're not even making sure the end tag matches, etc).

So you just need to edit the `end` rule to match ANYWHERE (not just at the start/end of a line) and then use a END_SAME_AS_BEGIN to make sure the end is the same as the beginning...

If you make a PR of what you have so far I can take a look.

eytienne:
#2658

eytienne:
I followed your advice, I was too focused on the rfc but we only need the level of precision of an highlighter.

eytienne:
Closed by #2658

PR #2658, opened 2020-08-18
enh(php) detect newer more flexible HEREdoc syntax 

eytienne:
Allows the use of the new indented heredoc : https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes


eytienne:
#2651 Resolves this issue

src/languages/php.js
```diff
@@ -18,18 +19,50 @@ export default function(hljs) {
...
+        begin: 'b"', end: '"'
+      }],
+      contains: [hljs.BACKSLASH_ESCAPE, SUBST]
+    }
+  );
+  var HEREDOC = {
+    begin: /<<<[ \t]*(?<marker>\w+)\n(?:.*\n)*?^[ \t]*\k<marker>\b/,
+    returnBegin: true,
+    contains: [
+      { begin: /<<<[ \t]*/ },
```
joshgoebel, line 52: I don't think se need any of this... move the <<< into the BEGIN match...  we shouldn't only need a simple being/end match here (inside the END_SAME_AS_BEGIN) helper, no?

joshgoebel: Ie, we don't need to find the end first...

src/languages/php.js
```diff
@@ -18,18 +19,50 @@ export default function(hljs) {
...
+  var DOUBLE_QUOTED = hljs.inherit(
+    hljs.QUOTE_STRING_MODE,
+    {
+      illegal: null,
+      variants: [{
+        begin: 'b"', end: '"'
+      }],
+      contains: [hljs.BACKSLASH_ESCAPE, SUBST]
+    }
+  );
```
joshgoebel, line 47: Pretty sure this is wrong and *not* the equivalent of what we were doing before, it's also MUCH harder to read. (it's unclear that there are two branches).  Pretty sure variants needs to spell out ALL the variants. You're also making every variant null, which is also incorrect, etc...

Variants are a complete list not "in addition to"....

I'd make another attempt here or just revert this to what we had before.

test/markup/php/heredoc.txt
```diff
@@ -1,14 +0,0 @@
...
-
-echo <<<EOT
-  string
-  EOT
-  still string
-EOT;
-
-array(<<<EOD
-foobar!
-EOD
```
joshgoebel, line 13: Don't we also need a test where the end is indented?

joshgoebel: Ok you moved it all to strings... I can live with that.

src/languages/php.js
eytienne: right, I missed that copying from ruby.js

src/languages/php.js
eytienne: Maybe I misunderstood thinking the variants of the main will result in the main + expansion for each variant, isn't it ? It was working as expected testing it, so how ? I wanted to reflect the fact that each string (single and double) has its own "binary" equivalent for PHP6 compatibility...

joshgoebel: > Maybe I misunderstood thinking the variants of the main will result in the main + expansion for each variant, isn't it?

No.  Variants are exclusive.  

> I wanted to reflect the fact that each string (single and double) has its own "binary" equivalent for PHP6 compatibility...

I think we can just do this with comments the way this was prior. Otherwise you're repeating a lot of the string bits over and over for no real big win...

eytienne: Ok so I noticed when retesting my code that binary strings were broken so I fixed it. I seems the variants where not expanded recursively, a bug maybe ? (Prior the substitution was not used in binary strings, surely not a big win because it surely affects nobody as it is a backward compatiblity residue but seems clearer...)

src/languages/php.js
```diff
@@ -18,18 +22,43 @@ export default function(hljs) {
...
+  };
+  var SINGLE_QUOTED = hljs.inherit(hljs.APOS_STRING_MODE, {
+    illegal: null,
+  });
+  var DOUBLE_QUOTED = hljs.inherit(hljs.QUOTE_STRING_MODE, {
+    illegal: null,
+    contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
+  });
+  var HEREDOC = {
+    begin: /<<<[ \t]*(?<marker>\w+)\n(?:.*\n)*?^[ \t]*\k<marker>\b/,
```
joshgoebel, line 40: Did we not discuss simplifying all this?

eytienne: yes i moved the "<<<" into the end_same_as_begin, it seems good now, the regex is clear with the named group marker, ready yo be merged ?

joshgoebel: My point is why do we need lines 40/41 at all? ALL we really need here is just the END_SAME_AS_BEGIN rule without any other rules, no?

eytienne: right, I missed that copying from ruby.js, fixed!

PR #2660, opened 2020-08-19
enh(csharp) add keywords `record` and `init` (C# 9.0)

Youssef1313:


joshgoebel:
Please show examples of how both of these are used... sometimes it's not just as simple as adding them to the keywords list.

Youssef1313:
```
using System;
public record MyRecord
{
    public string SomeMember { get; set; }
}
```

Not sure if it requires some complex rules or not.

joshgoebel:
Yes, it's more like a class.  See lines 148 or 159.  We'd expect that `MyRecord` was highlighted as a `title` as well, which requires a custom rule.

`init` looks like more a normal boring keyword, at least from https://csharp.christiannagel.com/2020/04/21/c-9-positional-or-nominal-creation/ this.

joshgoebel:
So we'd need a new rule for `record`, plus a simple expects test to prove that the rule works as expected.

Youssef1313:
I'm not familiar with making `rule`s. :(
May I keep this PR for `init` only and open an issue to track adding `record`, or you'd like to close it all together and open an issue to track both keywords?

joshgoebel:
> May I keep this PR for init only and open an issue to track adding record

This please.  Thought you should consider giving it a try, it's not hard.  I'm happy to help.

It'd look pretty much like:

```ts
      {
        beginKeywords: 'record', end: /[{;=]/,
        illegal: /[^\s:]/,
        contains: [
          TITLE_MODE,
          // are generics allowed?
          hljs.C_LINE_COMMENT_MODE,
          hljs.C_BLOCK_COMMENT_MODE
        ]
      },
```

joshgoebel:
Either way this would need a mention in CHANGES.md. :-)

Youssef1313:
That doesn't seem hard.

It looks like generics are allowed based on what I'm getting in sharplab.
I've updated *csharp.js* and will add the test and update *CHANGES.md* soon.

@joshgoebel Thanks for helping!

joshgoebel:
@Youssef1313 Thanks so much for your contribution!  Wasn't so hard, was it?

PR #2669, opened 2020-09-01
fix(bash) Less false positives for keywords in arguments ie, `--keyword-flag`

### LINKED ISSUE #2668: (shell) Shell tokenization doesn't treat hyphens as parts of words
**Describe the issue**

I've run into a case similar to #1740 and other issues. In my case, we had `--enable-foo` on our site, and highlightjs thinks that `enable` is a keyword.

**Sample Code to Reproduce**

<details><summary>minimal full reproduction</summary>

```html
<!DOCTYPE html>
<html lang="en">
<head>
<link rel="stylesheet"
      href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.2/styles/github.min.css">
<script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.2/highlight.min.js"></script>
</head>
<body>

<pre class="highlightjs highlight"><code data-lang="sh" class="language-sh hljs">--enable-https</code></pre>

<script>hljs.initHighlightingOnLoad()</script>

</body>
</html>
```

</details>

**Expected behavior**

In this specific case, highlightjs shouldn't treat `enable` as a keyword. More generally, when parsing shell scripts, I believe that `-` should be considered a "word character".
I'm aware from reading previous issues that when this comes up there's hesitance to try to do much with the bash parsing unless someone is ready to write a real parser. But I think hyphenate commands and options are a common case. Even if there are cases this might start to fail on which worked before (I can't think of any, but maybe that's my limited imagination), they're probably much more exotic than `--set-foo`, `--enable-bar`, and, `cmd-case-while-blahblahblah`.

**Additional context**

I'm okay with the idea of highlightjs closing out this issue if it's just too much the same as its predecessors.

If backwards compatibility is a major concern, perhaps this could be an option which is set by the user somehow?
`hljs.SetParserOption("shell", "hyphens-are-wordchars")`? (I have very little notion of how highlightjs works on the inside and how configurable the parsers are or ought to be by design.)

sirosen:
In bash, option, param, and command names may contain builtins as hyphen-delimited components. As in `foo --set-bar`.

In order to handle this, just add `-` to the pattern's set of word characters. Includes a new test for bash tokens containing keywords, which checks hyphens and underscores for a few common cases.

Against current `master`, the new test fails, which helps to confirm the fix.

resolves #2668

----

Aside: I really have to compliment you guys on the structure of the testsuite. Even with my very limited knowledge in this domain, adding a new failing test case and verifying it was a breeze. Thanks to everyone who works on this project!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ Big picture:
 
 Language Improvements:
 
+- fix(bash) Fewer false positives for keywords in arguments (#2669) [sirosen][]
 - fix(js) Prevent long series of /////// from causing freezes (#2656) [Josh Goebel][]
 - enh(csharp) Add `init` and `record` keywords for C# 9.0 (#2660) [Youssef Victor][]
 - enh(matlab) Add new R2019b `arguments` keyword and fix `enumeration` keyword (#2619) [Andrew Janke][]
@@ -36,6 +37,7 @@ Language Improvements:
 [ezksd]: https://github.com/ezksd
 [idleberg]: https://github.com/idleberg
 [eytienne]: https://github.com/eytienne
+[sirosen]: https://github.com/sirosen
 
 
 ## Version 10.1.1
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -83,7 +83,7 @@ export default function(hljs) {
     name: 'Bash',
     aliases: ['sh', 'zsh'],
     keywords: {
-      $pattern: /\b-?[a-z\._]+\b/,
+      $pattern: /\b-?[a-z\._-]+\b/,
       keyword:
         'if then else elif fi for while in do done case esac function',
       literal:
--- a/test/markup/bash/token-containing-keyword.expect.txt
+++ b/test/markup/bash/token-containing-keyword.expect.txt
@@ -1,0 +1,10 @@
+<span class="hljs-comment"># a keyword as part of an option</span>
+mycmd --disable-foo
+
+<span class="hljs-comment"># a keyword as part of a parameter</span>
+some-cmd set-some-setting
+some-cmd set_some_setting
+
+<span class="hljs-comment"># a keyword as part of command</span>
+check-case foo
+check_case foo
--- a/test/markup/bash/token-containing-keyword.txt
+++ b/test/markup/bash/token-containing-keyword.txt
@@ -1,0 +1,10 @@
+# a keyword as part of an option
+mycmd --disable-foo
+
+# a keyword as part of a parameter
+some-cmd set-some-setting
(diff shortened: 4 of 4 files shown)


joshgoebel:
**Yes, nice!**

Perhaps add an example of one or two of the `-a` (or whatever) single letter keywords just to make sure they are still highlighted properly. (I didn't see a test for that)

sirosen:
I just got to adding such a test, and I was surprised to see that it failed:
```
# test keyword usages with hyhens at their start
if [ $VAR -ne 0 ]; then echo 'hi'; fi
[ $FOO -eq 0 ] && [ ! -f $BAR ] || exit 2
```
became
```
<span class="hljs-comment"># test keyword usages with hyhens at their start</span>
<span class="hljs-keyword">if</span> [ <span class="hljs-variable">$VAR</span> -ne 0 ]; <span class="hljs-keyword">then</span> <span class="hljs-built_in">echo</span> <span class="hljs-string">&#x27;hi&#x27;</span>; <span class="hljs-keyword">fi</span>
[ <span class="hljs-variable">$FOO</span> -eq 0 ] &amp;&amp; [ ! -f <span class="hljs-variable">$BAR</span> ] || <span class="hljs-built_in">exit</span> 2
```

I'm reading and trying to understand why this is happening, but I'm not sure I'll be able to figure it out.

sirosen:
Actually, I'm not sure if that's a failure or not? These terms (e.g. `-f`) don't appear to be marked up today -- at least, not when I test. I'm not sure what the `_` section of the language spec, containing them, is for.
I'm happy to make relevant/related changes, but I need some direction on that.

egor-rogov:
Hmm... Options (like `-lt` etc) just don't work and it looks like they never did. That's because `\b` (word boundary) doesn't treat hyphens as a part of a word. We have to replace `\b` with explicit definition of a boundary.
On the other hand, if we do, we'll get rather ugly elements like `<span class="hljs-_">...</span>`, which was supposed, as I understand, as relevance boosters. But do we really what it?

joshgoebel:
@egor-rogov Creating a new issue so this doesn't get bogged down in that. 

@sirosen Thanks for looking into. :-)

And thanks for contributing!  It's truly appreciated.

joshgoebel:
@sirosen Can you please update CHANGES.md and then we can get this merged?

joshgoebel (approved):

sirosen:
I pushed a commit with the changelog update, but also did a squash-rebase so that it will merge cleanly.

joshgoebel:
> I pushed a commit with the changelog update, but also did a squash-rebase so that it will merge cleanly

PR #2672, opened 2020-09-04
fix(tooling) developer.html does not render highlighted language properly

egor-rogov:
Our `developer.html` tool correctly shows markup using specified language, but always displays rendering in autodetection mode.
To reproduce, enter e. g.
```
for i in *.sh; do echo $i; done
```
and choose Language: bash. The code is (erroneously) rendered for `nginx`, while markup is (correctly) for `bash`.

I'm not familiar with vue and somehow managed to fix the issue although in a disgusting way. There should be a better solution for sure.

### DIFF
--- a/tools/developer.html
+++ b/tools/developer.html
@@ -70,7 +70,7 @@
     </div>
     <div>
       <h3>Rendered in <span class="rendering_time">...</span> ms [<span class="rendering_stats">...</span>]</h3>
-      <highlightjs auto :code="code" />
+      <highlightjs :language="language" :code="code" />
     </div>
     <div>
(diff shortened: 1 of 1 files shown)


joshgoebel:
Oh wow, that was a bad mistake on my part. :-/  This isn't bad, but I'm sure there is a way to move that into a single method and then just call it from both...  I'll see.

joshgoebel:
@egor-rogov Back to you. :-)  If that looks good I will merge.

egor-rogov:
Yes, that's good, thanks!

joshgoebel (approved):

PR #2676, opened 2020-09-08
Introducing the new C# 9 keywords

IEvangelist:
Hi 👋🏼 highlight.js friends!

I'd like to introduce several new C# keywords. These are being added as part of C# 9.

- `init`
- `record`
- `and`
- `not`
- `or`
- `nint`
- `nunit`
- `with`

See our official docs: https://docs.microsoft.com/dotnet/csharp/whats-new/csharp-9

### DIFF
--- a/src/languages/csharp.js
+++ b/src/languages/csharp.js
@@ -13,13 +13,14 @@ export default function(hljs) {
       // Normal keywords.
       'abstract as base bool break byte case catch char checked const continue decimal ' +
       'default delegate do double enum event explicit extern finally fixed float ' +
-      'for foreach goto if implicit in init int interface internal is lock long ' +
+      'for foreach goto if implicit in int interface internal is lock long nint nunit ' +
       'object operator out override params private protected public readonly ref sbyte ' +
       'sealed short sizeof stackalloc static string struct switch this try typeof ' +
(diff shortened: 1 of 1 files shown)


joshgoebel:
Looks good. Could you add a CHANGES.md entry please?

Also, is there any chance you'd mind grouping all the "datatype" keywords as well (since you seem to know C#) and perhaps it'd be quick for you to do?

In the future we should probably highlight them as data types instead of keywords proper, what do you think?

IEvangelist:
Hi @joshgoebel - I can update the CHANGES.md, no problem. I'm not sure I follow the other suggestions - as I'm not that familiar with how some of the internals work in highlight.js.

- What is this `datatype` keyword you speak of and how do I group things as that?

joshgoebel:
>  I'm not sure I follow the other suggestions - as I'm not that familiar with how some of the internals work in highlight.js.

Well for starters I just meant grouping them literally, with a comment (like the keywords are already grouped a bit into categories)... but really we probably should highlight them as "type" or "built_in"... but getting them grouped first would be a step in that direction.

IEvangelist:
I'm going to close and reopen a new PR, properly forking the repo.

IEvangelist:
Hey @joshgoebel - see #2679

PR #2678, opened 2020-09-09
fix(parser) complete fix for resuming matches from same index

### LINKED ISSUE #2649: (parser) continueScanAtSamePosition can break highlighting if no other matches are found
**Describe the issue**
When accessing a reflected class object using `.class`, a string following it doesn't get highlighted as such.

java

highlight

**Sample Code to Reproduce**
For example, `ImmutablePair.of(stuff, "bar")` works fine, but `ImmutablePair.of(Stuff.class, "bar")` doesn't highlight `"bar"` as a string.  There's a live repro using highlightjs 10.1.2 here:  https://jsfiddle.net/chmxa1ze/

**Expected behavior**
The string should be highlighted as such.

**Additional context**
This may be related to #2428 and/or its fix.

joshgoebel:
Resolves #2649.

Both our prior definitions of resuming were "off the mark"... resuming properly actually means you have to briefly must have _two parallel lines of regex search_.

1. You have to resume the prior multi-match at the SAME [index] position ("resume" meaning only matching expressions that haven't been tried yet). It's possible one of those remaining expressions could still match at `index`. 
2. You also have to begin a full multi-match at the next [index+1] position.  This is to allow ALL the potential regex expressions a chance to match starting at `index+1`... 
3. Then you "merge" the results... meaning whichever matched first (smaller index) is the next actual match.

The prior solution ignored the need for number 2 above... so we resumed matching but ONLY looking for the remaining expressions (however far in the future) so we would miss a lot of other quite valid expressions.

Or another way to think of this might be a "rotating match".

Normally our regex matching is looking for say [A,B or C]:

```
match [A,B,C] at offset 0
```

But after we match B and decide to ignore it we really want to rotate the matches:
 [C,A,B(but not the same B)]. So we do this:

```ts
// we don't need to worry about A or B at 0, we already ruled them out
resume match [C] at offset +0
full match [A,B,C] at offset +1
```

Technically that last C is doing a little extra work potentially (in some cases it might match the exact same thing the first matcher does), but it shouldn't be too bad, as this is already an edge case.

---

An actual Java real-life case:

```java
ImmutablePair.of(Stuff.class, "bar");
23
```

For simplicity our rules in Java are: `[string, "class" (begin keyword), number]`.  Class having the magic "not proceeded by "." internal rule (which means the "class" match here will be ignored)... At that point technically we still need to "complete" the existing matcher and see if it might still match `number`... but if we do alone (without considering the full ruleset) then we end up skipping the string `"bar"` and only highlighting 23 (the first number we find)...

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
     "test": "mocha --globals document test",
     "test-markup": "mocha --globals document test/markup",
     "test-detect": "mocha --globals document test/detect",
-    "test-browser": "mocha --globals document test/browser"
+    "test-browser": "mocha --globals document test/browser",
+    "test-parser": "mocha --globals document test/parser"
   },
   "engines": {
     "node": "*"
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "test": "mocha test",
     "test-markup": "mocha test/markup",
     "test-detect": "mocha test/detect",
-    "test-browser": "mocha test/browser"
+    "test-browser": "mocha test/browser",
     "test-parser": "mocha test/parser"
   },
   "engines": {
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -498,20 +498,13 @@ const HLJS = function(hljs) {
           // considered for a potential match
           resumeScanAtSamePosition = false;
         } else {
-          top.matcher.lastIndex = index;
           top.matcher.considerAll();
         }
+        top.matcher.lastIndex = index;
 
         const match = top.matcher.exec(codeToHighlight);
         // console.log("match", match[0], match.rule && match.rule.begin)
 
-        // if our failure to match was the result of a "resumed scan" then we
-        // need to advance one position and revert to full scanning before we
-        // decide there are truly no more matches at all to be had
-        if (!match && top.matcher.resumingScanAtSamePosition()) {
-          advanceOne();
-          continue;
-        }
         if (!match) break;
 
         const beforeMatch = codeToHighlight.substring(index, match.index);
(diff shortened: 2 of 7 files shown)


joshgoebel:
I'm thinking perhaps we could simplify further in that I'm pretty sure the only RESUME match we care about would HAVE to be at position 0... so if the first matcher came back with non index === 0 match I think we could perhaps just ignore it completely in favor of the second.

src/highlight.js
```diff
@@ -498,20 +498,13 @@ const HLJS = function(hljs) {
...
+        top.matcher.lastIndex = index;
 
         const match = top.matcher.exec(codeToHighlight);
         // console.log("match", match[0], match.rule && match.rule.begin)
 
-        // if our failure to match was the result of a "resumed scan" then we
-        // need to advance one position and revert to full scanning before we
-        // decide there are truly no more matches at all to be had
-        if (!match && top.matcher.resumingScanAtSamePosition()) {
-          advanceOne();
```
egor-rogov, line 512: `advanceOne()` is no longer needed.
Otherwise it's all looks good!

egor-rogov (approved):

allejo (approved):

package.json
```diff
@@ -33,6 +33,7 @@
     "test-markup": "mocha test/markup",
     "test-detect": "mocha test/detect",
     "test-browser": "mocha test/browser"
+    "test-parser": "mocha test/parser"
```
joshgoebel, line 36:
```suggestion
    "test-detect": "mocha test/detect",
    "test-browser": "mocha test/browser",
    "test-parser": "mocha test/parser"
```

PR #2679, opened 2020-09-09
enh(csharp) Add all v9 keywords, and other missing keywords

IEvangelist:
Hi 👋🏼 highlight.js friends!

I'd like to introduce several new C# keywords. These are being added as part of C# 9.

- Added C# 9 keywords
  - `init`
  - `record`
  - `and`
  - `not`
  - `or`
  - `nint`
  - `nunit`
  - `with`
- Added other missing keywords (i.e.; `default` as literal, etc.)
- Improved organization
- Added `built_in` keywords

See our official docs: https://docs.microsoft.com/dotnet/csharp/whats-new/csharp-9

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -288,3 +288,4 @@ Contributors:
 - Alexandre Grison <a.grison@gmail.com>
 - Jim Mason <jmason@ibinx.com>
 - lioshi <lioshi@lioshi.com>
+- David Pine <david.pine.7@gmail.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-## Version 10.2.1
+## Version 10.2.1 (next up)
 
 Language Improvements:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-## Version 10.2.1 (next up)
+## Version 10.3.0 (next up)
 
 Language Improvements:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+## Version 10.2.1
+
+Language Improvements:
+
+- Add all C# 9 keywords, and other missing keywords (#2679) [David Pine][]
+
+[David Pine]: https://github.com/IEvangelist
+
 ## Version 10.2.0
 
 Parser Engine:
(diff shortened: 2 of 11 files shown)


joshgoebel:
Very nice.

IEvangelist:
> Very nice.

Hi again @joshgoebel - thank you very much! I've fixed all the broken tests that I caused, this should be good to go now. Thanks again.

IEvangelist:
Hi @joshgoebel - assuming this gets merged, how long until it is published as an official package update?

Youssef1313:
@IEvangelist FYI, if you're asking to get this on the official docs, I *think* docs uses [this fork](https://github.com/DuncanmaMSFT/highlight.js), so the fork needs to first get updated with upstream. (there should be additional steps which I don't know)

Youssef1313:
That sounds great! Thanks for letting me know.

joshgoebel (approved):

joshgoebel:
>  assuming this gets merged, how long until it is published as an official package update?

We do minor releases every 45 days or so and we just did a release.

DuncanmaMSFT:
> That sounds great! Thanks for letting me know.

Hey @Youssef1313 and @joshgoebel , right now we build a subset of languages into a single version of highlight.js that we use on Docs, including some of our custom ones. In the V10 world, with those languages stored elsewhere, is there a build command/process for producing that same single highlight.js file? I was looking at https://highlightjs.readthedocs.io/en/latest/building-testing.html, but didn't see any mention of 3rd party hosted language syntaxes

joshgoebel:
@DuncanmaMSFT See the 3rd_party_quickstart.  The normal build process will "just work" if additional language libraries are checked out out into `extra` inside of the HLJS folder.

joshgoebel:
@DuncanmaMSFT And no more IE11 with version 10 of course.

DuncanmaMSFT:
> @DuncanmaMSFT And no more IE11 with version 10 of course.

Thanks @joshgoebel , for both responses, this is great. The IE11 issue isn't **necessarily** a blocker for us, since the code snippets would still be there and the page would work, it just wouldn't be syntax highlighted and that doesn't seem like a blocker for that specific browser (< 5% of our traffic all up, but < 2% in an area of the site like /dotnet/)

Issue #2521, opened 2020-04-30
More nuance for match classes (class title? function title? etc.)


joshgoebel:
Table of some specific suggestions:

|Name|Description|
|---|---|
|title.class| A class name |
|title.function| A function name |
|title.namespace| A namespace name |
|comment.block| A block comment |
|meta.shebang| A shebang line for a script file |

---

**Is your request related to a specific problem you're having?**

Yes, I'd like to have a way to label/highlight names of a given type outside of their original context.

For example:

```js
class Cat extends Pet { ... 
```

Turns into `<class><title>Cat</title><keyword>extends</keyword><title>Pet</title>`. And if we wanted to style it a certain way we could target `.hljs-class .hljs-title`.  Although Pet is semantically different here and that's not indicated at all.

The same is true for functions:

```js
function cleanMyRoom() {}
```

Turns into `<function><title>cleanMyRoom</title...` and hence could be targeted by a theme with `.hljs-function .hljs-title`... but now we have the usage of these things outside of their definition:

```js
var kitty = new Cat();
Cat.prototype.pet = function() { ... }
```

In both cases [re: `Cat`]  we can easily figure out this is a "class" object and we could apply `title` to it... but is it a class? a namespace? a function name?  We've lost all context.

And for functions:

```js
let launchTheRockets = () => { ...
```

Technically `launchTheRockets` is a name/title here (and we could easily detect that and highlight)... _but it's not the function itself_ (which is `() => {...}`... so again we have a name without context... and we're forced to fall back to the generic `title`.

And this problem get worse for languages that also have namespaces that could also be syntactically inferred... all we have is `title` and perhaps context at the definition site - although we have no `namespace` class...

**The solution you'd prefer / feature you'd like to see added...**

I think we may need to consider some new classes, or expanding existing ones with sub classes:

- *Alternative 1:* New classes `class_name`, `function_name`, `namespace_name`
- *Alternative 2:* Expand existing classes with subclass: `title.function`, `title.class`, `title.namespace`... 

The way alternative 2 would unfolding during rendering CSS might look something like:

```html
<span class="hljs-title hljs-title-function">launchTheRockets</span>
```

IE, the existing "broader" class is preserved.  This would require no changes to any existing styles, while allowing new themes (or upgrades) to target items with more specificity. I'm a little more attached to system 2 as it seems more easily "backwards compatible".

**Any alternative solutions you considered...**

Nope, I outlined my two thoughts above.

**Additional context...**

[...]

PR #2680, opened 2020-09-10
enh(R) major overhaul for the R language grammar 

klmr:
This is a major overhaul of the lexer for R. The PR

* fixes the idiosyncratic R rule for identifiers
* fixes the list of keywords, which was previously just an ad-hoc list of some true keywords and some (but not all) functions from the base package
* adds a category of builtin functions
* adds a syntax rule to match R’s custom `%…%` infix operators
* adds support for ‘roxygen’ doc comments
* adds R 4.0 raw string literal (`r"(…)"`) support
* fixes some inaccuracies and missing features in the number literals, such as hexadecimal binary exponents (e.g. `0x1.ap+5` == 52).

Some of these changes were loosely based on the [R lexer for the “rouge” highlighter](https://github.com/rouge-ruby/rouge/blob/593f3c922270c68cb85e1b9761e8e3f1be790bfc/lib/rouge/lexers/r.rb), which was authored by me.

### To do:

* [ ] Support Unicode identifiers (**help wanted**; couldn’t get `\p{L}` or similar to work in the `IDENT_RE` regex).
* [x] Add test cases (I’ve got test cases but these still need to be polished)
* [x] Test auto-detection
* [x] <del>Support for detecting function declarations/params</del>
* [x] <del>Support for detecting function calls (?)</del>
* [x] <del>Support for S4/RC/R6 class declarations (?)</del>


joshgoebel:
>  Support Unicode identifiers (help wanted; couldn’t get \p{L} or similar to work in the IDENT_RE regex).

That's because nothing in our engine yet uses or supports the `u` flag to turn on Unicode... I'm not sure we'd be opposed to that, but I'm also curious if it's a lot slower or has any downsides. Also might be something we allowed enabling per-language with a language flag... (like `case_insensative`, etc)

If you wanted to play with that idea for R (to see how it worked) take a look at `mode_compiler` and `langRe` - the function that builds regexs for the target language...

src/languages/r.js
```diff
@@ -2,59 +2,105 @@
...
+          hljs.END_SAME_AS_BEGIN({ begin: /r"(-*)\(/, end: /\)(-*)"/ }),
+          hljs.END_SAME_AS_BEGIN({ begin: /r"(-*)\{/, end: /\}(-*)"/ }),
+          hljs.END_SAME_AS_BEGIN({ begin: /r"(-*)\[/, end: /\](-*)"/ }),
+          hljs.END_SAME_AS_BEGIN({ begin: /r'(-*)\(/, end: /\)(-*)'/ }),
+          hljs.END_SAME_AS_BEGIN({ begin: /r'(-*)\{/, end: /\}(-*)'/ }),
+          hljs.END_SAME_AS_BEGIN({ begin: /r'(-*)\[/, end: /\](-*)'/ }),
+          {begin: '"', end: '"', relevance: 0},
+          {begin: "'", end: "'", relevance: 0}
+        ],
+        relevance: 10
```
joshgoebel, line 48: Don't several different languages have `r` strings now?  Not sure relevance 10 should apply.  10 really means "this is pretty much exclusive to our language".

src/languages/r.js
```diff
@@ -2,59 +2,105 @@
...
+            'trigamma trunc unclass untracemem UseMethod xtfrm',
         },
         relevance: 0
       },
+
+      hljs.COMMENT(
+        '#',
+        '$',
+        {
+          returnBegin: true,
```
joshgoebel, line 91: If you want to match ROXYGEN you should just match it, not this match comment then return to begin.

src/languages/r.js
```diff
@@ -66,12 +112,10 @@ export default function(hljs) {
...
-        className: 'string',
-        contains: [hljs.BACKSLASH_ESCAPE],
-        variants: [
-          {begin: '"', end: '"'},
-          {begin: "'", end: "'"}
-        ]
+        // infix operator
+        begin: '%',
+        end: '%',
+        relevance: 10
```
joshgoebel, line 118: Also a lot of things have % so probably doesn't deserve any relevance boost.

src/languages/r.js
klmr: The syntax for these raw strings is different in Python. That said, C++’s syntax (while still different) overlaps with R’s. Ideally I’d like setting the relevance “high, but not exclusive” but the documentation says

> When setting an explicit value it’s recommended to use either 10 or 0.

I can skip this and use the default if that’s more appropriate.

src/languages/r.js
klmr: Yeah, that actually led to detection failures and is already removed — I just forgot to push those changes.

klmr:
>  Support for detecting function declarations/params

This turns out a lot harder than anticipated, so it’s probably beyond the scope of this PR. The reason is that, unlike in most languages, R function declarations aren’t prefixed by a keyword, so I can’t use `beginKeywords`.

For reference, I’ve got the following to work (where the red bit is the “[title](https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html)”):

Screenshot 2020-09-11 at 08 58 39

However, this comes at the expense of the following code (to handle all different identifier and assignment formats), and that *still* doesn’t handle all cases, notably when the declaration is broken across several lines (nor when users use a custom assignment operator or when the declaration is nested, e.g. inside `local(…)`). It also doesn’t handle semantically equivalent declarations such `map <- purrr::map`.

```javascript
const FUNCTION_PROLOGUE = '\\s*%?(<-|<<-|=)%?\\s*function\\s*\\(';

const FUNCTION_TITLE = function (begin) {
  return {
    className: 'title',
    begin: begin,
    end: /\s*%?(<-|<<-|=)%?/,
    relevance: 0,
    excludeEnd: true,
    endsParent: true
  };
};

…

{
  begin: `(${IDENT_RE})\\s*<-\\s*function\\s*\\(`,
  end: /$/,
  returnBegin: true,
  contains: [ FUNCTION_TITLE(/./) ],
  relevance: 10
},
{
  begin: `(${IDENT_RE})${FUNCTION_PROLOGUE}`,
  end: /$/,
  returnBegin: true,
  contains: [ FUNCTION_TITLE(/./) ],
},
{
  begin: '`.*?`' + FUNCTION_PROLOGUE,
  end: /$/,
  returnBegin: true,
  contains: [ FUNCTION_TITLE(/`.*?`/) ]
},
{
  begin: `'.*?'${FUNCTION_PROLOGUE}`,
  end: /$/,
  returnBegin: true,
  contains: [ FUNCTION_TITLE(/'.*?'/) ]
},
{
  begin: `".*?"${FUNCTION_PROLOGUE}`,
  end: /$/,
  returnBegin: true,
  contains: [ FUNCTION_TITLE(/".*?"/) ]
},
```

Ironically, implementing this for formal class declarations would be a lot easier. But since *functions*, not classes, are the fundamental unit of encapsulation in R it would be unwarranted to make the latter stand out more than the former.

klmr:
> Support for detecting function calls

As far as I can see there’s nothing in the CSS class reference for highlight.js that would be appropriate to style a function call site. So I’ll drop this idea as well.

src/languages/r.js
joshgoebel: I think that's there only to avoid people using entirely random values... we probably could add some further clarification (esp. for languages in core).

Grrr, it seems Python is flagging it's own r|ur|u strings as 10 as well... I may need to tone that down... in general bumping relevance for someone as common as a string is usually more likely to get you into more trouble than it's worth as string as such a common thing in so many languages.

IE, pretty soon the fact that your language has a few strings OVERWHELMS *every other detail* of that language, which is not a good thing.

Yes, lets just go with the default.

joshgoebel:
> However, this comes at the expense of the following code (to handle all different identifier and assignment formats),

Looks like about the kind of thing you'd have to do... although you could probably do that all with `variants` instead of a bunch of different rules.  I mean these are really all variants after all.

Also, I'd recommend look-ahead vs returnBegin.  See `regex#lookahead` for a helper function you can use.  Long-term we'd like to try and kill the use of returnBegin.  99% of the time a look-ahead is better.

> semantically equivalent declarations such map <- purrr::map.

Isn't that actually one of the easy cases?  `ident <- ident` or perhaps I don't understand the semantics?

joshgoebel:
> As far as I can see there’s nothing in the CSS class reference for highlight.js that would be appropriate to style a function call site.

See https://github.com/highlightjs/highlight.js/issues/2521.

This is something I'd love to improve.

src/languages/r.js
klmr: Ah, actually it seems I misunderstand the usage of that tag based on my reading of the C# lexer. The Java lexer interprets it completely differently from the former: C# treats the comment start (`///`) as a doctag. Java doesn’t, instead it treats actual doxygen tags (`@param` …) as doctags.

I’ll emulate Java, not C#.

joshgoebel: Oh, yes if you wanted the HTML output to show the nesting you'd need that, but that doesn't seem correct either. :)  Typically (in my understanding) "doctags" are something that appear "inside" of comments.

```
/** this is a comment
@type {string}  This is a doctag, etc.
*/ ditto
```

klmr:
> Also, I'd recommend look-ahead vs returnBegin.

Yeah I’ve been experimenting a bit with some other modes and found `returnBegin` quite problematic, it caused all kinds of errors that I found impossible to diagnose (for somebody more familiar with the code base they’r probably easy). Lookahead assertions on the other hand worked.

> > semantically equivalent declarations such map <- purrr::map.

> Isn't that actually one of the easy cases? `ident <- ident` or perhaps I don't understand the semantics?

There’s nothing in R’s syntax distinguishing functions from regular variables (they *are* regular variables). The way I understand `className: 'title'` this shouldn’t highlight all assignment, only function declarations (etc), right? I think in a language like R this kind of markup just doesn’t necessarily make sense.

joshgoebel:
> Yeah I’ve been experimenting a bit with some other modes and found returnBegin quite problematic

It causes the cursor to literally rewind (not advance)... so you have to be ABSOLUTELY sure that something in `contains` is going to IMMEDIATELY match (and advance the cursor) or you will have an infinite loop with the `returnToBegin` rule just matching itself over and over.  

Which I think we handle now, but it's going to result in your rule just breaking, not working as intended.

joshgoebel:
> There’s nothing in R’s syntax distinguishing functions from regular variables (they are regular variables). The way I understand className: 'title' this shouldn’t highlight all assignment, only function declarations (etc), right?

Ah, yes I see... yes `title` isn't appropriate if you don't know that the thing is a function.

klmr:
I’m relatively happy with the current state of things and as far as I’m concerned this is ready to be merged. Could you please have another look?

Adding Unicode identifiers can be left for another PR — should I open a separate issue to track this item?

test/markup/r/names.expect.txt
```diff
@@ -0,0 +1,34 @@
...
+
+<span class="hljs-literal">NA</span>, <span class="hljs-literal">NA_integer_</span>, <span class="hljs-literal">NA_real_</span>, <span class="hljs-literal">NA_character_</span>, <span class="hljs-literal">NA_complex_</span>, <span class="hljs-literal">NULL</span>, <span class="hljs-literal">NaN</span>, <span class="hljs-literal">Inf</span>
+
+<span class="hljs-comment"># Keywords</span>
+
+<span class="hljs-keyword">function</span>, <span class="hljs-keyword">while</span>, <span class="hljs-keyword">repeat</span>, <span class="hljs-keyword">for</span>, <span class="hljs-keyword">if</span>, <span class="hljs-keyword">in</span>, <span class="hljs-keyword">else</span>, <span class="hljs-keyword">next</span>, <span class="hljs-keyword">break</span>
+
+<span class="hljs-comment"># Not reserved</span>
+
+NULLa, NULL1, NULL., `NULL`, <span class="hljs-string">&#x27;NULL&#x27;</span>, NA_foo_, na_real_, Function, for.
```
joshgoebel, line 19: Is `NULL` not a string of some sort?

test/markup/r/numbers.expect.txt
```diff
@@ -0,0 +1,19 @@
+<span class="hljs-comment"># Numbers</span>
+
+<span class="hljs-number">0</span>, <span class="hljs-number">01</span>, <span class="hljs-number">08</span>, <span class="hljs-number">123456</span>, <span class="hljs-number">1256.701</span>, <span class="hljs-number">123e3</span>, <span class="hljs-number">123E+3</span>, <span class="hljs-number">1.23e-3</span>, <span class="hljs-number">1.23E3</span>, <span class="hljs-number">.25</span>, <span class="hljs-number">2.</span>
+
+<span class="hljs-comment"># Integers</span>
+
+<span class="hljs-number">123L</span>, -<span class="hljs-number">50L</span>
```
joshgoebel, line 7: Is the intention to leave - outside the number intentional?  Just checking.

src/languages/r.js
```diff
@@ -2,76 +2,148 @@
...
+              }
+            },
+            {
+              className: 'doctag',
+              begin: /@[a-zA-Z]+/
+            },
+            {
+              className: 'meta-keyword',
+              begin: /\\[a-zA-Z]+/,
+            }
```
joshgoebel, line 125: What is this trying to accomplish?

src/languages/r.js
```diff
@@ -2,76 +2,148 @@
...
+              begin: '@examples',
+              starts: {
+                contains: [
+                  { begin: /\n/ },
+                  {
+                    begin: /#'\s*(?=@[a-zA-Z]+)/,
+                    endsParent: true,
+                  },
+                  {
+                    begin: /#'\s?/,
```
joshgoebel, line 95: Is `\s?` or not important here?

src/languages/r.js
```diff
@@ -2,76 +2,148 @@
...
+                  }
+                ]
+              }
+            },
+            {
+              // Handle `@param` to highlight the parameter name following
+              // after.
+              className: 'doctag',
+              begin: '@param',
+              starts: {
```
joshgoebel, line 107: Was contains not adequate here?

src/languages/r.js
```diff
@@ -2,76 +2,148 @@
...
+  // R also supports quoted identifiers, which are near-arbitrary sequences
+  // delimited by backticks (`…`), which may contain escape sequences. See
+  // `test/markup/r/names.txt` for examples.
+  // FIXME: Support Unicode identifiers.
+  const IDENT_RE = /(?:(?:[a-zA-Z]|\.[._a-zA-Z])[._a-zA-Z0-9]*)|\.(?!\d)|`(?:[^`]|\\`)+`/;
 
   return {
     name: 'R',
     contains: [
-      hljs.HASH_COMMENT_MODE,
```
joshgoebel, line 15: Typically comments always lead (to eat strings inside comments, etc), is there a reason you changed the sequencing?

joshgoebel (changes requested):

joshgoebel:
@klmr Very nice work overall!

src/languages/r.js
```diff
@@ -2,76 +2,148 @@
...
-            'setGroupGeneric setClass ...',
+            'function if in break next repeat else for while',
           literal:
-            'NULL NA TRUE FALSE T F Inf NaN NA_integer_|10 NA_real_|10 NA_character_|10 ' +
-            'NA_complex_|10'
+            'NULL NA TRUE FALSE Inf NaN NA_integer_|10 NA_real_|10 ' +
+            'NA_character_|10 NA_complex_|10',
+          built_in:
+            // Builtin constants
+            'LETTERS letters month.abb month.name pi T F ' +
```
joshgoebel, line 48: What are T and F?

klmr: Built-in (but overridable) shortcuts for the literals `TRUE` and `FALSE`. For obvious reasons their use is discouraged. But they _do_ exist.

test/markup/r/names.expect.txt
klmr: No, `` `NULL` `` is a “quoted identifier” (plain `NULL` is a literal). It’s generally highlighted the same way as bare identifiers — i.e. not at all.

src/languages/r.js
joshgoebel: I suppose the ability to override is why they aren't considered literals, OK.

test/markup/r/numbers.expect.txt
klmr: It‘s intentional because that’s what the previous lexer did. It also makes lexing easier because I don’t have to disambiguate `a<-3` and `a< -3` (the former needs to be parsed as `a <- 3`). That said, I can change the rule if desired.

src/languages/r.js
joshgoebel: Nevermind I didn't see the `\\`? :-)  With all the escaping easy to miss.

test/markup/r/numbers.expect.txt
joshgoebel: It's fine as-is, was just making sure it was an intentional choice.

src/languages/r.js
klmr: Roxygen documentation has tags (`@tagname`) and pseudo-LaTeX formatting instructions (`\code{foo}`). I used `meta-keyword` for the latter — is that not appropriate?

Here’s an example snippet:

```r
#' @param foo is an \code{integer}
```

The intended (and actual) parse is as follows:

```xml
<comment>
  #'
  <doctag>@param</doctag>
  <meta>foo</meta>
  is an
  <meta-keyword>\code</meta-keyword>
  {integer}
<comment>
```

joshgoebel: Yeah, meta-keyword is appropriate there I think.  I just didn't see the `\\` so eht whole rule made no sense to me. :)

src/languages/r.js
klmr: Actually it isn’t, it’s left over from a refactor. Removed!

src/languages/r.js
```diff
@@ -2,76 +2,148 @@
...
+            }
+          ]
+        }
+      ),
+
+      hljs.HASH_COMMENT_MODE,
 
       {
-        // escaped identifier
-        begin: '`',
```
joshgoebel, line 63: Did you remove this on purpose?  It was probably a relevance boost... I'd consider bringing it back but using the 

```
`[IDENT]`
```

Pattern (with no className) so it can still serve as a relevance boost vs other langauges perhaps stealing these as "strings".  That's probably why it was there originally.

src/languages/r.js
klmr: Oh, right, comments can come first (though Roxygen comments need to come even before, otherwise they won’t be recognised). It’s also important that strings come before identifiers/keywords, since raw strings start with `r`, which would otherwise be parsed as an identifier.

And keywords, in turn, need to be *inside* the main `contains` rather than outside (as other languages seem to handle this). I tried it otherwise and it led to lots of incorrect parses for number literals that weren’t fixed by liberal application of word boundaries. There’s probably a simple fix but I spent some time trying to find it without success.

joshgoebel: > Oh, right, comments can come first (though Roxygen comments need to come even before, otherwise they won’t be recognised). It’s also important that strings come before identifiers/keywords, since raw strings start with r, which would otherwise be parsed as an identifier.

Sounds right.

> And keywords, in turn, need to be inside the main contains rather than outside (as other languages seem to handle this). I tried it otherwise and it led to lots of incorrect parses for number literals that weren’t fixed by liberal application of word boundaries. There’s probably a simple fix but I spent some time trying to find it without success.

Ok, this is a bit strange and at least deserves a comment in the source if we're leaving it this way... can't believe i didn't notice this :-) 

>  I tried it otherwise and it led to lots of incorrect parses for number literals that weren’t fixed by liberal application of word boundaries. 

Does defining all your numbers with word boundaries not help?  Could you provide a small examples of a few failing edge cases?

src/languages/r.js
klmr: The purpose is to have a new lexeme *after* `@param`, not nested inside. Otherwise all the formatting (background, bold …) of `doctag` also applies to it.

joshgoebel: Hmmm, I'm not sure whether to be pedantic here or not, but most of our other grammars do not think of doctags like this but rather like:

```
/**
@param {string} name - the name of the user
*/
````

```
<comment>/**
<doctag>@param <type>{string}</type> <variable>name</variable></doctag> - the name of the user
*/</comment>
```

Ie, the doctag includes the tag and it's metadata.  And some of our themes take advantage of this with `.doctag .type, .doctag .variable`, etc...

I think perhaps what you want here is nesting and `variable` instead of `meta`. Thoughts?

Comment (joshgoebel) [reply src/languages/r.js:107]: > And some of our themes take advantage of this with .doctag .type, .doctag .variable, etc...

Seems I may be a liar here (I can't find it), but the prior point still stands about how this is typically handled.  I'm open to be persuaded, but this feels like something we should sort out and be consistent across grammars - as long-term I wouldn't mind a more modular way to handle doctags in general that is perhaps multi-grammar if possible.

Comment (klmr) [reply src/languages/r.js:107]: > Ie, the doctag includes the tag and it's metadata.

Oh, in that case I will *definitely* change it.

> and `variable` instead of `meta`

That’s much better, yes. I overlooked that class.

**Comment by joshgoebel** [comment]
@klmr Any chance you'd want to be R maintainer?  We can add a fancy new `Maintainer: ` tag to the language...

File: src/languages/r.js (hunk 11 above)
Comment (klmr) [reply src/languages/r.js:63]: The previous version actually set `relevance: 0` so I removed it because it was no longer necessary, after including it into `IDENT_RE` to make some other stuff work (see 89a346e3c83791a87655a10527241b3396c37883).

But I’ve now put it back in since it’s once again necessary after moving the keywords to the top level.

File: src/languages/r.js (hunk 9 above)
Comment (klmr) [reply src/languages/r.js:15]: I actually managed to fix it now by restoring the mode for backtick-quoted identifiers and adding custom word boundaries to the beginning of all number modes (`(?<![a-zA-Z0-9._])`).

**Comment by klmr** [comment]
Thanks for the comments, they resulted in several improvements. I think I’ve addressed them all now.

> Any chance you'd want to be R maintainer?

Gladly. Should I change swap mine and Joe’s name at the beginning of the file?

**Comment by joshgoebel** [comment]
> Gladly. Should I change swap mine and Joe’s name at the beginning of the file?

I was suggesting a Maintainer tag as it implies more (that you're the goto person for R stuff) :-) .... if you JUST want to be author that's fine too, or take both...

**Comment by klmr** [comment]
> Wouldn't adding a contains to deal with the escapes make a lot more sense?

My thinking was that a single regex is probably faster than having multiple modes. But that’s probably not a bottleneck.

> seems the complexity would be lower if we let an ident be an ident and then treat "escaped ident" as a separate thing... no?

Yes, you’re right. My thinking was that merging them would cut down on duplication but in the end the full `IDENT_RE` including quoted identifiers is only used once. I actually found another bug in the expression through the refactor.

> I was suggesting a Maintainer tag

Oh, I misunderstood. Yeah, I’m happy with that too.

**Comment by joshgoebel** [comment]
> My thinking was that a single regex is probably faster than having multiple modes. But that’s probably not a bottleneck.

Clarity over speed is always my motto. :-)  And yes it's probably not a bottleneck... the core engine is merging all of this into one regular expression anyways (when it can)...

File: test/markup/r/names.expect.txt [hunk 12]
```diff
@@ -0,0 +1,34 @@
...
+
+<span class="hljs-comment"># Primitive built-ins and others</span>
+
+<span class="hljs-built_in">return</span>, <span class="hljs-built_in">switch</span>, <span class="hljs-built_in">sum</span>, stop, try
```
joshgoebel, line 19: Are stop and try meant to have no highlighting?

klmr: Yes. They’re defined in the ‘base’ R package but they’re not “special” (and that package has >1000 names, I guess we don’t want to threat them all as built-ins, right?). “special” names that get highlighting as built-ins are defined differently in R.

But I can see how this might be confusing, putting them into a separate section.

joshgoebel: Yes, makes one ask "What the frack are others?" lol.

joshgoebel (approved):

joshgoebel:
Looking good, please add a CHANGES.md entry!

joshgoebel:
@klmr Would like to make an attempt at summarizing this in 5-6 (or more) bullet points for the commit (squashed)?  I tired to whittle down what you had myself but it was just way too much. :-)

joshgoebel:
Or is the bullet list from this thread start sufficient?

klmr:
@joshgoebel Yup, I think the initial bullet list still describes the PR.

joshgoebel:
@klmr Thanks so much for your work on this!

joshgoebel:
@klmr Could you circle back soon and make a new PR that resolves the lookbehind issue?  The main issues I see are issues with detecting things like:

```
.32
bob.32
```

We may have to find another way or simply reduce the fidelity... You may be able to use a `on:begin` rule to do the look-behind for you...

klmr:
I’m confused — which lookbehind issue? Did you mean to post this on #2715, referring [the lookbehind discussion regarding contextual built-ins](https://github.com/highlightjs/highlight.js/pull/2715#discussion_r497813944)?

joshgoebel:
You're PR used negative look-behind, which is not yet supported by all major browsers, so we cannot use it yet.

But never-mind, I'm taking a whack at fixing it myself. :-)

joshgoebel:
If you have any thoughts:

https://github.com/highlightjs/highlight.js/pull/2745/commits/ba4e08321f397f218e8444f177fe5178437505a1

It's a little bit of a hack... and now I'm wondering if the rule doesn't need to be moved to the end of the stack since right now it would also match infix and escaped identifiers if their second character was `.` or numeric...

We can do this with on:begin but I worry about performance implications since numeric rules can be easily triggered.

klmr:
Ah damn, I had forgotten that. I *think* your fix should work.

The `on:begin` solution would obviously be cleaner. No idea about performance … the current regex is also not necessarily efficient since it has to match the lookahead redundantly.

Though the commented-out version of `on:begin` in your commit assumes the potential match isn’t at the first character in the text, otherwise the index access underflows, doesn’t it?

joshgoebel:
> Though the commented-out version of on:begin in your commit assumes the potential match isn’t at the first character in the text, otherwise the index access underflows, doesn’t it?

That's what the `> 0` check is for.

>  the current regex is also not necessarily efficient since it has to match the lookahead redundantly.

Huh?  Aren't we essentially doing the same thing a negative look-behind would be doing?

klmr:
I have to admit I don’t really know how lookbehind would be implemented. But I’d imagine that the engine would start matching the part after the lookbehind and, if it starts finding a match, it would go back and perform the lookbehind, which in our case is always only a single character wide.

Whereas with the current (positive lookahead solution), encountering a matching lexeme requires the engine to match the entire positive lookahead of the mode’s `begin`. It will then reset to the start of the match (+1, for the initial negative character class match) and match the whole rule again, for `end`.

So, when actually finding a valid hit, whereas the solution using a negative lookbehind performs on average a single redundant character match, the lookahead solution performs a much longer redundant match. But this is wholly intuition based and doesn’t account for how often each of these situations is encountered.

joshgoebel:
>  But I’d imagine that the engine would start matching the part after the lookbehind and, if it starts finding a match, it would go back and perform the lookbehind

I'm not 100% certain, but pretty sure that's not how it works at all based on a plain reading of:

https://www.regular-expressions.info/lookaround.html

Also thinking about the fact that you can have multiple look-behinds, in lots of different places in the regex... I can imagine sometimes it being a TON of work to process a whole regex body only to THEN find out your look-behind of "a" never was going to match in the first place... makes a lot more sense to check first.

IE, a lot like the way the `begin` rules are structured in my hack... if we had an easy way to move the cursor/index we could do the same thing... but since we have to match an arbitrary # of regex as a single spot (some without lookbehind, etc) I can't think of how we'd do that.

> It will then reset to the start of the match (+1, for the initial negative character class match) and match the whole rule again

That's true, though this is a common pattern we use in many grammars for complex expressions.  The question comes down to is the regex engine itself faster than raw Javascript code... I'm betting yes.

> Is there a way of benchmarking only the R language tests? 

We don't have any type of performance benchmarks currently that would help with these types of decisions.

> And how would this affect automatic language detection performance?

Auto-detect just tries every single language then the one that scores highest wins... so if suddenly R was significantly slower than before then auto-detect takes that hit as well.  We see a bit of the "average" case just be running the test suite, so that could detect HUGE regressions, but it wouldn't catch something more nuanced... like say an input that was 99% numbers suddenly being very slow when it would have been processed quickly before.

joshgoebel:
This is the kind of thing one could likely benchmark on Node.js to see how different a true look-behind vs a fake (JS based callback) was... since the Node.js engine DOES support look-behind (which is how your PR got by tests).  I might look into that since that would be very useful to know.  If the performance was adequate it might make sense to add a lookbehind helper to the MODES library... (since it's going to be at least a year before we could add real look-backs to the library).

joshgoebel:
The rulesets:

```js
var boundary = function(hljs) {
  return {
    contains: [
      { begin: /\b\d+/, className: "number" }
    ]
  }
};

var lookbehind = function(hljs) {
  return {
    contains: [
      { begin: /(?<![a-zA-Z0-9._])\d+/, className: "number" }
    ]
  }
};

var hack = function(hljs) {
  return {
    contains: [
      {
        begin: /([^a-zA-Z0-9._])(?=\d+)/,
        end: /\d+/,
        excludeBegin: true,
        className: "number"
      },
    ]
  }
};

var callback = function(hljs) {
  return {
    contains: [
      {
        begin: /\d+/,
        className: "number",
        "on:begin": (match, response) => {
          if (match.index) {
            // let priorChar = ;
            if (match.input[match.index - 1].match(/[a-zA-Z0-9._]/)) response.ignoreMatch();
          }
        },
      }
    ]
  }
};
```

```
\b boundary x 8,682 ops/sec ±1.25% (84 runs sampled)
negative lookbehind x 8,353 ops/sec ±1.52% (84 runs sampled)
hack with begin/end x 8,047 ops/sec ±0.56% (89 runs sampled)
with callback x 3,244 ops/sec ±0.64% (94 runs sampled)
```

That's a big enough difference that I think we should either find a way to live with `\b` or we stick with the begin/end hack until it proves problematic.

@egor-rogov @allejo  Any thoughts?

joshgoebel:
Although turns out it's a LOT closer if you mix \b with callbacks... the issue with pure callbacks is if you have something like:

```
booger12731237 162312636
```
12731237 fails to match, so the cursor is advanced one... so then "2731237" will match, but it will fail because immediately before it is a 1... then "731237" will match, etc... so the problem is the "backtracking"... the callback has to run 10 times instead of once because currently we have no skipMatch (advancing the cursor past the WHOLE match), only ignoreMatch.

But as soon as you add `\b` to "anchor" the match you reduce that amount of work significantly.  Though I suppose the "worse case" would still be content filled with something like "a5 " repeated over and over.

```
negative lookbehind x 8,275 ops/sec ±0.87% (92 runs sampled)
with callback x 7,478 ops/sec ±1.92% (88 runs sampled)
```

I'm thinking now perhaps we try the callback and see how it goes:

```js
        variants: [
          { begin: /\b0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*[pP][+-]?\d+i?/ },
          { begin: /\b0[xX][0-9a-fA-F]+([pP][+-]?\d+)?[Li]?/ },
          { begin: /(\b\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?[Li]?/ }
       ],
        "on:begin": (match, response) => {
          if (match.index > 0) {
            let priorChar = match.input[match.index - 1];
            if (priorChar.match(/[a-zA-Z0-9._]/)) response.ignoreMatch();
          }
        },
```

pkaminski:
@joshgoebel Looks like 10.3.0 was released with the lookbehinds in place, so it's breaking on load in Safari.  Any chance of a quick patch release to fix this?  Thanks.

joshgoebel:
Damn, cause that's in the other branch... yeah... 10.3.1 here we come.... sigh. :-)

src/languages/r.js
```diff
@@ -2,75 +2,154 @@
...
+
       {
-        // number
         className: 'number',
-        begin: "\\d+(?:\\.\\d*)?(?:[eE][+\\-]?\\d*)?i?\\b",
+        variants: [
+          // Special case: only hexadecimal binary powers can contain fractions.
+          { begin: /(?<![a-zA-Z0-9._])0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*[pP][+-]?\d+i?/ },
+          { begin: /(?<![a-zA-Z0-9._])0[xX][0-9a-fA-F]+([pP][+-]?\d+)?[Li]?/ },
+          { begin: /(?<![a-zA-Z0-9._])(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?[Li]?/ }
```
joshgoebel, line 136: Sorry I'm feeling stupid, remind me what edge case we're guarding against here that `\b` doesn't cover?  Is it just `_` or?

klmr: Essentially it’s just `.`, since `_` is part of `\w` and thus makes a word boundary, isn’t it? The purpose is to prevent matching in identifiers, which allow all of the characters in the negative lookbehind.

PR #2683, opened 2020-09-11
fix(csharp) prevent public/static/private incorrect flagged as title

### LINKED ISSUE #2681: (C#) Keyword before tuple return type is highlighted as a method name
**Describe the issue**

image

See https://docs.microsoft.com/en-us/dotnet/csharp/discards#tuple-and-object-deconstruction for example.

C#

**Sample Code to Reproduce**

```csharp
private static (int, int) GetOriginCoordinates()
{
    return (0, 0);
}
```

**Expected behavior**

static should be highlighted as a keyword (i.e. `hljs-keyword`), not as a function name (i.e. `hljs-title`)

To avoid merge conflicts, this should wait until https://github.com/highlightjs/highlight.js/pull/2679 is merged.

joshgoebel:
Closes #2681.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ Language Improvements:
 
 - Add all C# 9 keywords, and other missing keywords (#2679) [David Pine][]
 - Add `objective-c++` and `obj-c++` aliases for Objective-C [Josh Goebel][]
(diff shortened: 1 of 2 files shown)


egor-rogov (approved):
Looks good to me

PR #2685, opened 2020-09-11
enh(java) Add support for `record`

### LINKED ISSUE #2604: [Request] Adding `record` as keyword to Java

joshgoebel:
Resolves #2604.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,9 @@
 
 Language Improvements:
 
-- Add all C# 9 keywords, and other missing keywords (#2679) [David Pine][]
-- Add `objective-c++` and `obj-c++` aliases for Objective-C [Josh Goebel][]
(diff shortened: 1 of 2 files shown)


egor-rogov (approved):
Looks fine

PR #2689, opened 2020-09-11
Fixed FORTRAN 77 style comments

### LINKED ISSUE #2677: (Fortran) FORTRAN 77 style comments
**Describe the issue**
Two comment styles are supported, old `C` (FORTRAN 77) and new `!`. Syntax starting with `C`/`c` in column 1 are treated as comments. The `^C` comment indicator should be better replaced with `^C ` (appended whitespace) as a workaround.

Fortran.

`highlight`

**Sample Code to Reproduce**
```fortran
module test
    implicit none
    public :: foo
contains ! <-- is treated as comment because of `c` in first column
    function foo()
        integer :: foo
        foo = 1
    end function foo
end module test
```

**Expected behavior**
Syntax starting with character `c` should not be treated as comments.

**Additional context**
The comments in `fortran.js` are defined as:

```js
  const COMMENT = {
    variants: [
      hljs.COMMENT('!', '$', {relevance: 0}),
      // allow Fortran 77 style comments
      hljs.COMMENT('^C', '$', {relevance: 0})
    ]
  };
```
Perhaps, it is sufficient to replace `^C` with `^C ` or `^C\s`.

interkosmos:
Based on the recommendation in #2677, FORTRAN 77 style comments have been changed to not catch Fortran 90 keywords (except for variables simply named `c`).

Closes #2677.


joshgoebel:
Please add an additional example or two in the markup/fortran/comments test cases.

And we'll need an entry in CHANGES.md.

joshgoebel:
And of course you need to pass existing tests, like "C" on an empty line is probably a comment, etc...

joshgoebel:
`npm run test-markup` to just run the markup tests.

interkosmos:
I dropped the FORTRAN 77-style comments entirely, as I don’t see any way to achieve a reasonable result. I guess, it should be more important to highlight syntax correctly than to have better coverage at the cost of quirks and bugs.

joshgoebel:
> I dropped the FORTRAN 77-style comments entirely, as I don’t see any way to achieve a reasonable result. 

What was the problem you ran into with just adding \s?

We already have a rule for the `C = value` variant (for value Fortran 90 code)... you'd need a `^C$` exception (for lines with nothing by C, which seems like it could only be a comment, serving no purpose in Fortran 90)... and then `C\s` is a comment... am I missing something?

interkosmos:
I searched for old FORTRAN 77 code, and the variety of comment styles will make this feature useless (if not strict ANSI FORTRAN 77). The following comments would not be catched, for example:

``` 
     C non-standard comment (Watcom)
CC another comment
C========
C**********
```

joshgoebel:
```
     C non-standard comment (Watcom)
CC another comment
C========
C**********
```

All of those look like they could be handled pretty deftly with a regex... and again we don't have to handle EVERY case for it to be useful... the actual issue was to prevent breaking your code snippet with Fortran 90 code starting at position 0.

So I'd say we could easily detect `C**...`, `C==...`, `CC...`... and then either detect the Watcom comments or not... I'm more interested in hearing about false detection of code as comments than in Fortran 77 comments we fail to detect (esp if they are obscure or non-standard) :-)

interkosmos:
So, tried it again, and hope stupid me got it right this time. Excuse the inconveniences.

joshgoebel:
Much nicer, but doesn't need the Axapta stuff that somehow got mixed in... can you rebase/clean up to have just the Fortran piece?  If not let me know and I an probably clean it up.

Awesome work!

joshgoebel (approved):

PR #2692, opened 2020-09-13
enh(twig): keywords, filters, functions

### LINKED ISSUE #2422: (twig) common keywords not highlighted (and, or, same as, etc)
"and", "or", "is empty", "same as","in" etc. not highlighting.

florianjr:
- Update Twig keyword, filter, functions list and fixed highlighting
  inside tag
- Update AUTHORS.TXT
- Update CHANGES.md

Resolves #2422

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -289,3 +289,4 @@ Contributors:
 - Jim Mason <jmason@ibinx.com>
 - lioshi <lioshi@lioshi.com>
 - David Pine <david.pine.7@gmail.com>
+- Florian Förster <foerster.florianjr@gmail.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,10 +5,12 @@ Language Improvements:
 - Add all C# 9 keywords, and other missing keywords (#2679) [David Pine][]
 - Add `objective-c++` and `obj-c++` aliases for Objective-C [Josh Goebel][]
 - Update keyword list for Axapta (X++) (#2686) [Ryan Jonasson][]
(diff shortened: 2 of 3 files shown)


florianjr:
The automatic check fails because real django code is detected as twig. Shall I update the default.txt files?

joshgoebel:
Well, first which keyword is it that's causing the conflict with Django?  `u` looks a little funny... perhaps it should be scored 0?

Also, please don't auto-lint... (esp. using entirely different settings than our own eslint).  Please return the file to it's original structure and lets stick with JUST the keyword changes... that includes not changing the tag/template stuff as well... if changes are needed there that should be done separately (or at the very least as a separate commit - so it can be properly reviewed and discussed).

Thanks.

florianjr:
I unterstand the Part with my linter. I‘ll do this. But the changes in the templating part were needed, because otherwise only the first keyword in a Tag would be highlighted. I’ll put these changes in selber are commits. I won’t do a rebase, I’ll open a new and clean PR Thanks

joshgoebel:
> But the changes in the templating part were needed, because otherwise only the first keyword in a Tag would be highlighted. 

Yeah, just that had nothing to do with the original issue you raised and just makes reviewing the PR harder... one PR would also be ok if the work was split into two commits making it easier to review them independently... 

Although it's usually best to keep things like this separate since often they lead to discussions and then just slow down what should otherwise be a simple PR (adding a few keywords).  If the original PR had JUST included the keyword changes and nothing else it would have likely been merged in minutes.

With the tag thing I'm worried the intention was *purposely* to highlight only the first as a name... but I'd have to dig into that further... so it'd be much easier to get the keyword improvements moved and then review the other matter separately.

PR #2699, opened 2020-09-21
enh(twig): Update keyword list

florianjr:
- Update keyword, filter and tag list
- Update metafiles with my personal information

Resolves: #2422

### DIFF
--- a/src/languages/twig.js
+++ b/src/languages/twig.js
@@ -2,6 +2,7 @@
 Language: Twig
 Requires: xml.js
 Author: Luke Holder <lukemh@gmail.com>
+Contributors: Florian Förster <foerster.florianjr@gmail.com>
 Description: Twig is a templating language for PHP
 Website: https://twig.symfony.com
 Category: template
@@ -13,7 +14,7 @@ export default function(hljs) {
     begin: '\\(', end: '\\)'
   };
 
-  var FUNCTION_NAMES = 'attribute block constant cycle date dump include ' +
+  var FUNCTION_NAMES = 'attribute block constant country_timezones cycle date dump html_classes include ' +
                   'max min parent random range source template_from_string';
 
   var FUNCTIONS = {
@@ -28,17 +29,17 @@ export default function(hljs) {
   var FILTER = {
     begin: /\|[A-Za-z_]+:?/,
     keywords:
-      'abs batch capitalize column convert_encoding date date_modify default ' +
-      'escape filter first format inky_to_html inline_css join json_encode keys last ' +
-      'length lower map markdown merge nl2br number_format raw reduce replace ' +
-      'reverse round slice sort spaceless split striptags title trim upper url_encode',
+      'abs batch capitalize column convert_encoding country_name currency_name currency_symbol data_uri date date_modify default ' +
+      'escape filter first format format_currency format_date format_datetime format_number format_time html_to_markdown inky_to_html inline_css join json_encode keys language_name last ' +
+      'length locale_name lower map markdown_to_html merge nl2br number_format raw reduce replace ' +
+      'reverse round slice sort spaceless split striptags timezone_name title trim u upper url_encode',
     contains: [
       FUNCTIONS
     ]
   };
 
-  var TAGS = 'apply autoescape block deprecated do embed extends filter flush for from ' +
-    'if import include macro sandbox set use verbatim with';
+  var TAGS = 'apply autoescape block deprecated do embed extends flush for from ' +
+    'if import in include is macro sandbox set use verbatim with';
 
   TAGS = TAGS + ' ' + TAGS.split(' ').map(function(t){return 'end' + t}).join(' ');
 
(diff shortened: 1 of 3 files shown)


florianjr:
@joshgoebel here's is now my PR for the keywords. I'll open a dedicated one for the fixes soon :)

src/languages/twig.js
```diff
@@ -28,17 +29,17 @@ export default function(hljs) {
   var FILTER = {
     begin: /\|[A-Za-z_]+:?/,
     keywords:
-      'abs batch capitalize column convert_encoding date date_modify default ' +
-      'escape filter first format inky_to_html inline_css join json_encode keys last ' +
-      'length lower map markdown merge nl2br number_format raw reduce replace ' +
```
joshgoebel, line 33: Are `markdown` and `filter` removed/deprecated?  Are they still in use in any supported versions?  We try to be conservative in removing keywords if they could potentially still be in use in code snippets, etc.

src/languages/twig.js
```diff
@@ -28,17 +29,17 @@ export default function(hljs) {
...
-      'reverse round slice sort spaceless split striptags title trim upper url_encode',
+      'abs batch capitalize column convert_encoding country_name currency_name currency_symbol data_uri date date_modify default ' +
+      'escape filter first format format_currency format_date format_datetime format_number format_time html_to_markdown inky_to_html inline_css join json_encode keys language_name last ' +
+      'length locale_name lower map markdown_to_html merge nl2br number_format raw reduce replace ' +
+      'reverse round slice sort spaceless split striptags timezone_name title trim u upper url_encode',
```
joshgoebel, line 35: Please add `|0` to score the `u` as 0 relevancy.

joshgoebel (changes requested):

joshgoebel:
Looking good.  Almost there!

src/languages/twig.js
florianjr: Yes, they were removed.

src/languages/twig.js
florianjr: I don't know if this is what you mean :)

src/languages/twig.js
joshgoebel: But people are presumable still using Twig 2.0, yes?  The docs are still available.

https://twig.symfony.com/doc/2.x/filters/index.html

florianjr: Then I add it again, I just looked at the 3.0. Thanks :)

src/languages/twig.js
joshgoebel: Ype!

src/languages/twig.js
joshgoebel: Wait, I thought `filter` was a tag?

florianjr: It's both :)

florianjr: How should we classify it?

florianjr: Yes, it is. So it's alright.

src/languages/twig.js
```diff
@@ -28,17 +29,17 @@ export default function(hljs) {
...
+      'reverse round slice sort spaceless split striptags timezone_name title trim u|0 upper url_encode',
     contains: [
       FUNCTIONS
     ]
   };
 
-  var TAGS = 'apply autoescape block deprecated do embed extends filter flush for from ' +
-    'if import include macro sandbox set use verbatim with';
+  var TAGS = 'apply autoescape block deprecated do embed extends flush for from ' +
+    'if import in include is macro sandbox set use verbatim with';
```
joshgoebel, line 42: Sorry I was just looking at these more closely... aren't these keywords, not tags?  I don't see any `inend` or `isend` in the docs?

florianjr: I just thought about this. To be precise they're operators, but I think we could handle them as functions.

joshgoebel:
Sorry I'm doing a terrible job reviewing this... let me try again:

Perhaps we need a new OPERATORS array, stuff them in there (for documentation) then decide which rule they go into AFTER that...

> "and", "or", "is empty", "same as","in" etc. not highlighting.

This was from the original issue which say this resolves... but I don't see mention of `and`, or `or` anywhere... does this truly resolve the original issue?  Are those keywords just dead in 3.0 now?

florianjr:
No sorry, I'm just unexperienced. [Here](https://twig.symfony.com/doc/3.x/), under Operators you can see the current ones listed. We should make a new array for this, that's true.

joshgoebel:
> We should make a new array for this, that's true.

Yes, probably should be it's own array and then join it to the TAGS inside the `template-tag` rule on line 59 or so:

```diff
        className: 'template-tag',
        begin: /\{%/, end: /%}/,
        contains: [
          {
            className: 'name',
            begin: /\w+/,
-            keywords: TAGS,
+            keywords: TAGS + " " + OPERATORS.join(" "),
            starts: {
```

Or something.  Not sure if this applies also to variables or not?

joshgoebel:
Although to handle the grouped keywords ("is empty", "same as") you'll likely need your own mode:

```
GROUPED = {
  begin: (\b(is empty|same as)\b)
}
```

To use in addition to the normal more boring keywords...

joshgoebel:
Ping.

joshgoebel:
Closing for now until you have time to circle back and work on this.

PR #2701, opened 2020-09-22
enh(javascript) allow leading `#` for private class fields (stop treating as illegal)

### LINKED ISSUE #2700: (javascript) Fails to parse when using private class fields in JS
**Describe the issue**

Use of the (Stage 3) JavaScript feature private class fields results in the entire code block being skipped.

<details><summary>example output (high-res PNG)</summary>

private fields

</details>

`javascript`

`highlight`

**Sample Code to Reproduce**

```js
class Demo {
  #neat;
}
```

**Expected behavior**

See the GitHub-highlighted example above!

**Additional context**

chriskrycho:
As of an upcoming release of ECMAScript (almost certainly 2021), private class fields may be defined with this syntax:

```js
class Demo {
  #neato;
}
```

Accordingly, treating *all* non-shebang `#` as illegal markers for JavaScript is no longer viable. Instead, support `$`, `_`, and `[A-z]` as well as `!` as legal identifiers here.

Note that this will still fail in cases where Unicode characters are used as the first character of a private class field name.

Fixes #2700

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -63,7 +63,7 @@ export default function(hljs) {
     literal: ECMAScript.LITERALS.join(" "),
     built_in: ECMAScript.BUILT_INS.join(" ")
   };
-  const nonDecimalLiterals = (prefixLetters, validChars) => 
+  const nonDecimalLiterals = (prefixLetters, validChars) =>
     `\\b0[${prefixLetters}][${validChars}]([${validChars}_]*[${validChars}])?n?`;
   const noLeadingZeroDecimalDigits = /[1-9]([0-9_]*\d)?/;
   const decimalDigits = /\d([0-9_]*\d)?/;
@@ -204,7 +204,7 @@ export default function(hljs) {
     keywords: KEYWORDS,
     // this will be extended by TypeScript
     exports: { PARAMS_CONTAINS },
-    illegal: /#(?!!)/,
+    illegal: /#(?![$_A-z])/,
     contains: [
       hljs.SHEBANG({
         label: "shebang",
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 Language Improvements:
 
+- enh(javascript) allow `#` for private class fields (#2701) [Chris Krycho][]
 - fix(bash) enh(bash) allow nested params (#2731) [Josh Goebel][]
 - fix(python) Fix highlighting of keywords and strings (#2713, #2715) [Konrad Rudolph][]
 - fix(fsharp) Prevent `(*)` from being detected as a multi-line comment [Josh Goebel][]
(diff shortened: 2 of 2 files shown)


joshgoebel:
Right idea, but we can't just remove this.  It needs to be updated to be aware of what private fields look like. So I think that means it would be a # followed by anything other than a valid symbol `[a-z_]` would still be illegal?  So, same idea but update the negative look-ahead... the `!` check isn't needed anymore since we now have a feature to handle shebang lines individually.

The illegal checks are part of the reason our auto-detect is so fast and accurate... this type of check immediately prevents JS from flagging Ruby, Python, or MANY other languages that use `#` for commenting... instead when it sees them it can immediately stop processing and be sure they are not Javascript.

chriskrycho:
Thank you! That makes sense, and I figured we might need to iterate on this extremely naïve first pass. I'll check the actual spec text and make sure the negative lookahead matches that.

chriskrycho:
@joshgoebel worth note that if the rule you're using for highlighting is just expecting `[_A-z]`, it's going to miss a lot. The actual spec allows a full range of Unicode characters and also `$`. So `#$α` is totally a legitimate name for a private class field. And in a project I'm working on, we actually *do* use Greek letters in certain spots. 😬 Not sure how to tackle that with a regex that handles it via negative lookahead.

src/languages/javascript.js
```diff
@@ -279,6 +279,6 @@ export default function(hljs) {
 
       }
     ],
-    illegal: /#(?!!)/
+    illegal: /#(?![!$_A-z])/
```
joshgoebel, line 282: We should be able to remove the ! now since we deal with shebang separately now.  `!` isn't allowed in normal identified afaik... can it be used with private fields somehow?

joshgoebel: > it's going to miss a lot... and also $...

Well I meant "in general". :)  Adding $ as you did is great.

> The actual spec allows a full range of Unicode characters ... So #$α is totally a legitimate name for a private class field. And in a project I'm working on, we actually do use Greek letters in certain spots. 😬 Not sure how to tackle that with a regex that handles it via negative lookahead.

Well you'd just add it to the range, no?  But we don't support the new `/u` modifier stuff yet... so you're stick with hard coding like Java does:

```
var JAVA_IDENT_RE = '[\u00C0-\u02B8a-zA-Z_$][\u00C0-\u02B8a-zA-Z_$0-9]*';
```

I'm find with tackling that matter separately though.

joshgoebel: Could we just remove the `!` here and then add an entry to CHANGES.md and then I think this could be merged.  Thanks.

chriskrycho: Yep, I’ll knock that out this evening or (more likely) tomorrow. Thanks!

joshgoebel: This'll need to be rebased now/rejiggered (since I just merged the TS/JS merge)... but hopefully won't be hard since the change to JS is only a one liner. :-)

joshgoebel:
@chriskrycho I'm carrying this one to the home plate so I can get it merged!  Thanks for the PR!

joshgoebel (approved):

chriskrycho:
Ah, thank you, and so sorry for the delay. 2020 has been a *year*.

PR #2705, opened 2020-09-22
(package) Process 3rd party CSS during the build process

Greg-Smulko:
Process 3rd party CSS in a similar way as the JS files.

Tested by following instructions from https://github.com/highlightjs/highlight.js/blob/master/extra/3RD_PARTY_QUICK_START.md on https://github.com/highlightjs/highlightjs-tsql . 

The generated `ssms.min.css` was placed in both `.\extra\tsql\dist` and `.\build\styles`, and the resulted file can be inspected in https://github.com/highlightjs/highlightjs-tsql/pull/6 .

Disclaimer: I'm not happy about the code :( , but I'm afraid I wouldn't be able to do it properly for now. :( I'd really like to bring https://github.com/highlightjs/highlightjs-tsql to a usable state first...

### DIFF
--- a/tools/build_cdn.js
+++ b/tools/build_cdn.js
@@ -98,8 +98,27 @@ async function buildDistributable(language) {
   let distDir = path.join(language.moduleDir,"dist")
   log(`Building ${distDir}/${filename}.`)
   await fs.mkdir(distDir, {recursive: true});
-  fs.writeFile(path.join(language.moduleDir,"dist",filename), language.minified);
+  fs.writeFile(path.join(distDir,filename), language.minified);
 
+
+  let styleSourceDir = path.join(language.moduleDir, "src/styles")
+  glob.sync("*", { cwd: styleSourceDir }).forEach((file) => {
+    let sourceFile = path.join(styleSourceDir, file)
+    let destFileName = file;
+    let destFilePath;
+    log(`Processing ${sourceFile}.`)
+    if (file.endsWith(".css")) {
+      destFileName = file.replace(".css", ".min.css")
+      destFilePath = `../${distDir}/${destFileName}`
+      install_cleancss(sourceFile, destFilePath);
+    }
+    else { // images, backgrounds, etc
+      destFilePath = `../${distDir}/${destFileName}`
+      install(sourceFile, destFilePath);
+    }
+
+    install(`${process.env.BUILD_DIR}/${destFilePath}`, `styles/${destFileName}`);
+  })
 }
 
  async function buildCDNLanguage (language) {


joshgoebel:
I'm not sure WHY this is necessary though or a HLJS concern?  The whole build process for languages is _because languages are complex modules that HLJS expects to be built, distributed, and packaged in a VERY particular way_.  So we standardize the way so:

1. It just works.
2. It's clear how it's expected to be done. (if there is an "official" way)

CSS is just a static file... you can distribute it/process it/minify it HOWEVER you'd like... it (and any associated assets) really have nothing to do with Highlight.js at all - just copy them into your dist folder and be done.  If you like Less, use that.  If you like Stylus, use that... Most languages don't even distribute any CSS or assets.

Greg-Smulko:
I see your point.

My counterarguments are that:
1. hljs provides the whole build/test infrastructure out-of-the-box as described [here](https://github.com/highlightjs/highlight.js/blob/master/extra/3RD_PARTY_QUICK_START.md#packaging), so adding separate tasks to each 3rd party plugin that has CSS only to minify it seems to be a bit counterproductive (especially that hljs supports it anyway),
2. my understanding was that it's very deliberate that 3rd party plugins can be fully integrated with hljs once put into the `extra` folder - especially that they're processed and put into the [main `build` folder](https://github.com/highlightjs/highlight.js/blob/25dd12a5324f3554a2ec30116789861ec318ed5f/tools/build_cdn.js#L67) in addition to [their own `dist` folder](https://github.com/highlightjs/highlight.js/blob/25dd12a5324f3554a2ec30116789861ec318ed5f/tools/build_cdn.js#L76). I don't know the reasoning behind this decision, but if it was done on purpose, I don't understand why 3rd party JS files are supposed to be part of the main build output, but CSS files not.

joshgoebel:
> my understanding was that it's very deliberate that 3rd party plugins can be fully integrated ... I don't know the reasoning behind this decision

To make it easy for integrators to build "monoliths" that "just work".  To perhaps encourage the community to create larger HLJS bundles than just the grammars we distribute and support within "core".

> I don't understand why 3rd party JS files are supposed to be part of the main build output, but CSS files not.

Was never even considered until this very moment. :-)  Personally I'm not sure that encouraging themes to have their own "works best with" or "only works with" style is great for the community, but also not sure that's a good reason not to make this easier. :-)

---

So we can do this but if we're going to do this we need to do it right.  That means all 3 of the build scripts (cdn, browser, nodejs) would need to be updated and each of the build routines:

- `installDemoStyles`
- `installStyles`
- `buildNode` (should probably be pulled out into it's own `installStyles` :-)

The easiest way to do this is probably to factor out a common "allStyleFiles" helper that returns a FULL list (core + 3rd party) CSS and then use that as the iteration source for each of the 3 individual builds.

Want to take another pass at a complete fix here?

joshgoebel:
Closing for now.  Very happy to reopen when a new PR is available for review.

PR #2715, opened 2020-09-29
enh(Python) Some Python language improvements

### LINKED ISSUE #2713: (Python) Wrong highlighting of default arguments
**Describe the issue**
`None` should be the same colour in the function signature and function body (since it's the same object being referred to), but instead it's plain in in the signature and colored in the body.

Screenshots from [this Stack Overflow post][3] (dark and light mode):

[![dark][1]][1] [![light][2]][2]

`python`/`py`

I don't know what that means. I assume it's about how highlight.js is deployed, but I'm an end user, not a Stack Exchange dev.

**Sample Code to Reproduce**
```py
def my_func(working_list=None):
    if working_list is None: 
        working_list = []

    working_list.append("a")
    print(working_list)
```

**Expected behavior**
`None` should be the same colour in the function signature and function body. GitHub gets it right (see snippet above). And FWIW, here are screenshots from Atom, PythonAnywhere, and gedit, which are all using different highlighting schemes to GitHub:

[![atom][4]][4] [![PA][5]][5] [![gedit][6]][6]

**Additional context**
I originally posted this bug report [on Stack Overflow][7] and was directed here by [Martijn Pieters][8].

  [1]: https://i.stack.imgur.com/Yg9v5.png
  [2]: https://i.stack.imgur.com/pItJs.png
  [3]: https://stackoverflow.com/a/366430/4518341
  [4]: https://i.stack.imgur.com/antMh.png
  [5]: https://i.stack.imgur.com/8gMjO.png
  [6]: https://i.stack.imgur.com/qg0MF.png
  [7]: https://meta.stackoverflow.com/q/401572/4518341
  [8]: https://meta.stackoverflow.com/users/100297/martijn-pieters

klmr:
* Fix #2713
* Fix keywords and builtins
* Fix string prefixes

(I’m happy to pull these out into separate PRs if that’s more appropriate!)

# Details:

## Fix keywords and builtins

* Keywords now correspond exactly to the list in section 2.3.1 of the [Python 3.8 reference][1]. Reserved identifiers are not considered.
* Built-in literal constants now correspond exactly to those listed in the reference [2].
* Built-in functions correspond to those listed in the reference [3].
* <del>Built-ins aren’t highlighted directly after a dot; this corresponds to how the IPython REPL highlights e.g. `x.len()` and (incorrectly) `x. len()`.</del>

[1]: https://docs.python.org/3/reference/lexical_analysis.html#keywords
[2]: https://docs.python.org/3/library/constants.html
[3]: https://docs.python.org/3/library/functions.html

## Fix string prefixes

* Remove invalid 'ur' prefix combination
* Add 'br' prefix combination
* Make string prefixes case insensitive


src/languages/python.js
```diff
@@ -7,13 +7,19 @@ Category: common
 
 export default function(hljs) {
   var KEYWORDS = {
+    $pattern: /(?<!\.)\w+/,
```
joshgoebel, line 10: Can't use negative look-behinds yet, browser support isn't good enough.

joshgoebel: > Built-ins aren’t highlighted directly after a dot; t

You might have to look into `beginKeywords` if you insist on solving this - as it includes that support manually. (to not match keywords after `.`)  Or you could use the callbacks, but this currently isn't possible with the build-in `keyword` support.

src/languages/python.js
```diff
@@ -39,43 +39,43 @@ export default function(hljs) {
...
         relevance: 10
       },
       {
-        begin: /(u|r|ur)"/, end: /"/,
+        begin: /([uU]|[rR])"/, end: /"/,
         relevance: 10
       },
       {
-        begin: /(b|br)'/, end: /'/
+        begin: /([bB][rR]?|[rR][bB])'/, end: /'/
```
joshgoebel, line 67: Isn't this inconsistent with how you write out the combos below?  Why in once case do we use an optional and in the other case we just list all 3 combos explicitly?  I'm not seeing the distinction.

src/languages/python.js
```diff
@@ -12,12 +12,11 @@ export default function(hljs) {
...
-      'divmod id object sorted bin enumerate input oct staticmethod bool eval int open str ' +
-      'breakpoint exec isinstance ord sum bytearray filter issubclass pow super bytes float ' +
-      'iter print tuple callable format len property type chr frozenset list range vars ' +
-      'classmethod getattr locals repr zip compile globals map reversed __import__ complex ' +
-      'hasattr max round',
+      '__import__ abs all any ascii bin bool breakpoint bytearray bytes callable chr classmethod ' +
+      'compile complex delattr dict dir divmod enumerate eval exec filter float format frozenset ' +
+      'getattr globals hasattr hash help hex id input int isinstance issubclass iter len list ' +
+      'locals map max memoryview min next object oct open ord pow print property range repr ' +
+      'reversed round set setattr slice sorted staticmethod str sum super tuple type vars zip',
```
joshgoebel, line 19: While we're here can we please just expand these to one-per-line (see) in an array (see `ecmascript.js` for an example) and then join them for keywords for future maintainability...

joshgoebel (changes requested):

joshgoebel:
@klmr Thanks for the effort here!

joshgoebel:
Also, please add a few entries to `CHANGES.md`.

joshgoebel:
@klmr And if you have more time to improve Python I might suggest looking at: https://github.com/highlightjs/highlight.js/issues/2633 :-)

src/languages/python.js
klmr: I don’t think it’s very important, let’s skip it for now.

src/languages/python.js
klmr: Yeah, that’s an oversight.

src/languages/python.js
klmr: How’s this?

klmr:
I’m happy to take a look at #2633 but I think that should be a separate PR.

joshgoebel:
Of course.

src/languages/python.js
```diff
@@ -6,76 +6,194 @@ Category: common
...
+    '__debug__',
+    'Ellipsis',
+    'False',
+    'None',
+    'NotImplemented',
+    'True',
+  ].join(' ');
+
+  const KEYWORDS = {
+    keyword: RESERVED_WORDS,
```
joshgoebel, line 127: Please move the joins here (convention).  The constants above should be pure arrays.

src/languages/python.js
```diff
@@ -6,76 +6,194 @@ Category: common
...
+    built_in: BUILT_INS,
+    literal: LITERALS
   };
+
   var PROMPT = {
     className: 'meta',  begin: /^(>>>|\.\.\.) /
   };
+
   var SUBST = {
```
joshgoebel, line 136: Lets change all those to `const` now as well.

src/languages/python.js
joshgoebel: Awesome.  Just those two tiny changes I mentioned and then I think this is good to ship!

joshgoebel (approved):

joshgoebel (approved):

joshgoebel:
@klmr Thanks fo much for all the effort on this!

joshgoebel:
Well, grrr... I didn't mean to truncate all those great notes from the commit, grrrr...

klmr:
Ha, no worries. If somebody’s interested they can refer back to the linked PR for the notes.

PR #2719, opened 2020-09-30
enh(build) slimmer Docker image and add .dockerignore

kaczmarj:
This pull request proposes the following changes:

1. Use `node:12-slim` as the base image
2. Add some documentation to the Dockerfile
3. Add a `.dockerignore` file, which is similar to a `.gitignore` in that it tells Docker to not copy some files during the image build.

The result of number 1 is that the Docker image is 342 MB, down from 1.1 GB originally.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,8 +5,6 @@ __pycache__
 *.swp
 node_modules
 .project
-yarn.lock
-extra/
 
 # editors
 .idea/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+.DS_Store
+build
+docs/_build
+__pycache__
+*.swp
+node_modules
+.project
+yarn.lock
+extra/
+
+# editors
+.idea/
+.vscode/
+.Rproj.user
+
+# misc
+/work
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,22 @@
-FROM node:12
-# docker build -t highlight-js .
-RUN apt-get update && apt-get install -y nginx
+# Dockerfile for highlight.js
+#
+# Build image:
+#   docker build -t highlight-js .
+#
+# Run container:
+#   docker run --rm -it --publish 8080:80 highlight-js
+#
+#   And open a browser to http://localhost:8080
+
+FROM node:12-slim
+RUN apt-get update -qq \
+    && apt-get install --yes --no-install-recommends \
+        nginx \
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,17 @@
+.DS_Store
+build
+docs/_build
+__pycache__
+*.swp
+node_modules
+.project
+yarn.lock
+extra/
```
joshgoebel, line 9: I think someone might want this included, it's whole point is to be used by the build process. Thoughts?

.dockerignore
```diff
@@ -0,0 +1,17 @@
+.DS_Store
+build
+docs/_build
+__pycache__
+*.swp
+node_modules
+.project
+yarn.lock
```
joshgoebel, line 8: Is this not needed to insure the same set of dependencies are installed inside the container?

joshgoebel (changes requested):
Overall looks great!

.dockerignore
kaczmarj: The `package.json` and `package-lock.json` will ensure that the same set of dependencies is installed in the container. You don't need `node_modules` because that will be created with `npm install`. `yarn.lock` isn't used by `npm install` as far as I know.

.dockerignore
kaczmarj: Should `extra/` be available in the container? If it is a build artifact, I would leave it out and let the build process create it.

joshgoebel: No it's not a build artifact, it's a requirement.

.dockerignore
joshgoebel: It's used by people using yarn... If someone wanted to switch this I wouldn't want them to have to track down this and alter it also... including it does no harm.

joshgoebel:
So lets change those two things and then I think this is mergable.

.dockerignore
joshgoebel: Well technically both... because artifacts are added to the modules in place to make committing the distributables to GitHub super easy.

kaczmarj: So I will remove `yarn.lock` and `extra/` from the `.dockerignore`. Would that make this PR mergeable? (also I am resolving the other related conversation in lieu of this one)

kaczmarj, line 9:
```suggestion
```

joshgoebel (approved):

joshgoebel:
Yep, that's all that jumped out to me.  Great size reduction, thanks!

PR #2720, opened 2020-09-30
chore(css) Fix background color of the Lightfair theme

therealsujitk:
Added background colour `#fff` to the Lightfair theme to make it visible on dark backgrounds. As of now when this theme is used, some text is not visible on certain darker backgrounds.


joshgoebel:
I pinged the original author since their email is right there... I'd guess this is correct unless the intention might be for a slightly different shade of white... we'll see.

joshgoebel:
@therealsujitk Thanks so much!

Issue #2146, opened 2019-10-03
(cpp) syntax should be split into C and C++ syntaxes


mortie:
Currently, C and C++ are both highlighted the same way, but that's not ideal. For example, the words `string` and `vector` are highlighted as built-ins, which makes no sense in a C context.

marcoscaceres:
hhmm... yeah, might be worth splitting into two.

yyyc514:
It's also possible we should start with a "C-like" like Prism does.  My recent work on CPP/Arduino might make sense to look at as a potential pattern.

yyyc514:
> Currently, C and C++ are both highlighted the same way, but that's not ideal. For example, the words string and vector are highlighted as built-ins, which makes no sense in a C context.

But those things wouldn't really be using in a C context either, would they?

mortie:
@yyyc514 In a language where the words "vector" and "string" have no special meaning, they're pretty reasonable variable names.

The context I encountered this in was with tagged unions; something like

``` C
struct whatever {
    enum {
        WHATEVER_NUMBER,
        WHATEVER_STRING
    } tag;
    union {
        double number;
        char *string;
    } val;
};

yyyc514:
Now just having separate keyword lists would be pretty easy.

yyyc514:
@egor-rogov How would you feel if we prepared for this by simply making C a requirement of C++ but for v10 effectively they were still the exact same thing?  Ie a patch would only:

- Create a new `c` language file
- Move the content to `c`
- `cpp` depends on `c`
- Add some comments to explain this.

**This would get the "breaking" part of the change out of the way and avoid us having to release v11 just to do this later... Thoughts?**

yyyc514:
In my perfect world v10 would break a lot of things, then we'd go back to stable releases for a year or so... I don't want to get into the habit of breaking something with every release.  So I think that requires a little planning ahead.

egor-rogov:
I'm okay with it. Having separated keywords for these two languages sounds like a good idea.

ztane:
It doesn't really make sense to have C and C++ share the same code - there are almost more differences than there are similarities, it is not just about keywords. The preprocessor and the basic literal tokens and comments could use the same structure but that's about it. For example C does not have templates, so there is no point in trying to figure out if `x<y>z` is a template or not.

jleffler:
[...]

PR #2728, opened 2020-10-03
enh(cpp) Improve C++ class template headers

### LINKED ISSUE #2102: (cpp) grammar confused by complex Template << >> patterns
This code results in everything after `L':'` to be highlighted as if it was a string: **(Editor: This is now resolved, the second issue still standing).**

```cpp
template<typename T>
class PropertyChangedBase {
public:
    winrt::event_token PropertyChanged(const winrt::Windows::UI::Xaml::Data::PropertyChangedEventHandler &value)
    {
        return m_propertyChanged.add(value);
    }

    void PropertyChanged(const winrt::event_token &token)
    {
        m_propertyChanged.remove(token);
    }

protected:
    template<typename U>
    void compare_assign(U &value, const U &new_value, std::wstring_view name)
    {
        if (value != new_value)
        {
            value = new_value;
            name.remove_prefix(name.find_last_of(L':') + 1);
            m_propertyChanged(*static_cast<T *>(this), winrt::Windows::UI::Xaml::Data::PropertyChangedEventArgs(name));
        }
    }

private:
    winrt::event<winrt::Windows::UI::Xaml::Data::PropertyChangedEventHandler> m_propertyChanged;
};
```

On this one, it just completely stops highlighting after the second `is_streamable`

```cpp
namespace impl {
    template<typename T, typename = void>
    struct is_streamable : std::false_type { };

    template<typename T>
    struct is_streamable<T, std::void_t<decltype(std::declval<std::wostream &>() << std::declval<T>())>> : std::true_type { };

    template<typename T>
    inline constexpr bool is_streamable_v = is_streamable<T>::value;
}

// Disable overload for already valid operands.
template<class T, class = std::enable_if_t<!impl::is_streamable_v<const T &> && std::is_convertible_v<const T &, std::wstring_view>>>
std::wostream &operator <<(std::wostream &stream, const T &thing)
{
    return stream << static_cast<std::wstring_view>(thing);
}
```

Both pieces above is completely correct code which compiles, assuming the right headers are included

### LINKED ISSUE #2716: (cpp) Template parameters trigger class header rule
**Describe the issue**

When using the `template <class T>` syntax in C++, the `class` keyword misleads the parser into thinking it’s found a class header, and goes into class title mode, if there’s some `{` shortly afterwards. The relevant rule is

https://github.com/highlightjs/highlight.js/blob/5b87cc4d106059bee1c30f123e5a4053e0af9de4/src/languages/c-like.js#L227

`cpp`

`highlight`. When using `highlightAuto`, the code snippet is mis-categorised as `angelscript` and parsed differently still.

**Sample Code to Reproduce**

```cpp
template <class T> // comment
auto foo(T x) { … }
```

highlight.js marks this up as

```xml
<keyword>template</keyword> &lt;<class><keyword>class</keyword> <title>T</title>&gt; // <title>comment</title>
<title>auto</title> <title>foo</title>(<title>T</title> <title>x</title>) {</class> … }
```

**Expected behavior**

When replacing `class` with `typename` (which is 100% equivalent in this context), it’s marked up (more) correctly:

```xml
<keyword>template</keyword> &lt;<keyword>typename</keyword> T&gt; <comment>// comment</comment>
<function><keyword>auto</keyword> <title>foo</title><params>(T x)</params> </function>{ … }
```

**Additional context**

Present in master. Some earlier version used highlighted this correctly (presumably before support for class headers was added). Fundamentally the issue is that template parameter declarations aren’t handled at all by the lexer (by contrast, template argument lists *do* have rudimentary handling; i.e. usage rather than declaration).

A fix would probably require a separate mode for template parameter declarations, which is a hairy topic. In fact, correctly disambiguating the meaning of `>>` (can be right-shift operator *or* a terminating nested template argument list) is context-sensitive, and undecidable for the lexer (!) because its meaning can depend on the type of names it is applied to: `a<b<c>>d` can either be parsed as a variable declaration of `d` where the type has a nested template argument (list if `a` is a type), or as an expression involving less-than comparison and bit shifts, if `a` is a variable: https://godbolt.org/z/rM18oc

klmr:
This fixes #2716 and it fixes #2102<sup>1</sup>. In order to parse class and function headers correctly, it is insufficient to just skip greedily (or, worse, non-greedily) over `/<.*>/`, since C++ template declarations as well as template instantiations can contain nested template parameter lists. As a further complication, they can even contain the operators `<`, `>`, `<<` and `>>` (as part of non-type template parameters or their defaults (in declarations)).

This PR attempts to be a minimal change — a much more comprehensive refactor of the C and C++ language definitions is somewhat overdue (see #2173 and #2146). But this will take time, and in the meantime the improvement from this PR is probably “good enough”.

This PR is a WIP: there are some open issues that I’ve commented inside the changes.

---

<sup>1</sup> The failing code from the second issue isn’t added as a test case because hljs still fails to highlight it as a function declaration, but that’s unrelated to templates and rather because hljs currently doesn’t recognise operator overload definitions at all.

### DIFF
--- a/src/languages/c-like.js
+++ b/src/languages/c-like.js
@@ -202,7 +202,7 @@ export default function(hljs) {
   };
 
   var TEMPLATE_USE = {
-    begin: /</, end: />/,
+    begin: /</, end: /[>{;]/,
     keywords: CPP_KEYWORDS,
     contains: EXPRESSION_CONTAINS.concat([
       // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
@@ -222,35 +222,27 @@ export default function(hljs) {
   };
 
   var TEMPLATE_DECLARATION = {
-    beginKeywords: 'template',
+    begin: /template\s*</,
     // Add additional stops to stop runaway template declaration, e.g.
     // `template <bool b = 1 < 2> void f();`
     end: /[>;{]/,
-    contains: [
+    keywords: CPP_KEYWORDS,
+    contains: EXPRESSION_CONTAINS.concat([
+      // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
+      { begin: /<</ },
+      TEMPLATE_USE,
       {
-        begin: /</, end: />/,
-        endsWithParent: true,
-        endsParent: true,
+        begin: /\(/, end: /[);{]/,
         keywords: CPP_KEYWORDS,
-        contains: EXPRESSION_CONTAINS.concat([
+        contains: [
           // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
           { begin: /<</ },
-          TEMPLATE_USE,
-          {
-            begin: /\(/, end: /\)/,
-            keywords: CPP_KEYWORDS,
-            contains: [
-              // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
-              { begin: /<</ },
-              TEMPLATE_USE
-            ]
-          },
-        ]),
-        relevance: 10
+          TEMPLATE_USE
+        ]
       },
-      // FIXME: Set relevance of class declarations at this point to 10!
-      CLASS_DECLARATION,
-    ]
+      'self'
+    ]),
+    relevance: 10
   };
 
   return {
--- a/src/languages/c-like.js
+++ b/src/languages/c-like.js
@@ -105,6 +105,7 @@ export default function(hljs) {
       'atomic_bool atomic_char atomic_schar ' +
       'atomic_uchar atomic_short atomic_ushort atomic_int atomic_uint atomic_long atomic_ulong atomic_llong ' +
       'atomic_ullong new throw return ' +
+      'class struct' +
       'and and_eq bitand bitor compl not not_eq or or_eq xor xor_eq',
     built_in: 'std string wstring cin cout cerr clog stdin stdout stderr stringstream istringstream ostringstream ' +
       'auto_ptr deque list queue stack vector map set pair bitset multiset multimap unordered_set ' +
@@ -200,6 +201,58 @@ export default function(hljs) {
     ]
   };
 
+  var TEMPLATE_USE = {
+    begin: /</, end: />/,
+    keywords: CPP_KEYWORDS,
+    contains: EXPRESSION_CONTAINS.concat([
+      // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
+      { begin: /<</ },
+      'self'
+    ])
+  };
+
+  var CLASS_DECLARATION = {
+    className: 'class',
+    beginKeywords: 'class struct union', end: /[{;:]/,
+    contains: [
+      { beginKeywords: "final" },
+      TEMPLATE_USE,
+      TITLE_MODE
+    ]
+  };
+
+  var TEMPLATE_DECLARATION = {
+    beginKeywords: 'template',
+    // Add additional stops to stop runaway template declaration, e.g.
+    // `template <bool b = 1 < 2> void f();`
+    end: /[>;{]/,
+    contains: [
+      {
+        begin: /</, end: />/,
+        endsWithParent: true,
+        endsParent: true,
+        keywords: CPP_KEYWORDS,
+        contains: EXPRESSION_CONTAINS.concat([
+          // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
+          { begin: /<</ },
+          TEMPLATE_USE,
+          {
+            begin: /\(/, end: /\)/,
+            keywords: CPP_KEYWORDS,
+            contains: [
+              // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
+              { begin: /<</ },
+              TEMPLATE_USE
+            ]
+          },
+        ]),
+        relevance: 10
+      },
+      // FIXME: Set relevance of class declarations at this point to 10!
+      CLASS_DECLARATION,
+    ]
+  };
+
   return {
     aliases: ['c', 'cc', 'h', 'c++', 'h++', 'hpp', 'hh', 'hxx', 'cxx'],
     keywords: CPP_KEYWORDS,
@@ -222,15 +275,8 @@ export default function(hljs) {
         begin: hljs.IDENT_RE + '::',
         keywords: CPP_KEYWORDS
       },
-      {
-        className: 'class',
-        beginKeywords: 'class struct', end: /[{;:]/,
-        contains: [
-          { beginKeywords: "final" },
-          {begin: /</, end: />/, contains: ['self']}, // skip generic stuff
-          hljs.TITLE_MODE
-        ]
-      }
+      TEMPLATE_DECLARATION,
+      CLASS_DECLARATION,
     ]),
     exports: {
       preprocessor: PREPROCESSOR,
--- a/src/languages/c-like.js
+++ b/src/languages/c-like.js
@@ -232,11 +232,15 @@ export default function(hljs) {
       { begin: /<</ },
       TEMPLATE_USE,
       {
-        begin: /\(/, end: /[);{]/,
+        begin: /\(/, end: /\)/,
+        endsWithParent: true,
         keywords: CPP_KEYWORDS,
         contains: [
           // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
           { begin: /<</ },
+          // Match greater-than to prevent this from terminating the mode (via
+          // the parent).
+          { begin: />/ },
           TEMPLATE_USE
         ]
       },
--- a/src/languages/c-like.js
+++ b/src/languages/c-like.js
@@ -207,6 +207,19 @@ export default function(hljs) {
     contains: EXPRESSION_CONTAINS.concat([
       // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
       { begin: /<</ },
+      {
+        begin: /\(/, end: /\)/,
+        endsWithParent: true,
+        keywords: CPP_KEYWORDS,
+        contains: [
+          // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
+          { begin: /<</ },
+          // Match greater-than to prevent this from terminating the mode (via
+          // the parent).
+          { begin: />/ }
+          // TEMPLATE_USE
+        ]
+      },
       'self'
     ])
   };
--- a/test/markup/cpp/templates.txt
+++ b/test/markup/cpp/templates.txt
@@ -102,3 +102,7 @@ class E;
 
 template <int a, int b, bool c = a << b, class T> void g2();
 class F;
+
+template <int A = (B<C>), class T> class C;
+template <int A = B<(c > D<E>)>, class T> class C;
+template <int A = B<(c > D<(e > f)>)>, class T> class C;
--- a/test/markup/cpp/templates.txt
+++ b/test/markup/cpp/templates.txt
@@ -1,0 +1,104 @@
+template <template <typename X> class T> void f();
+
+template class A::B<1, 2> {};
+
+template void A::f(int);
+
+template <class T, std::size_t N = sizeof(T), class K = T>
+ostream& operator<< (ostream& os, const skg::Triplet<T> pt) ;
+void other_stuff_that_isnt_colored();
+
+
+template<class T>
+class My_vector { /* ... */ };
+
+template<class T = void>
+struct My_op_functor { /* ... */ };
+
+template<typename... Ts>
+class My_tuple { /* ... */ };
+
+
+template<class> class My_vector;
+template<class = void> struct My_op_functor;
+template<typename...> class My_tuple;
+
+
+template<typename T> concept C1 = true;
+template<typename... Ts> concept C2 = true; // variadic concept
+template<typename T, typename U> concept C3 = true;
+ 
+template<C1 T> struct s1;         // constraint-expression is C1<T>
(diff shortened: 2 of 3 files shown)


src/languages/c-like.js
```diff
@@ -105,6 +105,7 @@ export default function(hljs) {
       'atomic_bool atomic_char atomic_schar ' +
       'atomic_uchar atomic_short atomic_ushort atomic_int atomic_uint atomic_long atomic_ulong atomic_llong ' +
       'atomic_ullong new throw return ' +
+      'class struct' +
```
klmr, line 108: These keywords were missing generally: there are many situations (not just templates) in which they are valid and weren’t currently highlighted; `class/struct foo` is valid wherever `foo` is used as a type name, such as in declarations or in casts (`foo y = (class foo)(x)`).

src/languages/c-like.js
```diff
@@ -200,6 +201,58 @@ export default function(hljs) {
     ]
   };
 
+  var TEMPLATE_USE = {
+    begin: /</, end: />/,
+    keywords: CPP_KEYWORDS,
+    contains: EXPRESSION_CONTAINS.concat([
+      // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
```
klmr, line 208: I’m repeating this quite a lot. I’m wondering whether this could be simplified, or whether the `/<</` mode should be moved inside `EXPRESSION_CONTAINS`.

src/languages/c-like.js
```diff
@@ -200,6 +201,58 @@ export default function(hljs) {
...
+            contains: [
+              // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
+              { begin: /<</ },
+              TEMPLATE_USE
+            ]
+          },
+        ]),
+        relevance: 10
+      },
+      // FIXME: Set relevance of class declarations at this point to 10!
```
klmr, line 251: Can this be done without copying the definition of `CLASS_DECLARATION`? (It’s probably trivial but despite appearances to the contrary I don’t actually know JavaScript.)

test/markup/cpp/templates.txt
```diff
@@ -0,0 +1,104 @@
...
+s.template foo<T>();
+class C;
+
+
+// Hard mode
+
+template <int a, int b, bool c = (a > b), class T> void f1();
+template <int a, int b, bool c = (a >> b), class T> void f2();
+
+template <int a, int b, bool c = (a < b), class T> void g1();
```
klmr, line 94: This test is failing, and I don’t know why: I thought my rules would handle the nested parenthesised expression — and this works for the corresponding test case above, i.e. for `(a > b)`.

test/markup/cpp/templates.txt
```diff
@@ -0,0 +1,104 @@
...
+class C;
+
+template <int a, int b, bool c = a < b, class T> void g1();
```
klmr, line 97: This test is also failing, and there’s probably no fix, since it’s context sensitive: `a<b, class T>` *could* be a valid template parameter, if `a` were a template template parameter (i.e. if we had `template <typename> typename a` instead of `int a`). Of course in that case we’d be missing an additional terminating `>`, but I can’t see how to make the parser aware of this since it comes (potentially much) later.

klmr (commented):
Request for comment:

test/markup/cpp/templates.txt
joshgoebel: I'd guess the `<` in `a < b` is starting a new TEMPLATE, messing everything up.

src/languages/c-like.js
joshgoebel: It doesn't look that terrible to me...

> whether the /<</ mode should be moved inside EXPRESSION_CONTAINS.

It doesn't technically belong there since it's only relevant in the context of a template, so I think the more explicit way it is now is slightly better than throwing it in there with a comment.

src/languages/c-like.js
joshgoebel: `inherit`

test/markup/cpp/templates.txt
joshgoebel: You could do someone like this with a callback that looks ahead and does code analysis, but that's definitely crossing over into "parsing" territory I think so I don't think we'd want to do that.

src/languages/c-like.js
```diff
@@ -200,6 +201,58 @@ export default function(hljs) {
...
+
+  var TEMPLATE_DECLARATION = {
+    beginKeywords: 'template',
+    // Add additional stops to stop runaway template declaration, e.g.
+    // `template <bool b = 1 < 2> void f();`
+    end: /[>;{]/,
+    contains: [
+      {
+        begin: /</, end: />/,
+        endsWithParent: true,
```
joshgoebel, line 232: Just so you know this `endsWithParent` does not nest... it simply means that within the scope of a `<...>` that it will attempt to match not on the contains rules on line 235, but ALSO the end rule on line 228.  But it doesn't go any deeper than that unless you added `endsWithParent` at every level...

I'm not sure if this has implications for your runaway guard on line 227 or not...

src/languages/c-like.js
joshgoebel: What often happens (and why these were excluded here) is false positives on `bob.class()`... does C++ allow class or struct as method names on classes?

test/markup/cpp/templates.txt
```diff
@@ -0,0 +1,104 @@
...
+
+
+eval<A<int>> eA;
+eval<B<int, float>> eB;
+eval<C<17>> eC;
+eval<D<int, 17>> eD;
+eval<E<int, float>> eE;
+
+
+template <template <auto> class> void FA();  // note: C++17
```
joshgoebel, line 76: This isn't a template match because it has a ` ` after template so the `<` will never match because the rule will end when it hits a space (which doesn't match any of your TEMPLATE rules...

Adding a space eater to your begin matcher for `<` should help.

joshgoebel: Are you using the developer tool?  This kind of thing is a lot easier to see with the visual aids of block nesting.

src/languages/c-like.js
```diff
@@ -200,6 +201,58 @@ export default function(hljs) {
     ]
   };
 
+  var TEMPLATE_USE = {
+    begin: /</, end: />/,
+    keywords: CPP_KEYWORDS,
+    contains: EXPRESSION_CONTAINS.concat([
+      // Match left-shift to prevent it from creating a nested `TEMPLATE_USE`.
+      { begin: /<</ },
+      'self'
```
joshgoebel, line 210: This is going to get tripped up on `<` (less than) inside a template I think because you have no way to know if it's an operator or the start of a new TEMPLATE_USE, which is why your one test is failing I think.

joshgoebel: Also can `<<` Truly never be a nested template use?

test/markup/cpp/templates.txt
klmr: Why would whitespace be an issue here? It works fine for `TEMPLATE_DECLARATION`. That said, nesting templates are actually an issue I had forgotten to comment on: how would I do that? I’ve got a circular dependency and I can’t use `self`, since the mode is nested deeper (i.e. `TEMPLATE_DECLARATION` does *not* contain `TEMPLATE_DECLARATION` — but it contains a mode which should contain `TEMPLATE_DECLARATION`).

In fact, I do use the developer tool, and here’s what I’m currently seeing for this input (I’ve added additional class names for debugging purposes):

Screenshot 2020-10-04 at 14 47 18

Ideally we’d want this to have another `tpl-decl` inside the `tpl-params`.

src/languages/c-like.js
klmr: Yes: each template declaration starts with `template`, and each template instantiation starts with an identifier, followed by a single `<`.

test/markup/cpp/templates.txt
klmr: Ah, of course! I actually don’t know how to handle this: I thought about adding `)` to the `end` of the `TEMPLATE_USE`, mode; this works, but then the mode started by `(` won’t terminate, even when I also set `excludeEnd: true`.

src/languages/c-like.js
klmr: It does not: C++ keywords are reserved in *all* contexts. There are some context-dependent reserved words (e.g. `override` and `final`) but these aren’t part of this list.

src/languages/c-like.js
klmr: Good question! So, if I understand you correctly, I could equivalently remove `endsWithParent, and change the `end` rule to `/[>;{]/` (i.e. the same as the parent `end`)? In this case I should probably do that instead, and add the same end to the other contained rules.

test/markup/cpp/templates.txt
klmr: Adding the following to the `TEMPLATE_USE` mode seems to work:

```javascript
    'on:begin': (matchData, response) => {
      const rest = matchData.input.slice(matchData.index + 1);
      let parens = 0;
      let brackets = 1; // start with `<`
      for (let i = 0; i < rest.length; i++) {
        const c = rest.charAt(i);
        switch (c) {
          case '(': parens++; break;
          case ')': parens--; break;
          case '<': brackets++; break;
          case '>': brackets--; break;
        }
        if (brackets == 0) return;
        if (parens == -1) {
          response.ignoreMatch();
          return;
        }
      }
    },
```

However, this feels like overkill (potentially also inefficient?); it also *only* works in cases where the comparison expression is nested inside parentheses, i.e. it works for this test case but not the next one.

If this approach is taken at all, we’d probably want to add some emergency stop characters (`;`, `{`) to the `switch`.

joshgoebel: Yeah, this is too much... and definitely has performance implications.

src/languages/c-like.js
joshgoebel: Well if the desire is that they end with the parent you'd probably want to just use that... that's what it does - pull the parent "end" down into the children... I was just pointing out that it only goes one layer deep... so if it got "stuck" in a child of a child then your "guard" would not save you... so that needs to be considered and might result in adding some more "false positive" edge case tests.

test/markup/cpp/templates.txt
joshgoebel: Ah I see now, you don't have a class (which you shouldn't) - so I wasn't seeing it because I was only looking at the HTML. :-)  You have an `end` rule so it works... I missed.  If you didn't have and `end` the mode would immediately terminate if it could find no children.

> That said, nesting templates are actually an issue I had forgotten to comment on: how would I do that? I’ve got a circular dependency and I can’t use self, since the mode is nested deeper 

You'd have to do it manually (which will limit depth)... Or rejigger the rules so that say your template rule was say `template<` to `>`...  so that you could then use self...

test/markup/cpp/templates.txt
joshgoebel: I think the best maybe we can do here is liberal use of endsWithParent to make sure we escape if we hit a `{`}, line end, etc... and try to return to normal matching...

test/markup/cpp/templates.txt
klmr: The two nested modes are purely so that I could start with a (highlighted) keyword (`template`), followed by a non-keyword character. I initially had `begin: /template\s*</` but (I thought) that didn’t highlight `template` as a keyword.

But I can’t reproduce this now: it *does* highlight `template` correctly, so I guess I can simplify that nested mode and make it recursive.

src/languages/c-like.js
klmr: I think that works now. I needed to add a nested `{begin: />/}` to prevent a `>` inside a `(…)` from terminating the template declaration mode prematurely. That said, the behaviour on my unit tests is actually the same as before, I need to think of some example which works now and would previously have failed.

klmr: Damn, I think I’ve just kicked the can down the road now. I think this fundamentally doesn’t work because I would need to nest the parent mode inside the mode that handles `(…)`. With my newest change, the the two first lines parse correctly, but the template declaration in the third line ends prematurely, and the “class title” starts at `class T` rather than at `class C`:

```cpp
template <int A = (B<C>), class T> class C;
template <int A = B<(c > D<E>)>, class T> class C;
template <int A = B<(c > D<(e > f)>)>, class T> class C;
```

test/markup/cpp/templates.txt
joshgoebel: > that didn’t highlight template as a keyword. ... But I can’t reproduce this now:

It should indeed IF it's in the `keyword` object at the same level... `beginKeywords` hooks this up automatically... if you're writing the rule by hand you'd need both `begin` and `keyword`.

src/languages/c-like.js
joshgoebel:
```
  var TEMPLATE_USE = {};
  Object.assign(TEMPLATE_USE, {
    // .. can now use TEMPLATE_USE recursively
```

All you need is a reference. :-)  Though I'm not sure we do this anywhere else and not sure it's a pattern we should start using now though. ...

I'm guessing the nesting here could be unlimited but is there some practical limits for MOST C++ code as far as what you'd see in the real world?

joshgoebel:
Closing this in favor of #2752.  Thanks for all the hard work on this though - it's been appreciated!  Hopefully you've learned a bit about the grammars and parser that will help you if you continue to contribute. :-)

If we find more failure cases we can always re-review, but this was getting way to complicated.

klmr:
Yes, I’m definitely in favour of closing this for now and re-submitting another PR later (in particular, *after* splitting the definitions for C and C++), I just need to find some time …

PR #2731, opened 2020-10-03
WIP: enh(bash) allow nested params

### LINKED ISSUE #2730: (Bash) Nested parameter expansion not detected
**Describe the issue**
Nested parameter expansions are not detected. In this case, it causes a problem with the quoting. Screenshots from [this Stack Overflow answer](https://stackoverflow.com/a/3352015/4518341):
Screenshot showing code highlighting Screenshot showing "variable" token
Notice how the last close-brace is white as if it's unquoted, and the following lines are green as if they're part of a string.

bash/lang-sh - I'm not sure if they're equivalent or what. If it helps, the HTML class is `"lang-sh s-code-block hljs bash"`

I don't know what that means. I assume it's about how highlight.js is deployed, but I'm an end user, not a Stack Exchange dev.

**Sample Code to Reproduce**
```bash
# remove leading whitespace characters
var="${var#"${var%%[![:space:]]*}"}"
# remove trailing whitespace characters
var="${var%"${var##*[![:space:]]}"}"
```

**Expected behavior**
GitHub gets it right.
As well, here's Atom:
Screenshot showing code highlighting in Atom

**Additional context**
none

joshgoebel:
Resolves #2730

### DIFF
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -10,9 +10,14 @@ Category: common
 export default function(hljs) {
   const VAR = {};
   const BRACED_VAR = {
-    begin: /\$\{/, end:/\}/,
+    begin: /\$\{/,
+    end:/\}/,
     contains: [
-      { begin: /:-/, contains: [VAR] } // default values
+      "self",
+      {
+        begin: /:-/,
+        contains: [ VAR ]
+      } // default values
     ]
   };
   Object.assign(VAR,{
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 Language Improvements:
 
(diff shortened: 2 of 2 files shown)


src/languages/bash.js
```diff
@@ -10,9 +10,14 @@ Category: common
 export default function(hljs) {
   const VAR = {};
   const BRACED_VAR = {
-    begin: /\$\{/, end:/\}/,
+    begin: /\$\{/,
+    end:/\}/,
     contains: [
-      { begin: /:-/, contains: [VAR] } // default values
+      "self",
```
joshgoebel, line 16: Actual change on this line!

allejo (approved):

PR #2735, opened 2020-10-05
WIP: enh(javascript) highlight operators

joshgoebel:
Test:

```js
f = () => {}
a = 52
a > b
c < d
1 <= 2
5 >= 2
```


taufik-nurrohman:
Is this spec applies to other languages too?

joshgoebel:
Not sure what exactly you are asking...

taufik-nurrohman:
Marking operators is very new in _highlight.js_ as I know. Since the author has decided to not highlight variables of any syntax even if it looks unique and easy to be matched long long time ago (such as PHP variables). I assume that by adding this marker to the JavaScript language, it will lead other language contributors to mark their language-specific operators as well.

joshgoebel:
> Marking operators is very new in highlight.js as I know. Since the author has decided long long time ago, to not highlight any variables even if it looks unique and easy to be matched by regular expression (such as PHP variables). I assume that by adding this marker to the JavaScript language, it will lead other language contributors to mark their language-specific operators as well.

Yes.  I've clarified the variable thing in other threads also - that was never consistently enforced (we have plenty of languages that highlight variables) so I'm fine with highlighting variables if they are easy to match in some languages.

The same holds try with operators now... if they can be matched without adding a ton of complexity to the grammars then I'm open to it.  So far Javascript is not proving as easy as I thought.  But I know some languages will be easier and some harder.

test/markup/cos/embedded.expect.txt
```diff
@@ -2,4 +2,4 @@
  * Multiline comment
  */</span>
 &amp;sql(<span class="sql"><span class="hljs-keyword">SELECT</span> * <span class="hljs-keyword">FROM</span> Cinema.Film <span class="hljs-keyword">WHERE</span> <span class="hljs-keyword">Length</span> &gt; <span class="hljs-number">2</span></span>)
-&amp;js&lt;<span class="javascript"><span class="hljs-keyword">for</span> (<span class="hljs-keyword">var</span> i = <span class="hljs-number">0</span>; i &lt; <span class="hljs-built_in">String</span>(<span class="hljs-string">&quot;test&quot;</span>).split(<span class="hljs-string">&quot;&quot;</span>).length); ++i) { <span class="hljs-built_in">console</span>.log(i); }</span>&gt;
+&amp;js&lt;<span class="javascript"><span class="hljs-built_in">console</span>.log(i); <span class="hljs-built_in">localStorage</span>.set(<span class="hljs-string">&quot;fish&quot;</span>, <span class="hljs-string">&quot;bass&quot;</span>)</span>&gt;
```
joshgoebel, line 5: This was changed to REDUCE the amount of JS code inside `cos`.  `cos` tests should really not break because we make improvements to the JS grammar.

joshgoebel:
Well, I thought I could do this... I got divide vs regex sorted out again... I got JSX working (look for HTML before we look for operators)... but now I'm stuck at types for TypeScript... I'm really not sure how to simply prevent types from being registered as just operators:

```
booger : Array<number>;
# Array [less-than]number[greater-than]
```

`:` isn't great for context either since it can be used for properties (can JSX be a property?) and ternary not just typing... It's easy enough to detect let/const/var but then you also have class instance vars which can be declared with almost no boilerplate (as shown above).

sa-orlov (approved):

sa-orlov (approved):

PR #2738, opened 2020-10-06
fix(c-like) preprocessor directives not detected after else

### LINKED ISSUE #2737: C/C++ Sequential #ifdef preprocessor directives not highlighted consistently
**Describe the issue**

A complete bug report, including image, was drafted on StackOverflow, e.g.

https://meta.stackoverflow.com/q/401819/3422102

from the answer at https://stackoverflow.com/a/64217643/3422102

The problem is multiple #ifdef directives in C/C++ are not highlighted consistently. The first receives syntax highlighting, the remainder don't;

No clue what StackExhange uses. ...

**Sample Code to Reproduce**

```
    #include <stdio.h>
    #include <stdlib.h>
    #include <ctype.h>
    #include <errno.h>
    
    #define MAXC 1024
    
    int main (int argc, char **argv) {
        
        char buf[MAXC];     /* buffer to hold each line */
        /* use filename provided as 1st argument (stdin by default) */
        FILE *fp = argc > 1 ? fopen (argv[1], "r") : stdin;
        
        if (!fp) {  /* validate file open for reading */
            perror ("file open failed");
            return 1;
        }
        
        while (fgets (buf, MAXC, fp)) {             /* read each line into buf */
            char *p = buf, *ep = p;                 /* pointers to use with strtol() */
            for (;;) {                              /* loop continually */
                errno = 0;                          /* reset errno */
                long val = strtol (p, &ep, 0);      /* attempt conversion to long */
                if (p == ep) {                      /* check if no characters converted */
    #ifdef DEBUG
                    fprintf (stderr, "error: no digits converted in '%s'.\n", p);
    #endif
                    break;
                }
                else if (errno) {                   /* check for under/overflow */
    #ifdef DEBUG
                    fprintf (stderr, "error: over/under-flow in conversion or '%s'.\n", p);
    #endif
                    break;
                }
                else if (*ep && !isspace(*ep)) {    /* check next char for non-space */
    #ifdef DEBUG
                    fprintf (stderr, "error: mixed alphanumeric '%s'.\n", ep);
    #endif
                    break;
                }
                else
                    printf (" %ld", val);           /* good all-digit conversion */
                
                p = ep;                             /* update pointr to endpointer */
            }
            putchar ('\n');                         /* tidy up with newline */
        }
        
        if (fp != stdin)   /* close file if not stdin */
            fclose (fp);
    }

```

Note only the first 

```
 #ifdef 
```
is highlighted.

joshgoebel:
Resolves #2737.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 Language Improvements:
 
+- fix(c-like) preprocessor directives not detected after else (#2738) [Josh Goebel][]
 - fix(bash) enh(bash) allow nested params (#2731) [Josh Goebel][]
 - fix(python) Fix highlighting of keywords and strings (#2713, #2715) [Konrad Rudolph][]
 - fix(fsharp) Prevent `(*)` from being detected as a multi-line comment [Josh Goebel][]
--- a/src/languages/c-like.js
+++ b/src/languages/c-like.js
@@ -119,6 +119,7 @@ export default function(hljs) {
   };
 
   var EXPRESSION_CONTAINS = [
+    PREPROCESSOR,
     CPP_PRIMITIVE_TYPES,
     hljs.C_LINE_COMMENT_MODE,
     hljs.C_BLOCK_COMMENT_MODE,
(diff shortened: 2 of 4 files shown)


egor-rogov (approved):
Looks good.

Issue #2742, opened 2020-10-07
The testing suite does not run (on Windows)


schtandard:
**How to reproduce:**
- Install `node.js`.
- From `highlight.js`' root directory, run `npm install`, then `npm test`.
- The following messages appear:
```
> highlight.js@10.2.1 test D:\Projekte\highlight.js_parser\highlight.js
> mocha test


Error: Cannot find module '../build'
Require stack:
- D:\Projekte\highlight.js_parser\highlight.js\test\index.js
- D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\lib\esm-utils.js
- D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\lib\mocha.js
- D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\lib\cli\one-and-dones.js
- D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\lib\cli\options.js
- D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\bin\mocha
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:965:15)
    at Function.Module._load (internal/modules/cjs/loader.js:841:27)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (D:\Projekte\highlight.js_parser\highlight.js\test\index.js:3:14)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.exports.requireOrImport (D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\lib\esm-utils.js:20:12)
    at Object.exports.loadFilesAsync (D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\lib\esm-utils.js:33:34)
    at Mocha.loadFilesAsync (D:\Projekte\highlight.js_parser\highlight.js\node_modules\mocha\lib\mocha.js:427:19)
[...]

PR #2743, opened 2020-10-07
Clarify that one has to build before testing

schtandard:
Since installing the dependencies is already mentioned in the section about building, a reference to that section should be sufficient.

I also included a reference to `build_and_test`.

Related issue: #2742

### DIFF
--- a/docs/building-testing.rst
+++ b/docs/building-testing.rst
@@ -71,7 +71,7 @@ detection still works with your language definition included in the whole suite.
 
 Testing is done using `Mocha <http://mochajs.org/>`_ and the
 files are found in the ``test/`` directory. You can use the node build to
-run the tests in the command line with ``npm test`` after building_. (Using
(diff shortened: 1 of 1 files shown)


docs/building-testing.rst
```diff
@@ -71,8 +71,8 @@ detection still works with your language definition included in the whole suite.
 
 Testing is done using `Mocha <http://mochajs.org/>`_ and the
 files are found in the ``test/`` directory. You can use the node build to
-run the tests in the command line with ``npm test`` after installing the
-dependencies with ``npm install``.
+run the tests in the command line with ``npm test`` after building_. (Using
+``npm run build_and_test`` you can build and then test with one command.)
```
joshgoebel, line 75:
```suggestion
run the tests from the command line with ``npm test`` after building_. (Using
``npm run build_and_test`` you can build and then test with one command.)
```

PR #2745, opened 2020-10-08
enh(autodetect) multiple autodetect fixes

joshgoebel:
These changes were made using a modified test suite from https://github.com/andreasjansson/language-detection.el

Before:

```
67% correct. 80% first or second correct.
```

After: 

```
72% detection is correct.  84% first OR second detected language is correct.
```

So a 4-5% measurable improvement and there are probably also additional improvements "hidden" as in matches that are still not correct, but likely to be better than they were before - though that's a hard one to measure. (it's very hard to tell the difference between Lisp-like languages, etc)

There were a few common areas of focus:

- Removing incorrect `illegal` to allow auto-detect to work for languages that were being flagged as illegal
- Remove double scoring element of many `beginKeywords `rules... I'm wondering if this shouldn't be done at the parser level itself.
- Add illegal in a few places to make false positives less likely.
- Increase match precision of some rules (also guard against false positives)
- Reduce relevancy of rules that could potentially match almost anything (name rules for Lisp like languages, `||` in Ruby matching params when it could be OR or a concat operator in SQL, etc...)

### DIFF
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -49,7 +49,7 @@ export default function(hljs) {
   var AT_PROPERTY_RE = /@-?\w[\w]*(-\w+)*/ // @-webkit-keyframes
   var IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*';
   var RULE = {
-    begin: /(?:[A-Z_.-]+|--[a-zA-Z0-9_-]+)\s*:/, returnBegin: true, end: ';', endsWithParent: true,
+    begin: /([*]\s?)?(?:[A-Z_.\-\\]+|--[a-zA-Z0-9_-]+)\s*(\/\*\*\/)?:/, returnBegin: true, end: ';', endsWithParent: true,
     contains: [
       ATTRIBUTE
     ]
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -125,6 +125,7 @@ export default function(hljs) {
         illegal: /\S/,
         contains: [
           hljs.C_BLOCK_COMMENT_MODE,
+          { begin: /;/ }, // empty ; rule
           RULE,
         ]
       }
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -58,7 +58,7 @@ export default function(hljs) {
   return {
     name: 'CSS',
     case_insensitive: true,
-    illegal: /[=\/|'\$]/,
+    illegal: /[=|'\$]/,
     contains: [
       hljs.C_BLOCK_COMMENT_MODE,
       {
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -65,7 +65,7 @@ export default function(hljs) {
         className: 'selector-id', begin: /#[A-Za-z0-9_-]+/
       },
       {
-        className: 'selector-class', begin: /\.[A-Za-z0-9_-]+/
+        className: 'selector-class', begin: '\\.' + IDENT_RE
       },
       {
         className: 'selector-attr',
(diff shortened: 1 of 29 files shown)


joshgoebel:
Moving this to 10.4 also since it'll probably conflict with the UTF8 stuff and I still want to merge in our long lost CSS branch also...

joshgoebel:
This is definitely going to make more sense if you view the commits one at a time vs as a whole.

src/languages/tcl.js
```diff
@@ -40,14 +44,13 @@ export default function(hljs) {
...
         variants: [
           {
-            begin: '\\$(\\{)?(::)?[a-zA-Z_]((::)?[a-zA-Z0-9_])*\\(([a-zA-Z0-9_])*\\)',
-            end: '[^a-zA-Z0-9_\\}\\$]'
+            begin: '\\$(::)?[a-zA-Z_]+((::)?[a-zA-Z0-9_]+)*' + regex.optional(ARRAY_ACCESS),
           },
           {
-            begin: '\\$(\\{)?(::)?[a-zA-Z_]((::)?[a-zA-Z0-9_])*',
-            end: '(\\))?[^a-zA-Z0-9_\\}\\$]'
+            begin: '\\$\\{(::)?[a-zA-Z_]((::)?[a-zA-Z0-9_])*' + regex.optional(ARRAY_ACCESS) + '\\}',
```
egor-rogov, line 53: Hmm... This doesn't look right to me. Consider e. g. `$foo($bar)`: array index can be a full blown expression. You can have more fun:
```tcl
set foo bar 
set ${foo}(1) 42 ; # this is not captured by the grammar
puts $bar(1)
```
It has nothing to do with autodection fixes though.

src/languages/vbscript.js
```diff
@@ -7,7 +7,40 @@ Website: https://en.wikipedia.org/wiki/VBScript
...
+  'chr lbound msgbox ucase getlocale cos cdate cbyte rtrim join hour oct typename trim ' +
+  'strcomp int createobject loadpicture tan formatnumber mid ' +
+  'split  cint sin datepart ltrim sqr ' +
+  'time derived eval date formatpercent exp inputbox left ascw ' +
+  'chrw regexp cstr err').split(" ");
+  const BUILT_IN_OBJECTS = [
+    "server",
+    "response",
+    "request",
+    // take no arguments so can be caleld without ()
```
egor-rogov, line 28: typo: called

egor-rogov (approved):
Extensive work! Generally looks good, see some comments.

src/languages/css.js
```diff
@@ -49,7 +49,7 @@ export default function(hljs) {
   var AT_PROPERTY_RE = /@-?\w[\w]*(-\w+)*/ // @-webkit-keyframes
   var IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*';
   var RULE = {
-    begin: /(?:[A-Z_.-]+|--[a-zA-Z0-9_-]+)\s*:/, returnBegin: true, end: ';', endsWithParent: true,
+    begin: /([*]\s?)?(?:[A-Z_.\-\\]+|--[a-zA-Z0-9_-]+)\s*(\/\*\*\/)?:/, returnBegin: true, end: ';', endsWithParent: true,
```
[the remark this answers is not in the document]
joshgoebel: Common css hack.

src/languages/tcl.js
joshgoebel: Do you actually know TCL are you just looked it up?  I’ll go back and review this again because I think maybe I see some issues now but I’m not sure. It be really helpful to have some additional test to show what supposed to be possible here. On some of these languages I’m really shooting blind just looking at the examples I have and what the existing rules seem to do.

egor-rogov: I used to knew it... wrote several programs in Tck/Tk long ago. Now I can hardly remember anything.

PR #2746, opened 2020-10-10
fix(js) prevent runaway regex

joshgoebel:
Evidentally detecting a serious of comments one after another can
still lead to a runaway regex when you have a ton of comments in
a large file.

This should still cover most common cases.


joshgoebel:
I need to understand this better.  Regex can be so annoying. :)

egor-rogov (approved):

PR #2747, opened 2020-10-10
WIP: IBM RPG IV a.k.a. ILE RPG (total free) language support

KerimG:
Trying to create a language definition for IBM's RPG language. This particular definition is supposed to just include the total free variant. There are two more variants of the same programming language, a column-based variant

image

 and a mixed variant

image

RPG can also embed SQL. That is, SQL commands aren't just strings but actual _entities_. Not sure how to implement that in the language definition, yet.


egor-rogov:
Thanks so much for your contribution! Sadly, due to lack of maintainer time we no longer merge additional languages into the core library - but we'd love to have your grammar as a 3rd party language module if you'd be willing to maintain it over time. It's quite easy to convert what you've already got into a 3rd party module.

We can host that repository here at the highlightjs organization (and are happy to) or you can host it on your own GitHub. The key thing is you need to be willing to maintain it, deal with any future issues, etc - i.e. become the de-facto maintainer of the module. Once you have your repository all setup we will happily accept a PR to add a link to our list of supported languages here in the official repo.

Please read https://github.com/highlightjs/highlight.js/blob/master/docs/language-contribution.rst.

joshgoebel:
> RPG can also embed SQL. That is, SQL commands aren't just strings but actual entities. Not sure how to implement that in the language definition, yet.

You mean you can just write SQL inline?  You'd need a for sure way to detect the beginning and end of the SQL and then trigger a submode... like:

```
begin: /SELECT/
end: /;/,
sublanguage: "sql",
```

This might require more complex matching within the SQL to prefer it from finding `;` in say a string, etc...  You might look at `php-template` for an example of this.

KerimG:
Thanks for the advice, @joshgoebel . 

I looked at a few more embedded SQL statements and I think there's a relatively simple way to detect it. 

```

Exec SQL
  SELECT SYSTOOLS.HTTPGETBLOB(:gURL, '')
  INTO :gFile
  FROM SYSIBM.SYSDUMMY1;
```

All embedded SQL parts begin with `Exec SQL` and end with `;`, though I'm not sure how it works if the ending semicolon is reliable when creating more complex SQLs. I got a few at work and will check them out on Monday. Maybe the language spec has some clues also.

joshgoebel:
> I looked at a few more embedded SQL statements and I think there's a relatively simple way to detect it.

If it's always in an indented block you could key off of that also.

---

I'm closing this PR as it doesn't need to remain open.  When you have your grammar ready you can open a new PR updating SUPPORTED_LANGUAGES... or if you'd like to host that repo inside the `highlightjs` organization let us know and we can set that up for you.

PR #2752, opened 2020-10-15
fix(cpp) fix class/template parsing issues

### LINKED ISSUE #2102: (cpp) grammar confused by complex Template << >> patterns

### LINKED ISSUE #2716: (cpp) Template parameters trigger class header rule

joshgoebel:
This fixes #2716 and it fixes #2102.

This solves the problem simply by aborting the class mode whenever we see a `<`, `>` or `=`, and returning to the regular rule matching.  This is a much simpler solution that trying to process deeply nested templates and count subexpressions of `<` vs `(` - all the while trying to figure out what is perhaps simple a less than (or greater than) vs a template delimiter.

This also adds support for `enum struct` and `enum class`.  This actually mostly worked before but now the "class" block extends around the "enum" as well which is slightly more correct I think.

- Also adds a minor linting fix to quite a few tests regarding carriage return at the end of file.

See tests.

@klmr


src/languages/c-like.js
```diff
@@ -225,10 +225,9 @@ export default function(hljs) {
       },
       {
         className: 'class',
-        beginKeywords: 'class struct', end: /[{;:]/,
+        beginKeywords: 'enum class struct', end: /[{;:<>=]/,
```
klmr, line 228: This keyword list is still missing `union`.

src/languages/c-like.js
```diff
@@ -225,10 +225,9 @@ export default function(hljs) {
...
         contains: [
-          { beginKeywords: "final" },
-          {begin: /</, end: />/, contains: ['self']}, // skip generic stuff
+          { beginKeywords: "final class struct enum" },
```
klmr, line 230: And I don’t think `enum` is required here.

test/markup/cpp/template_complexity.txt
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// Disable overload for already valid operands.
+template<class T, class = std::enable_if_t<!impl::is_streamable_v<const T &> && std::is_convertible_v<const T &, std::wstring_view>>>
+std::wostream &operator <<(std::wostream &stream, const T &thing)
+{
+    return stream << static_cast<std::wstring_view>(thing);
+}
+
+enum struct DataHolder { }
```
klmr, line 16: It doesn’t really matter either way but we should probably add semicolons behind this and the following type definitions to make the test cases valid C++.

klmr (changes requested):
This still doesn’t handle attributes, e.g.

```cpp
struct [[deprecated]] foobar {};
```

Of course a simple fix would be to add `|\[\[` to `end` but handling them properly isn’t actually much harder: we might just need a nested rule `{begin: /\[\[/, end: /]]/, keywords: 'using'}`.

But I wouldn’t increase the scope of this PR, as such attributes still relatively rare in C++ code (and even rarer on type declarations).

src/languages/c-like.js
joshgoebel: Yeah, but doesn't hurt either. :-)

joshgoebel:
@klmr  Better?

klmr (approved):

Issue #2756, opened 2020-10-15
(JS, C++, C#, python, etc) Improper tokenization with variable names in languages that use Unicode 


EmNudge:
**Describe the issue**
Most modern languages use the Unicode spec to decide what constitutes a proper variable name. Thus, hangul filler characters, even though they seem to be whitespace, are valid variable names.
The variable name `thisﾠisﾠaﾠvariableﾠname` is a valid variable, although hl.js tokenizes the first word or so as separate from the rest.
In highlighting systems that are aware of this, like Chrome's dev tools, it is parsed correctly. The first snippet uses hangul fillers while the second one uses whitespace (which is why it throws an error):
![image](https://user-images.githubusercontent.com/24513691/96193673-a9f47700-0f16-11eb-89bc-f763692fde46.png)

**Which language seems to have the issue?**
Every language that follows the Unicode spec for variable names theoretically. I've only tested the ones mentioned in the title, but it might make more sense to just tokenize every language as if it were one of those that followed the unicode spec since this should be the default.

**Are you using `highlight` or `highlightAuto`?**
N/A

**Sample Code to Reproduce**
N/A

**Expected behavior**
hl.js should be using ID_Start and ID_Continue for the languages that use it, or at least use it as a default.
The regex to match this is rather large, but it can be shortened significantly in regex engines supporting the `u` flag (modern JavaScript)
```js
const matchIDStart = char => /\p{ID_Start}/u.test(char);
const matchIDContinue = char => /\p{ID_Continue}/u.test(char);
```
`\p{ID_Start}` matches any valid starting character in a variabe name
`\p{ID_Continue}` matches any valid non-starting character in a variable name (such as numbers and some variation selectors)

joshgoebel:
> hl.js should be using ID_Start and ID_Continue for the languages that use it, or at least use it as a default.

How might we know which languages those are? :-)  We have almost 200 languages...  Right now our keyword default is `\w` and we don't really have an "identifier" default... each language kind of provides their own identifier regex... I'm thinking this might be something we have to do one language at a time as it comes up... flipping just that language into "utf8 mode" and then making sure it's regex are all still valid, etc...

This might be quite an effort for the whole project.  Just a quick attempt to build all our regex with `/u` and then run just our markup test suite:

```
  108 passing (716ms)
  173 failing
```

Evidently the `/u` flag changes how regex strings are parsed and it's a LOT more strict about a lot of things that work just fine without it.  One example:

Valid before:

[...]

PR #2759, opened 2020-10-16
(chore) Clean up all regexs to be UTF-8 compliant/ready

joshgoebel:
Work toward #2756.

Cleans up a lot of incorrect (unnecessary escaped) regex and would not compile with the `u` flag.  After that makes some rather large performance improvements (with utf8 turned on at least) to `yaml` and `mipsasm`.  It looks like the mipasm rules have been wrong all alone... as far as I can determine they are intended to match a literal `.` (otherwise they are far too broad) but were matching any character - which seems to terribly slow down the whole grammar in `u` mode.

This consisted mostly of:

- Most unescaped `{` and `}`
- Lots of unneeded escapes for `-`, `<`, `>`, and others.
- Converting strings to regex if it made them simpler, easier to read (editor syntax coloring)

I plan to review the PR myself line by line, but can't imagine it'll be fun.  I did try to change the minimal amount necessary.  Often I turned strings into regex if it made them easier to read and see what I was doing.  Once in a while I touched a nearby regex.

I imagine the fact that all tests still pass is a pretty good indication this is 99% correct. :-)

Note: This doesn't actually turn on UTF8 anywhere... it just fixes all the regex so that if `u` is added inside the main mode compiler everything "just works".  It is still needed to be reviewed what else might need to be done on the road to UTF8 support.


joshgoebel:
No rush, I'm pushing this off until 10.4.  10.3 is big enough.

src/languages/qml.js
```diff
@@ -90,7 +90,7 @@ export default function(hljs) {
   // Find QML object. A QML object is a QML identifier followed by { and ends at the matching }.
   // All we really care about is finding IDENT followed by { and just mark up the IDENT and ignore the {.
   var QML_OBJECT = {
-    begin: QML_IDENT_RE + '\\s*{', end: '{',
+    begin: QML_IDENT_RE + /\s*\{/, end: /\{/,
```
joshgoebel, line 93: Need to review this.

src/languages/tap.js
```diff
@@ -22,7 +22,7 @@ export default function(hljs) {
       },
       // YAML block
       {
-        begin: '(\s+)?---$', end: '\\.\\.\\.$',
+        begin: /---$/, end: '\\.\\.\\.$',
```
joshgoebel, line 25: This was broken before `\s` vs `\\s` and when fixed seemed to break things so without any context going to go with the simpler rule for now.

src/languages/dsconfig.js
```diff
@@ -15,12 +15,12 @@ export default function(hljs) {
   };
   var UNQUOTED_PROPERTY = {
     className: 'string',
-    begin: '[\\w-?]+:\\w+', end: '\\W',
+    begin: /[\w\-?]+:\w+/, end: /\W/,
```
joshgoebel, line 18: I don't completely understand when `-` needs to be escaped inside a `[]` block since sometimes it seems to figure it out on context (or maybe it's my imagination).

src/languages/mipsasm.js
```diff
@@ -90,6 +90,6 @@ export default function(hljs) {
         relevance: 0
       }
     ],
-    illegal: '\/'
+    illegal: /\//
```
joshgoebel, line 93: Need to look at this carefully.

src/languages/prolog.js
```diff
@@ -51,15 +51,13 @@ export default function(hljs) {
...
     className: 'string', // 0'a etc.
-    begin: /0\'(\\\'|.)/
+    begin: /0'(\\'|.)/
   };
 
   var SPACE_CODE = {
-
     className: 'string',
-    begin: /0\'\\s/ // 0'\s
+    begin: /0'\\s/ // 0'\s
```
joshgoebel, line 60: This is greek to me, but I think it's correct... anyone know Prolog?  The "\s" is literal?

src/languages/yaml.js
```diff
@@ -102,7 +102,7 @@ export default function(hljs) {
       // Indentation of subsequent lines must be the same to
       // be considered part of the block
       className: 'string',
-      begin: '[\\|>]([0-9]?[+-])?[ ]*\\n( *)[\\S ]+\\n(\\2[\\S ]+\\n?)*'
+      begin: '[\\|>]([0-9]?[+-])?[ ]*\\n( *)[^\\n]+\\n(\\2[^\\n]+\\n?)*'
```
joshgoebel, line 105: Speed optimization... I think this is accomplishing the same thing...

src/languages/routeros.js
```diff
@@ -144,7 +144,7 @@ export default function(hljs) {
       }, //*/
 
       {
-        begin: '\\b(' + COMMON_COMMANDS.split(' ').join('|') + ')([\\s\[\(]|\])',
+        begin: '\\b(' + COMMON_COMMANDS.split(' ').join('|') + ')([\\s[(\\]|])',
```
joshgoebel, line 147: Maybe I broke this?  Not sure what it's supposed to be doing.

src/languages/lsl.js
```diff
@@ -31,29 +31,29 @@ export default function(hljs) {
         className: 'literal',
         variants: [
             {
-                begin: '\\b(?:PI|TWO_PI|PI_BY_TWO|DEG_TO_RAD|RAD_TO_DEG|SQRT2)\\b'
+                begin: '\\b(PI|TWO_PI|PI_BY_TWO|DEG_TO_RAD|RAD_TO_DEG|SQRT2)\\b'
```
joshgoebel, line 34: There are speed optimizations believe it or not the benchmark was faster without the ":" to mark it as a non capturing group.

src/languages/mipsasm.js
joshgoebel, line 93:
```suggestion
    // forward slashes are no allowed
    illegal: /\//
```

joshgoebel: Yep, it's the same as before.  Adding a comment to clarify.

src/lib/mode_compiler.js
```diff
@@ -24,7 +24,7 @@ export function compileLanguage(language) {
   function langRe(value, global) {
     return new RegExp(
       regex.source(value),
-      'm' + (language.case_insensitive ? 'i' : '') + (global ? 'g' : '')
+      'um' + (language.case_insensitive ? 'i' : '') + (global ? 'g' : '')
```
joshgoebel, line 27: This will be reverted before merge.

src/highlight.js
```diff
@@ -579,12 +579,17 @@ const HLJS = function(hljs) {
     @param {Array<string>} [languageSubset]
     @returns {AutoHighlightResult}
   */
+    let ts = {};
```
joshgoebel, line 582: This will be reverted before merge.

joshgoebel:
I've 99% review this myself but could still use another set of eyes. :)

src/languages/dsconfig.js
joshgoebel: Ah, seems simple enough... in cases where it could be ambiguous it now wants it to be escaped:

```
regex = /[a\-b]/u
```

But if you just always put it at the end of the character group then it's obvious it's not a range.

src/languages/routeros.js
joshgoebel: Pretty sure it's fine, added a comment to explain it.

src/languages/prolog.js
joshgoebel: This makes sense to me now I think.  Code might be:

```prolog
% \s literal (escaped)
0'\s
% character "b"
0'b
% character "'" (escaped)
0'\'
```

src/languages/dsconfig.js
allejo: From the changes I've seen up until this file, it looks like you've moved the `-` to the end of each `[]` or it was already there, do you want to keep that consistent?

src/languages/ebnf.js
```diff
@@ -9,7 +9,7 @@ export default function(hljs) {
 
     var nonTerminalMode = {
         className: "attribute",
-        begin: /^[ ]*[a-zA-Z][a-zA-Z-_]*([\s-_]+[a-zA-Z][a-zA-Z]*)*/
+        begin: /^[ ]*[a-zA-Z][a-zA-Z-_]*([\s\-_]+[a-zA-Z][a-zA-Z]*)*/
```
allejo, line 12:
```js
begin: /^[ ]*[a-zA-Z][a-zA-Z-_]*([\s\-_]+[a-zA-Z][a-zA-Z]*)*/
#                           ^       ^^
```

You're escaping one `-` here but not the other?

src/languages/xquery.js
```diff
@@ -106,7 +106,7 @@ export default function(hljs) {
 
   var ANNOTATION = {
     className: 'meta',
-    begin: /%[\w-:]+/
+    begin: /%[\w\-:]+/
```
allejo, line 109: hyphen consistency

src/languages/yaml.js
allejo: `\S` is anything _but_ whitespace. So `[\S ]` to me sounds like "anything but whitespace but also allow a single space character". I don't think tabs or spaces aren't covered by `\n`?

src/languages/dsconfig.js
joshgoebel: I'm happy now that I understand it.

src/languages/yaml.js
joshgoebel: But we're using "not \n"... The question to ask here is what was the original intent... aren't we trying to handle code like:

```
key: |
  multi-string
  value
```

Ie, some number of spaces (matching, hence the `\2`) then content (up until a new line)...  I believe `[^\n]` matches the intention, yes?

joshgoebel: The full expression is `[\S ]+`... not space or space, repeated at least once... and this naturally excludes new lines... so I (I believe) wrote an equivalent and faster version.

src/languages/ebnf.js
joshgoebel: Doesn't need to be, but I'll move them both to the end to make this consistent.

src/languages/gams.js
```diff
@@ -78,7 +78,7 @@ export default function (hljs) {
       ASSIGNMENT,
       {
         className: 'comment',
-        begin: /([ ]*[a-z0-9&#*=?@>\\<:\-,()$\[\]_.{}!+%^]+)+/,
+        begin: /([ ]*[a-z0-9&#*=?@><:\-,()$[\]_.{}!+%^]+)+/,
```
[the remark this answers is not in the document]
joshgoebel: - The `[` does not need to be escape inside a `[` block... it's clear it's not opening a new block because it's inside one.
- `<` is not a character that's has ever needed escaping (so says my linter), and it's now invalid in `/u` mode.

> Like technically /{/ doesn't need to be escaped, but this PR is escaping it anyways.

Very different `{` **must be escaped** for the stricter UTF-8, which is the whole point of this PR.  `[` does not need to be.  The overall preference here is not to use unnecessary escapes anymore.

src/languages/less.js
```diff
@@ -118,13 +118,13 @@ export default function(hljs) {
...
       IDENT_MODE('keyword',  'all\\b'),
-      IDENT_MODE('variable', '@{'  + IDENT_RE + '}'),     // otherwise it’s identified as tag
+      IDENT_MODE('variable', '@\\{'  + IDENT_RE + '\\}'),     // otherwise it’s identified as tag
       IDENT_MODE('selector-tag',  INTERP_IDENT_RE + '%?', 0), // '%' for more consistent coloring of @keyframes "tags"
       IDENT_MODE('selector-id', '#' + INTERP_IDENT_RE),
       IDENT_MODE('selector-class', '\\.' + INTERP_IDENT_RE, 0),
       IDENT_MODE('selector-tag',  '&', 0),
       {className: 'selector-attr', begin: '\\[', end: '\\]'},
-      {className: 'selector-pseudo', begin: /:(:)?[a-zA-Z0-9\_\-\+\(\)"'.]+/},
+      {className: 'selector-pseudo', begin: /:(:)?[a-zA-Z0-9_\-+()"'.]+/},
```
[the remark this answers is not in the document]
joshgoebel: In this regard I don't think it matters, plus without a linter rule this is hard to enforce over time in any case.  I'm happy if it's escaped properly where-ever it's found.

src/languages/stylus.js
```diff
@@ -122,7 +122,7 @@ export default function(hljs) {
     'video'
   ];
 
-  var LOOKAHEAD_TAG_END = '(?=[\\.\\s\\n\\[\\:,])';
+  var LOOKAHEAD_TAG_END = '(?=[.\\s\\n\\[:,])';
```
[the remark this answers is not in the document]
joshgoebel: Thanks, hard to see all these. :)

src/languages/lisp.js
```diff
@@ -6,9 +6,9 @@ Category: lisp
 */
 
 export default function(hljs) {
-  var LISP_IDENT_RE = '[a-zA-Z_\\-\\+\\*\\/\\<\\=\\>\\&\\#][a-zA-Z0-9_\\-\\+\\*\\/\\<\\=\\>\\&\\#!]*';
+  var LISP_IDENT_RE = '[a-zA-Z_\\-+\\*\\/<=>&#][a-zA-Z0-9_\\-+*\\/<=>&#!]*';
   var MEC_RE = '\\|[^]*?\\|';
-  var LISP_SIMPLE_NUMBER_RE = '(\\-|\\+)?\\d+(\\.\\d+|\\/\\d+)?((d|e|f|l|s|D|E|F|L|S)(\\+|\\-)?\\d+)?';
+  var LISP_SIMPLE_NUMBER_RE = '(-|\\+)?\\d+(\\.\\d+|\\/\\d+)?((d|e|f|l|s|D|E|F|L|S)(\\+|-)?\\d+)?';
```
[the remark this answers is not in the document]
joshgoebel: It's used with `+` and strings later, so it doesn't work if it's converted to a regex... I'd need to start using the `regex` tools everywhere... 

If someone wanted to focus on "regex not string" a lot more as a separate initiative I'd be game for that, but I only tried to do that here in cases where it made sense and was trivial to do - the scope of this PR is already huge.

src/languages/gams.js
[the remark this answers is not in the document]
allejo: > Very different `{` **must be escaped** for the stricter UTF-8, which is the whole point of this PR. `[` does not need to be. The overall preference here is not to use unnecessary escapes anymore. _They've always been flagged by the linter, it just hasn't been a high priority before now._

ah thanks for the explanation! Didn't realize that `{` had to be escaped for UTF-8 support. I just thought since it wasn't ambiguous it didn't need to be escaped.

---

> I'm a bit confused on this one. Is the removal of `\\` here correct? Is it trying to escape `<` or is `\` an actual character in this `[]` block?

Can confirm this? Being inside of a `[]` and regex, it looks to be like `\\` is for an `\` character and not trying to escape `<`. i.e. if this were a string originally, then I agree that it should be removed because it'd be `\<` but this has always been a regex so it doesn't look to me like it's escaping the `<`.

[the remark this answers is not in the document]
joshgoebel: Oh yikes, yeah I think you are right. :)  I put it back!

src/languages/prolog.js
allejo: Yea, looks like `0'` is the way of getting the character code of a char. So `0'\s` returns `32`. And you are correct, `0''` is valid, it returns `39`

Screenshots and confirmation courtesy of @Adwitiya-Singh

image

More info, https://www.swi-prolog.org/pldoc/man?section=charescapes

joshgoebel: > And you are correct, 0'' is valid, it returns 39

Valid or invalid?  Although I suppose the regex currently allows it either way, LOL... and I'm not even sure if that is bad or not.

src/languages/yaml.js
allejo: Taking this snippet out of context, I got the following screenshots:

image
image
image

`\S` is short for [`[^ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes)

At the very least, shouldn't `\t` be inside alongside `\n` too?

src/languages/prolog.js
allejo: My bad, I misread your original comment 😅 `0''` is _valid_ as seen in the screenshot. But since the regex allows it, works for me

src/languages/yaml.js
joshgoebel: That's clear enough, but I think the prior behavior is WRONG... (unless you want to school me on a finer detail of YAML here - which is possible since YAML is insane)... the intention AFAIK is to go all the way to the end of the line... I'm not sure tabs should break a string sequence... I would expect them to simply be registered as part of the content.

If that is *incorrect* due to some insanity of YAML then I believe you would be right in that we should add `\t`.

allejo: Ahhh I see what this regex is supposed to do now. I was just comparing the semantics of diffs between regexes. You are correct, there should not be a `\t` in here and previous behavior was wrong 👍

allejo, line 105:
```suggestion
      begin: '[\\|>]([0-9]?[+-])?[ ]*\\n( +)[^\\n]+\\n(\\2[^\\n]+\\n?)*'
```

While we're talking about the behavior of this regex in this PR, should it be one or more spaces for the indentation? Having no indentation is invalid.

```yaml
key: |
multi-string
value
```

joshgoebel: In this case (and what I know of YAML) I'd agree but I'm not certain what's going on with all the other `[\\|>]([0-9]?[+-])?` there... are there some other variants we're not considering?  I think the 0-9 is to allow offsets/extra indents but I'm not sure what `\\` and `>`... is the \` is just needless escaping then that leaves only the `>`... does that change anything?

allejo: My favorite resource for deciphering YAML multi-line values... https://yaml-multiline.info/

- You can use 1-9 for indentation if the first line's indentation won't match the rest of the lines

According to yamllint.com, the `\` is valid,

this:

```yaml
key: \
  multi-string
  value
```

becomes this:

```yaml
key: "\\ multi-string value"
```

joshgoebel:
Anything else? :)

allejo (approved):
Just don't forget to revert [this](https://github.com/highlightjs/highlight.js/pull/2759/files#r511647261) but this looks good to me

PR #2761, opened 2020-10-16
fix(c/cpp) line continuation in single-line comments

### LINKED ISSUE #2757: (C) \ not respected
**Describe the issue**

`C`

`highlight`

**Sample Code to Reproduce**

https://jsfiddle.net/gm16s8nf/
```c
// this is a comment, but since I added a slash, \
this is part of it too
```

**Expected behavior**

In C, a '\\' character extends the current lined (effectively negating the following `\n` )
Since the second line is also part of the comment, it should be highlighted as such.

**Additional context**

klmr:
Resolves #2757 

### Changes

Use a custom definition for single-line comments in `c-like` instead of deferring to `hljs.C_LINE_COMMENT_MODE`. The only difference is a new nested mode that eats `\\\n`.

### Checklist

### DIFF
--- a/src/languages/c-like.js
+++ b/src/languages/c-like.js
@@ -18,6 +18,9 @@ export default function(hljs) {
   function optional(s) {
     return '(?:' + s + ')?';
   }
+  var C_LINE_COMMENT_MODE = hljs.COMMENT('//', '$', {
+    contains: [{begin: /\\\n/}]
+  });
   var DECLTYPE_AUTO_RE = 'decltype\\(auto\\)'
   var NAMESPACE_RE = '[a-zA-Z_]\\w*::'
   var TEMPLATE_ARGUMENT_RE = '<.*?>';
@@ -80,7 +83,7 @@ export default function(hljs) {
         begin: /<.*?>/, end: /$/,
         illegal: '\\n',
       },
-      hljs.C_LINE_COMMENT_MODE,
+      C_LINE_COMMENT_MODE,
       hljs.C_BLOCK_COMMENT_MODE
     ]
   };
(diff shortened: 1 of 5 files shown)


src/languages/c-like.js
```diff
@@ -18,6 +18,9 @@ export default function(hljs) {
   function optional(s) {
     return '(?:' + s + ')?';
   }
+  var C_LINE_COMMENT_MODE = hljs.COMMENT('//', '$', {
+    contains: [{begin: /\\\n/}]
```
joshgoebel, line 22: Should this be `\\$` since we're always in multi-line mode?

src/languages/c-like.js
```diff
@@ -18,6 +18,9 @@ export default function(hljs) {
   function optional(s) {
     return '(?:' + s + ')?';
   }
+  var C_LINE_COMMENT_MODE = hljs.COMMENT('//', '$', {
```
joshgoebel, line 21:
```suggestion
  }
  // added for historic reasons because `hljs.C_LINE_COMMENT_MODE` does 
  // not include such support nor can we be sure all the grammars depending
  // on it would desire this behavior
  var C_LINE_COMMENT_MODE = hljs.COMMENT('//', '$', {
```

src/languages/c-like.js
klmr: `$` doesn’t work here (presumably since it’s a zero-width assertion, and the newline isn’t consumed?)

joshgoebel: Duh. :-)

klmr: In fact, `\\(?=$).` would work — but only in single-line mode, since `.` otherwise fails to match a newline.

joshgoebel (approved):

PR #2770, opened 2020-10-18
chore(dev) add theme picker to the developer tool

joshgoebel:
Adds theme picker to tools/developer.html, makes it easier to see how
various themes look when developing/improving language grammars.

### Checklist
- [x] Added markup tests, or they don't apply because this change does not touch grammars...

### DIFF
--- a/tools/developer.html
+++ b/tools/developer.html
@@ -226,8 +226,7 @@
         var editor = $(this).parents('.editor');
         editor.toggleClass('visible-structure');
 
-        localStorage.setItem(key_STRUCTURE,
-          editor.hasClass('visible-structure'));
+        localStorage.setItem(key_STRUCTURE, editor.hasClass('visible-structure'));
       });
     });
 
--- a/tools/developer.html
+++ b/tools/developer.html
@@ -66,6 +66,100 @@
         <button id="update-highlighting">Update highlighting</button>
         <button id="show-structure">Show/hide structure</button>
         Language: <select class="languages"><option value="">(Auto)</option></select>
+        Theme: <select class="theme">
+          <option>default.css</option>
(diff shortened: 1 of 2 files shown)


tools/developer.html
```diff
@@ -108,37 +219,56 @@ <h3>Markup</h3>
...
-            localStorage.getItem(key_SOURCE),
-            localStorage.getItem(key_LANG)
-          ];
+          let code = localStorage.getItem(key_CODE)
+          if (!code) return null;
+
+          return {
+            code,
+            language: localStorage.getItem(key_LANGUAGE),
+            theme: localStorage.getItem(key_THEME),
```
allejo, line 253:
```suggestion
            theme: localStorage.getItem(key_THEME) || 'default.css',
```

The page loads without a default style if we rely solely on local storage.

tools/developer.html
```diff
@@ -108,37 +219,56 @@ <h3>Markup</h3>
...
-        SourceStore.save(source, language);
+        saveSettings();
       });
 
       $('.editor button#show-structure').click(function(e) {
         var editor = $(this).parents('.editor');
         editor.toggleClass('visible-structure');
+
+        localStorage.setItem(key_STRUCTURE,
+          editor.hasClass('visible-structure'));
```
allejo, line 230:
```suggestion
        localStorage.setItem(key_STRUCTURE, editor.hasClass('visible-structure'));
```

not sure why this was on two lines

allejo (changes requested):

tools/developer.html
joshgoebel: Not sure it mattered, but ok. :)

Issue #2751, opened 2020-10-14
Research: more nuanced RegEx highlighting


binyamin:
### Is your request related to a specific problem you're having?
I have a block of code with a rather long regex. The entire regex is one color.

### The solution you'd prefer / feature you'd like to see added
Break a regex into smaller components, maybe with an enclosing `.hljs-regexp` class. Those components might be general, or regex-specific.

### Additional context
(Theme: Nord)
**In Highlight.js**
![image](https://user-images.githubusercontent.com/39805353/96049208-b230b180-0e45-11eb-9ca3-10c6b51e2b07.png)
**In VS Code**
![image](https://user-images.githubusercontent.com/39805353/96049126-8f060200-0e45-11eb-9288-a47d89b496a6.png)

joshgoebel:
This would likely fall under https://github.com/highlightjs/highlight.js/issues/2500.

> maybe with an enclosing .hljs-regexp class. 

Yes, what would be needed is actual subclasses (ie, `regex.subgroup`) (in a parser sense, not necessary CSS)... such as [forgive my naming]:

- sub grouping ()
- character group []
- count modifier (`*`,`?`,`+`)
- escape sequences?
- pins (`^`, `$`)

We don't have a way to codify such subclasses yet, but I'm open to figuring that out - it just hasn't become necessary yet.

My worry here is this would require pretty accurate almost real parsing of regex (which traditionally we try to avoid because of the complexity and maintenance) - and that may or may not be possible. If someone came up with a rough prototype I'd definitely review it.  But if the parsing can't be fully accurate (ie, it misidentifies things regularly or simply can't handle some regex) then I'm not very interested in trying to fix regex parsing bugs in perpetuity.  And of course then themes would need to be altered to support this on a case by case basis unless the missing rules could be added programmatically.

**ALSO worth noting:** This could easily be done with a 3rd party plugin if it's decided this does not belong in core.  You'd simply redefine `hljs.REGEXP_MODE` to be whatever you desired [adding the rich detail you're requesting vs custom css classes] -  and then be responsible for providing a CSS theme that provided the new styles (or make that a task for the user).

joshgoebel:
@binyamin You want to try and see what you can come up with?

binyamin:
@joshgoebel I'd love to. I don't know the code at all, but I can take a look.

joshgoebel:
Actually it may be much simpler and you might not need a full parser at all...  

[...]

PR #2771, opened 2020-10-18
WIP: enh(parser) Better regular expression internals highlighting

klmr:
This PR adds a new language definition for regular expressions.

#2751 discusses adding more nuanced regex highlighting as a nested mode to existing languages. By contrast, this PR concerns highlighting regex as a standalone language (though of course it might be worth investigating how to enable it as a nested mode).

Notably, [Stack Overflow used to support syntax highlighting of regular expressions, specifically](https://meta.stackexchange.com/a/184109/1968) but now no longer does.

This is work in progress, and there are some outstanding questions, notably

* What’s the canonical name *vs.* aliases? (At the moment: `regex`; aliases: `re` , `regexp`).
* Which class names to use to highlight individual regex features?
    * The ones I’m currently using are (not necessarily good) suggestions
* How extensive should the highlighting be anyway?
* What regex dialects do we support?
    * My attempt was to capture as many different dialects as possible — PCRE, POSIX and other languages.
* How to implement groups such that parentheses, as well as modifier sequences (`?…` directly after the opening parenthesis) are somehow highlighted? My initial attempt via `starts` didn’t work (commented out code).
* Should character class ranges (e.g. `[0-9A-F]`) be highlighted specially?

### Missing features

I don’t necessarily plan to implement these in this PR unless there’s interest, since they’re fairly esoteric:

* [if-then-else conditionals](https://www.regular-expressions.info/conditional.html) (`(?(re)then|else)` etc.)
* [subroutines](https://www.regular-expressions.info/subroutine.html) (`(?(DEFINE)(?'name're))`, `(?&name)`)
* [free-spacing regex](https://www.regular-expressions.info/freespacing.html) (`x` mode)
* character class set operations (`[class-[sub]]`, `[class&&[intersect]]`, `[class&&intersect]`)
* [escaping via `\Q…\E`](https://www.regular-expressions.info/characters.html#special) (`\Q` and `\E` are currently highlighted as regular escape sequences, the stuff in between gets no special treatment)
* [ASCII escape sequence tokens `\cX` and octal escape sequences](https://www.regular-expressions.info/nonprint.html) (how widespread are these?)
* The lexer does not try to guess if it’s dealing with the basic POSIX dialect, i.e. `|`, `+` etc. are always treated as special, and `\|`, `\+` etc. are always treated as escape sequences

### Checklist

### DIFF
--- a/src/languages/regex.js
+++ b/src/languages/regex.js
@@ -52,11 +52,11 @@ export default function (hljs) {
   const GROUP_SPECIALS = [
     // Named group
     'P?<[^>]+>',
-    "'['>'+'",
-    // Zero-length assertions
-    '[|>=!]|<=|<!|',
+    "'[^']+'",
+    // Zero-length assertions, branch reset & atomic
+    '[|>=!]|<=|<!',
     // Modifiers applied inside group
-    `${MODIFIERS}:`,
+    `(?:${MODIFIERS}):`,
     ':'
   ];
 
--- a/src/languages/regex.js
+++ b/src/languages/regex.js
@@ -65,30 +65,14 @@ export default function (hljs) {
     begin: `\\(\\?(?:${MODIFIERS})\\)`
   };
 
-  let GROUP = {};
-  GROUP = {
-    className: 'group',
-    begin: `\\((?:\\?(?:${GROUP_SPECIALS.join('|')}))?`, end: /\)/,
-    // starts: {
-    //   // begin: /./, end: /\)/,
-    //   // excludeBegin: true,
-    //   contains: [
-    //     ESCAPE_SEQUENCE,
-    //     CHARACTER_CLASS,
-    //     META,
-    //     COMMENT,
-    //     MODIFIER,
-    //     GROUP
-    //   ]
-    // }
-    contains: [
-      ESCAPE_SEQUENCE,
-      CHARACTER_CLASS,
-      META,
-      COMMENT,
-      MODIFIER,
-      'self'
-    ]
+  const GROUP_BEGIN = {
+    className: 'keyword',
+    begin: `\\((?:\\?(?:${GROUP_SPECIALS.join('|')}))?`
+  };
+
+  const GROUP_END = {
+    className: 'keyword',
+    begin: /\)/
   };
 
   return {
@@ -101,7 +85,8 @@ export default function (hljs) {
       META,
       COMMENT,
       MODIFIER,
-      GROUP
+      GROUP_BEGIN,
+      GROUP_END
     ]
   };
 };
--- a/src/languages/regex.js
+++ b/src/languages/regex.js
@@ -45,7 +45,7 @@ export default function (hljs) {
     begin: /[.+*?|^$]|\{[^}]+\}/,
   };
 
-  const COMMENT = hljs.COMMENT(/\(#/, /\)/);
+  const COMMENT = hljs.COMMENT(/\(\?#/, /\)/);
 
   const MODIFIERS = '-?(?:[ictsmnpdJUbqX^]|xx?)';
 
--- a/src/languages/regex.js
+++ b/src/languages/regex.js
@@ -67,7 +67,15 @@ export default function (hljs) {
 
   const GROUP_BEGIN = {
     className: 'keyword',
-    begin: `\\((?:\\?(?:${GROUP_SPECIALS.join('|')}))?`
+    begin: /\(/,
+    starts: {
+      contains: [
+        {
+          className: 'meta',
+          begin: `\\?(?:${GROUP_SPECIALS.join('|')})`
+        }
+      ]
+    }
   };
 
   const GROUP_END = {
--- a/src/languages/regex.js
+++ b/src/languages/regex.js
@@ -78,6 +78,7 @@ export default function (hljs) {
   return {
     name: 'regex',
     aliases: ['re', 'regexp'],
+    disableAutodetect: true,
 
     contains: [
       ESCAPE_SEQUENCE,
--- a/src/languages/regex.js
+++ b/src/languages/regex.js
@@ -42,7 +42,7 @@ export default function (hljs) {
 
   const META = {
     className: 'keyword',
-    begin: /[.+*?|^$]|\{[^}]+\}/,
+    begin: /[.+*?|^$]|\{(?:\d+(?:,\d*)?|,\d+)\}/,
   };
 
   const COMMENT = hljs.COMMENT(/\(\?#/, /\)/);
--- a/src/languages/regex.js
+++ b/src/languages/regex.js
@@ -1,0 +1,107 @@
+/*
+Language: regex
+Description: Regular expressions 
+Author: Konrad Rudolph <konrad.rudolph@gmail.com>
+*/
+
+export default function (hljs) {
+  const ESCAPE_SEQUENCE = {
+    className: 'built_in',
+    begin: [
+      // Unicode script/block/category
+      /\\[pP](?:[^{]|\{\^?[^}]+\})/,
+      // Unicode character
+      /\\[ux](?:[0-9a-fA-F]{1,4}|\{[0-9a-fA-F]{1,4}\})/,
+      // Backreferences
+      /\\k(?:'-?\d+'|<-?\d+>|{-?\d+})/,
+      /\\g(?:'-?\d+'|<-?\d+>|{-?\d+}|-?\d+)/,
+      // Everything else
+      /\\./
+    ].map(r => r.source).join('|')
+  };
+
+  const CHARACTER_CLASS = {
+    className: 'string',
+    begin: /\[\^?]?/,
+    end: /]/,
+    contains: [
+      {
+        className: 'literal',
+        variants: [
+          // POSIX named character class.
+          { begin: /\[:\^?/, end: /:]/ },
+          // POSIX collation sequences.
+          { begin: /\[\./, end: /\.]/ },
+          // POSIX character equivalence.
+          { begin: /\[=/, end: /=]/ }
+        ]
+      },
+      ESCAPE_SEQUENCE
+    ]
+  };
+
+  const META = {
+    className: 'keyword',
+    begin: /[.+*?|^$]|\{[^}]+\}/,
+  };
+
+  const COMMENT = hljs.COMMENT(/\(#/, /\)/);
+
+  const MODIFIERS = '-?(?:[ictsmnpdJUbqX^]|xx?)';
+
+  const GROUP_SPECIALS = [
+    // Named group
+    'P?<[^>]+>',
+    "'['>'+'",
+    // Zero-length assertions
+    '[|>=!]|<=|<!|',
+    // Modifiers applied inside group
+    `${MODIFIERS}:`,
+    ':'
+  ];
+
+  const MODIFIER = {
+    className: 'meta',
+    begin: `\\(\\?(?:${MODIFIERS})\\)`
+  };
+
+  let GROUP = {};
+  GROUP = {
+    className: 'group',
+    begin: `\\((?:\\?(?:${GROUP_SPECIALS.join('|')}))?`, end: /\)/,
+    // starts: {
+    //   // begin: /./, end: /\)/,
+    //   // excludeBegin: true,
+    //   contains: [
+    //     ESCAPE_SEQUENCE,
+    //     CHARACTER_CLASS,
+    //     META,
+    //     COMMENT,
+    //     MODIFIER,
+    //     GROUP
+    //   ]
+    // }
+    contains: [
+      ESCAPE_SEQUENCE,
+      CHARACTER_CLASS,
+      META,
+      COMMENT,
+      MODIFIER,
+      'self'
+    ]
+  };
+
+  return {
+    name: 'regex',
+    aliases: ['re', 'regexp'],
+
+    contains: [
+      ESCAPE_SEQUENCE,
+      CHARACTER_CLASS,
+      META,
+      COMMENT,
+      MODIFIER,
+      GROUP
+    ]
+  };
+};


src/languages/regex.js
```diff
@@ -65,30 +65,14 @@ export default function (hljs) {
...
-    //   ]
-    // }
-    contains: [
-      ESCAPE_SEQUENCE,
-      CHARACTER_CLASS,
-      META,
-      COMMENT,
-      MODIFIER,
-      'self'
-    ]
```
klmr: Actually we probably don’t need to match nesting groups. Instead, we just need to match the beginning and end of groups. If the group delimiters don’t match, so what?

joshgoebel:
> #2751 discusses adding more nuanced regex highlighting as a nested mode to existing languages.  By contrast, this PR concerns highlighting regex as a standalone language (though of course it might be worth investigating how to enable it as a nested mode).

This was quite intentional. :-)  As a general rule we don't merge new grammars at the present time because of lack of maintainer time to review and maintain them properly going forward.  So I was far more interested in what improving our regex support inside of other languages might look like vs defining a whole new grammar.

So I think for this to work as a grammar it would ALSO need to be a submode used to improve existing grammars, making it more of an extension to all our languages than a stand-alone new grammar.  And the auto-detect seems problematic (and slow), so that would likely simply be disabled.  Sadly I think this "dual usage" leads to problems of its own (with CSS naming) - but perhaps those are solvable.

> What’s the canonical name vs. aliases? (At the moment: regex; aliases: re , regexp).

Long name, filename is canonical. So here `regex` as in `regex.js`.  So at the moment seems right to me.

> Which class names to use to highlight individual regex features?

For usage to improve existing grammars we need new subclasses inside the `regex` namespace... ie "regex.escape" (`hljs-regex-escape`), etc... this requires manual (or programatic) alteration of existing styles... and it's likely (for many themes) those colors would be related to the color of the regex in general... I think this would tend to be at cross purposes to how you might visually want to highlight the regex as part of a stand-alone grammar - if there are NO other program elements and could just focus on having the full range of palette available for just the regex itself... I think those two highlighting modes would be very different. (for starters you could argue if they should even use `regex.escape` vs `subst` or some of the more normal classes)

Of course even if we handled only the "submode" case well it's possible someone like SO could (with a VERY small amount of code/CSS) provide their own workable "stand-alone" use case - but unless we could find a way to consolidate the all CSS issues that might mean that we never ship a "regex" language with the core language itself but only a REGEX_MODE.

Anyways to sum up the submode is the goal.  Another grammar would be just a side benefit if all the issues could be sorted out. And even as a submode I think we may need some custom support added to the library.  I'm worried about auto-detect performance when now every stray `/` could lead to a whole other set of expensive regex... so to me it seems optimal that the regex submode would be special in that it would NOT apply when doing auto-detect itself... but only AFTER the language was determined in perhaps a second pass... although depending on the length of the code that might be even worse.

You've started out here a bit more ambitious than I was hoping. :-)

> How extensive should the highlighting be anyway?

This would seem to also hinge on stand-alone vs submode, wouldn't it?  I think for just embedded use than even highligting just 2-4 common things would be SUPER great...  where-as if this was a stand-alone language (with a full color palette available in a theme) it'd be hard to know when you might stop...

> What regex dialects do we support?

Ugh. :-)  At this point I'm inclined to say "keep it minimal"... but again all the above questions might also have an impact here as well.

> How to implement groups such that parentheses, as well as modifier sequences (?… directly after the opening parenthesis) are somehow highlighted? My initial attempt via starts didn’t work (commented out code).

I think you'd match the whole sequence as a group and then figure it out with submodes if possible so matchin `(?=` would be separate from matching `(`, but that's just off the top of my head.

> Should character class ranges (e.g. [0-9A-F]) be highlighted specially?

Dunno.

joshgoebel:
One of the easiest programatic changes we can make to existing CSS is brightness and hue shifts... so if we were to create a few new CSS classes in such fashion that might argue for a very minimal amount of initial classes... ones we could easily auto-generate... or at least the high impact ones (escapes?) got auto-generated styles and then more esoteric things could be added by theme designers as needed but otherwise would default to just the plain jain `regexp` highlighting style.

joshgoebel:
May latest thoughts here are how I'm thinking about expanded the CSS and themes:

https://github.com/highlightjs/highlight.js/issues/2521

src/languages/regex.js
```diff
@@ -0,0 +1,101 @@
+/*
+Language: regex
+Description: Regular expressions 
+Author: Konrad Rudolph <konrad.rudolph@gmail.com>
+*/
+
+export default function (hljs) {
+  const ESCAPE_SEQUENCE = {
+    className: 'built_in',
```
joshgoebel, line 9: Seems clear this would be `regex.escape`?

src/languages/regex.js
```diff
@@ -0,0 +1,101 @@
...
+      // Backreferences
+      /\\k(?:'-?\d+'|<-?\d+>|{-?\d+})/,
+      /\\g(?:'-?\d+'|<-?\d+>|{-?\d+}|-?\d+)/,
+      // Everything else
+      /\\./
+    ].map(r => r.source).join('|')
+  };
+
+  const CHARACTER_CLASS = {
+    className: 'string',
```
joshgoebel, line 24: `regex.character_class`? (I wonder if we should care about abbreviations with modern gzip and the fact that most web traffic today should be at least gzipped...

src/languages/regex.js
```diff
@@ -0,0 +1,101 @@
...
+    begin: /\[\^?]?/,
+    end: /]/,
+    contains: [
+      {
+        className: 'literal',
+        variants: [
+          // POSIX named character class.
+          { begin: /\[:\^?/, end: /:]/ },
```
joshgoebel, line 32: The question would be if something like this would get different coloring so:

```
[a-z]
[:letter]
``

The :letter is obviously quite different in context.

File: src/languages/regex.js [hunk 5]
```diff
@@ -0,0 +1,101 @@
...
+      ESCAPE_SEQUENCE
+    ]
+  };
+
+  const META = {
+    className: 'keyword',
+    begin: /[.+*?|^$]|\{(?:\d+(?:,\d*)?|,\d+)\}/,
+  };
+
+  const COMMENT = hljs.COMMENT(/\(\?#/, /\)/);
```
joshgoebel, line 48: Regex can have comments? LOL...  now we have a clear conflict with `regex.comment` vs `comment`. :-)

src/languages/regex.js
```diff
@@ -0,0 +1,101 @@
...
+    // Zero-length assertions, branch reset & atomic
+    '[|>=!]|<=|<!',
+    // Modifiers applied inside group
+    `(?:${MODIFIERS}):`,
+    ':'
+  ];
+
+  const MODIFIER = {
+    className: 'meta',
+    begin: `\\(\\?(?:${MODIFIERS})\\)`
```
joshgoebel, line 65: Not sure I know what is happening here... example? If they are literally modifiers then `regex.modifier` doesn't seem so terrible...

src/languages/regex.js
```diff
@@ -0,0 +1,101 @@
...
+  const META = {
+    className: 'keyword',
+    begin: /[.+*?|^$]|\{(?:\d+(?:,\d*)?|,\d+)\}/,
+  };
+
+  const COMMENT = hljs.COMMENT(/\(\?#/, /\)/);
+
+  const MODIFIERS = '-?(?:[ictsmnpdJUbqX^]|xx?)';
+
+  const GROUP_SPECIALS = [
```
joshgoebel, line 52: And again not sure I'm familiar here. :-/

joshgoebel: Oh I see now.  Just group types, although I'm not familiar with some of these.

src/languages/regex.js
```diff
@@ -0,0 +1,101 @@
...
+  ];
+
+  const MODIFIER = {
+    className: 'meta',
+    begin: `\\(\\?(?:${MODIFIERS})\\)`
+  };
+
+  const GROUP_BEGIN = {
+    className: 'keyword',
+    begin: /\(/,
```
joshgoebel, line 70: `regex.group`?

src/languages/regex.js
klmr: Note that named POSIX classes are nested: `[a-zA-Z]` (roughly) corresponds to `[[:alpha:]]`, and `[a-zA-Z0-9]` would be `[[:alpha:][:digit:]]` (or `[[:alnum:]]`)). But really, the closer correspondence is between `[[:digit:]]` and `\d`; i.e. shorthand character classes. Doesn’t mean we have to highlight them identically, or even that we have to treat them as special anyway.

src/languages/regex.js
klmr: Yeah, `regex.modifier` would be apt. I was going to add explanations via test cases but I haven’t been doing this systematically yet. Here’s an example:

For example, `((?i)foo)bar` matches `foo` in a case-insensitive manner, followed by `bar` case-sensitively. It’s the equivalent of regular modifiers (`/foobar/i`), except applied to a subexpression only.

src/languages/regex.js
klmr: Going back to the example of `((?i)foo)bar`, an alternative way of writing this in some regex dialects is `(?i:foo)bar`. That’s what we capture here.

src/languages/regex.js
joshgoebel: OMG, does that local `i` mod work in JS regex? :-)

joshgoebel:
I'd suggest we sort thru some of the logistical issues here first before you added a bunch of tests (or effort).  Although I suppose you could always publish this as a 3rd party grammar in any case - in which case the tests would work just fine.  But if this doesn't land as a full-blown grammar then the whole testing framework we have now for full-blown grammars won't apply.

klmr:
> This was quite intentional. :-) As a general rule we don't merge new grammars at the present time because of lack of maintainer time to review and maintain them properly going forward. So I was far more interested in what improving our regex support inside of other languages might look like vs defining a whole new grammar.

Ah, hm. I actually saw the other issue only after I had already started on the PR due to the feature regression on Stack Overflow. I’m fine with maintaining this language definition as an independent project if there’s no interest in merging it for now (though I doubt Stack Overflow will use it then). If the PR somehow helps with creating a “regex submode”, even better.

That said, I’d also be happy with a combination submode/independent language if that’s somehow possible.

> And the auto-detect seems problematic (and slow), so that would likely simply be disabled.

Yeah I don’t see a way of making auto-detect work at all here. Neither as an independent language nor as a submode.

> > Which class names to use to highlight individual regex features?
> 
> For usage to improve existing grammars we need new subclasses inside the `regex` namespace... ie "regex.escape"

Given that regex is sufficiently dissimilar from “regular” programming languages, wouldn’t namespaced CSS names also make sense for a top-level language definition? I’d actually prefer this to abusing inappropriate existing CSS classes.

> You've started out here a bit more ambitious than I was hoping. :-)

We could definitely simplify the definition. A comprehensive implementation just seemed easy enough. I don’t think the current form is very complex.

> > How to implement groups such that parentheses, as well as modifier sequences (?… directly after the opening parenthesis) are somehow highlighted? My initial attempt via starts didn’t work (commented out code).
> 
> I think you'd match the whole sequence as a group and then figure it out with submodes if possible so matchin `(?=` would be separate from matching `(`, but that's just off the top of my head.

That question actually predated the current version of my code. I no longer think matching the whole group really helps. Instead, my current version matches the beginning and end of a group in separate modes, and the begin mode starts a mode to recognise `?…` sequences (lookaround, modifiers etc.).

src/languages/regex.js
klmr: Doesn’t look like it. I usually refer to the (incredibly broken) www.regular-expressions.info. It has the most comprehensive cross-language reference I’m aware of, and also lists (albeit now outdated) details about the scope of each implementation, e.g. [for JavaScript](https://www.regular-expressions.info/javascript.html).

joshgoebel:
> That said, I’d also be happy with a combination submode/independent language if that’s somehow possible.

Provided we can sort out the goal posts being in different places for each... this would be a nice way to get regex "back"... though they depending on how we packaged it they could till decide not to build it.  I haven't got a feel for their decisions in that regard yet - though it's a fair point they had it before... but perhaps as time goes back that may mean nothing regarding adding it back.

> nor as a submode.

Well as a submode (or just regular mode) it could "weigh in" like all modes do on relevancy... but then I think the relevance would need to be in fractional units... since to this point a "regex" has been treated a single unit... earning a score of 1... so suddenly if a regex opened a new mode and started adding all sorts of it's own relevancy it's quickly overwhelm other grammars... this same problem exists with a separate grammar.

> Given that regex is sufficiently dissimilar from “regular” programming languages, wouldn’t namespaced CSS names also make sense for a top-level language definition? I’d actually prefer this to abusing inappropriate existing CSS classes.

Well, we wouldn't abuse them, we'd alias them (so we'd need a new system for that)... so you'd specify maybe "regex.escape" but then we'd have a "escape" => "string" mapping table in the language for it to communicate how to best work with other themes.  The problem becomes one of context and themes for example INSIDE of C++ the `regex.escape` makes sense... but inside of a "Regular Expression" grammar "regex" is wholly redundant... and if you say the semantics are the same, so just ignore the redundancy then you still have a theming problem.

How do themes specify the styling for regex as a stand-alone grammar?  Now you end up with:

```css
/* when used in any other language */
.regex-escape

/* when used inside the regex specific grammar */
.regex .regex-escape
```

This is not a burden we want to force on themes before (although very few have still chosen to do it to add nuance to different languages).  This is far more likely because of poor decisions we'd made about classification rather than the fact the a theme truly wants a keyword in C++ to be a different color than a keyword in say Pascal.

> We could definitely simplify the definition. A comprehensive implementation just seemed easy enough. I don’t think the current form is very complex.

On it's own no. :-) But vs our ~10 line definition (see `REGEXP_MODE`) of what a regex is currently, yes. :-)


> That question actually predated the current version of my code. I no longer think matching the whole group really helps. Instead, my current version matches the beginning and end of a group in separate modes, and the begin mode starts a mode to recognise ?… sequences (lookaround, modifiers etc.).

Yes, For `[]` I think you need to know you're inside a group (because things can have contextual meaning?) but for `()` I don't think that's the case... so yes I'd just match the begin an end as entirely different things.

joshgoebel:
So I think next steps to first enhancing our existing regexp support inside our existing languages:

- Move the regex stuff into src/languages/lib/regexp.js and refactor slightly to export the main mode
- import it into modes.js and use it inside `REGEXP_MODE` 
- decide on the 3-5 most important subcomponents of regex
- auto-generation of CSS for existing themes for the new subcomponents of regex

Are you interested in working on the auto-generation of CSS piece as well, or just the grammar pieces?

---

This would be the simplest possible thing we could do (I think) to get the new regex support everywhere at once, so I'm inclined to try it first. :-)

src/languages/regex.js
joshgoebel: Yeah, I couldn't get it working now trying both syntaxes...

joshgoebel:
My vision for submodes:

```
export const REGEXP_MODE = {
  begin: /(?=\/[^/\n]*\/)/, // make sure we have a whole regex
  contains: [{
    className: 'regexp',
    begin: /\//,
    end: /\/[gimuy]*/,
    illegal: /\n/,
    submode: regex.submode // imported from the regex module
  }]
};
```

Submode essentially takes the place of "contains" (it's own contains will be used).  Where `regex.submode` is essentially a "language like" object.  Submodes would be compiled (once) independently the first time a language references them but then a copy of the top mode would be made and "grafted" into place in the currently language. By that I mean that what is happening is you're really "merging" two modes (the mode that is submode and the mode it's being embedded into - because you're using some pieces of each).  This is necessary because the "multi-matcher regex" for the submode will need to be recompiled with the embedded context in mind... ie the submode matcher is going to look like when compiled:

```
[ ...submode.contains, embedded_context.end_re, embedded_context.illegal_re]
```

The name "submode" may need to be rethought, it's close to `merge_mode` in all honesty the way I'm thinking about it lately. *sighs* Although now I think we still have the problem of case in-sensitive vs not-insensitive... because the multi-matcher regex  might have to mix  both modes... the submode being say sensitive (as regex is) but the embedded context being case in-sensative...  This would be a good time for those per group params that some regex implementations support. :-/

joshgoebel:
I think this needs more thought... I forgot about languages where you can hop in and out of the sublanguage, like ruby where you can do variable substitution inside regex:

```rb
irb(main):001:0> name = "bob"
irb(main):002:0> re = /\s#{name}/
irb(main):003:0> re
=> /\sbob/
```

joshgoebel:
@egor-rogov @allejo I'm not sure how to proceed here... I tried to turn/focus this towards improving our internal regex support for all languages (which I initially though would be easier) and considered this as a "submode" rather than a sublanguage (to avoid the known issues with sub-languages - and give me some reason to focus on submodes)... but I think I'm stuck.  The possible case sensitive/insensitive border between languages makes mixing regex (at the border of the submode) a real problem that I can't see a reasonable fix for [it's mentioned above in more detail]. (without adding a lot more complexity to the MultiMatcher and potentially at a measurable cost of performance.)

So then this either needs to simply die, become a 3rd party module, or become an official sublanguage (also making it a new grammar).  Even if we choose the latter (it's a win for all languages in a way) we still have the issue of styling/theming... using it as a self-container grammar would necessitate different styling than using it as a sublanguage. IE, used alone I'd expect to see a wide range of theme colors come to into play for the various bits (as with most grammars) - but used in an existing theme I might expect it to much more closely following the theming of the actual `regex` style, with perhaps more subtle variation for the internal pieces, etc.  

Any thoughts here? (on any of the above)?

joshgoebel:
The one nice thing about using a sublanguage would be a nice border of "is it built in or not"... I wouldn't make it a requirement or anything... so if someone wanted enhanced regex highlighting, they'd build the library with `regexp` support... if not, they wouldn't... so that way it's an easy opt-out for someone... although I suppose this should also be possible with submodes, if we figured out the hard questions.  The big difference being whether you get a whole new "highlight context" (sublanguage) or whether you're living inside the existing context (submode).

michaelblyons:
> Notably, [Stack Overflow used to support syntax highlighting of regular expressions, specifically](https://meta.stackexchange.com/a/184109/1968) but now no longer does.

While I am enthusiastic about _this_ effort to parse regular expressions, the old one you refer to [didn't actually do anything interesting](https://meta.stackoverflow.com/a/374389/241211).

joshgoebel:
I don't really see a route forward here for morphing this into a submode, and after some thought it may simply be "too much" for a sublanguage.  If we just inserted this this randomly into Javascript (or any grammar with generic regex support) then now we're potentially highlighting POSIX character sets (and other things), despite those maybe not even existing in the parent language... perhaps that wouldn't come up super often, but this problem is avoided entirely with a MUCH smaller submode that is greatly reduced in scope to only try to handle really common things like brackets, quantifiers, etc... ie, regex 101 stuff. That alone would still provide a massive boost in readability.

Submodes (which would potentially allow you to customize the regex logic for the grammar it was being inserted into) might solve the scope problem, but this isn't possible with sublanguages - and I can't think of any easy path forward with submodes based on the case sensitive vs case insensitive border issue. And that's just one of the hard problems here... It seems there are simply too many hard problems to solve here at the moment.  

I seems such a reduced scope mode (to subtly improve individual grammars) is exactly the opposite of what you'd want if the goal was to make this a proper first-class grammar of it's own (which was the original goal of this PR)... these are two quite different problems - and likely require very different solutions.  I don't see a great way right now to make them a single workable solution.

@klmr Do you want to move forward with publishing this as a 3rd party grammar module?

joshgoebel:
> the old one you refer to didn't actually do anything interesting.

OMG, surely you're joking... is that really what they had before for regex?  Yeah, that's almost the same as saying they don't support it at all. :-)

klmr:
> Do you want to move forward with publishing this as a 3rd party grammar module?

Yeah, that’s what I will do. Unfortunately I’m in the middle of another project with a tight deadline that’s taking up all of my free time at the moment. Once that’s done I’ll be able to devote some time to publishing the regex language as a third-party module (and I’ll also get back to some of the other language definitions I’ve dabbled in).

PR #2773, opened 2020-10-19
add `nim` class for nim (update SUPPORTED_LANGUAGES.md to reflect current behaviour)

pietroppeter:
Nim as class name has been added some time ago, see: https://github.com/highlightjs/highlight.js/commit/0c99483e5bf7e033f906ae65df38ebed8e06d9f4

I guess SUPPORTED_LANGUAGES.md was not updated contextually, this fixes it.

### Changes

add nim as class for Nim language in SUPPORTED_LANGUAGES.md

### Checklist

I think this is just syncing documentation with current behaviour, so I would say none of the stuff below probably applies.

### DIFF
--- a/SUPPORTED_LANGUAGES.md
+++ b/SUPPORTED_LANGUAGES.md
@@ -122,7 +122,7 @@ Languages that listed a **Package** below are 3rd party languages and are not bu
 | NSIS                    | nsis                   |         |
 | Never                   | never                  | [highlightjs-never](https://github.com/never-lang/highlightjs-never) |
 | Nginx                   | nginx, nginxconf       |         |
-| Nim                     | nimrod                 |         |
+| Nim                     | nim, nimrod            |         |
 | Nix                     | nix                    |         |
 | Object Constraint Language | ocl                 | [highlightjs-ocl](https://github.com/nhomble/highlightjs-ocl)        |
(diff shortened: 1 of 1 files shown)


pietroppeter:
I also have a related question: would `hljs.registerLanguage('nim', ...` work to add support for nim language or should `hljs.registerLanguage('nimrod', ...` be used?

**Note**: this is what it would take to add support for nim in stackoverflow, see: https://meta.stackoverflow.com/questions/402194/add-support-to-stack-overflow-for-highlighting-nim-code

joshgoebel (approved):

joshgoebel:
> I also have a related question: would hljs.registerLanguage('nim', ... work to add support for nim language or should hljs.registerLanguage('nimrod', ... be used?

The name is actually irrelevance there, it's just a label that... when someone manually wants to highlight something later they need to use the same label... it's just a "mapping"... that said the convention is to use the same name as the file in the filesystem so `nim.js` turns into `registerLanguage('nim'` when using the default built tools.

In order for SO to support Nim they have to make a decision to add it to their build script - so far they seem hesitant to continue to add new languages as every language they add makes the site a small bit slower for users.

pietroppeter:
Thanks for the explanation, all clear!

And yes, I understand the constraint on supporting additional language highlighting  for SO.

Have a good day!

PR #2779, opened 2020-10-22
add unpkg cdn instructions.

### LINKED ISSUE #2774: [Proposal] add unpkg cdn
There is a cdn for everything on npm: [unpkg](https://unpkg.com/)

This are some [data](https://twitter.com/mjackson/status/1315718264035573761) from last month:

The past month, http://unpkg.com served
- 50 billion requests
- 2.4 billion unique users
- 550 TB of data
- 99.8% cache hit rate

Since highlight.js can be installed from npm or yarn.

I would like to add instructions to use the cdn in the readme.

luiscobits:
### Changes

add unpkg cdn instructions in readme.md
resolves #2774 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,8 +240,8 @@ A prebuilt version of highlight.js bundled with many common languages is hosted 
 **unpkg** ([link](https://unpkg.com/browse/highlight.js/))
 
 ```html
-<link rel="stylesheet" href="https://unpkg.com/highlight.js/styles/default.css">
-<script src="https://unpkg.com/highlight.js/lib/core.js"></script>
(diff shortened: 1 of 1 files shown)


joshgoebel:
Hmmm... Ok this is problematic because the core NPM package is not intended to be used via CDN... `core` does not include the common stack, etc...

You need to reference our CDN npm package instead... (which will have the traditional `highlight.min.js` file... and we really should show pinning the current version in the link - like we do with both the other CDN examples.

luiscobits:
Thanks for the guidance, and thank you for the link to the cdn assets.
I will change the code, and push a new commit.

joshgoebel (approved):

PR #2781, opened 2020-10-23
[julia] Update keywords, built_in and literals for julia 1.X.

fredrikekre:
### Changes

This updates the lists of `keyword`, `built_in` and `literal` for Julia 1.X. Julia 1.X was released over 2 years ago so (IMO) it should be safe to remove the old tokens that are not applicable any more, but I don't know what your policy is for such things.

### Checklist
- [x] Tests: I did not find any tests for Julia syntax and adding this is beyond my capabilities (Edit: found this https://github.com/highlightjs/highlight.js/blob/master/test/detect/julia/default.txt so I will update that).

### DIFF
--- a/src/languages/julia.js
+++ b/src/languages/julia.js
@@ -2,87 +2,330 @@
 Language: Julia
 Description: Julia is a high-level, high-performance, dynamic programming language.
 Author: Kenta Sato <bicycle1885@gmail.com>
-Contributors: Alex Arslan <ararslan@comcast.net>
+Contributors: Alex Arslan <ararslan@comcast.net>, Fredrik Ekre <ekrefredrik@gmail.com>
 Website: https://julialang.org
 */
 
 export default function(hljs) {
   // Since there are numerous special names in Julia, it is too much trouble
   // to maintain them by hand. Hence these names (i.e. keywords, literals and
-  // built-ins) are automatically generated from Julia v0.6 itself through
+  // built-ins) are automatically generated from Julia 1.5.2 itself through
   // the following scripts for each.
 
-  // ref: http://julia.readthedocs.org/en/latest/manual/variables/#allowed-variable-names
+  // ref: https://docs.julialang.org/en/v1/manual/variables/#Allowed-Variable-Names
   var VARIABLE_NAME_RE = '[A-Za-z_\\u00A1-\\uFFFF][A-Za-z_0-9\\u00A1-\\uFFFF]*';
+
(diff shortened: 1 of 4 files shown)


fredrikekre:
CC @mortenpi @tlienart, you might be interested in this.

joshgoebel:
Thanks for this!  While we're at can we go ahead and move the words into one-word per line arrays to include the visibility and ease of future maintenance?  IE:

```
LITERALS = [
// one word per line
]
```

```
keywords : {
  literal: LITERALS.join(" "),
  // ...
```

You can see examples with javascript.js and ecmascript.js...

joshgoebel:
>  it should be safe to remove the old tokens that are not applicable any more, but I don't know what your policy is for such things.

We'd usually leave them if there is a chance they're still in use in the world or old code snippets but since these lists are auto-generated I don't mind if we drop the older keywords since as you say it's been quite some time.

fredrikekre:
Thanks, I updated the PR.

test/detect/julia/default.txt
```diff
@@ -1,21 +1,5 @@
 ### Types
 
-# Old-style definitions
-
```
joshgoebel, line 4: If we're going to cut all this we need to add in 5-10 lines of code that uses some of the new keywords as well.

src/languages/julia.js
```diff
@@ -2,95 +2,328 @@
...
+    'PROGRAM_FILE',
+    'QuickSort',
+    'RoundDown',
+    'RoundFromZero',
+    'RoundNearest',
+    'RoundNearestTiesAway',
+    'RoundNearestTiesUp',
+    'RoundToZero',
+    'RoundUp',
+    'VERSION',
```
joshgoebel, line 97:
```suggestion
    'VERSION|0',
```

joshgoebel: Sees like this could be found in almost anything, to lowering it's relevance.

joshgoebel:
Awesome.  Almost there! :-)

src/languages/julia.js
fredrikekre: Alright, this is for autodetection? Isn't this needed for e.g. `for` which exists in almost any language?

fredrikekre:
Added some new stuff to `test/detect`, and also noticed that some duplicates where included by my scripts, so fixed that too.

joshgoebel (approved):

joshgoebel:
@fredrikekre Thanks so much for contributing!

PR #2784, opened 2020-10-24
Fix CSS Function-Like Bug

taufik-nurrohman:

### Checklist
- [ ] Tested.

### DIFF
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -19,6 +19,7 @@ export default function(hljs) {
           hljs.APOS_STRING_MODE,
           hljs.QUOTE_STRING_MODE,
           hljs.CSS_NUMBER_MODE,
+          FUNCTION_LIKE,
         ]
       }
     ]
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -6,7 +6,7 @@ Website: https://developer.mozilla.org/en-US/docs/Web/CSS
 
 /** @type LanguageFn */
 export default function(hljs) {
-  var FUNCTION_LIKE = {
+  let FUNCTION_LIKE = {
     begin: /[\w-]+\(/, returnBegin: true,
     contains: [
       {
@@ -24,7 +24,7 @@ export default function(hljs) {
       }
     ]
   }
-  var ATTRIBUTE = {
+  let ATTRIBUTE = {
     className: 'attribute',
     begin: /\S/, end: ':', excludeEnd: true,
     starts: {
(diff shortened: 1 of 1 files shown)


joshgoebel:
This will need to be based on the css_consistency long-lived branch [no master] (and need to be confirmed to work with CSS, less, stylus, and scss as well.  See `tools/css`.  IE, you'd add a markup tests, run `tools/css` to copy the test to all the grammars that can support CSS literal code, and then confirm the fix for all 4.

taufik-nurrohman:
https://github.com/highlightjs/highlight.js/pull/2791

taufik-nurrohman:
> See `tools/css`.

👀 ?

> IE, you'd add a markup test, run `tools/css` to copy the test to all the grammars that can support CSS literal code, and then confirm the fix for all 4.

Ugh, that&rsquo;s out of my competency.

joshgoebel:
Just a utility you run from the command prompt:

```
./tools/css
```

> Ugh, that’s out of my competency.

Regarding the toolchain stuff is actually pretty simple (assuming you have Node.js setup, etc), perhaps give it a shot.

1. Update `test/markup/css/css_consistency.txt` and `test/markup/css/css_consistency.expect.txt` with a small example
2. Run `./tools/css` to copy the test files verbatim to the other CSS like languages
3. Build the node.js version of the library (so we can run tests) - `node ./tools/build.js -t node`
4. Run the markup tests `npm run test-markup`

That's the minimal stuff you need to do if you want to continue to work on CSS as for consistency purposes the entire CSS stack (css, scss, stylus, less ) now needs to be worked on as a unit (to ensure it remains consistent).

taufik-nurrohman:
Can&rsquo;t see `test/markup/css/css_consistency.txt` anywhere to be updated. Do I need to make it myself?

Can I name it `css-function.txt`?

joshgoebel:
>Can’t see test/markup/css/css_consistency.txt anywhere to be updated. 

First, you need to be working on the css_consistency branch. https://github.com/highlightjs/highlight.js/tree/css_consistency

```
% ls -l test/markup/css/*cons*                                                                 2.7.1
-rw-r--r--  1 jgoebel  staff  1705 Oct 27 04:36 test/markup/css/css_consistency.expect.txt
-rw-r--r--  1 jgoebel  staff   562 Oct 27 04:36 test/markup/css/css_consistency.txt
```

It's probably simplest to start with that branch and then copy/paste your changes over since a lot has changed.

> Can I name it css-function.txt?

No, there is purposely a single file now for all tests that should be "consistent" across all our CSS like grammars (most of which support raw CSS as they are supersets).  If that file became too large we could consider updating the `css` script and having multiple `consistency*` files (grouping features), but I don't think that's necessary just yet. It's still quite small.

taufik-nurrohman:
Sorry, haven&rsquo;t pull it yet. Now I see it.

joshgoebel:
FYI: I've since decided long-lived branches (for us) are a not so great idea (without a clear plan they are just likely to be forgotten and fall into disrepair)... so `css_consistency` will probably get merged into 10.4... so in the future hopefully this won't cause any further confusion. :-)

PR #2791, opened 2020-10-26
Fix CSS Function-Like Bug

taufik-nurrohman:

### Checklist
- [ ] Tested.

### DIFF
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -6,7 +6,7 @@ Website: https://developer.mozilla.org/en-US/docs/Web/CSS
 
 /** @type LanguageFn */
 export default function(hljs) {
-  var FUNCTION_LIKE = {
+  const FUNCTION_LIKE = {
     begin: /[\w-]+\(/, returnBegin: true,
     contains: [
       {
@@ -19,11 +19,12 @@ export default function(hljs) {
           hljs.APOS_STRING_MODE,
           hljs.QUOTE_STRING_MODE,
           hljs.CSS_NUMBER_MODE,
+          FUNCTION_LIKE
         ]
       }
     ]
   }
-  var ATTRIBUTE = {
+  const ATTRIBUTE = {
     className: 'attribute',
     begin: /\S/, end: ':', excludeEnd: true,
     starts: {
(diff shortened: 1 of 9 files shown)


joshgoebel:
This seems misnamed... so far all it does is update the media types and some linting?

taufik-nurrohman:
My bad. Just mistakenly dropped the first commit via rebase.

joshgoebel:
Rebase be a powerful tool, like a lightsaber, but ya gotta be careful. :)

test/markup/css/css_consistency.txt
```diff
@@ -20,3 +20,13 @@ a:visited { color: blue; }
...
+
+div {
+  color: rgb(var(--r), var(--g), var(--b));
+  background: linear-gradient( var(--black), var(--white) );
+  border-color: rgb(
+    var(--r),
+    var(--g),
+    var(--b)
+  );
+}
```
joshgoebel, line 32: Of course you still need to update the expected output (the `expect` file) as well as running `./tools/css` to copy the tests to all 3 of the other CSS like grammars - which may or may not need their own changes to pass.

taufik-nurrohman: Don&rsquo;t worry. I&rsquo;m on my way.

joshgoebel: Ah ok. :)

taufik-nurrohman: IMG_20201027_172117

joshgoebel: It's not a node executable... it's bash... you just run it directly. (sorry if I said otherwise before, can't recall)

taufik-nurrohman: Done.

src/languages/css.js
```diff
@@ -21,14 +21,15 @@ export default function(hljs) {
           hljs.APOS_STRING_MODE,
           hljs.QUOTE_STRING_MODE,
           hljs.CSS_NUMBER_MODE,
+          FUNCTION_LIKE
```
joshgoebel, line 24: This can't work because FUNCTION_LIKE doesn't exist yet, not until after it's created - it can't refer to itself.  So you first have to create it, then add it to the inside `contains`... but I'm wondering why this is necessary and if we couldn't just collapse the rule and remove `returnBegin`... then you'd just use `contains: "self"` as the rule would self-nest.

It seems wrong to match EVERY possible function as a built_in (feel like we've had this discussion before)... if we just matched actual built-ins you could do this with `beginKeyword` instead... and simplify as I mentioned above.  If not then as I said you'll need for first create the mode, then modify it to reference itself.

joshgoebel: IE, something like:

```js
const MODE = { ... } // create it
MODE.contains[1].contains.push(MODE)  // modify it to point to itself
```

It's definitely worth considering if we could simplify the rule though.

joshgoebel: Another question worth asking is whether we 100% need to know we're inside a function.. if not you may not need the nesting at all... just match `var(`, color var... match `foo(`, color foo... There might be no need to open a mode, look for the final `)`, etc...

taufik-nurrohman: > just match `var(`, color `var`... match `foo(`, color `foo`, etc...

Looks simpler this way 👍

joshgoebel:
Closing this is favor of the other open PR.  Thanks for your help on this!

PR #2792, opened 2020-10-26
enh(javascript) Added `node-repl` for Node.js REPL sessions

nagayev:
Issue: none
### Changes

Add javascript-repl language with tests.

### Checklist
- [X] Added myself to `AUTHORS.txt`, under Contributors


src/languages/javascript-repl.js
```diff
@@ -0,0 +1,30 @@
+/*
+Language: Python REPL
```
joshgoebel, line 2: Is it now? :)

test/markup/javascript-repl/sample.txt
```diff
@@ -0,0 +1,6 @@
+> a=1
+1
+> b=2
+2
+> 
+>
```
joshgoebel, line 6: This needs a matching `test/markup/javascript-repl/sample.expect.txt` file... you can run the markup tests after you've built the library with:

```
npm run build && npm run test-markup
```

src/languages/javascript-repl.js
```diff
@@ -0,0 +1,30 @@
...
+Requires: javascript.js
+Author: Marat Nagayev <nagaevmt@yandex.ru>
+Category: common
+*/
+
+export default function(hljs) {
+    return {
+      name: 'Javascript REPL',
```
joshgoebel, line 10:
```suggestion
      name: 'JavaScript REPL',
```

Proper casing.

test/markup/javascript-repl/sample.txt
joshgoebel: And please add at least one `...` line as well.

joshgoebel (changes requested):
Awesome!  Almost good to go.

joshgoebel:
I'm fine with adding a `repl` grammar in general since we did this recently with Python also and it's my opinion that the repl's are better broken out than being mixed with the main languages (though we still have some mixed).

Though I wonder if the REPL isn't really `node-repl`... we have Deno also, but that's Typescript?  Also is there a SpiderMonkey JS run-time repl also?  @egor-rogov @allejo Any thoughts here?  Would be nice to avoid the need for two repls (one for JS and one for TS)...  So just wondering about two things:

- The exact name for this grammar.
- JS/TS support/Deno support, etc...
- Other JS runtime REPLS... (SpiderMonkey, etc.)

While we're here might as well try and think about the bigger picture.

src/languages/javascript-repl.js
nagayev: haha, copypaste

nagayev:
@joshgoebel Is it okey?
sample.txt
```
> for(let i=0;i<5;i++){
... console.log(i);
... }
0
1
2
3
4
undefined
> 

```

nagayev:
@joshgoebel I will fix CI build tommorow.

nagayev:
@joshgoebel How can I generate correct sample.exprect.txt?

joshgoebel:
> How can I generate correct sample.exprect.txt?

Well technically your'e supposed to write it by hand :-) ...but if you're very careful there is a commented line in `test/markup/index.js` that will overwrite ALL the tests with what the library THINKS the output should be.

nagayev:
@joshgoebel tkanks!

joshgoebel:
Not sure you got that merge/rebase correct...

nagayev:
@joshgoebel Why?
I just copypasted AUTHORS and CHANGES from master and inserted myself.

joshgoebel:
I see a bunch of php/ruby stuff now that has nothing to do with this PR.

nagayev:
@joshgoebel Please check my final version.

joshgoebel:
Looks nice.  My only concern is still the naming (there is really not general JS repl, it's the node repl), so I'm curious if anyone will chime in with any thoughts there.

src/languages/javascript-repl.js
```diff
@@ -0,0 +1,30 @@
...
+Language: JavaScript REPL
+Requires: javascript.js
+Author: Marat Nagayev <nagaevmt@yandex.ru>
+Category: common
+*/
+
+export default function(hljs) {
+    return {
+      name: 'JavaScript REPL',
+      contains: [
```
joshgoebel, line 11:
```suggestion
      name: 'JavaScript REPL',
      aliases: ["js-repl"],
      contains: [
```

IF we keep the name I think this would be desirable.

nagayev:
@joshgoebel 
Hmm.
I was thinking of this name, node-repl. So I can rename language and maybe no one thinks otherwise.

joshgoebel:
>I was thinking of this name, node-repl. So I can rename language and maybe no one thinks otherwise.

But I reallly do not want `node-repl`, `spidermonkey-repl`, etc... If there are other repl's I'd probably want to kill them with this same stone (if possible)... which makes more of an argument for `js-repl`. :-)  I'd just like to see if any of the other core team members have any thougths.

nagayev:
@joshgoebel No response from another members of core team...
So isn't it time to merge?
My PR regularly has conflicts because it has been open for quite a long time.
So I suggest merging this PR after resolving conflict with AUTHORS and, if necessary, changing the js-repl to a different name.

joshgoebel:
> My PR regularly has conflicts because it has been open for quite a long time.

@nagayev Not a worry anymore.  I can handle any final merge conflicts as it's just the changelog and authors.  Zero more effort required on your part.

I tagged this 10.4 so it will go into the next release.  There isn't a hurry to merge until then.  I've still been noodling on the name myself as well.

nagayev:
@joshgoebel All right

joshgoebel:
Since you were working on master you may want to hard reset your master to upstream/mater now that this is merged to make future contributions easier.  And I'd recommend branching first in the future to avoid the need for this step. :)

I also removed this from `common` as that is a different discussion.

Thanks so much!  This will go out in 10.4 (later this month prolly).

PR #2796, opened 2020-10-28
add support for nested functions within CSS

### LINKED ISSUE #2674: (css) Nesting Function-Like Bug
**Describe the issue**

`css`

`highlight`

**Sample Code to Reproduce**

~~~ .css
body {
  text-align: left;
  font-size: 16px;
  background: linear-gradient( var(--black), var(--white) );
  color: rgb(var(--r), var(--g), var(--b));
}
~~~

https://jsfiddle.net/f27o1gyt/

khamer:
Fixed nested CSS functions - like `var(--example, rgb(255, 255, 255))`, etc. Resolves #2674.

### Changes
This change, while it looks a bit awkward, allows CSS FUNCTION_LIKE patterns to be self referencing and contain themselves. Because it's nested a level, I wasn't able to find a way to use the 'self' keyword.

### Checklist
- [x] Added new markup test files custom-properties-and-functions.txt

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -293,3 +293,4 @@ Contributors:
 - Tom Wallace <thomasmichaelwallace@gmail.com>
 - Michael Newton <miken32@github>
 - Fredrik Ekre <ekrefredrik@gmail.com>
+- Kevin Hamer <kevin.hamer@gmail.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 Language Improvements:
 
 - enh(julia) Update keyword lists for Julia 1.X (#2781) [Fredrik Ekre][]
+- fix(css) Fix highlighting of nested css functions (#2674) [Kevin Hamer][]
 
 Dev Improvements:
 
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -25,6 +25,8 @@ export default function(hljs) {
       }
     ]
   };
+  FUNCTION_LIKE.contains[1].contains.push(FUNCTION_LIKE); // self-reference for nested functions
+
   var VENDOR_PREFIX= {
     begin: /-(webkit|moz|ms|o)-/
   };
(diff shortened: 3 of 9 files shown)


taufik-nurrohman:
@joshgoebel

khamer:
Hrm, looks like the failures above are related to how quotes are being encoded, and happens without my PR too. I can update the tests there too - unless that's a bug, and the quotes aren't supposed to be encoded?

joshgoebel:
The quotes should be encoded... that changed after this branch was split off... I rebased this onto master just the other day but perhaps I forgot to run the tests to make sure I didn't miss something like that.  So it's safe to update the tests to have encoded quotes.

joshgoebel:
A minimal test case for nested css functions needs to be added to the `css/css_consistency` markup test then cloned to all the CSS-like languages.  We need to make sure this works in all 4, not just CSS.

You can use the `./tools/css` script to update the tests after you change just the `css` test.

joshgoebel:
Ping. Anyone still here?

taufik-nurrohman:
Maybe you could just merge my changes with this?

joshgoebel:
They could - did the other grammars already work correctly with 0 changes?  I'm not certain that was ever established...

khamer:
Hey, I am - and no, it doesn't work with the other languages as-is, as I believe the other languages have existing ways of highlighting function-like things. I'll do a sweep today and see which of the languages I think actually need to have function support added from my PR, as I don't think they all do.

joshgoebel:
> as I believe the other languages have existing ways of highlighting function-like things

My understanding (and correct me if I'm wrong) is they are all super-sets of CSS meaning they might have other [newer] ways to write a function, but it's possible to still use the lower-level CSS way - and if so then we need to also highlight that (and do it exactly the same as CSS).

joshgoebel:
Closing this from inactivity until sometime has time to actually pursue it - at which time it can be re-opened.

PR #2797, opened 2020-10-28
Add Julia to SUPPORTED_LANGUAGES.md

fredrikekre:
According to https://meta.stackoverflow.com/questions/274371/what-is-syntax-highlighting-and-how-does-it-work StackOverflow uses the SUPPORTED_LANGUAGES list. Would be great to get Julia in there. Since there is Julia support (https://github.com/highlightjs/highlight.js/blob/master/src/languages/julia.js) I wonder if it is just an oversight that it is not included in this list or is there something else needed to be an officially supported language? Thanks.

### DIFF
--- a/SUPPORTED_LANGUAGES.md
+++ b/SUPPORTED_LANGUAGES.md
@@ -95,6 +95,7 @@ Languages that listed a **Package** below are 3rd party languages and are not bu
 | Java                    | java, jsp              |         |
 | JavaScript              | javascript, js, jsx    |         |
 | Jolie                   | jolie, iol, ol         | [highlightjs-jolie](https://github.com/xiroV/highlightjs-jolie) |
+| Julia                   | julia, julia-repl      |         |
 | Kotlin                  | kotlin, kt             |         |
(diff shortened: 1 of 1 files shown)


joshgoebel (approved):

joshgoebel:
>  I wonder if it is just an oversight that it is not included in this list or is there something else needed to be an officially supported language? 

Was Julia supported (correctly) with the previous highlighter on SO?

fredrikekre:
No I think it used some generic highlighter, so this will be a huge improvement.

joshgoebel:
You could create a question on `meta` asking for Julia support to be added but so far they seem hesitant to add more language because of the increasing bundle size.  I'm not sure we are going to make progress here until they switch to a strategy of lazy-loading the grammars as needed.  If you ask I'll upvote it.

fredrikekre:
Thanks, I asked here: https://meta.stackexchange.com/questions/355928/julia-syntax-highlighting

joshgoebel:
Yeah, I think i commented: https://github.com/joshgoebel/se_highlightjs/ :-)

PR #2803, opened 2020-10-30
fix(js/ts) Prevent while/if/switch from falsly matching as functions

joshgoebel:
When adding support for detecting what appear to be class functions:

```js
class Dog {
  bark(args) {
    loud = true
  }
}
```

...we seem to have missed that this also matches if, while, switch, etc... so those need to be coded as exceptions to the rule.

### Changes

Adds an exception mode to catch these first before the function rule matches them.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 Language Improvements:
 
-- fix(js/ts) Prevent while/if/switch from falsly matching as functions () [Josh Goebel][]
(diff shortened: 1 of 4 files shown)


allejo (changes requested):
- You're missing `for`
  image

- Do we want to account for the following?
  image

joshgoebel:
> Do we want to account for the following?

Is that even possible?  If so that is DUMB.  I don't think there is any way we can account for it.

allejo (approved):
should the changelog entry and unit tests be updated as well before this is merged? otherwise, this looks good to me

Issue #1993, opened 2019-02-26
Unnecessary dependencies installed with 9.15.x


Filipoliko:
Hi,
I ran into a problem with the latest version of highliht.js. You have moved all the devDependencies to dependencies. I checked the packages and dont see any reason why this should have been done, since all of them are used in tools, or tests only and they are not necessary dependencies to use the highlight.js.

This update caused us an error in build, when using gulp-less package to install less. For some reason your dependency gear-lib forced our project to use less version 1.7.5 (instead of 2.x || 3.x versions from gulp-less) and this made our build fail. As workaround we freezed the less version in our project, but it would be great if you moved these dependencies back to devDependencies.

Or is there any reason for this move?

Thanks for hearing me out and for your work!

mzyil:
I'm also getting async v0.8.0 installed as a dependency of gear-lib which breaks functionality of different modules. The broken module (botbuilder-unit) didn't have async as its dependency because it was a dependency of its dependency. Talk about bad practices...

marcoscaceres:
Ok, will get those moved to devDependencies.

realityking:
I appreciate the idea of making the build script accessible via the npm package. Have you considered turning the CLI into its own npm package? It would cut down on dependencies for those folks who don't need it.

I know time is scarce for this project, I'm happy to help get this set up.

marcoscaceres:
I've sent a PR for this https://github.com/highlightjs/highlight.js/pull/1995

@realityking, that's a good suggestion. However, I think it might be best to not bother with the builder as the js files can be imported into projects (allowing for other bundlers to do their thing)... I'd be more interested investing time into converting the code into ES6 modules.

PR #2808, opened 2020-10-31
(VB.Net) Date literals, Directives, Comments, test cases

### LINKED ISSUE #2775: (VB.Net) Doesn't highlight date literals e.g. #7/26/1970#
**Describe the issue**
Date literals are not highlighting correctly

VB.Net (and Stack Overflow uses VB.Net to highlight VBA and VBscript as well, which have the same syntax and so same problem)

Not using auto detection, using language tags
...

**Sample Code to Reproduce**
```vbnet
Sub TestDateLiteral()
    Dim Anne As Person, Bill As Person
    Set Anne = Person.Factory("Anne", "F")                  'No Date Literal
    Set Bill = Person.Factory(#7/26/1970#, "Bill", "M")     'With Date Literal
End Sub
```
https://jsfiddle.net/waz4fLxd/2/

### LINKED ISSUE #2851: (VB.Net) Incorrect highlighting of Rem Comments
**Describe the issue**
The legacy [Rem/REM statement](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/rem-statement) is an alternative to the `'` apostrophe literal for single line comments

Using `highlight` (via tag associations on SO)

**Sample Code to Reproduce**
https://jsfiddle.net/zv9k1hox/

**Expected behavior**
Rem is treated as a comment and doesn't affect highlighting of subsequent lines:
```vbnet
     Rem Tab to be replaced with \t
    If InStr(Content, vbTab) > 0 Then Content = Replace(Content, vbTab, "\t")
    
    Rem Double quote to be replaced with \"
    If InStr(Content, Chr(34)) > 0 Then Content = Replace(Content, Chr(34), "\" & Chr(34))
    Set EscapeJSON = Self
End Property

Public Function ToString(Optional ContentWrapper As String = "") As String
        ToString = ContentWrapper & this.Content & ContentWrapper
End Function

Public Property Get Content() As String
    Content = this.Content
End Property

Public Property Let Content(ByVal Value As String)
    this.Content = Value
End Property
```

**Additional context**
Triggered by [this](https://codereview.stackexchange.com/q/251995/146810) question on Code Review

### LINKED ISSUE #2875: (vb.net) Comments need to be reworked - `<` in comment breaks highlighting
**Describe the issue**
A simple comment including a smaller than sign `<` will make all following code lines formatted as comments.

The issue is with VB.NET (probably but not tested the same issue with vbscript).

Using highlight (via tag associations on SO)

**Sample Code to Reproduce**
```vb
'rule 1: column A allows only text length < 9
Set AffectedCells = Intersect(Target, Target.Parent.Range("A:A"))
```
as shown in this filddle: https://jsfiddle.net/pcrcoding/4bcxaot0/

**Expected behavior**
I expect lines that start with a `'` or `REM ` to handle the entire line until line break as coment and end comment formatting at the line break. The expected behaviour would be the one seen in this post at *"Sample Code to Reproduce"*

**Additional context**
Also check https://github.com/highlightjs/highlight.js/issues/2851 for the `REM `.

Hirse:
Resolves #2775, resolves #2875, resolves #2851

### Changes
- Adds support for [Date data types](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/data-types/date-data-type), see #2775
- Adds support for `REM`-comments and fixes `'''` doctags, see #2851 and #2875
- Switches from C number mode to custom number mode to support VB.net specifics
  - Hex (`&H`), Oct (`&O`), and binary (`&B`) prefixes
  - Separating digits with underscores: `90_946`
  - Type suffixes: `123UI` (unsigned integer)
- Improves [Directives](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/directives/) detection and adds support for `Enable`, `Disable`, and `Then` keywords
- Adds more markup test for vbnet
- Fixes [imports statements](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/imports-statement-net-namespace-and-type) in detect file

### Checklist
- [x] Added markup tests
- [x] ~Added myself to `AUTHORS.txt`, under Contributors~ Already there 😉


Hirse:
The Microsoft Docs site seems to use a fork of HLJS (v9) which adds support for date literals ... 
Microsoft Docs - VB.Net date literals

... but breaks Directives (by looking for the ending `#`).
Microsoft Docs - VB.Net Directives

Both of which should now work as expected.
HLJS VB.Net after fix

CHANGES.md
```diff
@@ -7,6 +7,7 @@ Language Improvements:
 - enh(ruby) Match numeric literals per language documentation [Richard Gibson][]
 - enh(javascript) Match numeric literals per ECMA-262 spec [Richard Gibson][]
 - enh(php) highlight variables (#2785) [Taufik Nurrohman][]
+- enh(vbnet) Highlight date literals (#2775) [Jan Pilzer][]
```
joshgoebel, line 10: You can expand this or add a few more lines, this PR does more than just this.

src/languages/vbnet.js
```diff
@@ -6,6 +6,26 @@ Website: https://docs.microsoft.com/en-us/dotnet/visual-basic/getting-started/
...
 export default function(hljs) {
+  const commentMode = hljs.COMMENT(
+    '\'',
+    '$',
+    {
+      returnBegin: true,
+      contains: [
+        {
+          className: 'doctag',
+          begin: '\'\'\'|<!--|-->',
```
joshgoebel, line 17: This is confusing, what are the actual variants here?  Do comments sometimes start with `'''` or do they always start with `'` and then later have a `'''` comment inside them, or both?  If possible it'd be nice to flatten this into a more typical `contains` style rule.  Barring that use a look-ahead (there are utils in `regex.js`) to do the match.. `returnBegin` is a bit out of favor now since look-aheads are simpler to use.

 Depending on where this goes doctag might be easier to read as a variant also.

joshgoebel: I see you just moved it, so if you have no idea what I'm talking about you can leave it as-is. :-)

src/languages/vbnet.js
```diff
@@ -6,6 +6,26 @@ Website: https://docs.microsoft.com/en-us/dotnet/visual-basic/getting-started/
 */
 
 export default function(hljs) {
+  const commentMode = hljs.COMMENT(
```
joshgoebel, line 9:
```suggestion
  const COMMENT = hljs.COMMENT(
```

src/languages/vbnet.js
```diff
@@ -33,30 +53,18 @@ export default function(hljs) {
...
-      ),
+      commentMode,
       hljs.C_NUMBER_MODE,
       {
         className: 'meta',
-        begin: '#', end: '$',
-        keywords: {'meta-keyword': 'if else elseif end region externalsource'}
+        begin: '#(?:const|disable|else|elseif|enable|end|externalsource|if|region)\\b', end: '$',
+        keywords: {'meta-keyword': 'const disable else elseif enable end externalsource if region'},
+        contains: [commentMode]
```
joshgoebel, line 62:
```suggestion
        contains: [COMMENT]
```

src/languages/vbnet.js
```diff
@@ -33,30 +53,18 @@ export default function(hljs) {
...
-          ]
-        }
-      ),
+      commentMode,
       hljs.C_NUMBER_MODE,
       {
         className: 'meta',
-        begin: '#', end: '$',
-        keywords: {'meta-keyword': 'if else elseif end region externalsource'}
+        begin: '#(?:const|disable|else|elseif|enable|end|externalsource|if|region)\\b', end: '$',
```
joshgoebel, line 60: There are the only possible tags?

You don't need the `?:`, doesn't matter if we capture or not.

src/languages/vbnet.js
```diff
@@ -33,30 +53,18 @@ export default function(hljs) {
...
+        keywords: {'meta-keyword': 'const disable else elseif enable end externalsource if region'},
+        contains: [commentMode]
+      },
+      {
+        className: 'literal',
+        begin: '#', end: '#',
```
joshgoebel, line 66: This is very broad, is there no format guidelines for the dates inside here?

src/languages/vbnet.js
joshgoebel: Should we also be matching the start of a line here anywhere?  Or can these occur in the middle of a line?

src/languages/vbnet.js
```diff
@@ -33,30 +53,18 @@ export default function(hljs) {
...
+        illegal: '\\n'
```
joshgoebel, line 67:
```suggestion
        illegal: /\n/
```

Prefer regex when you can use it.

test/detect/vbnet/default.txt
```diff
@@ -1,5 +1,5 @@
-Import System
-Import System.IO
+Imports System
+Imports System.IO
```
joshgoebel, line 2: Is this just new syntax?

src/languages/vbnet.js
joshgoebel: We likely need to handle a few specific date variants or reduce this to 0 relevancy (or both).  Some common formats could get 1 relevancy and then `#.*#` could default to 0 relevance if it does not look like a date.

joshgoebel: Right now something like `###########################` in markdown would count as 20 VB literals, which is not good at all.

joshgoebel:
https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/data-types/date-data-type

> You must enclose a Date literal within number signs (# #). You must specify the date value in the format M/d/yyyy, for example #5/31/1993#, or yyyy-MM-dd, for example #1993-5-31#. You can use slashes when specifying the year first. This requirement is independent of your locale and your computer's date and time format settings.

So I think that means we can have two variants for each of the possible date literals.

test/detect/vbnet/default.txt
Hirse: I don't know about new, but I can only find `imports` in the [documentation](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/imports-statement-net-namespace-and-type) and the list of keywords:
https://github.com/highlightjs/highlight.js/blob/6698b8a5959c398568c7c437fe028a6e3167c776/src/languages/vbnet.js#L18

joshgoebel:
This could land in 10.4 if you have time to finish it up soon.

Hirse:
I'm working on it. 😄

src/languages/vbnet.js
Hirse: As far as I understand, general comments start with a single `'`, but there are special [Documentation Comments](https://docs.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/documentation-comments) that start with `'''` and contain XML tags.

src/languages/vbnet.js
Hirse: Removed matching group.

I think the statements can be indented, but not after other content.

src/languages/vbnet.js
Hirse: Done, added more specific regex.

src/languages/vbnet.js
```diff
@@ -1,63 +1,114 @@
...
+  /**
+   * Float: /-?\b\d[\d_]*((\.[\d_]+(E[+-]?[\d_]+)?)|(E[+-]?[\d_]+))[RFD@!#]?/
+   * Integer (base 10): /-?\b\d[\d_]*((U?[SIL])|[%&])?/
+   * Integer (base 16): /&H[\dA-F_]+((U?[SIL])|[%&])?/
+   * Integer (base  8): /&O[0-7_]+((U?[SIL])|[%&])?/
+   * Integer (base  2): /&B[01_]+((U?[SIL])|[%&])?/
+   */
+  const NUMBER = {
+    className: 'number',
+    begin: /((&[HOB])[\dA-F_]+|(-?\b\d[\d_]*))((\.[\d_]+)?(E[+-]?[\d_]+)?)?((U?[SIL])|[RFD@!#%&])?/,
```
joshgoebel, line 41: [Maintainability] Please beak these into separate rules using `variants` as you have them shown above.

src/languages/vbnet.js
```diff
@@ -1,63 +1,114 @@
...
+  };
+
+  /**
+   * Date Literals consist of a date or a time or both separated by whitespace, surrounded by #.
+   * Date can be in M/D/YYYY or YYYY-MM-DD format.
+   * A time value may be specified either using a 24-hour value (H:mm[:ss]) or a 12-hour value (h[:mm[:ss]] A)
+   */
+  const DATE = {
+    className: 'literal',
+    begin: /# *(((\d+[-/]){2}\d+)|((\d+(:\d+){1,2} ?(AM|PM)?)|(\d+ ?(AM|PM)))|((\d+[-/]){2}\d+ ((\d+(:\d+){1,2} ?(AM|PM)?)|(\d+ ?(AM|PM))))) *#/,
```
joshgoebel, line 29: [Maintainability] Ditto on `variants` here... if there are a few different possibilities (it looks like it) use variants or `regex.either`...  I'd expect to at least see two variants:

m/d/y vs y/m/d

src/languages/vbnet.js
```diff
@@ -1,63 +1,114 @@
...
+ * Language: Visual Basic .NET
+ * Description: Visual Basic .NET (VB.NET) is a multi-paradigm, object-oriented programming language, implemented on the .NET Framework.
+ * Authors: Poren Chiang <ren.chiang@gmail.com>, Jan Pilzer
+ * Website: https://docs.microsoft.com/dotnet/visual-basic/getting-started/
+ */
 
 export default function(hljs) {
+  const CHAR = {
+    className: 'string',
+    begin: /"(""|[^/n])"C\b/,
```
joshgoebel, line 11: What is the first case here?

src/languages/vbnet.js
```diff
@@ -1,63 +1,114 @@
...
+    begin: /((&[HOB])[\dA-F_]+|(-?\b\d[\d_]*))((\.[\d_]+)?(E[+-]?[\d_]+)?)?((U?[SIL])|[RFD@!#%&])?/,
+    relevance: 0,
+  };
+
+  const LABEL = {
+    className: 'symbol',
+    begin: /^\w+:$/,
+  };
+
+  const COMMENT = hljs.COMMENT(/'(?!'')|\bREM\b/, /$/);
```
joshgoebel, line 54: Please use variants here also, not `|`.

src/languages/vbnet.js
joshgoebel: Much better though!

src/languages/vbnet.js
```diff
@@ -1,63 +1,114 @@
...
+        begin: /<\/?/,
+        end: />/,
+        contains: [hljs.PHRASAL_WORDS_MODE],
+      },
+    ],
+  });
+
+  const DIRECTIVES = {
+    className: 'meta',
+    begin: /#(const|disable|else|elseif|enable|end|externalsource|if|region)\b/,
```
joshgoebel, line 69: Does this only exist at the beginning of a line or?   Could this rule be tighter?

src/languages/vbnet.js
Hirse: Char can either be a single character `"a"C` or an escaped double quote `""""C`.

joshgoebel:
Awesome this is looking pretty good.

src/languages/vbnet.js
joshgoebel: Can we explain with a comment please.

src/languages/vbnet.js
joshgoebel: You can see many other grammars for examples.

src/languages/vbnet.js
Hirse: I think it can be indented, but changing to something like `/#[\t ]*(const|...` would make the whitespace part of it.

src/languages/vbnet.js
```diff
@@ -1,63 +1,114 @@
...
-        }
-      ),
-      hljs.C_NUMBER_MODE,
-      {
-        className: 'meta',
-        begin: '#', end: '$',
-        keywords: {'meta-keyword': 'if else elseif end region externalsource'}
-      }
-    ]
+      CHAR,
```
joshgoebel, line 108:
```suggestion
      CHARACTER,
```

Prefer to use full words for readability vs abbreviations. (obivously need to change the top also)

src/languages/vbnet.js
joshgoebel: I dislike matching the whole line, but I say lets do it for specificity:

```
 begin: '^\s*#(const|disable|else|elseif|enable|en...
```

Ie, something like that... you could add a TODO comment to go back and hit this with `beforeMatch:` when that gets merged.

src/languages/vbnet.js
joshgoebel: Yeah, see my comment in the wrong place... lets do it now and comment a TODO for fixing with `beforeMatch` after that's merged.

joshgoebel:
I think you're fixing some of the other VB issues (comments?) though so perhaps look thru them and update this issue at the top if it closes more than one?

src/languages/vbnet.js
```diff
@@ -1,73 +1,118 @@
...
+   * Integer (base  8): /&O[0-7_]+((U?[SIL])|[%&])?/
+   * Integer (base  2): /&B[01_]+((U?[SIL])|[%&])?/
+   */
+  const NUMBER = {
+    className: 'number',
+    begin: /((&[HOB])[\dA-F_]+|(-?\b\d[\d_]*))((\.[\d_]+)?(E[+-]?[\d_]+)?)?((U?[SIL])|[RFD@!#%&])?/,
+    relevance: 0
+  };
+
+  const LABEL = {
```
joshgoebel, line 49: Are these labels as in "jump labels" like in assembly?  `goto FIRE_ROCKET` type label?

src/languages/vbnet.js
Hirse: There are two date formats (m/d/y and y-m-d) and two time formats (12h and 24h).
They are both optional (d, t, or dt) for 8 total combinations.
Any preference for listing variants?

src/languages/vbnet.js
Hirse: Yep.

Although it seems like they don't have to be on a line by themselves:
https://docs.microsoft.com/en-ca/dotnet/visual-basic/programming-guide/program-structure/how-to-label-statements

src/languages/vbnet.js
joshgoebel: 4 is probably fine... just add some visiblity

- 12h
- 24h
- YMD time (you can use `regex.concat` and `regex.either` to make this nicer
- DMY time (you can use `regex.concat` and `regex.either` to make this nicer

src/languages/vbnet.js
```diff
@@ -1,73 +1,118 @@
...
+    className: 'symbol',
```
joshgoebel, line 50: I'd like to start calling these what they are, label is semantic and has meaning in a lot of languages.

```suggestion
    className: 'label',
```
Then you add this to the top level of the language:

```
    classNameAliases: {
      label: 'symbol'
    }
```

src/languages/vbnet.js
```diff
@@ -1,73 +1,160 @@
...
+
+  /**
+   * Date Literals consist of a date, a time, or both separated by whitespace, surrounded by #.
+   */
+  const DATE = {
+    className: 'literal',
+    variants: [
+      {
+        // Date: #M/D/YYYY# or #YYYY-MM-DD#
+        begin: /# *((\d+[-/]){2}\d+) *#/
```
joshgoebel, line 39: Too broad be specific and match as precisely as we can:

```
// MDY
\d{1,2}-\d{1,2}-\d{4}
```

Etc.

src/languages/vbnet.js
```diff
@@ -1,73 +1,160 @@
...
+    begin: /^\w+:/
+  };
+
+  const COMMENT = hljs.COMMENT(null, /$/, {
+    variants: [
+      {
+        begin: /'(?!'')/
+      },
+      {
+        begin: /\bREM\b/
```
joshgoebel, line 90: Pretty sure REM needs a space after it, no?  Many things other than spaces would count as a word boundary but I think you need a space for it to be a comment? Or am I mistaken?

Probably an issue with the \b at the beginning to.

- A space is required between the REM keyword and comment.
- You can put a REM statement alone on a line, or you can put it on a line following another statement. The REM statement must be the last statement on the line. If it follows another statement, the REM must be separated from that statement by a space.

I think this might be a case for now where we need `\sREM ` even though it's annoying.

src/languages/vbnet.js
```diff
@@ -1,73 +1,160 @@
...
+
+  const LABEL = {
+    className: 'label',
+    begin: /^\w+:/
+  };
+
+  const COMMENT = hljs.COMMENT(null, /$/, {
+    variants: [
+      {
+        begin: /'(?!'')/
```
joshgoebel, line 87: This shouldn't be needed, solve this with ordering instead, just list DOC comment before comments so they get the match.

src/languages/vbnet.js
Hirse: Split and mores specific regexes.

src/languages/vbnet.js
Hirse: Absolutely correct.
Changed to `([\t ]|^)REM(?!\S)`

src/languages/vbnet.js
```diff
@@ -28,23 +28,41 @@ export default function(hljs) {
...
-   */
+  /** Date Literals consist of a date, a time, or both separated by whitespace, surrounded by # */
   const DATE = {
     className: 'literal',
     variants: [
       {
-        // Date: #M/D/YYYY# or #YYYY-MM-DD#
-        begin: /# *((\d+[-/]){2}\d+) *#/
+        // #YYYY-MM-DD# (ISO-Date)
+        begin: /# *\d{4}(-\d\d){2} *#/
```
joshgoebel, line 37: Is `2000/1/1` not allowed?

src/languages/vbnet.js
```diff
@@ -28,23 +28,41 @@ export default function(hljs) {
...
       },
       {
-        // Time: #H:mm[:ss]# (24h) or #h[:mm[:ss]] A# (12h )
-        begin: /# *((\d+(:\d+){0,2} *(AM|PM)?)) *#/
+        // #M/D/YYYY# (US-Date)
+        begin: /# *(\d{1,2}\/){2}\d{4} *#/
```
joshgoebel, line 41: 1/1/2000 ?

joshgoebel: Oh I see, I think it's harder to read this way. :)

src/languages/vbnet.js
```diff
@@ -102,6 +109,18 @@ export default function(hljs) {
...
   });
 
+  const COMMENT = hljs.COMMENT(null, /$/, {
+    variants: [
+      {
+        begin: /'(?!'')/
+      },
+      {
+        // TODO: Use `beforeMatch:` for leading spaces
+        begin: /([\t ]|^)REM(?!\S)/
```
joshgoebel, line 119: Hmmm... How is "not not space" clearer than "space". IE, `\s`?

joshgoebel: Oh wait... thinking.

joshgoebel: Yeah, why not a positive lookahead matching \s?

src/languages/vbnet.js
Hirse: The docs seems inconclusive:
> You must specify the date value in the format M/d/yyyy, for example #5/31/1993#, or yyyy-MM-dd, for example #1993-5-31#.

The format would say "No", but the example says "Yes".

Changed it to be more inclusive though.

src/languages/vbnet.js
Hirse: https://github.com/highlightjs/highlight.js/blob/24872afce1b2d69d806e88e6b8f706114955db5e/test/markup/vbnet/literals.txt#L9

src/languages/vbnet.js
Hirse: Because it's late. 😆 

Removed the double negative.

Thanks for the quick feedback. If you have more comments, I'll take care of them tomorrow.

src/languages/vbnet.js
```diff
@@ -1,73 +1,179 @@
...
+    className: 'number',
+    relevance: 0,
+    variants: [
+      {
+        // Float
+        begin: /-?\b\d[\d_]*((\.[\d_]+(E[+-]?[\d_]+)?)|(E[+-]?[\d_]+))[RFD@!#]?/
+      },
+      {
+        // Integer (base 10)
+        begin: /-?\b\d[\d_]*((U?[SIL])|[%&])?/
```
joshgoebel, line 80:
```suggestion
        begin: /\b\d[\d_]*((\.[\d_]+(E[+-]?[\d_]+)?)|(E[+-]?[\d_]+))[RFD@!#]?/
      },
      {
        // Integer (base 10)
        begin: /\b\d[\d_]*((U?[SIL])|[%&])?/
```

We can't do this, no way to tell the difference between `-3` and `2 - 3` so we do not include `-` as part of numbers.

src/languages/vbnet.js
```diff
@@ -1,73 +1,179 @@
...
+        begin: /<\/?/,
+        end: />/
+      }
+    ]
+  });
+
+  const COMMENT = hljs.COMMENT(null, /$/, {
+    variants: [
+      {
+        begin: /'(?!'')/
```
joshgoebel, line 115:
```suggestion
        begin: /'/
```

src/languages/vbnet.js
```diff
@@ -1,73 +1,179 @@
...
-Description: Visual Basic .NET (VB.NET) is a multi-paradigm, object-oriented programming language, implemented on the .NET Framework.
-Author: Poren Chiang <ren.chiang@gmail.com>
-Website: https://docs.microsoft.com/en-us/dotnet/visual-basic/getting-started/
-*/
+/**
+ * Language: Visual Basic .NET
+ * Description: Visual Basic .NET (VB.NET) is a multi-paradigm, object-oriented programming language, implemented on the .NET Framework.
+ * Authors: Poren Chiang <ren.chiang@gmail.com>, Jan Pilzer
+ * Website: https://docs.microsoft.com/dotnet/visual-basic/getting-started/
+ */
```
joshgoebel, line 6: Please revert the comment stylistic changes.  The block must be a normal /* */ comment without any other decoration.

joshgoebel (changes requested):
I think those 3 things are it.  Looking pretty awesome. :-)

src/languages/vbnet.js
Hirse: That makes sense.

I modeled my regex after the `C_NUMBER_RE` which starts with `-?`:
https://github.com/highlightjs/highlight.js/blob/3a2934d7e7af99631527bf55465fa229522735b9/src/lib/modes.js#L8

joshgoebel: Oh wow, that would be a mistake.  I created an issue for it.

src/languages/vbnet.js
```diff
@@ -1,9 +1,9 @@
...
- * Language: Visual Basic .NET
- * Description: Visual Basic .NET (VB.NET) is a multi-paradigm, object-oriented programming language, implemented on the .NET Framework.
- * Authors: Poren Chiang <ren.chiang@gmail.com>, Jan Pilzer
- * Website: https://docs.microsoft.com/dotnet/visual-basic/getting-started/
- */
+/*
+Language: Visual Basic .NET
+Description: Visual Basic .NET (VB.NET) is a multi-paradigm, object-oriented programming language, implemented on the .NET Framework.
+Authors: Poren Chiang <ren.chiang@gmail.com>, Jan Pilzer
+Website: https://docs.microsoft.com/en-us/dotnet/visual-basic/getting-started/
```
joshgoebel, line 5: FYI: I had no issues with this URL change if it was linking to a more general page...

Only issue was with the stylistic * stuff. :)

joshgoebel (approved):
Are the float tests I added all still valid?  We had cases we weren't covering.

I simplified the date/time stuff a little with the helpers we have.

Hirse:
Yes, the tests look good.

joshgoebel:
@Hirse Thanks for all the hard work and closing 3 issues!!!

PR #2790, opened 2020-10-26
enh(javascript) Match numeric literals per ECMA-262 spec

gibson042:
https://tc39.es/ecma262/#sec-literals-numeric-literals
https://tc39.es/ecma262/#sec-additional-syntax-numeric-literals

### Changes
Brought JavaScript numeric literal variants in line with the [spec](https://tc39.es/ecma262/#sec-literals-numeric-literals), just like was done for Python in #2780 and Ruby in #2788.

### Checklist
- [x] Added markup tests
- [x] Updated the changelog at `CHANGES.md`
- [x] Added myself to `AUTHORS.txt`, under Contributors

PR #2788, opened 2020-10-25
enh(ruby) Match numeric literals per language documentation

gibson042:
https://docs.ruby-lang.org/en/2.7.0/doc/syntax/literals_rdoc.html#label-Numbers

Fixes #2766

### Changes
Replaced overly-basic numeric regular expression with specific variants to match the [Ruby language documentation](https://docs.ruby-lang.org/en/2.7.0/doc/syntax/literals_rdoc.html#label-Numbers).

### Checklist
- [x] Added markup tests
- [x] Updated the changelog at `CHANGES.md`
- [x] Added myself to `AUTHORS.txt`, under Contributors

PR #2780, opened 2020-10-23
enh(python) Match numeric literals per the language reference

gibson042:
I started to work on #2766, but realized that the Python foundation I planned to use was itself incomplete, so I'm testing the waters with this and can follow up if it is accepted.

### Changes
Replaced generic numeric regular expressions with specific ones to match the [Python language reference](https://docs.python.org/3/reference/lexical_analysis.html#numeric-literals).

### Checklist
- [x] Added markup tests, or they don't apply here because...
- [x] Updated the changelog at `CHANGES.md`
- [x] Added myself to `AUTHORS.txt`, under Contributors

PR #2812, opened 2020-10-31
enh(java) Match numeric literals per Java Language Specification

gibson042:
https://docs.oracle.com/javase/specs/jls/se15/html/jls-3.html#jls-3.10

### Changes
Brought Java numeric literal variants in line with the language specification, just like was done for Python in #2780 and Ruby in #2788 and JavaScript in #2790.

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -27,44 +27,38 @@ export default function(hljs) {
       },
     ]
   };
-  /**
-   * A given sequence, possibly with underscores
-   * @type {(s: string | RegExp) => string}  */
-  var SEQUENCE_ALLOWING_UNDERSCORES = (seq) => regex.concat('[', seq, ']+([', seq, '_]*[', seq, ']+)?');
-  var JAVA_NUMBER_MODE = {
+
+  // https://docs.oracle.com/javase/specs/jls/se15/html/jls-3.html#jls-3.10
+  var decimalDigits = '[0-9](_*[0-9])*';
+  var frac = `\\.(${decimalDigits})`;
+  var hexDigits = '[0-9a-fA-F](_*[0-9a-fA-F])*';
+  var NUMBER = {
     className: 'number',
     variants: [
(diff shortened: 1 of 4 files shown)


src/languages/java.js
```diff
@@ -27,44 +27,38 @@ export default function(hljs) {
...
+      { begin: `(\\b(${decimalDigits})((${frac})|\\.)?|(${frac}))` +
+        `[eE][+-]?(${decimalDigits})[fFdD]?\\b` },
+      // excluding ExponentPart
+      { begin: `\\b(${decimalDigits})((${frac})[fFdD]?\\b|\\.([fFdD]\\b)?)` },
+      { begin: `(${frac})[fFdD]?\\b` },
+      { begin: `\\b(${decimalDigits})[fFdD]\\b` },
+
+      // HexadecimalFloatingPointLiteral
+      { begin: `\\b0[xX]((${hexDigits})\\.?|(${hexDigits})?\\.(${hexDigits}))` +
+        `[pP][+-]?(${decimalDigits})[fFdD]?\\b` },
```
joshgoebel, line 49: So the `p` part is decimal vs hex?

joshgoebel:
So `_` can be repeated over and over in Java?

src/languages/java.js
gibson042: Yep. [`HexadecimalFloatingPointLiteral`](https://docs.oracle.com/javase/specs/jls/se15/html/jls-3.html#jls-HexadecimalFloatingPointLiteral) includes a mandatory `BinaryExponent`, which is a case-insensitive `p` followed by an optionally signed (decimal) integer.

gibson042:
> So `_` can be repeated over and over in Java?

Yep: https://docs.oracle.com/javase/7/docs/technotes/guides/language/underscores-literals.html (emphasis mine):
> In Java SE 7 and later, **any number** of underscore characters (_) can appear anywhere between digits in a numerical literal.

joshgoebel:
What do people use that for in practice? LOL.

joshgoebel (approved):

PR #2813, opened 2020-10-31
enh(parser) use negative look-ahead for beginKeywords

joshgoebel:
### Changes

Just realized we only need the callback for the look-behind, we can use actual
regex to look-ahead.

### Checklist
- [x] Existing test coverage


test/api/index.js
```diff
@@ -1,16 +1,17 @@
 'use strict';
 
 describe('hljs', function() {
-  require('./ident');
-  require('./underscoreIdent');
-  require('./number');
-  require('./cNumber');
+  require('./autoDetection');
+  require('./beginKeywords');
```
joshgoebel, line 5: We were missing this before, so this fixes that.

egor-rogov (approved):
Looks good.
(suddenly ran out of free time, will try to catch up slowly...)

PR #2814, opened 2020-11-01
enh(diff) enhance diffs generated by `git format-patch`

fbezdeka:
### Changes
Improving the highlighting of patches generated with `git format-patch`.

- Mark lines starting with "index " as meta-strings (please check if this is the right class to use...)
- Mark lines starting with "diff --git" as comments
- Improve highlighting of hunk identifier lines (lines starting with @@). It may happen that something follows behind the closing "@@" pair.

### Checklist

### DIFF
--- a/src/languages/diff.js
+++ b/src/languages/diff.js
@@ -29,7 +29,8 @@ export default function(hljs) {
           {begin: /^-{3}/, end: /$/},
           {begin: /^\*{3} /, end: /$/},
           {begin: /^\+{3}/, end: /$/},
-          {begin: /^\*{15}$/ }
+          {begin: /^\*{15}$/ },
+          {begin: /^diff --git/, end: /$/},
         ]
       },
       {
--- a/src/languages/diff.js
+++ b/src/languages/diff.js
@@ -16,7 +16,7 @@ export default function(hljs) {
         className: 'meta',
         relevance: 10,
         variants: [
-          {begin: /^@@ +-\d+,\d+ +\+\d+,\d+ +@@$/},
+          {begin: /^@@ +-\d+,\d+ +\+\d+,\d+ +@@/},
           {begin: /^\*\*\* +\d+,\d+ +\*\*\*\*$/},
           {begin: /^--- +\d+,\d+ +----$/}
         ]
--- a/src/languages/diff.js
+++ b/src/languages/diff.js
@@ -21,6 +21,12 @@ export default function(hljs) {
           {begin: /^--- +\d+,\d+ +----$/}
         ]
       },
+      {                                                                             
+         className: "meta-string",                                                  
+         variants: [                                                                
+           {begin: /^index/, end: /$/ }                                             
+         ]                                                                          
+      },
       {
         className: 'comment',
         variants: [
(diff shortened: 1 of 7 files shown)


test/markup/diff/meta.expect.txt
```diff
@@ -0,0 +1,2 @@
+<span class="hljs-meta">@@ -1,8 +1,7 @@</span>
+<span class="hljs-meta">@@ -1,8 +1,7 @@</span> postfix
```
joshgoebel, line 2: The name of the test should really reflect what the content is or what it's testing, not the class used... so `line_range` might be a better name here, etc.  Same comment on the other meta-string test...

test/markup/diff/meta-string.txt
```diff
@@ -0,0 +1 @@
+index revA..revB file_mode
```
joshgoebel, line 1: This should preferably be a real example exactly as it would appear in an actual generated diff.

src/languages/diff.js
```diff
@@ -16,11 +16,17 @@ export default function(hljs) {
...
         relevance: 10,
         variants: [
-          {begin: /^@@ +\-\d+,\d+ +\+\d+,\d+ +@@$/},
+          {begin: /^@@ +\-\d+,\d+ +\+\d+,\d+ +@@/},
           {begin: /^\*\*\* +\d+,\d+ +\*\*\*\*$/},
           {begin: /^\-\-\- +\d+,\d+ +\-\-\-\-$/}
         ]
       },
+      {                                                                             
+         className: "meta-string",
```
joshgoebel, line 25: What is your thinking that this should be meta-string rather than meta or comment?

joshgoebel:
Awesome, at a glance this seems a nice little tweak.

src/languages/diff.js
fbezdeka: I just tried to get a result that is as close as possible to the `vi` highlighting. But if you prefer `comment` or `meta` I'm fine with it.

image

joshgoebel: My vim doesn't do that... it's all comments... I say lets make it comments for now.  I don't see any good reason it needs to be super emphasized.

test/markup/diff/meta.expect.txt
fbezdeka: I will merge both test into a single "git-patch" test. Will take some time but I will come up with a rebase...

joshgoebel: That would be much better! Thanks.

src/languages/diff.js
fbezdeka: Strage. vim 8.1 (on Debian) and 8.2 (on Fedora) does...

fbezdeka:
```bash
$ grep "IndexLine" /usr/share/vim/vim82/syntax/diff.vim 
syn match diffIndexLine	"^index \x\x\x\x.*"
hi def link diffIndexLine	PreProc
```

joshgoebel: I don't doubt you, but just because vim does it doesn't mean we have to do follow.   I took a glance at both ways and just don't think the line deserves such emphasis.

joshgoebel (approved):

joshgoebel:
@fbezdeka Thanks so much for PR!

PR #2816, opened 2020-11-02
fix(vue) Vue.js component should not require `hljs` global to function properly

### LINKED ISSUE #2815: Vue.js plugin "ReferenceError: hljs is not defined"
A Vue.js plugin should be self-contained and not require an external script.
Currently however, the Vue.js component for highlight.js depends on `hljs` but `hljs` is not in scope in its definition.
https://github.com/highlightjs/highlight.js/blob/master/src/plugins/vue.js#L24

If a Vue.js user uses npm or yarn to install highlight.js, and follows the instructions in the section [Using with Vue.js](https://github.com/highlightjs/highlight.js#using-with-vuejs) they will receive the following error in their browser.
`[Vue warn]: Error in render: "ReferenceError: hljs is not defined"`
The line that produces this exception is this one:
https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L26

Proposal:
1. Put the contents of the [vue.js plugin file](https://github.com/highlightjs/highlight.js/blob/master/src/plugins/vue.js) inside a function that accepts `hljs` as an argument and returns `VuePlugin`. Make the file export that function.
2. Change the [plugins/vue.js import](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L14) to import this new function instead of `VuePlugin`
3. Change [this location](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L871) to invoke the wrapper function.

Kylebrown9:
Makes changes as described in the linked issue to improve the usability of the Vue.js plugin by wrapping `Component` and `VuePlugin` with a function so that `hljs` can be brought into scope.

Resolves #2815

### Changes
1. Put the contents of the Vue.js plugin file inside an exported function that accepts `hljs` as an argument and returns VuePlugin. 
2. Change the `plugins/vue.js` import to import this new function instead of `VuePlugin`
3. Change this location to invoke the wrapper function.

### Checklist


joshgoebel:
IE: https://github.com/highlightjs/highlight.js/pull/2818

Thoughts?

Kylebrown9:
I am using the `npm` build through `yarn`.
The statement "`hljs` should be global" feels very prescriptive especially in a Vue.js / node environment, where we have the tools to manage our dependencies and their scopes.

joshgoebel:
> The statement "hljs should be global" feels very prescriptive especially in a Vue.js / node environment, where we have the tools to manage our dependencies and their scopes.

I'm not talking about a node environment. :)  _Or are you talking about server-side rendering of Vue.js?_

Kylebrown9:
I hit this error in browser rendering of Vue.js using webpack and node.

src/plugins/vue.js
```diff
@@ -5,59 +5,63 @@ function hasValueOrEmptyAttribute(value) {
...
-  // template: `<pre><code :class="className" v-html="highlighted"></code></pre>`
-};
+  };
 
-export const VuePlugin = {
-  install(Vue) {
-    Vue.component('highlightjs', Component);
-  }
-};
+  return { Component, VuePlugin };
```
joshgoebel, line 66: What does exporting component do for us since we never import it anywhere?  I would understand the intent in a larger sense, but this file isn't distributed anywhere - it's only built inside of HLJS.

Kylebrown9: I don't know that there is one and I'd be happy to change that.
I just did it to make my change more closely resemble the original, which did export `Component` event though it was never referenced.

joshgoebel: > I just did it to make my change more closely resemble the original, which did export Component event though it was never referenced.

HAHA, so it's my fault... far to easy to slap `export` next to things in the middle... it pops more when you put them all in a single place. No need to change anything.

src/plugins/vue.js
```diff
@@ -5,59 +5,63 @@ function hasValueOrEmptyAttribute(value) {
...
+      return createElement("pre", {}, [
+        createElement("code", {
+          class: this.className,
+          domProps: { innerHTML: this.highlighted }})
+      ]);
+    }
+    // template: `<pre><code :class="className" v-html="highlighted"></code></pre>`
+  };
+  
+  export const VuePlugin = {
```
joshgoebel, line 60:
```suggestion
  const VuePlugin = {
```

joshgoebel: Can't export except at top-level.

joshgoebel:
Do you purposely have the ability for maintainers here to update your branch disabled?  I don't have any of my usual tools at my disposal...

Kylebrown9:
Nope, can you point me towards the settings for that?

joshgoebel:
Usually it's along the right hand side... but I think there is a setting in your settings somewhere that decides the default whether you allow maintainers to push to PRs you make or not...

Kylebrown9:
Huh, that option is not present on this PR for me
https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests
image

joshgoebel:
Please finish up the PR checklist (this doesn't need tests I don't think) and bring this back up-to-date with master and I think we're good to go!

joshgoebel (approved):

joshgoebel:
@Kylebrown9 Thanks so much for contributing!

PR #2818, opened 2020-11-02
Should npm library in browser has same behavior as web build (exposing global `hljs`)?

joshgoebel:
Whether using the web build or `npm` build inside the browser the same
behavior should exist - ie `hljs` is defined and accessible on `window`.

### Thoughts

I'd suggest that the larger eco-system of Highlight.js plugins/grammars/extensions expect this to be true and not all of them even go to the trouble of building a `npm` package or exposing a module... sometimes a plugin is no more than a few lines of code that you copy and paste wherever is convenient.

### Checklist
- [x] Added markup tests, or they don't apply here not a grammar change.


joshgoebel:
The issue that led to the opening of this is being resolved by injecting the hljs variable into the Vue plugin (which seems to be the correct solution given that Vue can run on the server-side also, where we have no existing legacy of having a global).  I'm leaning towards closing this for now (as ~"globals bad") unless someone has any "pro" arguments they'd like to share.

If this comes up again (outside of the Vue context) I'd suggest perhaps a doc fix suggesting that if someone is using a build pipeline but wants to use "simple plugins" (without modification) that refer to a `hljs` global that they need to just define this global themselves in their codebase.

joshgoebel:
Closing as not a good idea right now (if ever). :-)

PR #2820, opened 2020-11-02
(Swift) Support raw strings

### LINKED ISSUE #2819: (Swift) Incorrect highlighting of raw strings
[Editor: Dropping some notes for whoever picks this up] 

This needs to also take into account: 

- [x] SUBST (inline code) inside raw strings
- [x] the different forms of escaping inside raw strings
- [x] the different syntax of variable substitution (we should probably highlight this, right?)
- [x] extend the string markup tests quite a bit to go along with this
- [ ] could we handle arbitrary number of #### with END_SAME_AS_BEGIN? Worth it?

---

The following is a single raw string:

```swift
##"Use #" and "# to delimit a raw string."##
```

However, it's incorrectly highlighted as two separate strings. My guess is that highlight.js thinks the quotes in the middle close a string, then start a new one?

In general, the hashes around raw strings aren't highlighted in the same color as the string, so perhaps raw strings aren't supported at all yet?

svanimpe:
Resolves #2819 

### Changes

Add support for raw strings with up to three leading/trailing hashes. In theory, raw strings can have any number of hashes, as long as they’re balanced, but more than one is rarely used.

### Checklist

### DIFF
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -67,6 +67,9 @@ export default function(hljs) {
     variants: [
       {begin: /"""/, end: /"""/},
       {begin: /"/, end: /"/},
+      {begin: /#"/, end: /"#/},
+      {begin: /##"/, end: /"##/},
+      {begin: /###"/, end: /"###/},
     ]
   };
   var NUMBERS = {
--- a/test/markup/swift/strings.txt
+++ b/test/markup/swift/strings.txt
@@ -1,0 +1,5 @@
+var string = """
+    var a = not actually code
+"""
+
+print(##"Use #" and "# to delimit a raw string."##)
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -296,3 +296,4 @@ Contributors:
 - Fredrik Ekre <ekrefredrik@gmail.com>
 - Jan Pilzer <Hirse@github>
 - Jonathan Sharpe <mail@jonrshar.pe>
+- Steven Van Impe <steven.vanimpe@icloud.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Language Improvements:
 - enh(java) Match numeric literals per Java Language Specification [Richard Gibson][]
 - enh(php) highlight variables (#2785) [Taufik Nurrohman][]
 - fix(python) Handle comments on decorators (#2804) [Jonathan Sharpe][]
+- enh(swift) Highlight raw strings in Swift (#2819) [Steven Van Impe][]
 
 Dev Improvements:
 
--- a/test/markup/swift/multiline-string.expect.txt
+++ b/test/markup/swift/multiline-string.expect.txt
@@ -1,3 +1,0 @@
-<span class="hljs-keyword">var</span> string = <span class="hljs-string">&quot;&quot;&quot;
-    var a = not actually code
-&quot;&quot;&quot;</span>
--- a/test/markup/swift/multiline-string.txt
+++ b/test/markup/swift/multiline-string.txt
@@ -1,3 +1,0 @@
-var string = """
-    var a = not actually code
-"""
--- a/test/markup/swift/strings.expect.txt
+++ b/test/markup/swift/strings.expect.txt
@@ -1,0 +1,5 @@
+<span class="hljs-keyword">var</span> string = <span class="hljs-string">&quot;&quot;&quot;
+    var a = not actually code
+&quot;&quot;&quot;</span>
+
+<span class="hljs-built_in">print</span>(<span class="hljs-string">##&quot;Use #&quot; and &quot;# to delimit a raw string.&quot;##</span>)


src/languages/swift.js
```diff
@@ -67,6 +67,9 @@ export default function(hljs) {
     variants: [
       {begin: /"""/, end: /"""/},
       {begin: /"/, end: /"/},
+      {begin: /#"/, end: /"#/},
+      {begin: /##"/, end: /"##/},
+      {begin: /###"/, end: /"###/},
```
joshgoebel, line 72: Can such strings contain unescaped newlines?

joshgoebel: Same subst and backlash escape rules apply to raw strings as regular strings or is that different?

svanimpe: No, but you can have raw multiline strings (wrapped in `#"""` and `"""#`). I also tried those and they highlight correct for me, with the rules above. Same goes for strings with interpolation or escaped characters.

joshgoebel: I'm confused what makes them "raw"... if this has no bearing on the escaping, etc?

svanimpe: FYI, here are all the code examples in my document that use raw strings:

```swift
print(#"Use \t to insert a tab."#)

print(#"""
Use \t to insert a tab. The following example:

print("\tHello")

prints:

    Hello
"""#)

print(##"Use #" and "# to delimit a raw string."##)

print(#"""
Use \t to insert a tab. The following example:

print("\tHello")

prints:

\#tHello
"""#)

print(#"Hi \#(player), your score is: \#(score)."#)
```

These all highlight correctly with the new rules.

svanimpe: What makes them raw is that they interpret "special characters" (such as `\t`) as normal character sequences (backslash and t) instead of their special meaning (tab). However, you can still use escape sequences and string interpolation by adding the same number of hashes after the backslash.

joshgoebel:
> That same # symbol can now be used inside strings, to mark special characters. For example, if you want to use string interpolation, you should now use \#(variableName) rather than just \(variableName), like this:

I wonder if we should highlight such things?

joshgoebel:
> When you use # with a string it affects the way Swift understands special characters in the string: \ no longer acts as an escape character, so \n literally means a backslash then an “n” rather than a line break

So it sounds like the `contains` needs to be changed for raw strings to not include escapes...

src/languages/swift.js
joshgoebel: I think we need more markup tests... sounds like there is more to be done here... at the very least we don't match normal escapes (as with every other string)... so we should have a negative test case for that... (showing that it workes properly, NOT highlighting what would be an escape)

And we should have some test cases of this new variant of escaping (and then decide whether to handle it or not)... if not then a comment in the test would explain, etc.

test/markup/swift/strings.txt
```diff
@@ -0,0 +1,5 @@
+var string = """
+    var a = not actually code
+"""
+
+print(##"Use #" and "# to delimit a raw string."##)
```
joshgoebel, line 5: If we handle three variants (#, ##, ###) in the language then we'd optimally want all 3 listed here in markup tests also.

svanimpe:
I'm afraid this is about as much as I can contribute in the short term, so I'm hoping you can take that up.

I even had to install NPM specifically for this, which should tell you how familiar I am with JS codebases 😅

svanimpe:
But even if you don't, I'm hoping this small fix can at least make it in. At the moment, raw strings are highlighted wrong. This would at least highlight them as a string, even if it's not perfect.

I also don't see any difference with normal Swift strings. They also highlight as one big hljs-string, even when they contain escapes or interpolated values.  For example:

```swift
"1.\tACE\t976\n2.\tBOB\t952\n3.\tCU2\t897"
```

becomes:

```html
<span class="hljs-string">"1.\tACE\t976\n2.\tBOB\t952\n3.\tCU2\t897"</span>
```

and

```swift
"Hi \(player), your score is: \(score)."
```

becomes:

```html
<span class="hljs-string">"Hi \(player), your score is: \(score)."</span>
```

joshgoebel:
> I also don't see any difference with normal Swift strings.

It's not visual.  Right now an extra escape would throw everything off:

```
#" \"#
```

This sting never ends because `\"` is an escape sequence... so your rule will never find the `"#` terminating sequence it's looking for, etc...

joshgoebel:
And you found a bug, `subst` should indeed be highlighted:

Screen Shot 2020-11-02 at 3 10 42 PM

joshgoebel:
> I'm afraid this is about as much as I can contribute in the short term, so I'm hoping you can take that up.

If not I, someone else.  Thanks for the effort! :)

svanimpe:
Thanks for having Swift support in the first place 😅

joshgoebel:
Closing this for now until someone has time to circle back to it and complete it fully.  The issue remains open of course (with link to this PR).

PR #2821, opened 2020-11-03
chore(ci) Add a GitHub actions workflow for our CI

joshgoebel:
Quick stab at a GHA CI workflow.  Travis CI has been ridiculously flaky lately.

### Checklist
- [x] Updated the changelog at `CHANGES.md` (Not needed)

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,46 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 15.x]
+        build-how: ["node", "browser", "browser -n"]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install
+    # normal browser build
+    - if: ${{ contains(matrix.build-how, 'browser') }}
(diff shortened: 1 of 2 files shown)


joshgoebel:
@allejo If we could do the browser builds as part of the matrix that would be even better... I think we could do that with a second key and `if`, but I'm not sure.  Though it already completes in just about a minute, which is great.  But faster is better with CI. :)

```
      matrix:
        node-version: [12.x, 14.x, 15.x]
        build_how: ["node", "browser", "browser -n"]
```

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,49 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    # normal browser build
+    - run: npm install
+    - if: ${{ contains(matrix.build-how, 'browser') }}
```
allejo, line 30: just FYI, an `if` should be attached to a "step" and not by itself

allejo: also, `${{ }}` notation is automatically assumed in an `if` so you can just do contains

joshgoebel: > just FYI, an if should be attached to a "step" and not by itself 

Seems to work with multiple ifs inside `steps` doing different things.

joshgoebel:
Screen Shot 2020-11-02 at 8 18 54 PM

Ok, yikes... that's a bit verbose... maybe I don't want such a complex matrix?  Any thoughts?

allejo (approved):
I'm okay with the output of the matrix, it's thorough with everything that's officially supported/tested

allejo:
I love how the CI only takes a minute on GHA :heart:

joshgoebel:
Is this not going to work on all the existing PRs though because the action doesn't exist on those branches?  That part confuses me.

allejo:
That's the behavior that I've seen, yes. Current PRs will need to have `master` merged into them or rebased.

joshgoebel:
> Current PRs will need to have master merged into them or rebased.

Ah, but typically that's just one click anyways. :-)  That'll work.

PR #2829, opened 2020-11-04
enh(swift) Match numeric literals per language reference

### LINKED ISSUE #2824: (Swift) Incorrect highlighting floating-point literals
The regular expression for matching numbers doesn't seem to handle exponents properly. For example, `1e-10` is a valid floating-point literal, but only the `10` highlights as a number.

I'm not entirely familiar with your regex syntax, but my guess is that this part: `[\\d_]+(\\.[\\deE_]+)?` means "one or more digits, followed by an optional part containing a decimal point and one or more digits or the letter `e`".  If that's correct, that  would explain why `12.1e12` does highlight correctly.

There may be a few more issues with this regex:

- Does it handle negative exponents?
- It seems to allow multiple letters `e/E` mixed in with the decimals. There should be only one `e/E` between the base and the exponent.

For reference, here is the grammar on floating-point literals:
https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID416

gibson042:
https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#grammar_numeric-literal

Fixes #2824

### Changes
Brought Swift numeric literal variants in line with the language reference.

### Checklist
- [x] Added markup tests


src/languages/swift.js
```diff
@@ -69,12 +69,30 @@ export default function(hljs) {
...
+  var decimalDigits = '([0-9]_*)+';
+  var hexDigits = '([0-9a-fA-F]_*)+';
+  var NUMBER = {
       className: 'number',
-      begin: '\\b([\\d_]+(\\.[\\deE_]+)?|0x[a-fA-F0-9_]+(\\.[a-fA-F0-9p_]+)?|0b[01_]+|0o[0-7_]+)\\b',
-      relevance: 0
+      relevance: 0,
+      variants: [
+        // decimal floating-point-literal (subsumes decimal-literal)
+        { begin: `-?\\b(${decimalDigits})(\\.(${decimalDigits}))?` +
```
joshgoebel, line 81: Has this not come up before or have I missed this on every other PR?

How does this know the difference between a unary - and a - operator?

```
a = -3
// vs
5-3
```

The `-` in the latter can't be part of the number without a proper full parser.

gibson042: In the other languages, leading `-` is never part of the number (it's just a preceding unary operator)—but that's not the case with Swift. You're right that this won't be fully solved, so it's necessary instead to pick a "good enough" behavior. How do you feel about lookbehind? Without using it, I think this is the best tradeoff, but with it, there could be an assertion that the leading `-` is not preceded by a literal or identifier character (e.g., something like ```((?<!"#*|[\\w\\xA8-\\uFFFD]\`?)-)?\\b(${decimalDigits})…```, assuming these are non-Unicode regular expressions).

joshgoebel: Can't use look-behind, it's not supported on Safari yet.  There are a few hacks (fake look-behind, callbacks, etc.) we could use but in other grammars we just take the expedient route and just do not include `-` as part of the number literal. 

> In the other languages, leading - is never part of the numeric literal (it's just a preceding unary operator)—but that's not the case with Swift. 

I'm not sure this is relevant for us.  

> Without using it, I think this is the best tradeoff

This will always mis-highlight `3-2` (which is bad and just looks weird) and it's _inconsistent_ with the way we handle this is every other grammar.  Plus `-` should really (in most cases) be highlighted as an operator (with our higher fidelity initiative).  _For things like this the core library strives for overall consistency among grammars rather than absolute adherence to the quirks of any particular language._  Highlighting a snippet of JavaScript vs a snippet of Swift side by side should not result in different behavior for unary literals.

So for consistency let's drop the `-?` from all the rules and you can include a TODO note about the behavior and mention lookbehind - and this matter can be revisited in the future when we have better tools at our disposal.

gibson042: OK, that makes sense.

src/languages/swift.js
```diff
@@ -69,12 +69,31 @@ export default function(hljs) {
...
+
+        // hexadecimal floating-point-literal (subsumes hexadecimal-literal)
+        { begin: `\\b0x(${hexDigits})(\\.(${hexDigits}))?` +
+          `([pP][+-]?(${decimalDigits}))?\\b` },
+
+        // octal-literal
+        { begin: '\\b0o([0-7]_*)+\\b' },
+
+        // binary-literal
+        { begin: '\\b0b([01]_*)+\\b' },
```
joshgoebel, line 93:
```suggestion
        { begin: /\b0o([0-7]_*)+\b/ },

        // binary-literal
        { begin: /\b0b([01]_*)+\b/ },
```

joshgoebel (approved):

joshgoebel:
@gibson042 Awesome!

joshgoebel:
@gibson042 Any chance you'd want to dig into Kotlin next? ;-)  There are some polynomial backtracking issues in the number regex, so it could use a rework.

joshgoebel:
Nevermind, I think perhaps we can just use the existing Java rule... the comments seem to indicated that at least. ;)

gibson042:
The relevant entry point for Kotlin appears to be https://kotlinlang.org/docs/reference/grammar.html#literalConstant , and I can't tell at a glance if it corresponds exactly with any other language, but suspect that it does not:
* `IntegerLiteral` is a single decimal digit or a nonzero digit followed by any mixture of digits and underscores (e.g., adjacent underscores are acceptable) that ends with a digit.
* `HexLiteral` is a case-insensitive "0x" followed by a nonempty sequence of hexadecimal digits with arbitrary underscores in the middle.
* `BinLiteral` is analogous, a case-insensitive "0b" followed by a nonempty sequence of binary digits with arbitrary underscores in the middle.
* `LongLiteral` suffixes either `IntegerLiteral`, `HexLiteral`, or `BinLiteral` with a case-insensitive "L".
* `UnsignedLiteral` suffixes either `IntegerLiteral`, `HexLiteral`, or `BinLiteral` with a case-insensitive "U" and optional case-insensitive "L".
* `DoubleLiteral` is either
  * an optional `DecDigits` (nonempty sequence of decimal digits with arbitrary underscores in the middle) followed by a "." followed by `DecDigits` followed by an optional `DoubleExponent` (case-insensitive "E" then optional ASCII sign then `DecDigits`), or
  * `DecDigits` followed by `DoubleExponent`.
* `FloatLiteral` suffixes either `DoubleLiteral` or `DecDigits` with a case-insensitive "F".

Feel free to open an issue and assign it to me.

PR #2831, opened 2020-11-04
is yarn any faster?

joshgoebel:
### Changes

Maybe switch to yarn since it's the way things are going.

### Checklist


sonic2kk:
Why was this closed? Was yarn not much faster? As you said it seems to be the way things are going, plus if someone comes along in future and asks the same question it would be good to document why you opted not to switch.

joshgoebel:
> Was yarn not much faster? 

Was slower, but could be margin of error... definitely nothing worth writing home about.

> As you said it seems to be the way things are going

I just read another article that said the opposite so since there was no pressing reason to do this I closed the issue.

Issue #2520, opened 2020-04-30
Add some type of look-behind support


joshgoebel:
**Is your request related to a specific problem you're having?**

Yes, there is no easy way to do look behind or to say: 

> match this, but only if immediately proceeded by that...

It's possible, but it gets ugly fast. For example here's how to match say `instanceof [class]` and do it _with no ambiguity_: (where [class] is the only part I want highlighted)

```js
      {
        begin: regex.concat('(?=', /\binstanceof /, IDENT_RE, ')'),
        contains: [
          { begin: /\binstanceof /,
            keywords: KEYWORDS,
            starts: {
              contains: [
                { begin: IDENT_RE, className: "title"}
              ],
            }
          },
        ],
      },
```

First we match the whole thing with a look-ahead... then our mode starts and our first rule has to re-include keywords from above (because now the context has changed), and then we have to match `instance of` and then start a new mode [for `IDENT_RE`] with `starts` to avoid the chance of our two expressions being similar enough that the first rule might fire for the second.

**The solution you'd prefer / feature you'd like to see added...**

Well if JS just supported look behind, we'd be all set but since it doesn't... Sample code from Javascript we're trying to match:

```js
if (str instanceof Token) { ...
// and
var token = new Token()
```

And how we might go about doing it:

```js
 { 
  skipBefore: /(\binstanceof[ ]+)/,
  begin: IDENT_RE,
  className: "title"
}, {
  skipBefore: /(\bnew[ ]+)/,
  begin: IDENT_RE,
  className: "title" 
},
// and for comparison: look-ahead, which is already possible;
{
  begin: regex.concat(IDENT_RE, /(?=\.(constructor|prototype))/),
  className: "title"
},
```

This is actually only a 5 line patch. Building this into the parser means the mode context doesn't change, so your existing keyword scope "just works"... and these become very simple rules.

**Potential Caveats**

1. It might have interactions with `returnBegin`... I'm not sure why you'd ever want to use them both... but we could either:
   - have `returnBegin` disable skipBefores "skip" ability, making it more of a "prefix"
   - throw an error for grammar developers and just say these two things don't mix.

2. The naming isn't perfect... technically this is `matchBeforeAndSkip`... I hesitate to name it `lookbehind` though because that's NOT what it's doing... it's still looking forward the whole time.  If another rule has consumed the content already this can't match it by "looking backwards", because there is no backwards looking going on.

**Any alternative solutions you considered...**

[...]

PR #2834, opened 2020-11-05
enh(parser) Compiler extensions (for grammars + plugins)

### LINKED ISSUE #2794: Proposal: Make it easier to specify (and comment) a list of illegals

Yes, illegal can become hard to read when it includes multiple items.

I saw this in routeros which brought this to my attention (again).  

```js
      { // illegal constructs
        variants: [
          { begin: /^@/, end: /$/ },               // dns
          { begin: /\/\*/, end: /\*\// },          // -- comment
          { begin: /%%/, end: /$/ },               // -- comment
          { begin: /^'/, end: /$/ },               // Monkey one line comment
          { begin: /^\s*\/[\w-]+=/, end: /$/ },    // jboss-cli
          { begin: /\/\//, end: /$/ },             // Stan comment
          { begin: /^\[</, end: />\]$/ },          // F# class declaration?
          { begin: /<\//, end: />/ },              // HTML tags
          { begin: /^facet /, end: /\}/ },         // roboconf - лютый костыль )))
          { begin: '^1\\.\\.(\\d+)$', end: /$/ }   // tap
        ],
        illegal: /./
      },
```

But this is very verbose and you have to think about it for a minute to realize what it's doing.

Note: Do not focus on if this is a GOOD idea for routeros to have such a long list (it may not be)... that's not a concern here... there are definitely cases where a list of 2-3 illegals makes very good sense.  I only show this as one example of how you'd do this today (get a more readable illegal list) without any core library changes.

Instead I'd like to formalize that as:

```js
{
    illegal: [
      /^@/,      // dns
      /\/\*/,    // C style multi-line comment
      /%%/,      // some other comment...
      // ...
    ],
}
```

And these would be auto-joined with regex.either...  The older single regex or string format would still be supported.  But if we saw an array here when we're call `regex.either` on it...

### LINKED ISSUE #2833: enh(parser) Compiler extensions for grammars and plugins

It can be very hard to do some very simple things like look-behind, sequences, etc...  and if one grammar finds a great helper function, there is no super easy way to share that with other grammars.  Lack of nice abstractions leads to some ugly code making it's way into the library:

```
          // The below rules all eat an extra character in front (for the
          // look-behind check) and then exclude it from the match, but I think
          // in many cases this will work out just fine.
          {
            begin: /([^a-zA-Z0-9._])(?=0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*[pP][+-]?\d+i?)/,
            end: /0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*[pP][+-]?\d+i?/,
            excludeBegin: true
          },
```

This pattern is repeated in several places in different forms to work around the fact that we have no look-behind currently.

There has been discussion of syntactic sugar for these things in the past (ie new keys but ones that only smartly compile down to existing ones, without truly adding any new features to the core parser).  I'd like to allow plugins (or grammars themselves) to extend mode compilation in this fashion.  This would alway plug-in authors to define their own syntactic-sugar on top of our existing modes without having to resort to code like: (which is harder to write and maintain)

```js
contains: [
  EXTEND({ /* mode details here */ },
  EXTEND({ /* mode details here */ },
  EXTEND({ /* mode details here */ },
]
```

I'd also like to use this to extend the core parser itself... adding newer functionality as small modular components - keeping the core parser minimal and simple (perhaps even reducing it's scope).  

Let's define an extension to allow something like:

```js
      {
        className:"booger",
        first: /boo ger /, // first match
        then: /de lic ious/, // followed by (this is what should be highlighted as `booger`)
        keywords: "boo lic",
      },
```

Which would match code like `boo ger de lic ious` and also flag "boo" and "lic" as keywords anywhere they appear.  

What would it could look like in a grammar:

```js
    extensions: [
      (mode, parent) => {
        if (!mode.first) return;

        mode.begin = regex.concat(mode.first, regex.lookahead(mode.then));
        mode.starts = {
          contains: [{
            className: mode.className,
            begin: mode.then,
            keywords: mode.keywords
          }]
        };
        mode.className = null;
      }
    ],
```

Which would compile down to native mode:

```js
      {
        begin: /boo ger (?=de lic ious)/,
        keywords: "boo lic",
        starts: {
          contains: [
            {
              className:"booger",
              begin: /de lic ious/,
              keywords: "boo lic"
            }
          ]
        }
      },
```

**Potential downsides**

You could of course get crazy with this and end up with unrecognizable grammars, but I think that's also true with using helper functions (and some of our grammars border on this already).  This can be allowed (or limited) as desired in the core library because of PR review... and if 3rd party languages are allowed a little creativity to try truly crazy things... I'm not sure what the real harm is... since again we're just allowing sugar that builds on existing parser features.  So they can do it now if they want, just with helper functions.

I think this freedom to experiment with the "mode grammar" itself outside core might result in some really neat ideas... and hopefully some nice abstractions that perhaps could find their way back into core.

You could of course just make the core parser more and more complex but I'd really like to avoid this and keep it simple (when possible) and this seems a nice way to do that but would still allow all sorts of patterns we commonly see to be expressed much more simply.

Useful extensions over time would become "blessed" and then before part of the core library and all grammars would have access to them by default (in addition to being able to add their own).  At that time we'd of course add them to the docs, etc... 

This is a much nicer way of taking common patterns we're already using and making them easier to write.

joshgoebel:
Closes #2794.  Works towards #2520. Closes #2833.

### Changes

Implements #2833 with the key `compilerExtensions` on a grammar which is an 
array of extensions that takes the mode (and it's parent) as input.

Inside the compiler (for all grammars):

- More modular pipeline that users the new extensions internally also
- Adds support for `illegal` being an array of illegal expressions seamlessly compiled into a single regex

R grammar:

- Testing out the `beforeMatch` (alternative to look-behind) as a candidate for future inclusion in the compiler.

<del>It also adds two plugin hooks that take the same input:

- <del>before:compileEarly
- <del>before:compileLate

_The plugin hooks were removed for now._

Overall this allows grammars (and plugins) to add syntactic sugar to their own definitions.

### Checklist
- [x] Added markup tests, or they don't apply here
- [x] Added tests to guarantee plugin extensions are called properly by compiler
- [x] Document the new plugin options


joshgoebel:
Ping.

src/languages/r.js
```diff
@@ -128,47 +153,30 @@ export default function(hljs) {
       },
       {
         className: 'number',
+        relevance: 0,
+        beforeMatch: /([^a-zA-Z0-9._])/, // not part of an identifier
         variants: [
           // TODO: replace with negative look-behind when available
           // { begin: /(?<![a-zA-Z0-9._])0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*[pP][+-]?\d+i?/ },
           // { begin: /(?<![a-zA-Z0-9._])0[xX][0-9a-fA-F]+([pP][+-]?\d+)?[Li]?/ },
           // { begin: /(?<![a-zA-Z0-9._])(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?[Li]?/ }
```
allejo, line 162: Should this comment be updated? I see `(?<![a-zA-Z0-9._])` being removed down below

allejo (approved):
This looks good to me. Some comments have become dated though and should be updated for accuracy

src/lib/mode_compiler.js
```diff
@@ -300,10 +282,27 @@ export function compileLanguage(language) {
...
 
-    mode.keywords = mode.keywords || mode.beginKeywords;
+    [
+      EXT.beginKeywords,
+      // do this last so compiler extensions that come earlier have access to the
+      // raw array if they wanted to perhaps manipulate it, etc.
+      EXT.compileIllegal,
+      // default to 1 relevance if not specified
+      EXT.compileRelevance
+    ].forEach(ext => ext(mode, parent));
```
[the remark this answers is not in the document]
joshgoebel: last-ish, lol... I can reword.

src/languages/r.js
joshgoebel: No, it's moved to `beforeMatch`... if we get look-behind then the commented out code would be what we'd prefer to use.

src/lib/mode_compiler.js
```diff
@@ -320,20 +319,13 @@ export function compileLanguage(language) {
       throw new Error("ERR: Prefer `keywords.$pattern` to `mode.lexemes`, BOTH are not allowed. (see mode reference) ");
     }
 
+    keywordPattern = keywordPattern || mode.lexemes || /\w+/;
+
     // `mode.lexemes` was the old standard before we added and now recommend
     // using `keywords.$pattern` to pass the keyword pattern
-    cmode.keywordPatternRe = langRe(mode.lexemes || keywordPattern || /\w+/, true);
+    cmode.keywordPatternRe = langRe(keywordPattern, true);
```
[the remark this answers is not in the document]
joshgoebel: Agree. Done.

src/languages/r.js
allejo: ahhhh I see now, I had gotten the look-behind confused with what you moved to `beforeMatch` and are using in this language extension. me like

docs/plugin-api.rst
```diff
@@ -130,3 +130,41 @@ language
   The language determined from the class attribute (or undefined).
 
 It returns nothing.
+
+
+before:compile(mode, parent)
```
egor-rogov, line 135: There are actially `before:compileEarly` and `before:compileLate`, right?

src/lib/compiler_extensions.js
```diff
@@ -0,0 +1,83 @@
...
+ * `match` to match a single expression for readability
+ * @type {CompilerExt}
+ */
+export function compileMatch(mode, _parent) {
+  if (!mode.match) return;
+  if (mode.begin || mode.end) throw new Error("begin & end are not supported with match");
+
+  mode.begin = mode.match;
+  delete mode.match;
+}
```
egor-rogov, line 74: This new `match` mode should be documented.

src/lib/compiler_extensions.js
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Allow `illegal` to contain an array of illegal values
+ * @type {CompilerExt}
+ */
+export function compileIllegal(mode, _parent) {
+  if (!Array.isArray(mode.illegal)) return;
+
+  mode.illegal = regex.either(...mode.illegal);
+}
```
egor-rogov, line 62: It should be documented that `illegal` new accepts an array.

egor-rogov (commented):
It's really cool. The feature can be easily abused, but it's a powerful tool.
I mentioned some points regarding the documentation.

docs/plugin-api.rst
joshgoebel: I need to rework all the docs (which is why this isn't merged).  No one has asked for this as a plugin so I'm probably going to rip that out completely (plus I'm not sure a world where to install grammar xyz you also need plugin 123 and 234 is a great place to end up, so we can push off that decision into the future)... so the new capability is just that languages can define their own sugar... and then inside the library we can "test" new extensions in a single language first... and if that goes well they can be promoted to the compiler.

So these docs are probably moving into `compilerExtensions` inside the mode docs.

joshgoebel: Moved docs into the mode reference.

docs/mode-reference.rst
```diff
@@ -89,6 +91,50 @@ disableAutodetect
...
+sugar to make reading and writing grammars easier.  **Note: This is heavily
+dependent upon compiler internals and may NOT be stable from minor release to
+minor release.** *It is currently recommended only for 1st party grammars.* The
+intention is that we use grammars to "test" out new compiler extensions and if
+they perform well promote them into the core library.
+
+mode
+  The incoming mode object
+
+parent
```
egor-rogov, line 109: Perhaps `parentMode` is better, to match above `(mode, parentMode) -> {}`.

egor-rogov (approved):
I really like it.
The feature is great, the added docs are clear.

docs/mode-reference.rst
joshgoebel: Done.

PR #2836, opened 2020-11-06
chore(code) Stricter linting (and guarantee ES6 syntax for time being)

joshgoebel:
Get a little more serious about our linting...

### Changes

- Sets linter to ES6, not ES2018
- Adds GitHub action:
  - Enforces linting of main source (src/* and src/lib/*)
  - Lints grammar files (src/languages/*) just to make sure no ES2018 syntax has crept in

### Checklist
- [x] Updated the changelog at `CHANGES.md` (not needing, more of a code chore)


.eslintrc.lang.js
```diff
@@ -0,0 +1,18 @@
...
+    node: true,
+    mocha: true
+  },
+  extends: [],
+  globals: {
+    Atomics: "readonly",
+    SharedArrayBuffer: "readonly"
+  },
+  parserOptions: {
+    ecmaVersion: 2015,
```
puzrin, line 14: AFAIK, that's the same as `es6: true` (`parserOptions` in current state can be dropped)

joshgoebel: Thanks.  I believe it is, but I also like it being explicit.

joshgoebel:
Merging as there should be nothing controversial here and merging is going to avoid merge conflicts later since this changes a bunch of stuff all over just to make the linter happy.

PR #2844, opened 2020-11-11
(chore) Deprecate the `requireLanguage` API

### LINKED ISSUE #2275: Add `baseLangage` to language grammars
From my tweaked language checker:

```
┌────────────────────┬────────────────────┬──────────┬────────────────────┬──────────┬────────────────────┐
│ expected           │ actual             │ score    │ 2nd best           │ score    │ info               │
├────────────────────┼────────────────────┼──────────┼────────────────────┼──────────┼────────────────────┤
│ cpp                │ cpp                │ 16       │ arduino            │ 16       │ Relevance match.   │
├────────────────────┼────────────────────┼──────────┼────────────────────┼──────────┼────────────────────┤
│ cpp                │ cpp                │ 25       │ arduino            │ 25       │ Relevance match.   │
├────────────────────┼────────────────────┼──────────┼────────────────────┼──────────┼────────────────────┤
│ xml                │ xml                │ 46       │ django             │ 46       │ Relevance match.   │
├────────────────────┼────────────────────┼──────────┼────────────────────┼──────────┼────────────────────┤
│ xml                │ xml                │ 7        │ django             │ 7        │ Relevance match.   │
├────────────────────┼────────────────────┼──────────┼────────────────────┼──────────┼────────────────────┤
│ xml                │ xml                │ 3        │ django             │ 3        │ Relevance match.   │
```

These aren't surprising, but also aren't very helpful.  They would be reported as errors except for the load order that's enforced by the requires... and the fact that we magically prefer the "first match" even if the scores are the same.

I'd like to add an explicit key to explain these relationships to the parser.

So that Arduino would be a `superset: 'cpp'` and Django would be a `superset: 'xml'`.  I'm open to better naming.  If two languages are detected to be related in this way then the auto-detect would simply ignore the superset in cases of EQUAL matches, only preferred it if it had a HIGHER match (which is what you'd expect if it was the superset, and not the original).

Also _if you're looking to correct an incorrect match_ it's not helpful to know the second match to cpp was Arduino, they are essentially the same (not going to color differently, etc).  It would be useful to know the second match (OTHER than Arduino) was say Pascal.

That's the change I'm suggested here if no one is opposed.  Should only be a few lines of code and adding the appropriate attribute to a few grammars.

joshgoebel:
### Changes

- Entire library now no longer uses any run-time dependencies (no grammar will break because another is missing)
- Add docs for fixMarkup deprecation
- Add docs for requireLanguage deprecation
- Remove `Requires:` meta-data used only for run-time dependencies.
- Gzip size is only ~230 bytes larger than before 
- [x] Fix auto-detect tests now that the equivalency between `cpp` and `arduino` is no longer hidden by the load order. 
- adds `supersetOf` to resolve conflicts between very similar languages (C++, Arduino).  The base language wins in the case of a tie. ie `arduino` is now a `supersetOf` `cpp`. (closes #2275)

Note: `Requires: ` still stays overall as it's used to clue the build process about "loose" (non-breaking) sublanguage dependencies between languages.  This dependency is purposely loose and therefore still a thing.

### Checklist

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -608,6 +608,7 @@ const HLJS = function(hljs) {
       // relevance while preserving the original ordering - which is how ties
       // have historically been settled, ie the language that comes first always
       // wins in the case of a tie
+      return 0;
     });
 
     const [best, secondBest] = sorted;
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -787,6 +787,9 @@ const HLJS = function(hljs) {
     @returns {Language | never}
   */
   function requireLanguage(name) {
+    console.warn("requireLanguage is deprecated and will be removed entirely in the future.");
+    console.warn("Please see https://github.com/highlightjs/highlight.js/pull/2844");
+
     const lang = getLanguage(name);
     if (lang) { return lang; }
 
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -592,7 +592,7 @@ const HLJS = function(hljs) {
 
     const sorted = results.sort((a, b) => {
       // sort base on relevance
-      if (a.relevance != b.relevance) return b.relevance - a.relevance;
+      if (a.relevance !== b.relevance) return b.relevance - a.relevance;
 
       // always award the tie to the base language
       // ie if C++ and Arduino are tied, it's more likely to be C++
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -592,7 +592,17 @@ const HLJS = function(hljs) {
 
     const sorted = results.sort((a, b) => {
       // sort base on relevance
-      return b.relevance - a.relevance;
+      if (a.relevance != b.relevance) return b.relevance - a.relevance;
+
+      // always award the tie to the base language
+      // ie if C++ and Arduino are tied, it's more likely to be C++
+      if (a.language && b.language) {
+        if (getLanguage(a.language).supersetOf === b.language) {
+          return 1;
+        } else if (getLanguage(b.language).supersetOf === a.language) {
+          return -1;
+        }
+      }
 
       // otherwise say they are equal, which has the effect of sorting on
       // relevance while preserving the original ordering - which is how ties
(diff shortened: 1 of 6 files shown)


src/highlight.js
```diff
@@ -787,6 +798,9 @@ const HLJS = function(hljs) {
     @returns {Language | never}
   */
   function requireLanguage(name) {
+    console.warn("requireLanguage is deprecated and will be removed entirely in the future.");
+    console.warn("Please see https://github.com/highlightjs/highlight.js/pull/2844");
```
allejo, line 802: Should there be some sort of check here for checking whether highlight.js is running a production environment?

allejo (approved):
Looks fine to me

src/highlight.js
joshgoebel: The idea is that everyone can see it. It only shows up in console so it’s not disruptive. And after this PR it’s also immediately actionable so it’s not something that people have to wait till version 11. Someone can see it immediately make changes to that code base to make it go away. 

I’m pretty sure this is consistent with the other deprecation notices.

joshgoebel: I do certainly wish there were some better way to handle this but like I can’t really think of what it would be too of my head.

allejo: I don't really feel strongly about this either way. My initial thought was someone using a plug-in in their CMS that had a bundled version of highlight.js or something similar and fixing the warning was beyond their control until the plug-in maintainer updated it.

Closest thing I can think of is webpack's support for `process.env.node_env === 'production'` but I can't say I've ever bothered looking for a browser-only solution other than having `*.production.min.js` and `*.development.js` files

joshgoebel: There is minified or not (perhaps could have a flag based on the build type) but there is no guarantee one is using minified in production and un-minified for development work.  Unless you're hacking on the library you're probably just using the minified version everywhere.

Issue #2825, opened 2020-11-03
Proposal: Language "Champions"


joshgoebel:
**Is your request related to a specific problem you're having?**

It would be nice to have a way to spotlight people who go out of their way to maintain/make significant contributions to our first party grammars.  Also have a "go to" person to route language specific issues to that hopefully is a bit of an expert in that language (meaning they bring a ton of language context to the table that the core team doesn't have, etc).

We could have a list in AUTHORS.txt or make a new document... open to suggestions.

**The solution you'd prefer / feature you'd like to see added...**

Come up with a "title" of some sort, have an official list... I'm open to other ideas.  Any way to add a little "prestige" might be nice.

**Any alternative solutions you considered...**

Nope.

**Additional context...**

https://github.com/highlightjs/highlight.js/issues/1678

joshgoebel:
@egor-rogov  @allejo 

Pinging a few people who've made large grammar contributions lately.  @halirutan  @schtandard @klmr

klmr:
I like the idea of having a “go to” person, and I’d be happy to fulfil this role for the R language definition (plus potentially others in the future).

halirutan:
It was always my belief that when I put my name on it, I'm responsible for it. It doesn't mean that I have to do all future work/fixes, but I always tried to be approachable when issues pop up so that I can help or give advice. For the Wolfram Language, I have quite a collection of highlighters in my bag:

- [JPrettify](https://github.com/halirutan/JPrettify) that I needed back when Joomla was a thing
- [Google-Prettify](https://github.com/halirutan/WolframLanguage-Google-Prettify) which is used on the [official Wolfram Community site](https://community.wolfram.com/) and was used on StackExchange until 2020
- [Rouge](https://github.com/rouge-ruby/rouge/pull/854) which is a highlighter for Jekyll
- [an IntelliJ plugin for the Wolfram Language](https://wlplugin.halirutan.de/) which of course is much more than a simple highlighter and which has a full-fledged parser and can do context-sensitive highlighting.
- [And now Highlight.js](https://github.com/highlightjs/highlight.js/pull/2706)

That being said, I don't need a champions-club to feel responsible but it surely is a very nice idea. The big problem with such things is always where to put it so that anyone even sees it. Because right now, we already have a champions-club and it's called `AUTHORS.txt` 😂 

![image](https://user-images.githubusercontent.com/511683/98433792-a3e35780-20ca-11eb-8226-369940054f2d.png)

[...]

Issue #2578, opened 2020-05-26
Why do SQL keywords show up in my compiled javascript?


raniesantos:
This is the code I use to register languages.

```js
import hljs from 'highlight.js/lib/core';

[
    'xml', 'css', 'scss', 'javascript', 'json',
    'php', 'php-template', 'apache', 'yaml',
    'markdown', 'handlebars', 'bash', 'shell',
].forEach((language) => {
    hljs.registerLanguage(language, require(`highlight.js/lib/languages/${language}`));
});

hljs.initHighlightingOnLoad();
```

This is a screenshot of the content of the compiled js file.

![Screenshot from 2020-05-26 14-32-41](https://user-images.githubusercontent.com/8528269/82867667-d6a2f480-9f5d-11ea-8481-03cd1d7510c7.png)

yyyc514:
Are you referring to just COMMENT?  Rest of it looks pretty normal to me.

yyyc514:
Wait is that a screen shot of a text editor or how HLJS is highlighting the content? That's unclear from your phrasing.

raniesantos:
2nd is a text editor screenshot.

raniesantos:
Basically, I took a peek at the compiled js and saw that this block of code is there even though I didn't register pgsql.

yyyc514:
We can't highlight SQL if we don't include a list of keywords in the compile Javascript.  That is normal.

raniesantos:
I don't have SQL in the languages I looped through though. Why is that?

yyyc514:
Hard to say.  I'd say you are pulling in the whole shebang somewhere else in your build process.

yyyc514:
Try removing pieces one by one and see if you can find what element is pulling it in.

raniesantos:
Yeah I'll double check later, I'm not at my machine at the moment.

yyyc514:
Sorry for being so spacey at first I didn't read your question clearly. :)

raniesantos:
Removing the loop and having multiple language registration lines fixed it.

But it bothers me that I can't loop this repeating code.

I'm going to try other ways of looping through it for now.

yyyc514:
It's possible your build system can't deal with the dynamic way you're loading that... many of them just statically analyze the code... so you can't "run" a loop during compiling.

yyyc514:
Closing.  Doesn't not seem to be a Highlight.js specific issue (looks like build system issue).  If you learn more, we can re-open.

[...]

PR #2845, opened 2020-11-11
(swift) improvements: support `some`, `@main`; Fix `#` keywords .

mportiz08:
General enhancements and a bug fix for highlighting Swift code.

### Changes
Adds highlighting support for new Swift 5.1 and 5.3 features that are heavily utilized in SwiftUI code:

1. [`some` keyword][1] (Introduced in Swift 5.1)
2. [`@main` attribute][2] (Introduced in Swift 5.3)

Fixes a bug where keywords that start with the `#` character (like `#endif` as an example) are not matched.

[1]: https://github.com/apple/swift-evolution/blob/main/proposals/0244-opaque-result-types.md
[2]: https://github.com/apple/swift-evolution/blob/main/proposals/0281-main-attribute.md

### Checklist
- [X] Added markup tests
- [X] Added myself to `AUTHORS.txt`, under Contributors

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -307,3 +307,4 @@ Contributors:
 - Florian Bezdeka <florian@bezdeka.de>
 - Patrick Scheibe <patrick@halirutan.de>
 - Kyle Brown <kylebrown9@github>
+- Marcus Ortiz <mportiz08@gmail.com>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,9 @@ Language Improvements:
   - Matching of named-characters aka special symbols like `\[Gamma]`
   - Updated list of version 12.1 built-in symbols
   - Matching of patterns, slots, message-names and braces
+- fix(swift) Handle keywords that start with `#` [Marcus Ortiz][]
+- enh(swift) Match `some` keyword [Marcus Ortiz][]
+- enh(swift) Match `@main` attribute [Marcus Ortiz][]
 
 Dev Improvements:
 
@@ -52,6 +55,7 @@ New themes:
 [Michael Rush]: https://github.com/rushimusmaximus
 [Patrick Scheibe]: https://github.com/halirutan
 [Kyle Brown]: https://github.com/kylebrown9
+[Marcus Ortiz]: https://github.com/mportiz08
 
 
 ## Version 10.3.1
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -10,6 +10,9 @@ Category: common, system
 
 export default function(hljs) {
   var SWIFT_KEYWORDS = {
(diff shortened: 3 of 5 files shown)


joshgoebel:
@svanimpe Could you take a quick look here.  Does this look good to you?  I think the changes look good, I just don't know the language itself.

@mportiz08  https://github.com/highlightjs/highlight.js/issues/2847

Do you know anything about `c`?   If not and you agree would you mind yanking that out as part of this PR also?

svanimpe:
I wonder if @mattt would be willing to help out here. Perhaps https://github.com/SwiftDocOrg/swift-doc can be used to generate a list of global functions in the Standard Library?

joshgoebel:
There are a number of Swift issues open right now (many of them of the more trivial variety) if a Swift champion were to show up on the scene. :)  https://github.com/highlightjs/highlight.js/issues/2825

joshgoebel:
> Perhaps https://github.com/SwiftDocOrg/swift-doc can be used to generate a list of global functions in the Standard Library?

I do worry that what we have now is perhaps someones list of "frequently used" rather than ever having been "all globals" (I really have no idea though), but now I'm curious.  Swift is shipped as part of the `common` group so that makes me think about it's ultimate size perhaps a bit more than other grammars...  like if we were suddenly going to add 1000 keywords that might be something we need to look at.  Right now though Swift is a pretty small grammar, so we certainly have room to grow.

joshgoebel:
If we're going to totally rewrite the keywords section overall I'd save that for another PR though and also switch to the new one keyword per line array style to make future maintenance easier.

mportiz08:
> Do you know anything about `c`? If not and you agree would you mind yanking that out as part of this PR also?

I don't, but that is really interesting/surprising—I'm honestly not aware of `c` being a keyword on its own. It looks like it was introduced with e5c1b0c59014a9f35c55fb7de0bcc29ab4b752a6.

joshgoebel:
@natecook1000  You added `c` to the list of Swift keywords in e5c1b0c.  Can you shed any light? :)

joshgoebel:
Since we don't have any immediate answer on `c` I don't see a need for that issue to hold up this work.  I glanced at the links provided and this looks entirely reasonable. 

Thanks for the contribution!

joshgoebel (approved):

joshgoebel:
If you'd perhaps like to continue helping out with Swift: #2578 #2819

Now quite as low-hanging fruit, but close.

svanimpe:
W.r.t. the built-ins: I've noticed that methods like `Array.count` also highlight because `count` is listed as a built-in, even though it's not a global function anymore.

joshgoebel:
Well if the language has changed then count be perhaps removed but ultimately this will be solved when the Swift grammar adds property access and method dispatch since that rule matching would win prevent the keyword from matching, etc.

PR #2850, opened 2020-11-12
chore(build) NPM publish of cdn-assets and highlight.js via GHA

joshgoebel:
### Changes

- GitHub action to build the library and publish to NPM (whether beta or release)
- Also updates `cdn-release` and publishes to NPM  (`cdn-assets`)

### Checklist

- [x] Polish (final cleanups)
- [x] When not dev/beta should use the appropriate `10-stable` branch, etc.
- [x] Test pushing real tag to highlight.js (10.4.0-beta0 or some such) and see if it goes all the way thru the process
- [ ] Which identify to use for secret tokens (GitHub and NPM)?  (if we need a new identity a separate issue may be spun off to track that). **Created new issue.**
- [ ] Before using for release we'll need to alter the website release scripts (works for beta since the website build scripts ignore betas).  **Understood.**

---

- [x] Updated the changelog at `CHANGES.md` (not needed)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,133 @@
+name: Release
+
+on:
+  # pull_request:
+    # branches: [ master ]
+  push:
+    tags:
+      - "*beta*"
+      - "*pre*"
+      # - "1[0-9]+.[0-9]+.[0-9]+"
+
+jobs:
+  prerelease:
+    name: Pre-release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout highlight.js
+        uses: actions/checkout@v2
+
+      - name: Tag is ${{ github.ref }}.
+        # we have to repeat ourselves here since the environment is not actually updated
+        # until the next step executes, so we can't access $TAG_NAME yet
+        run: |
+          echo "TAG_NAME=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+          echo "MAJOR_VERSION=$(echo ${GITHUB_REF/refs\/tags\//} | cut -d'.' -f1)" >> $GITHUB_ENV
+      - name: Make sure we are pushing a tag...
+        if: ${{ !contains(github.ref,'refs/tags/') }}
+        run: false
+        # run: echo "TAG_NAME=0.0.0-test0" >> $GITHUB_ENV
+
+      - if: ${{ contains(github.ref, 'beta') || contains(github.ref, 'pre') }}
+        run: |
+          echo "NPM_TAG=beta" >> $GITHUB_ENV
+          echo "RELEASING=beta" >> $GITHUB_ENV
+
+      - if: ${{ !(contains(github.ref, 'beta') || contains(github.ref, 'pre')) }}
+        run: |
+          echo "NPM_TAG=latest" >> $GITHUB_ENV
+          echo "RELEASING=stable" >> $GITHUB_ENV
+
+      - name: match-tag-to-package-version
+        uses: geritol/match-tag-to-package-version@0.0.2
+        env:
+          TAG_PREFIX: refs/tags/ # Optional, default prefix refs/tags/
+
+      - name: Use Node.js 15.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 15.x
+      - name: Build Node.js package
+        run: |
+          npm install
+          node ./tools/build.js -t node
+          npm test
+
+      - name: Publish highlight.js to NPM
+        id: publish
+        uses: JS-DevTools/npm-publish@v1
+        with:
+          check-version: true
+          token: ${{ secrets.NPM_TOKEN }}
+          package: ./build/package.json
+          tag: ${{ env.NPM_TAG }}
+
+      - if: steps.publish.outputs.type != 'none'
+        run: |
+          echo "Version changed: ${{ steps.publish.outputs.old-version }} => ${{ steps.publish.outputs.version }}"
+
+        # if stable release
+      - name: Stable Release
+        if: env.RELEASING == 'stable'
+        run: echo "BRANCH_NAME=${MAJOR_VERSION}-stable" >> $GITHUB_ENV
+        # else (beta)
+      - name: Beta Release
+        if: env.RELEASING == 'beta'
+        run: echo "BRANCH_NAME=master" >> $GITHUB_ENV
+      - name: Confirm release is either stable or beta
+        if: ${{ !(env.RELEASING == 'stable' || env.RELEASING == 'beta') }}
+        run: |
+          echo We seem to be releasing `${RELEASING}`.
+          false
+
+      - name: Checkout cdn-release
+        uses: actions/checkout@v2
+        with:
+          repository: 'highlightjs/cdn-release'
+          path: 'cdn-release'
+          token: ${{ secrets.CDN_REPO_TOKEN }}
+          ref: ${{ env.BRANCH_NAME }}
+
+      - name: Build CDN package
+        run: node ./tools/build.js -t cdn :common
+
+      - name: Commmit & Push cdn-release ${{ env.TAG_NAME }}
+        working-directory: ./cdn-release
+        run: |
+          rm -r ./build
+          mv ../build/ ./build/
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add ./build/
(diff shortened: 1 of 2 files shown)


joshgoebel:
@allejo I plan to add back the commented out lines (to make sure the tag matches package.json and to run tests)... but any other thoughts?

This is just the core NPM library... CDN repo is still a whole other ball of wax.

joshgoebel:
Screen Shot 2020-11-11 at 11 23 54 PM

I'm not sure if we just use github action or if we need to create a whole new GitHub identity for this... it bothers me a little that the personal token I had to use to get this to work belongs to me personally... I think for best practices sake we'd want a token from a user that ONLY had access to the single repo `cdn-assets`, right?

The same problem seems to exist for NPM where tokens are also per-person.

Thoughts?

.github/workflows/release.yml
```diff
@@ -0,0 +1,109 @@
+name: Release
+
+on:
+  pull_request:
+    branches: [ master ]
```
joshgoebel, line 5: This of course comes out, was just using for testing.

joshgoebel:
@isagalaev I'm also considering switching us to `v10.0.0` tags for versions... (adding the `v`)... I was reading thru your website build scripts right now and I actually they might just work:

- You use `parse_version` which returns a canonical version (to a degree).  IE, when you pass in `v10` it knows how to deal with that and when you ask for `str(version)` (which is what you pass down to most things) the `v` is gone.
- The fact that you bail for beta versions (checking `is_prerelease`).  There would have been issues with `10.0.0a0` vs `10.0.0` (tag vs number from CHANGES.md) - not to mention the canonical short version not matching the long version (a0 vs alpha0). But since you don't care about betas, it's irrelevant.
- Actually we may need a tiny change for where you do `git checkout $version` to put the `v` back in front.

Does this sound about right?

joshgoebel:
Ok I deployed 10.4.0-beta0 with this, seems to work pretty well now. :-)

.github/workflows/release.yml
```diff
@@ -0,0 +1,133 @@
...
+        uses: actions/checkout@v2
+
+      - name: Tag is ${{ github.ref }}.
+        # we have to repeat ourselves here since the environment is not actually updated
+        # until the next step executes, so we can't access $TAG_NAME yet
+        run: |
+          echo "TAG_NAME=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+          echo "MAJOR_VERSION=$(echo ${GITHUB_REF/refs\/tags\//} | cut -d'.' -f1)" >> $GITHUB_ENV
+      - name: Make sure we are pushing a tag...
+        if: ${{ !contains(github.ref,'refs/tags/') }}
```
allejo, line 27:
```suggestion
        if: !contains(github.ref,'refs/tags/')
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,133 @@
...
+        run: false
+        # run: echo "TAG_NAME=0.0.0-test0" >> $GITHUB_ENV
+
+      - if: ${{ contains(github.ref, 'beta') || contains(github.ref, 'pre') }}
```
allejo, line 31:
```suggestion
      - if: contains(github.ref, 'beta') || contains(github.ref, 'pre')
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,133 @@
...
+        run: |
+          echo "NPM_TAG=beta" >> $GITHUB_ENV
+          echo "RELEASING=beta" >> $GITHUB_ENV
+
+      - if: ${{ !(contains(github.ref, 'beta') || contains(github.ref, 'pre')) }}
```
allejo, line 36:
```suggestion
      - if: !(contains(github.ref, 'beta') || contains(github.ref, 'pre'))
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,133 @@
...
+        # if stable release
+      - name: Stable Release
+        if: env.RELEASING == 'stable'
+        run: echo "BRANCH_NAME=${MAJOR_VERSION}-stable" >> $GITHUB_ENV
+        # else (beta)
+      - name: Beta Release
+        if: env.RELEASING == 'beta'
+        run: echo "BRANCH_NAME=master" >> $GITHUB_ENV
+      - name: Confirm release is either stable or beta
+        if: ${{ !(env.RELEASING == 'stable' || env.RELEASING == 'beta') }}
```
allejo, line 78:
```suggestion
        if: !(env.RELEASING == 'stable' || env.RELEASING == 'beta')
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,133 @@
+name: Release
+
+on:
+  # pull_request:
+    # branches: [ master ]
```
allejo, line 5: What are your intentions of having this run on PRs against master? Or was this just for debugging?

.github/workflows/release.yml
```diff
@@ -0,0 +1,133 @@
...
+      - name: Checkout highlight.js
+        uses: actions/checkout@v2
+
+      - name: Tag is ${{ github.ref }}.
+        # we have to repeat ourselves here since the environment is not actually updated
+        # until the next step executes, so we can't access $TAG_NAME yet
+        run: |
+          echo "TAG_NAME=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+          echo "MAJOR_VERSION=$(echo ${GITHUB_REF/refs\/tags\//} | cut -d'.' -f1)" >> $GITHUB_ENV
+      - name: Make sure we are pushing a tag...
```
allejo, line 26: Debugging purposes or when would this workflow run when not a tag?

allejo (commented):
> I'm not sure if we just use github action or if we need to create a whole new GitHub identity for this... it bothers me a little that the personal token I had to use to get this to work belongs to me personally... I think for best practices sake we'd want a token from a user that ONLY had access to the single repo `cdn-assets`, right? If we do create a whole new identify then it might make sense to use that for these commits.
> 
> The same problem seems to exist for NPM where tokens are also per-person.
> 
> If we start needing a few more things like this is anyone opposed to 1Password? [1Password/1password-teams-open-source](https://github.com/1Password/1password-teams-open-source) Then anyone needing access to these things could be added to 1Password and any team secrets would be stored there. I don't expect this to be a frequent need at all, just I hate to own all the secrets myself or have them tied to me personally.

+1

Personal Access Tokens (PATs) are tied to your account so if ever stolen it'll have access to all other orgs you belong to. So definitely in favor of a dedicated service account.

I'm cool with using 1Password for Open Source! I didn't know they offered this :heart:

> @allejo I plan to add back the commented out lines (to make sure the tag matches package.json and to run tests)... but any other thoughts?

Overall this looks good! I really like this, thanks so much for doing this!

.github/workflows/release.yml
joshgoebel: It doesn’t like this and my editor color is it funny I think that some type of weird YAML thing.

.github/workflows/release.yml
joshgoebel: This one can be removed because it doesn’t have the!

.github/workflows/release.yml
joshgoebel: No intention it’ll be ripped out that’s what I was testing it before I switch to pushing actual tags.

.github/workflows/release.yml
joshgoebel: It’s just a fail safe.

.github/workflows/release.yml
joshgoebel: Or is the trick that you surround the whole expression in parentheses I’m not sure I tried that

allejo: I think it may just be some weird highlighting with editors, according to GitHub's spec...

> When you use expressions in an if conditional, you may omit the expression syntax (`${{ }}`) because GitHub automatically evaluates the if conditional as an expression.
>
> Source: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions

joshgoebel: Yes you told me that and sometimes it works but yet most of the examples I see even on GitHub’s on site I have the brackets all the time.

joshgoebel: I spent way too long on this one so I’m inclined to leave it alone for now since it works.

PR #2852, opened 2020-11-12
fix(misc) Misc fixes from looking at LGTM

joshgoebel:
### Changes

- minor linting fixes
- fix(r) fixed keywords not properly spaced
- fix(javascript) final solid fix for runaway regex
- fix(livescript) fix potential catastrophic backtracking

### Checklist

### DIFF
--- a/src/languages/livescript.js
+++ b/src/languages/livescript.js
@@ -115,7 +115,7 @@ export default function(hljs) {
         {
           // regex can't start with space to parse x / 2 / 3 as two divisions
           // regex can't start with *, and it supports an "illegal" in the main mode
-          begin: /\/(?![ *])(\\\/|.)*?\/[gim]*(?=\W)/
+          begin: /\/(?![ *])(\\.|[^\\\n])*?\/[gim]*(?=\W)/
         }
       ]
     },
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,9 @@ New Languages:
 
 Language Improvements:
 
+- fix(r) fixed keywords not properly spaced (#2852) [Josh Goebel][]
+- fix(javascript) fix potential catastrophic backtracking (#2852) [Josh Goebel][]
+- fix(livescript) fix potential catastrophic backtracking (#2852) [Josh Goebel][]
 - bug(xml) XML grammar was far too imprecise/fuzzy [Josh Goebel][]
 - enh(xml) Improve precision to prevent false auto-detect positives [Josh Goebel][]
 - fix(js/ts) Prevent for/while/if/switch from falsly matching as functions (#2803) [Josh Goebel][]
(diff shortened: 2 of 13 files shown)


joshgoebel:
@allejo @egor-rogov If either of you find more time I created a board of what's holding up 10.4:

https://github.com/highlightjs/highlight.js/projects/2

egor-rogov (approved):
Look good

src/languages/livescript.js
```diff
@@ -115,7 +115,7 @@ export default function(hljs) {
         {
           // regex can't start with space to parse x / 2 / 3 as two divisions
           // regex can't start with *, and it supports an "illegal" in the main mode
-          begin: /\/(?![ *])(\\\/|.)*?\/[gim]*(?=\W)/
+          begin: /\/(?![ *])(\\.|[^\\\n])*?\/[gim]*(?=\W)/
```
allejo, line 118: I'm having a hard time understanding this change?

- `\\\/|.` => `\/` OR any character
- `(\\.|[^\\\n])` => `\` followed by any character OR any character except `\` or newline

Right? Or what's going on here?

joshgoebel: The intention here is to properly deal with escape sequences so that the regex isn’t detected to terminate prematurely.  

https://lgtm.com/projects/g/highlightjs/highlight.js/snapshot/b9e6761338f38476f7bd97f167f4033cb9c29789/files/src/languages/livescript.js?sort=name&dir=ASC&mode=heatmap#x22e678cf171c270f:1

I’m still coming to grasp with some of these complicated scenarios but the problem with the original regular expression is that there are two different ways that it could match what looks like a potential terminating forward slash. It could match up with the first literal match or you can also match with the dot / match. You have to remember it’s not just linear. Every time it hits that either or it has to explore both branches. So you’re creating a tree effectively.  And what happens if you have a bunch of the wrong pattern in a row is that tree just gets deeper and deeper and deeper and the number of possibilities scales towards infinity.  

The reason these are insidious is that computer today are so fast and if ultimately there is a match to be found then hopefully it finds it before the recursion gets too deep. The problem comes when there is no match so it’s stuck looking forever and then you get too deep and it’s game over. 

You can explore this with regex101 If you put it in PHP mode and often you have to add a $ to the very end to force it to lock the end of the regex forcing it down a road where its sure to fail which will lock it up. 

The second expression removes this ambiguity. If we find a backslash then we always consume the next character after it. 
And if we don’t find a backsplash then we allow everything except \ or carriage return. The disallowing the backslash is a key part.   This prevents the decision tree where it could go to different directions.

joshgoebel: https://www.rexegg.com/regex-explosive-quantifiers.html

joshgoebel: Basically you have to be really careful with | or qualifiers since  quantifier is really just a complex or.

allejo: I got the gist of the potential problem with the original regex here in theory, but I'd be lying if I said I actually understand this in-depth, so I'll trust LGTM and you on this :smile:

allejo (approved):

PR #2853, opened 2020-11-12
(chore) eslint --fix against most grammars

### LINKED ISSUE #2854: Cleanup: Bring all the grammars up to a standard level of Lint compliance
This is part of the larger task of auditing all the grammars for incorrect or exponential backtracking regex.  To make those PRs simpler and easier to review we first clean them up with `eslint --fix`.

I've been against these type of sweeping changes in the past without good reason, but if we have to touch every single file anyways to clean up the regex (so it can be properly audited) then we may as well lint them first to make that job easier.

- [x] A-F https://github.com/highlightjs/highlight.js/pull/2853
- [x] G, H, I
- [x] J, K, L
- [x] M, N, O
- [x] P, Q, R
- [x] S, T, U
- [x] V, W, X
- [x] Y, Z

joshgoebel:
Closes #2854.

### Changes

- Runs `eslint --fix` on all grammars that have no outstanding PRs
- [x] Add markup test for every language (to regression test this and more importantly the coming changes) - #2862 needs to be merged first and then this rebase

### Checklist


joshgoebel:
This was mostly automated work with just some tiny manual cleanups.

egor-rogov (approved):
OK
(I haven't look through all the changes, just several files)

PR #2861, opened 2020-11-13
(shell) Support multiline commands

### LINKED ISSUE #2860: (shell) multi-line command (using line continuation \) are not highlighted correctly
**Describe the issue**

Multi-line command (using line continuation `\`) are not highlighted correctly.

`shell`

**Sample Code to Reproduce**

```
$ docker run \
      --publish=7474:7474 --publish=7687:7687 \
      --volume=/neo4j/data:/data \
      --volume=/neo4j/plugins:/plugins \
      --volume=/neo4j/conf:/conf \
      --volume=/logs/neo4j:/logs \
      --user="$(id -u neo4j):$(id -g neo4j)" \
      --group-add=$groups \
      neo4j:3.4
```

**Expected behavior**

The entire command should be highlighted. Currently, only ` docker run \` is recognized.

Mogztter:
resolve #2860 

### Changes

Add support for multi-line commands using the line continuation `\`.

### Checklist
- [x] Added markup tests


Mogztter:
> ```
> end: '(?=[^\\\\]\\s*$)', subLanguage: 'bash'
> ```
> 
> Does that fix the test?

I think you have the right idea but this regular expression does not work. It actually excludes the last character.
The following:

```
$ /bin/sh
```

now produces:

```html
<span class="hljs-meta">$</span><span class="bash"> /bin/s</span>h
```
(notice the letter "h" is outside the "bash" span)

Mogztter:
@joshgoebel I think we need to use a negative lookbehind. Basically the line must not end with a line continuation, so the end of a line must not be preceded by zero or more spaces and the character "\\".
Unfortunately, lookbehind assertions are not supported in a few browsers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Browser_compatibility

joshgoebel:
So let's include the last character and use the look-ahead to find the end of line.

```js
end: /[^\\](?=\s*$)/,
```

Thoughts?

Mogztter:
> Thoughts? I pushed my changes.

Oh nice, if it works then :+1: 
I will add a few tests to make sure that it's working as expected (and also rebase/resolve the conflict) then we are good to go!

test/markup/shell/command-continuation-alt.txt
```diff
@@ -0,0 +1,5 @@
+> /bin/cat \.travis.yml\
+ -b | head -n1
+     1	language: node_js
+> echo 'hello'
+hello
```
joshgoebel, line 5: How is this diff from the first one?  If there are just two tests for the same thing you don't need a second file, just make one larger test.

Mogztter: There's no space before `\` (line continuation) and there's a `\` as part of the command (not at the end of the line).
I also added a simple command at the end to make sure that the regular expression is not too greedy.

I merged everything in `command-continuation.txt`.

joshgoebel (approved):

CHANGES.md
```diff
@@ -32,6 +23,7 @@ New Languages:
 
 Language Improvements:
 
+- enh(shell) Add support for multiline commands with line continuation `\` (#2861) [Guillaume Grossetie][]
```
Mogztter, line 26: @joshgoebel I'm confused, this language improvement is not part of the 10.4.0-beta1 right?

joshgoebel: Correct really they heading should probably change. Betas aren’t the same as releases though they’re rolling so things just keep getting added until we have a release

joshgoebel: So really the header and should be 10.4 rolling or some such. I’ll give it some thought. I only updated it because we have to update the header for production releases but that’s not true for beta releases so we have more flexibility there.

PR #2865, opened 2020-11-15
enh(autodetect) tcl: improve autodetection

joshgoebel:
Pull TCL auto-detect changes into their own PR for closer inspection

### Checklist


joshgoebel:
Refer to thoughts here: https://github.com/highlightjs/highlight.js/pull/2745#discussion_r523469461

joshgoebel:
Screen Shot 2020-12-18 at 4 49 42 AM

Upon taking a closer look at this I'm not sure array index should be part of variable at all... line 2 there looks "correct" to me... with () being unhighlighted and 1 being highlighted as a number.

I'm less sure what the correct answer is when the variable is wrapped in `{}`:

```
${ns::other::name(32)}
```

joshgoebel:
Or perhaps include number inside ${}:

Screen Shot 2020-12-18 at 5 30 10 AM

joshgoebel:
I've pretty much lost the thread here but I think the original idea was the make the rules more precise and less likely to match false-positives... plus the rules are pretty unusual as written now IMHO...

allejo:
Sorry I'm just getting around to this one, what do you want to do this? Just classify everything as variables?

> Or perhaps include number inside ${}:
> 
> Screen Shot 2020-12-18 at 5 30 10 AM

I think this one looks like it makes the most sense

sa-orlov (approved):

PR #2870, opened 2020-11-15
enh(asciidoc) support unconstrained syntax

### LINKED ISSUE #2869: (asciidoc) Unconstrained text formatting is not correctly highlighted
**Describe the issue**

AsciiDoc supports constrained and unconstrained syntax. For instance, when using the bold syntax, you can mark a word or phrase as bold by enclosing it in a single set of asterisks (e.g., \*word*) (constrained). You can mark bounded characters (i.e., characters within a word) as bold by enclosing them in a set of double asterisks (e.g., char\**act**ers) (unconstrained).

`asciidoc`

**Sample Code to Reproduce**

```
bold **un**constrained text.

italic __un__constrained text.

monospace ``un``constrained text.
```

**Expected behavior**

It should produce the following result:

```html
bold <span class="hljs-strong">**un**</span>constrained text.

italic <span class="hljs-emphasis">__un__</span>constrained text.

monospace <span class="hljs-code">``un``</span>constrained text.
```

**Additional context**

https://asciidoctor-docs.netlify.app/asciidoc/current/text/

Mogztter:
### Changes

resolves #2869

### Checklist
- [x] Added markup tests


src/languages/asciidoc.js
```diff
@@ -18,8 +18,8 @@ export default function(hljs) {
         '^/{4,}\\n',
         '\\n/{4,}$',
         // can also be done as...
-        //'^/{4,}$',
-        //'^/{4,}$',
+        // '^/{4,}$',
+        // '^/{4,}$',
```
joshgoebel, line 22: Is this related to the unconstrained change

src/languages/asciidoc.js
```diff
@@ -104,7 +104,13 @@ export default function(hljs) {
         begin: '^(NOTE|TIP|IMPORTANT|WARNING|CAUTION):\\s+',
         relevance: 10
       },
-      // inline strong
+      // inline unconstrained strong
+      {
+        className: 'strong',
+        begin: /\*{2}/,
```
joshgoebel, line 110: Is it not possible to escape an * here?

src/languages/asciidoc.js
Mogztter: Not really, eslint was unhappy with the missing space after `//` but not mandatory I can revert this change if you want.

joshgoebel: Oh comments, duh... I missed that at just a glance... no this is fine as it's minor.

src/languages/asciidoc.js
Mogztter: you can escape but I'm not sure I quite understand the existing rule:

```js
// allow escaped asterisk followed by word char
contains: [
  {
    begin: '\\\\*\\w',
    relevance: 0
  }
]
```

The following should not be highlighted:

```
this is a \*strong* word.
```

```
this is a str*\*on**g word.
```

But the last example is pretty uncommon (and fragile). In this case, it would be better to use substitution:

```
:stars: **
this is a str{stars}on{stars}g word.
```

Since the AsciiDoc specification is under-way, I would ignore edge cases for now because I'm not even sure what is the expected result.

Here's a few examples, with the input and the rendered result:

escaped

more

As you can see it's quite complex...

src/languages/asciidoc.js
```diff
@@ -104,7 +104,13 @@ export default function(hljs) {
...
+        end: /(\n{2}|\*{2})/
+      },
```
joshgoebel, line 112: Also I think we're going to need a variant here for the \n possibility, it's too "wide"... it can match too much, making it not worthy of relevance.
```suggestion
      {
        className: 'strong',
        begin: /\*{2}/,
        variants: [
          { end: /\n{2}/, relevance: 0 },
          { end: /\*{2}/ }
        ]
        
      },
```

joshgoebel: Hmmm, although now that I think of it `** anything **` is also VERY broad... can these be used without text?  ALl your examples show them surrounding text immediately if so we should add that as a condition here with a look ahead:

```
begin: /\*{2}(?=\w+)/
```

Thoughts?

src/languages/asciidoc.js
joshgoebel: > I'm not sure I quite understand the existing rule:

It's simply allowing \* inside of bold so that it doesn't trigger "end bold".  I'm not sure why the requirement of "one character after"... that might be specific to asciidoc.  I brought it up since we already handle this for unconstrained so it seems we should also for constrained unless for some reason it's only really used in one type but not the other.

> with the input and the rendered result:

We're less interested in the rendered result.  _We are not a rendering engine._  We're trying to highlight the semantics of the language.  So for some things that means strong won't be strong at all, it might be red or blue.

```
this is a str**o**n**g word.
```

This would be broken with your current PR as the `**g` would be bold (if we can find two linebreaks anywhere else).

```
this is a \*strong* word.
```

We'd need to add a general rule to eat a `\*` sequence then when it's found to avoid triggering any of the other rules.  This should be a test case no doubt.

---

We should really start here with which cases are reasonable to handle, add markup tests. Then see if we can make those tests green. :)

src/languages/asciidoc.js
Mogztter: since it's unconstrained, you can have spaces, a single newline...:

```
A sentence ** with words, and more words!  
using bold style :)**.
```

result

src/languages/asciidoc.js
joshgoebel: The markup tests should include the negative cases too, to make sure we properly handle them, etc...

src/languages/asciidoc.js
joshgoebel: Yes, I glanced at your link now and roughly understand they are different.  Here I'm speaking only of auto-detect... we don't just need to know how to highlight Asciidoc but we also have to avoid confusing it with say Javascript... so every rule (that's too generic) needs to have it's relevance tuned or eliminated to avoid false positives.  For example a JS snippet with tons of `**` math formulas might be flagged as Asciidoc... and that should be avoided if possible.

But lets focus on getting them working first I guess. :)

src/languages/asciidoc.js
Mogztter: > We should really start here with which cases are reasonable to handle, add markup tests. Then see if we can make those tests green. :)

💯 I will add tests.

> We're less interested in the rendered result. We are not a rendering engine. 

I know, it was just easier to share every variations.

> The markup tests should include the negative/edge cases too, to make sure we properly handle them, etc...

👍

src/languages/asciidoc.js
```diff
@@ -110,16 +110,16 @@ export default function(hljs) {
...
       },
       // inline constrained strong
       {
         className: 'strong',
         // must not follow a word character or be followed by an asterisk or space
         // must not be preceded by the escape character \
-        begin: '(?<!\\\\)\\B\\*(?![\\*\\s])',
-        end: '(\\n{2}|\\*)',
+        begin: /(?<!\\)\B\*\w/,
+        end: /\*/,
```
joshgoebel, line 122: Nope, can't use look-behind, not supported by all browsers yet. Sorry.  You could look at making this a chain (with `starts`) or rebase onto #2834 and use the syntactic sugar to the same end.  But none of this is true look-behind... we can only look-forward at this time.  So in order to do something like this you'll need to consume the prior character (ie. another mode can't have already consumed it).

joshgoebel: > or rebase onto #2834 and use the syntactic sugar to the same end.

That's what I'd recommend if you know how.

joshgoebel: Or like I suggested earlier capture the `\*` pattern as it's own previous rule (to prevent this rule from firing).

Mogztter: > Or like I suggested earlier capture the \* pattern as it's own previous rule (to prevent this rule from firing).

:+1:

Mogztter: Please note that I'm only updating strong rules.
Monospaced and emphasis are quite similar but first I want to make strong rules right (then apply the same rules to other formatting marks).

src/languages/asciidoc.js
```diff
@@ -113,12 +113,17 @@ export default function(hljs) {
         begin: /\*{2}[\w\s\n]+/,
         end: /\*{2}/
       },
+      // escaped formatting marks
+      {
+        begin: /\\/,
+        end: /[*_`]/
```
joshgoebel, line 119: This should all be a single `begin` regex... you don't want to match `\\blah blah blah*`.

joshgoebel (changes requested):

src/languages/asciidoc.js
Mogztter: Initially that's what I did but I got the following exception:

```
Error: 0 width match regex
```

I'm now using `begin` with a `contains`. I added additional tests to cover these cases:

```
\\blah blah *blah*.

\\blah blah blah*.
```

src/languages/asciidoc.js
```diff
@@ -107,18 +107,38 @@ export default function(hljs) {
         begin: '^(NOTE|TIP|IMPORTANT|WARNING|CAUTION):\\s+',
         relevance: 10
       },
-      // inline strong
+      // inline unconstrained strong
+      {
+        className: 'strong',
+        begin: /\*{2}[\w\s\n]+/,
```
joshgoebel, line 113: So word space or newline... what is this mean to be excluding again?

src/languages/asciidoc.js
```diff
@@ -107,18 +107,38 @@ export default function(hljs) {
...
+        }]
+      },
+      // inline constrained strong
       {
         className: 'strong',
         // must not follow a word character or be followed by an asterisk or space
-        begin: '\\B\\*(?![\\*\\s])',
-        end: '(\\n{2}|\\*)',
+        // must not be preceded by the escape character \
+        begin: /\B\*(\w\n?\w)+(?!\n\n)/,
```
joshgoebel, line 128: This looks strange.  What is the exact intentional of this rule?  Trying to prevent `**` from wrapping across paragraphs?

src/languages/asciidoc.js
joshgoebel: When you have something like this that isn't obvious at a glance usually adding a comment can help future maintainers.

src/languages/asciidoc.js
joshgoebel:
```
        begin: /\\[*_`]/,
```

Would work... `0 width` is only triggered when you have a rule that matches something 0 width, which is generally a no-no.

Mogztter:
Hey @joshgoebel thanks for your prompt reviews!
I'm still experimenting and trying various regular expressions and I don't want to bother you.
Maybe I should mark this pull request as draft for now?

src/languages/asciidoc.js
Mogztter: You are correct, an empty line (two consecutive newline characters) indicates a new paragraph.

_not bold_
```
*bob was

a wild man
```

vs

_bold_
```
*bob was
a wild man*
```

The last example is equivalent to:

```
*bob was a wild man*
```

joshgoebel:
I was hoping to maybe prevent you from traveling down some dead-end roads but if you just want to explore on your own and then ping me later, that's fine.  One concern I always have is that it's *possible* to do many things with our engine that we don't actually encourage and there is a certain line where you start "parsing" code instead of merely pattern matching it.  And we try not to cross into that territory. 

I worry getting the rules perfect around a single `*` might be hard and that makes me wonder how badly we were doing it before. :-)

You picked a non-trivial problem for starters. :-)

src/languages/asciidoc.js
joshgoebel: One note here then I'll shut up until you ping me.  We also have a Slack now if that might interest you. (see top of README).

```
       begin: /\B\*(\w\n?\w)+(?!\n\n)/,
```

This is probably way to narrow (and on the other end grabbing way to much).  I presume that other things can exist inside bold - like links, italics, other inline parts of the grammar.  So this rule would fail to take any of that into account (from what I can see).  Plus if you keep extending it to match more and more you run the risk of it matching things that aren't really the "begin" and "end" - which means because this rule ate them, nothing else can match them.  IE, optimally begin and end should "eat"/match as little as possible.  Usually (for begin) this is accomplished with look-ahead.

What you really want here (I think) is:

```ts
begin: /\B\*/,
unlessMiddleContains: /\n\n/ // sadly, this doesn't exist
contains: [ // ... ],
end : /\*/
```

We have `illegal`, but that will just abort.  We can't "backtrack".  So usually what you do (in simpler cases is try to cover the WHOLE expression with a lookahead):

```ts
begin: concat(/\B\*/, lookahead(contentWithoutAsteriskOrPairOfNewlinesEndingInAsterisk)),
contains: [ //... ],
end : /\*/
```

IE, your begin does all the work... I'm just not sure what that regex might look like off the top of my head. :-)

joshgoebel:
```
 begin: /\B\*(\w\n?\w)+(?!\n\n)/,
```

Also since you have no way to pin this I'm pretty sure it will just rewind:

```
*this is bold
but you'd think

that it should not be*
```

[I believe] the regex will rush forward to `think`, see the new paragraph (not a match!) and then just rewind to `bold\n`, which is a match, so the bold rule is a match.

Mogztter:
> I was hoping to maybe prevent you from traveling down some dead-end roads but if you just want to explore on your own and then ping me later, that's fine.
> One concern I always have is that it's possible to do many things with our engine that we don't actually encourage and there is a certain line where you start "parsing" code instead of merely pattern matching it. And we try not to cross into that territory.

You are probably right.
I will keep the pull request active and explicitly request your review when there's something interesting. Are you OK with that?

> I worry getting the rules perfect around a single * might be hard and that makes me wonder how badly we were doing it before. :-)

I adopted a new strategy!
I've implemented basic rules without all the subtle variations. I think the result is quite good :smile: 
The regular expressions are relatively simple to read and all the tests are passing.

> You picked a non-trivial problem for starters. :-)

I like challenges :1st_place_medal:

src/languages/asciidoc.js
Mogztter: Indeed it's working... I could have sworn that I tried that :flushed:

src/languages/asciidoc.js
Mogztter: The code was overly complex and not even correct... let's forget about it, I was just tinkering :factory_worker:

src/languages/asciidoc.js
```diff
@@ -107,17 +107,38 @@ export default function(hljs) {
...
+      },
+      // escaped formatting marks (i.e., \* \_ or \`)
+      {
+        begin: /\\[*_`]/
+      },
+      // inline contrained string (single line)
+      {
+        className: 'strong',
+        // must not precede or follow a word character
+        begin: /\B\*\w+\*\B/,
```
joshgoebel, line 129: Too tight.

```asciidoc
I know they say that *"bob loves jemma"*.
```

Why not `[^\n]*`?

joshgoebel: I'm not sure these single expressions aren't going to cut it... we'll need separate begin and end tags to handle things like:

```asciidoc
Does *Bob _love_ Gemma*?
```
Which we should handle (we do for Markdown).    Or even the proper handling of `\*` inside bold, which you removed.  Generally we try not to merge a PR that removes features unless there are no other options.

joshgoebel: In generally anytime you try and match the middle it's hard.  In case like this you'd want to be super flexible, like I suggest earlier... for example if we're matching a single line bold:

```
// doesn't match all sorts of things, `"` just to name one.
 begin: /\B\*\w+\*\B/,
// matches everything except a new line, the one thing we know isn't allowed
 begin: /\B\*[^\n]+?\*\B/,
```

But again you'd have to split this into begin end still:

```js
begin: /\B\*(?=[^\n]+?\*\B)/, // look-ahead to make sure it closes
contains: OTHER_STUFF,
end: /\*\B/
```

Of course this gets harder since the regex also has to handle escapes, like `\*`...

joshgoebel (changes requested):

src/languages/asciidoc.js
```diff
@@ -107,17 +107,38 @@ export default function(hljs) {
...
-        contains: [{
-          begin: '\\\\*\\w',
-          relevance: 0
-        }]
+        begin: /\*\*(.+?)\*\*/
+      },
+      // inline unconstrained strong (multi-line)
+      {
+        className: 'strong',
+        begin: /\*{2}(\s\S+)*\*{2}/,
```
joshgoebel, line 118: This of course as the same issues with begin/end... we need access to `contains`  so we can properly highlight markup that appears inside of bold.

joshgoebel:
> The regular expressions are relatively simple to read and all the tests are passing.

Simple is good, but I think it only works so well because you've just covered the very basic cases.  I'm not convinced yet we can solve the `*` problem 100%, but if we can it's going to be with something gnarly like I mentioned earlier:

```
/\*(([^*\n\\]|\\[^\n])+\n)+([^*\n\\]|\\[^\n])*\*/

ie matching:

CONTENT = [ content up to \n (and handling escapes)]
LINE = CONTENT [\n]

"*"LINE
LINE(possibly repeated multiple time)
CONTENT"*"
```




You can play around on https://regex101.com/.  It's super useful for stuff like this.

joshgoebel:
Screen Shot 2020-11-17 at 2 53 39 PM

The regex I pasted earlier.

joshgoebel:
Made it a bit easier to see with named groups:

Screen Shot 2020-11-17 at 2 59 50 PM

src/languages/asciidoc.js
Mogztter: > Of course this gets harder since the lookahead regex also has to correctly handle escapes, like \*...

Yes that's why I thought it would be simpler to match everything in `begin`... :unamused: 

> we'll need separate begin and end tags to handle things like (emphasis inside bold)

Unless I missed something, it currently does not work.
It would be nice to support it but we should probably to do it in another pull request.

Mogztter: > Or even the proper handling of \* inside bold, which we supported previously.

I removed it because I don't see why we need it:

```
This stuff sure is *gr*e*e*dy*!
```

This stuff sure is <strong>\*gr\*e\*e\*dy\*</strong>!

> I'm not convinced yet we can solve the * problem 100%,

I don't think it will be possible unless we create a parser, that's why I'm pruning uncommon syntax and/or syntax with undefined behavior.

I think it's important that `This is not \*strong*!` works.
Please note that `This is *str\*ong*!` will produce: This is <strong>\*str\\\*ong*</strong>!

Mogztter: We should probably port tests from Asciidoctor: https://github.com/asciidoctor/asciidoctor/blob/master/test/substitutions_test.rb

```
\**Git**Hub
```

Result: <strong>*Git</strong>*Hub

```
\*a few strong words*
```

Result:  \*a few strong words\*

As you can see `\*` inside bold is not tested.

joshgoebel: > Or even the proper handling of * inside bold, which we supported previously.

I was referring to `\*`, sorry for not being clearer.  Someone went to the trouble of supporting it before making me reason perhaps it's more important than you think. :-)  Obviously * inside bold doesn't require special treatment if it isn't triggering the end of the bold rule or causing any issues.

> Yes that's why I thought it would be simpler to match everything in begin... 

Which is the right idea sort of but you have to cover the whole expression and do it with a look-ahead... so that the actual MATCH portion of begin and end is "*" and "*".  But that's not the largest issue - see my points below on regressions.

> I don't think it will be possible unless we create a parser

The regex I gave you was working for simple cases, did you try it or play around with it in regex101?  See the attached snap...

> It would be nice to support it but we should probably to do it in another pull request.

Oh I'm not trying to say accomplish everything in this PR (far from it!), but I'm also not wanting to merge anything that I know is still broken or is going to make future work harder. So regarding broken lets go back to my prior example:

```
Does *Bob _love_ Gemma*?
```

This probably sort of works thanks to _ being in \w... So my apologies for not using a better example:

```
Does *Bob -love- Gemma*? (broken -)
Does *Bob "love" Gemma*? (broken ")
Does *Bob love Gemma? Truly?* (broken ?)
```

Your PR would break all these that worked previously, would it not?  It's possible that I could be persuaded to drop `\*` for the short-term, but even if that were true I still think there is work to be done here before this would be mergeable.  Go back and look at my comments regarding `[^\n]` vs `\w`, etc... if we resolved these type issues we might have something workable here.

joshgoebel:
So if we just want to focus on minimum to merge I think we need to focus on:

```
      // inline contrained string (single line)
      {
        className: 'strong',
        // must not precede or follow a word character
        begin: /\B\*\w+\*\B/,
      },
      // inline constrained strong (multi-line)
      {
        className: 'strong',
        // must not precede or follow a word character
        begin: 
```

We need more tests cases and those rules need to be widened so as to not fail with simple content in between the `**` pair (examples above in my most recent responses).  Once that's working there are some minor consistency issues we could go over or I can maybe just clean up myself.

src/languages/asciidoc.js
Mogztter: > I was referring to \*, sorry for not being clearer. Someone went to the trouble of supporting it before making me reason perhaps it's more important than you think. :-) Obviously * inside bold doesn't require special treatment if it isn't triggering the end of the bold rule or causing any issues.


Let me check with Dan (who wrote the regular expression 8 years ago). Maybe at the time constrained formatting was not greedy... or maybe I'm missing something!

> The regex I gave you was working for simple cases, did you try it or play around with it in regex101? See the attached snap...

Not yet, I will give it a try tonight.

> So regarding broken lets go back to my prior example...
> This probably sort of works thanks to _ being in \w... So my apologies for not using a better example...
> Your PR would break all these that worked previously, would it not?

Let me try these examples on master but I'm pretty sure they do not work as expected today.
But rest assured, I plan to make them work but one step at a time :wink: 

> Go back and look at my comments regarding [^\n] vs \w, etc... if we resolved these type issues we might have something workable here.

:+1:

src/languages/asciidoc.js
```diff
@@ -9,6 +9,37 @@ Category: markup
...
+  const P_Nl = unpack_hex_range('16EE-16F02160-21822185-218830073021-30293038-303AA6E6-A6EF')
+  const P_Nd = '0-9' + (unpack_hex_range('0660-066906F0-06F907C0-07C90966-096F09E6-09EF0A66-0A6F0AE6-0AEF0B66-0B6F0BE6-0BEF0C66-0C6F0CE6-0CEF0D66-0D6F0DE6-0DEF0E50-0E590ED0-0ED90F20-0F291040-10491090-109917E0-17E91810-18191946-194F19D0-19D91A80-1A891A90-1A991B50-1B591BB0-1BB91C40-1C491C50-1C59A620-A629A8D0-A8D9A900-A909A9D0-A9D9A9F0-A9F9AA50-AA59ABF0-ABF9FF10-FF19'))
+  const P_Pc = unpack_hex_range('005F203F20402054FE33FE34FE4D-FE4FFF3F')
```
joshgoebel, line 28: Oh wow.  Two steps forward one step back. :-)  This might be a bit much. :-)

Context: https://github.com/highlightjs/highlight.js/issues/2756

We're still up in the air on UTF8 but if we're headed there we should do it properly, not with this kind of hack.

src/languages/asciidoc.js
```diff
@@ -110,30 +141,47 @@ export default function(hljs) {
...
+      // constrained preceded by ":", ";", or "}".
+      {
+        begin: /[:;}][*_`](?!\*)/
+      },
+      // inline constrained string (single line)
       {
         className: 'strong',
         // must not precede or follow a word character
-        begin: /\B\*\w+\*\B/,
+        begin: `\\B\\*(\\S|\\S${CC_ANY}*?\\S)\\*(?!${CG_WORD})`,
```
joshgoebel, line 176: Can we pull out all the UTF-8 and just let this be \w for now?

src/languages/asciidoc.js
```diff
@@ -110,30 +141,47 @@ export default function(hljs) {
...
+      },
+      {
+        begin: /\\\\_{2}[^\n]*_{2}/
+      },
+      {
+        begin: /\\\\`{2}[^\n]*`{2}/
+      },
+      // constrained preceded by ":", ";", or "}".
+      {
+        begin: /[:;}][*_`](?!\*)/
```
joshgoebel, line 170: Remind me what this is doing again? Is this a false positive guard or?

src/languages/asciidoc.js
```diff
@@ -110,30 +141,47 @@ export default function(hljs) {
...
+      // escaped constrained formatting marks (i.e., \* \_ or \`)
       {
         begin: /\\[*_`]/
       },
-      // inline contrained string (single line)
+      // escaped unconstrained formatting marks (i.e., \\** \\__ or \\``)
+      // must ignore until the next formatting marks
+      // this rule is not 100% compliant with Asciidoctor but we are entering undefined behavior territory...
+      {
+        begin: /\\\\\*{2}[^\n]*\*{2}/
```
joshgoebel, line 160: Wait does this syntax literally mean "it's marked as bold, but ignore it and don't make it bold"?

src/languages/asciidoc.js
joshgoebel: You could leave a TODO comment if you wanted.

src/languages/asciidoc.js
joshgoebel: Also shouldn't this be CC_ANY?

src/languages/asciidoc.js
```diff
@@ -107,17 +138,55 @@ export default function(hljs) {
...
+      {
+        className: 'strong',
+        // must not precede or follow a word character
+        begin: `\\B\\*(\\S|\\S${CC_ANY}*?\\S)\\*(?!${CG_WORD})`,
+      },
+      // inline constrained strong (multi-line)
+      {
+        className: 'strong',
+        // must not precede or follow a word character
+        begin: /\B\*(\w\n?)*\*(?!\w)/,
```
joshgoebel, line 182: I'm so confused about constrains vs unconstrained... should this work:

```
*does bob
-really-
know best?*
```

src/languages/asciidoc.js
joshgoebel: Oh and you don't need to turn them all into strings either, just use `regex.concat`.

src/languages/asciidoc.js
Mogztter: OK, I will use `\w` and remove the associated case.
We are using these unicode ranges in Asciidoctor.js to translate `\p{Word}` (Ruby).

src/languages/asciidoc.js
Mogztter: Yes, we do not allow a constrained formatting mark to be preceded by ":", ";" or "}". I've added the following test cases to make sure that bold/strong is not applied:

```
There's a colon:*directly* before the starting formatting mark.

There's a semi-colon directly before the starting formatting mark &ndash;*2018*

There's a closing curly bracket directly {before}*the starting formatting mark*.
```

src/languages/asciidoc.js
Mogztter: > Wait does this syntax literally mean "it's marked as bold, but ignore it and don't make it bold"?

It's another guard when unconstrained bold is escaped using `\\`. Take a look at the test case:

```
\\**--anything goes **
```

I've also added an even more complex test case:

```
\\**--anything goes ** with * a **lot** of stars**
```

> Also shouldn't this be CC_ANY?

Yes

> Oh and you don't need to turn them all into strings either, just use regex.concat.

Oh really? Could you please provide an example, I've never seen `regex.concat` :exploding_head:

src/languages/asciidoc.js
Mogztter: > I'm so confused about constrains vs unconstrained... shouldn't this work:

```
*does bob
-really-
know best?*
```

The above should produce bold.

> I'm wondering if perhaps this is really just the same as the ** except for the before/end conditions? If so then that regex would probably work here also I think?

Close but not quite the same: https://asciidoctor.org/docs/user-manual/#when-should-i-use-unconstrained-quotes

> Consider the following questions:
> Is there a letter, number, underscore directly outside the formatting marks (on either side)?
> Is there a colon, semi-colon, or closing curly bracket directly before the starting formatting mark?
> Is there a space directly inside of the formatting mark?
> If you answered “yes” to any of these questions, you need to switch to unconstrained (double formatting) quotes.

See: https://github.com/asciidoctor/asciidoctor/blob/dd843d59b1e49f29d68386bfdf0438ce11b42577/lib/asciidoctor.rb#L447-L449
Please keep in mind that we are also doing some work in https://github.com/asciidoctor/asciidoctor/blob/dd843d59b1e49f29d68386bfdf0438ce11b42577/lib/asciidoctor/substitutors.rb#L75

Mogztter: I'm also using a less greedy regular expression than Asciidoctor because we want to stop when the paragraph ends.

src/languages/asciidoc.js
```diff
@@ -107,17 +107,55 @@ export default function(hljs) {
...
+      },
+      {
+        begin: /\\\\_{2}[^\n]*_{2}/
+      },
+      {
+        begin: /\\\\`{2}[^\n]*`{2}/
+      },
+      // constrained preceded by ":", ";", or "}".
+      {
+        begin: /[:;}][*_`](?!\*)/
```
joshgoebel, line 139:
```suggestion
      // constrained formatting mark may not be preceded by ":", ";" or "}".
      // match these so the constrained rule doesn't see them
      {
        begin: /[:;}][*_`](?!\*)/
```

src/languages/asciidoc.js
joshgoebel: > The above should produce bold.

But your rule doesn't match the `-`, or am I missing something?

> Close but not quite the same

Unless I'm missing something all I read there relates directly to what qualifies the beginning and ending of the match though, yes?  Matching the "middle" content could use the same (or similar) wider rule so that it wouldn't get hung up on details like `-` or punctuation, etc.  So really we're trying to structure the regex into 3 segments:

- beginning
- middle
- end

It seems for both `*` and `**` the middle rules are the same, no?  Only the begin and end conditions change?

src/languages/asciidoc.js
joshgoebel: Try to pretend the reader of the grammar doesn't know Asciidoctor. :)

src/languages/asciidoc.js
joshgoebel: Yeah, we have access to the `/u` flag in JS and if we're going to start matching `\p` stuff that's what we should be using, but right now there are some performance concerns.  If you know anything about it and wanted to contribute to that thread help would be welcome. :-)

src/languages/asciidoc.js
joshgoebel: It's in `lib/regex.js`...  just utility functions.

joshgoebel: > It's another guard when unconstrained bold is escaped using \\. Take a look at the test case: 

OK if really it's like a paired "don't apply any highlighting on the inside", ie "anything goes"... (kind of like a comment in a programming language) but why are there 3 variants of it? :-)  Asciidoc is strange. :)

src/languages/asciidoc.js
Mogztter: > But your rule doesn't match the -, or am I missing something?

Oh yes, you're right, sorry I didn't understand.
I've added this case and updated the regular expression accordingly.

> Unless I'm missing something all I read there relates directly to what qualifies the beginning and ending of the match though, yes?

No, this rule is about what is inside the formatting marks: "Is there a space directly inside of the formatting mark?"

```
this is * a bold move*

this is ** a bold move**
```

Only the second sentence will produce bold.

> So really we're trying to structure the regex into 3 segments:

I guess since the formatting marks are delimiters, so:

- beginning: delimiter
- middle: some content
- end: delimiter

In theory, that should be easy but in practice it's not because "*" can be used in different *context*:

```
****
sidebar
****

* foo
* bar
** baz

Read the footnotes*.

E = mc*2*

E = *mc*2

Use `/*` and `*/` for multiline comments.
```

The last example is not correctly interpreted by Asciidoctor :sweat_smile: 

Anyway, we should be careful about "as WIDE as possible".

joshgoebel (approved):
Add changelog entry please.

Mogztter:
> Add changelog entry please.

@joshgoebel Done.


Once it's merged, I will also open a follow-up pull request to apply the same unconstrained rules to emphasis and monospaced.

joshgoebel (approved):

Issue #2559, opened 2020-05-15
(parser) Deprecate `useBR`


joshgoebel:
From isagalaev in #2529:

> I was walking by, and while I didn't read through the entire thing I wanted to say this. There were a few features added to the core at users' requests when it was all small an nimble. These days I would say it makes much more sense to be opinionated and remove corner cases like useBR, tabWidth and several different spellings of "lang-", "language-" as language prefixes in the class name (as far as I remember there is a very precise recommendation in HTML5 to use "language-"). Everyone with special cases would be expected to do pre- or post-processing.

Very early on we were perhaps a bit too open and allowed some things to slip into core that perhaps should not have - and today they are also things that could easily be handled with our new plugin API.

`useBR` is a messy one for me, so I'm picking on it first.  I think it's obvious the "best/blessed" path for using HLJS is with pre/code tags... that gives you the best chance of preserving your code in markup without doing ugly things like inserting HTML line breaks in it in the first place.  Those who really need to use a different element and need to do `<BR>` > `\n` translate can simply use a plugin to accomplish the same thing in the future.

- [x] v10.1 Deprecate `useBR`
- [ ] Provide plugin?
- [ ] v11 Remove `useBR`

egor-rogov:
+1

joshgoebel:
> Those who really need to use a different element and need to do <BR> > \n translate can...

This is not encouraged.  `pre/code` is the blessed path for multi-line code blocks.  Someone wanting to restore this functionality using plugins could look at `before:highlightBlock` and `after:highlightBlock`.

I've changed this issue to "deprecate" and am closing it with 980832e3d38ce56ab50ec7359492952d2dde5237.

my-small-white:
58/5000 
I used this library and found that the code block was in one line and could not be solved until I saw the community discussion and realized that the
problem was caused. Thank you very much for solving my problem

Issue #2558, opened 2020-05-15
Version 11.0 checklist


joshgoebel:
This is still a ways away but I wanted to start a list of the breaking changes we intend to introduce so we don't forget anything later.

- [x] rip out beforeMatch in lue of multi-match?
- [x] `.title--class` or `.title__class` #3081
- [x] Throttle deprecation logging #3069
- [x] Remove `lexemes`

---

- [x] Pull Vue plugin out of main build
- [x] Remove aliases that are "also works with" not true aliases #3049
- [x] Remove `fixMarkup` (deprecated via #2534)
- [x] Remove `useBR` #2559
- [x] Remove `tabReplace` (deprecated via #2874)
- [x] Remove `php3`, 4, 5, 6, etc. aliases
- [x] Remove `sql_more` grammar
- [x] `relevance` returned by highlight is allowed to be a floating point number now
- [x] Remove HTML merging (deprecated via #2889)
- [x] Consider status of ESM support in Node #2600 
- [x] rename `regex.join` to indicate it's internal
- [x] remove `re` key from element.result (highlightBlock)
- [x] remove the old `htmlbars` stub
- [x] remove the old `c-like` stub
- [x] remove schoolbook or remove it's funky background [ref](https://github.com/highlightjs/highlight.js/pull/2923)
- [x] change default padding of all themes to be more generous
- [x] rename `second_best` to `secondBest`
- [x] remove apache, http, nginx, properties, coffeescript from `:common` https://github.com/highlightjs/highlight.js/issues/2848
- [x] move `CSS_NUMBER_MODE` into `css-shared` (it's too specific to be a general rule)
- [x] review notes on https://github.com/highlightjs/highlight.js/issues/2945 and consider making some attributes of Result "private"... ie, `__top` vs `top`... since those are intended more for debugging rather than public use.
- [x] `initHighlighting*` needs to be aliased to `highlightAll()`

taufik-nurrohman:
- [ ] `hljs.listAliases()` &tarr; to get the complete list of aliases (with a map to the original language maybe?)
- [x] `hljs.listLanguages()`
- [ ] `hljs.listNames()` &rarr; to get the complete list of language names for human to be used in select box, etc.
- [x] `hljs.registerAlias()`
- [x] `hljs.registerLanguage()`

What do you think?

yyyc514:
What good is `listNames` without the identifiers?  I'd think if anything `allLanguages` or something would be more helpful.

```js
var options = 
[...]

PR #2873, opened 2020-11-17
chore(plugins) move useBR, tabReplace, and HTML merging into internal plugins

### LINKED ISSUE #2874: (parser) Deprecate: `tabReplace` parser option
From isagalaev in #2529:

> I was walking by, and while I didn't read through the entire thing I wanted to say this. There were a few features added to the core at users' requests when it was all small an nimble. These days I would say it makes much more sense to be opinionated and remove corner cases like useBR, tabWidth and several different spellings of "lang-", "language-" as language prefixes in the class name (as far as I remember there is a very precise recommendation in HTML5 to use "language-"). Everyone with special cases would be expected to do pre- or post-processing.

Very early on we were perhaps a bit too open and allowed some things to slip into core that perhaps should not have - and today they are also things that could easily be handled with our new plugin API.

- [ ] v10.5 Deprecate `tabReplace`
- [x] Provide plugin?
- [ ] v11 Remove `tabReplace`

---

### Plugin Code

**Consider:** Before resorting to this you should probably consider the much simpler [CSS `tab-size` property](https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size).

```js
const TAB_REPLACE_RE = /^(<[^>]+>|\t)+/gm;
const tabReplacePlugin = {
  "after:highlightBlock": ({ result }) => {
    result.value = result.value.replace(TAB_REPLACE_RE, (m) =>
      m.replace(/\t/g, options.tabReplace)
    );
  }
};

// how to use it
hljs.addPlugin(tabReplacePlugin);
```

joshgoebel:
Move non happy path features into internal plugins for later breaking
out into external plugins.

### Related

- [x] Resolves #2874. Deprecate `tabReplace` support issue / plugin
- [x] #2559 Deprecate `useBR` support issue / plugin
- [x] #2889 Deprecate HTML merging support issue / plugin
- #2558 Version 11 plans

### Checklist


joshgoebel:
@allejo @egor-rogov And with this I'd like to consider adding both `tabReplace` and HTML merging to the chopping block for v11.  `useBR` is already on the way out.  Users who need them can obviously get the job done by using plugins.

IE in the core library we support the happy path: 

- we take content from a <pre> block (allowing for meaningful `\n`, etc)
- we highlight content "as-is" - we do not transform it (tabs, new lines, spacing, etc... all untouched)
- any HTML inside the `<pre>` will be silently dropped 

All the below are edge cases:

- You have HTML mixed with your code that you want co-mingled (use HTML merging plugin)
- You have tabs rather than spaces in your text and want them transformed into spaces. (use a plugin or use CSS `tab-size`)

Thoughts?

joshgoebel:
Screen Shot 2020-11-17 at 2 53 39 PM

The regex I pasted earlier.

CHANGES.md
```diff
@@ -60,15 +52,37 @@ Dev Improvements:
...
 - *StackOverflow Dark* by [Jan Pilzer][]
 - *StackOverflow Light* by [Jan Pilzer][]
 
+
+Deprecations (*these will be mentioned in every release*):
+
+- HTML "merging" is deprecated. (#2873) [Josh Goebel][]
+  - HTML inside `<pre>` blocks will no longer be magically merged back into the
+  highlighted code's HTML result - it will instead be silently removed.
+  - Use a plugin if you truly need this functionality
```
Hirse, line 66: Is there a link to the plugin to make this more actionable?

joshgoebel: There is no plugin atm. Someone who really strongly needs the feature would need to contribute one.  Perhaps it should say "search for a plugin"?  I haven't even gotten around to making an issue thread for this one.  The existing code for that is not trivial for cutting and pasting... I also really don't want to encourage it's continued usage.  If a 3rd party author wants to bundle it into a plugin (the architecture changes here show this is possible) and then support it long-term, that would be fine. Otherwise I'd like to see the feature RIP.

It really never should have been added to core.

joshgoebel: https://github.com/highlightjs/highlight.js/issues/2889

sa-orlov (commented):
Ок

sa-orlov (commented):
Hi

sa-orlov (commented):
Hi

PR #2878, opened 2020-11-19
quietier post-install, smarter deprecation warning

joshgoebel:


allejo (approved):

src/highlight.js
```diff
@@ -535,6 +540,17 @@ https://highlightjs.org/
    * @property {boolean} illegal - indicates whether any illegal matches were found
   */
   function highlight(languageName, code, ignore_illegals, continuation) {
+    if (!hideUpgradeWarning()) {
```
Romakita, line 543: you  should removed this log. In pure node.js context and a CLI (not browser), the console.log will be run every time the code is executed. 
The warning on the post-install is enough, I think.

Romakita (changes requested):

src/highlight.js
allejo: Isn't that what setting `showedUpgradeWarning = true` in the next few lines is for?

Romakita: Your variable is just store in memory. A CLI start process, run and done process each time you want to run a command.
Example:
Capture d’écran 2020-11-19 à 16 43 29

Romakita: You can see the log appear each time I ran `tsed init .`. In my case,
I don't use directly the hightlight.js (it's https://github.com/helpers/helper-markdown/blob/master/package.json#L29) so I can change the warning behavior at runtime.

joshgoebel: You can control this either via the HLJS options hash or with an environment variable now.  Your really should file an issue with the upstream to upgrade to v10.

Romakita: But again isn't my job to contact all maintainers for that to avoid an unexpected behavior ^^

Romakita: I think you don't realize the side effect you had introduce just with this choice to print a small warning at the runtime.

Romakita: >  And this will only be shown now if you are actually highlighting something, not just requiring the library.
I had the screen as example to said you that is not the truth !

Romakita: Ok for the env variable, but just imagine if you have to configure all variables environment for each modules you use as dependencies to disable a simple warning ....

Romakita: I may be a bit heavy, but I find it hard to understand why you need to inform the devs about the end of support so many times and especially at runtime when you already have a log in postinstall.

allejo: Because we _really_ want people to get off of highlight.js 9.x and tools like Yarn ignore post-install messages.

Romakita: Yes I understand your need, I'm facing the same problem when I want to deprecated a npm module. That's not why I'm going to impose depreciation logs on the runtime ^^.

joshgoebel: > I think you don't realize the side effect you had introduce just with this choice to print a small warning at the runtime.

Correct, because I didn't realize Node.js treats warnings the same as ERRORS. If I had it would never have shipped as a warning.

> I had the screen as example to said you that is not the truth !

It will be the truth in 9.18.5. :-)  It won't be shown just requiring the library - it'll only show up if you're USING the library - in which case you should know about the potential security implications.

> Yarn ignore post-install messages.

Correct.  Best way to see it is cover it multiple places.

>  I find it hard to understand why you need to inform the devs about the end of support so many times

There are real security implications for those who wish to continue to use the old library.  Anyone using it to highlight user generated code is possible opening themselves (or their sites) up to ReDOS attacks or similar in the future - and possibly other things.  100s of bugs (some serious) have been fixed since v10 was released back in March I think?  If someone isn't actually highlighting anything, this might not matter - hence now only showing the warning when someone USES the highlighter.

Romakita: > It will be the truth in 9.18.5. :-) It won't be shown just requiring the library - it'll only show up if you're USING the library - in which case you should know about the potential security implications.
yes I totally understand that. But if the node.js module use that, the log will be displayed. Wait and see :). 

Thanks for you answers and your patience :)

joshgoebel:
This was merged into 9-18-stable.  Closing.

src/highlight.js
joshgoebel: No problem.  Thanks for yours.  We really didn't expect to break anyones day with the `warn`... I'm hoping 9.18.5 is much better.

Does your project not highlight code at all?

Romakita: Not directly. I use handlebars which uses handlebars-helpers which uses markdown-helper which uses highlight.js. And in my code I don’t use highlight (just handlebars) but it’s probably used by the helpers for a good reason^^

joshgoebel: Dependency hell. :) Ugh.

Romakita: Yes I try to have less dependencies but for a CLI is very complicated ^^

joshgoebel: Not saying it isn't, but dependencies (esp. in Node.js land), ugh. :-)

PR #2893, opened 2020-11-27
fix(http) avoid recursive sublanguage and tighten rules

joshgoebel:
Currently it's too easy to trigger recursive sublanguages with 
a file that contains multiple paragraph breaks (\n\n). `http`
can potentially fire over and over resulting in auto-detect on
the remaining content with all grammars over and over.

This tightens up the HTTP rules a lot to avoid this recursion
as well as reduce a lot of false positives.

### Checklist


allejo (approved):
```html
<span class="hljs-attribute">Host</span><span class="hljs-punctuation">: </span>example.org
```

It feels kinda weird to me that the space is considered part of the "punctuation."

but this looks good :+1:

joshgoebel:
> It feels kinda weird to me that the space is considered part of the "punctuation."

I no longer spend extra effort on spacing issues IF they would be difficult to solve.

PR #2900, opened 2020-12-02
(chore) Clean up linting

Hirse:
### Changes
- Remove JSHint config
- ~Switch ESLint config to JSON (for JSON-schema support)~
- Change `npm` lint script to include `src/plugins` folder
  - Fix some linting issues in the vue plugin
  - Fix some linting issues in the `merge_html` plugin
- Add `lint-languages` npm script
- Clean up lint workflow
  - Use latest node
  - Use `npm ci` for faster and more reliable dependency installs
  - Use new `npm lint` scripts
- Add new ESLint configs for `test` and `tools`, which are not included in CI, but help during development

### Checklist
- [x] Added markup tests, or they don't apply here because it's not a code change
- [ ] Updated the changelog at `CHANGES.md` - NA?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,27 +2,21 @@ module.exports = {
   env: {
     browser: true,
     es6: true,
-    node: true,
-    mocha: true
+    node: true
   },
   extends: [
     "eslint:recommended",
     "standard"
   ],
-  globals: {
-    Atomics: "readonly",
-    SharedArrayBuffer: "readonly"
-  },
   parserOptions: {
     ecmaVersion: 2015,
     sourceType: "module"
   },
-  // parser: '@typescript-eslint/parser',
   plugins: [
     "@typescript-eslint"
   ],
   rules: {
-    "no-var": 1,
+    "no-var": "warn",
     "init-declarations": ["error", "always"],
     "array-callback-return": "error",
     "block-scoped-var": "error",
@@ -55,32 +49,47 @@ module.exports = {
       files: ["src/languages/*.js"],
       rules: {
         "no-unused-expressions": "off",
-        // // languages are all over the map and we don't want to
-        // // do a mass edit so turn off the most egregious rule violations
+        // languages are all over the map and we don't want to
+        // do a mass edit so turn off the most egregious rule violations
         // indent: "off",
-        camelcase: 0,
-        "no-control-regex": 0,
-        "no-useless-escape": 0,
-        "comma-dangle": 1,
+        camelcase: "off",
+        "no-control-regex": "off",
+        "no-useless-escape": "off",
+        "comma-dangle": "warn",
         "array-bracket-spacing": ["error", "always"
-        //   {
-        //     objectsInArrays: true
+          // {
+          //   objectsInArrays: true
           // }
         ],
-        // "object-curly-spacing": 1,
+        // "object-curly-spacing": "warn",
         // "key-spacing": "off",
-        // "array-bracket-spacing": [1],
-        "array-bracket-newline": [1, {
+        // "array-bracket-spacing": ["warn"],
+        "array-bracket-newline": ["warn", {
           multiline: true,
           minItems: 2
         }],
-        "array-element-newline": 1,
+        "array-element-newline": "warn",
         "object-curly-newline": [1, {
           minProperties: 1
         }],
         "object-property-newline": [2,
           { allowAllPropertiesOnSameLine: false }
         ]
+      }
+    },
+    {
+      files: ["test/**/*.js"],
+      env: {
+        mocha: true
+      },
+      parserOptions: {
+        ecmaVersion: 2021
+      }
+    },
+    {
+      files: ["tools/**/*.js"],
+      parserOptions: {
+        ecmaVersion: 2021
       }
     }
   ]
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,7 @@ module.exports = {
     ecmaVersion: 2015,
     sourceType: "module"
   },
+  // parser: '@typescript-eslint/parser',
   plugins: [
     "@typescript-eslint"
   ],
@@ -83,13 +84,13 @@ module.exports = {
         mocha: true
       },
       parserOptions: {
-        ecmaVersion: 2021
+        ecmaVersion: 2018
       }
     },
     {
       files: ["tools/**/*.js"],
       parserOptions: {
-        ecmaVersion: 2021
+        ecmaVersion: 2018
       }
     }
   ]
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,16 +6,16 @@ name: Lint
 on:
   pull_request:
     branches:
-      - master
+    - master
 
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
-        with:
-          node-version: "*" # latest
-      - run: npm ci
-      - run: npm run lint
-      - run: npm run lint-languages
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: "*" # latest
+    - run: npm ci
+    - run: npm run lint
+    - run: npm run lint-languages
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,28 +4,18 @@
 name: Lint
 
 on:
-#   push:
-    # branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches:
+      - master
 
 jobs:
-  build:
-
+  lint:
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [15.x]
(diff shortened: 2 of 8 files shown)


.eslintrc.js
```diff
@@ -1,86 +0,0 @@
...
-          minItems: 2
-        }],
-        "array-element-newline": 1,
-        "object-curly-newline": [1, {
-          minProperties: 1
-        }],
-        "object-property-newline": [2,
-          { allowAllPropertiesOnSameLine: false }
-        ]
-      }
```
joshgoebel, line 83: Where did all these settings go?

.github/workflows/lint.yml
```diff
@@ -4,28 +4,16 @@
...
-
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [15.x]
-
     steps:
     - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
```
joshgoebel, line 23: Let's please add back the version to the GHA output even if we don't need technically need `strategy`.  It's valuable when seeing there is an issue to immediately know which Node version is being used.  

It should be easy to fetch from the environment or a GHA variable I'd think?

.github/workflows/lint.yml
```diff
@@ -4,28 +4,16 @@
...
-      uses: actions/setup-node@v1
+    - uses: actions/setup-node@v1
       with:
-        node-version: ${{ matrix.node-version }}
-    - run: npm install
-    - name: Lint the main source
-      run: npx eslint src/*.js src/lib/*.js
-    - name: Lint the language grammars
-      run: npx eslint --no-eslintrc -c .eslintrc.lang.js src/languages/**/*.js
+        node-version: * # latest
```
joshgoebel, line 17: Or maybe this already spits out the version?  I'm not sure.

.eslintrc.js
Hirse: Those were set to only apply to `src/languages/*.js`, which is overwritten in the `src/languages/.eslintrc.json` file.

joshgoebel: Right, pretty sure this breaks the prior (desirable behavior):

- Linting of grammars (for CI purposes) is loose - a grammar lint issue should NOT how up a PR.
- Linting of grammars during development is stricter. (so that improvements can be made when they make sense) 

This behavior would need to be kept.

.github/workflows/lint.yml
Hirse: Apparently, it doesn't show it in the PR checks-list
image
and neither in the actions summary
image
but it would be available in the log
image

I'm fine reverting to `strategy` if you prefer seeing the version in all those places.

.eslintrc.js
Hirse: Ah, I missed that distinction. 
Will restore the desired behavior.

.github/workflows/lint.yml
joshgoebel: What is doing now may be sufficient. This is probably less important for linting than our main CI test.

Hirse:
@joshgoebel would you mind giving this another look?

joshgoebel:
What is the advantage of JSON again?  It's very hard to review because JSON forces every single line to change when really nothing has changed.

Hirse:
> What is the advantage of JSON again? 

JSON can be validated with a schema. In some editors (I'm using VSCode) this is built-in and gives a great experience:
image

> It's very hard to review because JSON forces every single line to change when really nothing has changed.

I don't quite understand what you mean here.

joshgoebel:
> JSON can be validated with a schema.

Yes, that's true... though _technically_ there is no reason it couldn't do the same exact thing based on the name of the file and using the default export... but whatever.

> I don't quite understand what you mean here.

Instead of the diff showing what actually changed (a few lines) it just shows 200 lines removed, 200 lines added, etc... making it very hard to review.  So I have to scroll thru the file manually comparing one line at a time visually... because I'm pretty sure at first glance you changed a few things and removed some of my comments, etc...  And that is why I haven't found the time to sort this out yet.  It's on the list to get to though.

Hirse:
I've reverted the change to `json` (for this PR) and minimized the changes for easier comparison and reviewing.

.eslintrc.js
```diff
@@ -2,27 +2,21 @@ module.exports = {
...
-    mocha: true
+    node: true
   },
   extends: [
     "eslint:recommended",
     "standard"
   ],
-  globals: {
-    Atomics: "readonly",
-    SharedArrayBuffer: "readonly"
```
joshgoebel, line 14: I think I just pulled these from elsewhere, they aren't needed?

.eslintrc.js
```diff
@@ -2,27 +2,21 @@ module.exports = {
...
-  },
   parserOptions: {
     ecmaVersion: 2015,
     sourceType: "module"
   },
-  // parser: '@typescript-eslint/parser',
```
joshgoebel, line 20: Please leave, sometimes I turn this on for stricter linting of TS stuff when developing, sometimes it gets in the way.

.eslintrc.js
```diff
@@ -2,27 +2,21 @@ module.exports = {
...
   plugins: [
     "@typescript-eslint"
   ],
   rules: {
-    "no-var": 1,
+    "no-var": "warn",
```
joshgoebel, line 19: Nice!

.github/workflows/lint.yml
```diff
@@ -4,28 +4,18 @@
...
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
-      with:
-        node-version: ${{ matrix.node-version }}
-    - run: npm install
-    - name: Lint the main source
-      run: npx eslint src/*.js src/lib/*.js
-    - name: Lint the language grammars
-      run: npx eslint --no-eslintrc -c .eslintrc.lang.js src/languages/**/*.js
+      - uses: actions/checkout@v2
```
joshgoebel, line 15: Was this just entirely broken before and not even working?  Trying to understand the indent changes.

package.json
```diff
@@ -24,7 +24,8 @@
   "types": "./types/index.d.ts",
   "scripts": {
     "mocha": "mocha",
-    "lint": "eslint -c .eslintrc.js src/*.js src/lib/*.js",
+    "lint": "eslint src/*.js src/lib/*.js src/plugins/*.js",
```
joshgoebel, line 27: Does this also need to revert since we switched back to JS or was the `-c` implied all along and I'm just being too explicit?

.eslintrc.js
```diff
@@ -55,33 +49,48 @@ module.exports = {
...
         ]
       }
+    },
+    {
+      files: ["test/**/*.js"],
+      env: {
+        mocha: true
+      },
+      parserOptions: {
+        ecmaVersion: 2021
```
joshgoebel, line 86: Does Node 10 support all 2021 ECMA stuff?  Technically we only test against 12+ but 10 is still a supported release and as far as I know we work on 10 (we should perhaps CI it as well)... so since this code should work on Node 10 we need to set ecmaVersion accordingly. Thoughts?

joshgoebel:
> and minimized the changes for easier comparison and reviewing.

OMG, 100x better.  **Thank you.**

.eslintrc.js
Hirse: No, the globals section defines variables that are used from the global scope. (In demo, that would be `hljs` for example.)
Since these two are not used in the code base, there is no need to specify them. (They might get defined in the environments like `window` and `document` would when using `browser` environment.)

See also: https://eslint.org/docs/user-guide/configuring#specifying-globals

.eslintrc.js
joshgoebel: I'd have approved 2018 without any thought but 2021 is a bit "fresh" so I wonder...... :-)

Hirse: I don't actually know. I'm fine setting it to `"ecmaVersion": 2018` which should cover Node 10 (released in 2018), but wouldn't suggest going lower to not get errors on `async` for example.

There is an ESLint plugin (`eslint-plugin-node`) that offers a rule (`node/no-unsupported-features`) to test for supported/unsupported node features directly, but that feels a bit like overkill since this is only the test code.

.github/workflows/lint.yml
Hirse: No, either indentation works. Reverting to align with the GitHub actions documentation

.eslintrc.js
joshgoebel: Oh that might be nice for the browser though, I'll have to check it out.  Lets do 2018 then if you don't know.

package.json
Hirse: Yes, by default ESLint searches for a config file named `.eslintrc` (with various endings).

See https://eslint.org/docs/user-guide/configuring#using-configuration-files-1

.eslintrc.js
Hirse: Done.

joshgoebel (approved):

joshgoebel (approved):

joshgoebel:
@Hirse Thanks!

PR #2902, opened 2020-12-02
(swift) New string grammar

### LINKED ISSUE #2819: (Swift) Incorrect highlighting of raw strings

svanimpe:
This is a rewrite of the grammar for strings in Swift. I’ve implemented the official grammar (https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#grammar_string-literal) and included support for raw strings with up to three hash signs.

I’m submitting this PR as a WIP to get some early feedback, as this is my first contribution. If the grammar looks good, I'll add a suite of unit tests and update the changelog/contributor list.

Fixes #2819 and replaces #2820.


src/languages/swift.js
```diff
@@ -96,7 +82,114 @@ export default function(hljs) {
...
+    contains: [RAW3_ESCAPED_CHARACTER, RAW3_INTERPOLATION]
+  };
+  var STRING = {
+    className: 'string',
+    variants: [
+      MULTILINE_STRING,
+      RAW1_MULTILINE_STRING,
+      RAW2_MULTILINE_STRING,
+      RAW3_MULTILINE_STRING,
+      SINGLE_LINE_STRING,
```
joshgoebel, line 182: We'd really want to do this a bit more programmatically though:

```suggestion
    variants: [
      MULTILINE_STRING,
      RAW_MULTILINE_STRING("#"),
      RAW_MULTILINE_STRING("##"),
      RAW_MULTILINE_STRING("###"),
      SINGLE_LINE_STRING,
```

Etc. Then if we needed to support more or less of these in the future it would be a one line change... This same thinking would apply to single line string as well.

src/languages/swift.js
```diff
@@ -96,7 +82,114 @@ export default function(hljs) {
...
+    className: 'subst',
+    begin: /\\##[0\\tnr"']|\\##u\{[0-9a-fA-F]{1,8}\}/
+  };
+  var RAW3_ESCAPED_CHARACTER = {
+    className: 'subst',
+    begin: /\\###[0\\tnr"']|\\###u\{[0-9a-fA-F]{1,8}\}/
+  };
+  var ESCAPED_NEWLINE = {
+    className: 'subst',
+    begin: /\\[\u0009\u0020]*(?:[\u000a\u000d]|\u000d\u000a)/
```
joshgoebel, line 105: Is this just `\r` and `\n`?  If so leets please use those common expressions instead.

I'm not sure what the 9 and 20 are? Can you clarify?

src/languages/swift.js
```diff
@@ -96,7 +82,114 @@ export default function(hljs) {
...
+    className: 'subst',
+    begin: /\\[0\\tnr"']|\\u\{[0-9a-fA-F]{1,8}\}/
+  };
+  var RAW1_ESCAPED_CHARACTER = {
+    className: 'subst',
+    begin: /\\#[0\\tnr"']|\\#u\{[0-9a-fA-F]{1,8}\}/
+  };
+  var RAW2_ESCAPED_CHARACTER = {
+    className: 'subst',
+    begin: /\\##[0\\tnr"']|\\##u\{[0-9a-fA-F]{1,8}\}/
```
joshgoebel, line 97: Could we use `variants` to make these entirely different syntaxes a little more visible?

src/languages/swift.js
joshgoebel: You may find `regex.concat` useful to help build your children regexes you'll need (see `regex.js`).

joshgoebel (changes requested):
Awesome.  Already looking much better than the first pass! :-)

svanimpe:
@joshgoebel I've pushed a new version that incorporates your feedback.

src/languages/swift.js
```diff
@@ -120,7 +153,7 @@ export default function(hljs) {
           },
           {
             className: 'params',
-            begin: /\(/, end: /\)/, endsParent: true,
+            begin: /\(/, end: /\)/,
```
joshgoebel, line 156: What is the context here again?

src/languages/swift.js
```diff
@@ -96,7 +84,52 @@ export default function(hljs) {
...
+    ]
+  });
+  var ESCAPED_NEWLINE = (rawDelimiter = "") => ({
+    className: 'subst',
+    begin: concat(/\\/, rawDelimiter, /[\t ]*(?:[\r\n]|\r\n)/)
+  });
+  var INTERPOLATION = (rawDelimiter = "") => ({
+    className: 'subst',
+    begin: concat(/\\/, rawDelimiter, /\(/),
+    end: /\)/
```
joshgoebel, line 103:
```suggestion
    className: 'subst',
    keywords: SWIFT_KEYWORDS,
    begin: concat(/\\/, rawDelimiter, /\(/),
    end: /\)/
```

Don't we need keywords here?

src/languages/swift.js
```diff
@@ -96,7 +84,52 @@ export default function(hljs) {
...
+  var ESCAPED_CHARACTER = (rawDelimiter = "") => ({
+    className: 'subst',
+    variants: [
+      { begin: concat(/\\/, rawDelimiter, /[0\\tnr"']/) },
+      { begin: concat(/\\/, rawDelimiter, /u\{[0-9a-fA-F]{1,8}\}/) }
+    ]
+  });
+  var ESCAPED_NEWLINE = (rawDelimiter = "") => ({
+    className: 'subst',
+    begin: concat(/\\/, rawDelimiter, /[\t ]*(?:[\r\n]|\r\n)/)
```
joshgoebel, line 98: Much easier to read.

Is this variation and nuance in ordering necessary though? Is the order truly sometimes \n\r?  Or could this be simplified to `\r?\n` as is more common?

src/languages/swift.js
svanimpe: That's something I was wondering too. The current grammar is a bit too general when it comes to keywords, and often misidentifies them. In this case, String interpolation only supports expressions, not statements, so most of the keywords in `SWIFT_KEYWORDS` shouldn't occur here.

src/languages/swift.js
svanimpe: That's just what the official grammar says:

```
line-break → U+000A
line-break → U+000D
line-break → U+000D followed by U+000A
```

So `\r?\n` is not correct, as a single `\r` is allowed too.

src/languages/swift.js
joshgoebel: Yes, I was flagging the obvious change in behavior this will have.

>  String interpolation only supports expressions, not statements, so most of the keywords in SWIFT_KEYWORDS shouldn't occur here.

If that's the case then this removal sounds reasonable...  (although if it breaks any of the existing tests it'll need to be looked at more carefully).

joshgoebel:
> If the grammar looks good, I'll add a suite of unit tests and update the changelog/contributor list.

I think this is looking pretty good, feel free to finish it up.

svanimpe:
Ready for a final review!

CHANGES.md
```diff
@@ -7,6 +7,7 @@ New Languages:
 Language improvements:
 
 - enh(makefile): Add `make` as an alias (#2883) [tripleee][]
+- enh(swift) Improved grammar for strings (#2819) [Steven Van Impe][]
```
joshgoebel, line 10: You need to add link on line ~25 also.

svanimpe: Done

svanimpe:
Is anything blocking this? I see "Changes requested" but no unresolved conversations.

joshgoebel (approved):

joshgoebel:
@svanimpe Nope. All good!

PR #2903, opened 2020-12-03
enh(asciidoc) Support level 5 sections

### LINKED ISSUE #2868: (asciidoc) Section titles level 5 (i.e. using 6 "=" characters)
**Describe the issue**

Section titles level 5 (i.e. using 6 "=" characters) are not recognized as section titles.

`asciidoc`

**Sample Code to Reproduce**

```
= Document Title (Level 0)

== Level 1 Section Title

=== Level 2 Section Title

==== Level 3 Section Title

===== Level 4 Section Title

====== Level 5 Section Title

== Another Level 1 Section Title
```

https://asciidoctor-docs.netlify.app/asciidoc/current/sections/titles-and-levels/#section-level-syntax

**Expected behavior**

`====== Level 5 Section Title` should be highlighted as a section title.

**Additional context**

For reference:

- Asciidoctor regular expression: https://github.com/asciidoctor/asciidoctor/blob/697d7171830f3dc757175bc29b7eb528d4cdb4b5/lib/asciidoctor/rx.rb#L239
- AsciiDoc documentation: https://asciidoctor-docs.netlify.app/asciidoc/current/sections/titles-and-levels/#section-level-syntax

il3ven:
Resolves #2868 

### Changes
Updated regexp for sections in `src/languages/asciidoc.js`

Before
```
begin: '^(={1,5}) .+?( \\1)?$'
```
After
```
begin: '^(={1,6})[ \t].+?([ \t]\\1)?$'
```

1. Section titles can contain `======`  that is why changed `(={1,5})` to `(={1,6})`
1. A `=`  can be followed by a space or a tab. So, added `[ \t]`.

Changes are based upon https://github.com/asciidoctor/asciidoctor/blob/697d7171830f3dc757175bc29b7eb528d4cdb4b5/lib/asciidoctor/rx.rb#L239

### Checklist


joshgoebel:
I'm confused the site you linked to has `(=={0,5})` and seems to indicate `==` is Level 1, `====` level 2, etc...

I'm not sure why they have the 0... since is a line really a title without any `====`?  But given their rule it would seem that

```
========== (== * 5)
```

Is also a possibility, no?

joshgoebel:
The examples on their doc site seem to contradict the whole `=={0,5}` also... so frustrating.

test/markup/asciidoc/default.expect.txt
```diff
@@ -62,4 +62,11 @@ I&#x27;ll join that party, too.
 <span class="hljs-bullet">. </span>one thing (yeah!)
 <span class="hljs-bullet">. </span>two thing <span class="hljs-code">`i can write code`</span>, and <span class="hljs-code">`more`</span> wipee!
 
+<span class="hljs-section">= Document Title (Level 0)</span>
+<span class="hljs-section">== Level 1 Section Title</span>
+<span class="hljs-section">=== Level 2 Section Title</span>
+<span class="hljs-section">==== Level 3 Section Title</span>
+<span class="hljs-section">===== Level 4 Section Title</span>
+<span class="hljs-section">====== Level 5 Section Title</span>
```
joshgoebel, line 70: Can we please add a test case showing the suffix version as well and that it also works?

joshgoebel (changes requested):

test/markup/asciidoc/default.expect.txt
il3ven: I have added them

joshgoebel (approved):

il3ven:
> `========== (== * 5)` Is also a possibility, no?

No. 

I think the regular expression `=={0,5}` means that after a `=` there can be 0 to 5 `=` signs. 

`======` is a valid section as in this case `=` is followed by 5 `=`. However, `=======` is not a valid section as the `=` is followed by 6 `=`.

I think a reason they used `=={0,5}` instead of `={1,6}` is to differentiate sections from document title.

```
= Document Title (Contains a single = )
```

Mogztter:
Thanks @il3ven for taking care of it.

You got it right.
`= Document Title` is a document title not a section title.

> The examples on their doc site seem to contradict the whole =={0,5} also... so frustrating.

It's two equal signs (`==`) followed by 0 up to 5 equal signs.
So effectively between `==` (2 equal signs) and `======` (6 equal signs).

joshgoebel:
> I think the regular expression =={0,5} means that after a = there can be 0 to 5 = signs.

It totally does, and I know that. To group the == you'd need (==) of course.  That whole `==` is just infuriating and confusing. :-)

> I think a reason they used =={0,5} instead of ={1,6} is to differentiate sections from document title.

`=={0,5}` and `={1,6}` have exactly the same behavior.  I can't see any reason to prefer `==` so I have no idea why they chose to write it that way.

> = Document Title is a document title not a section title.

I don't think we care, it'll still match as a "section" for us, no?

> It's two equal signs (==) followed by 0 up to 5 equal signs. So effectively between == (2 equal signs) and ====== (6 equal signs).

Nope.

```
=                 A single = 
={0,5}          0 to 5 =s
```

So `=` would match.

Mogztter:
Ah ah ah I got tricked 😬

> So = would match.

On second thought,  you can have section title level 0 (i.e., a single `=`) when using a book doctype.
Maybe the reasoning was to match the AsciiDoc definition, you can have level 0 to level 5 section titles.

Issue #2847, opened 2020-11-11
(Swift) Letter c is a built-in?


svanimpe:
In the following snippet:

```swift
func max(_ a: Double, _ b: Double, _ c: Double) -> Double {
  if a > b {
    return a > c ? a : c
  } else {
    return b > c ? b : c
  }
}
```

The parameter c is incorrectly highlighted as a built-in.

svanimpe:
This seems to result from https://github.com/highlightjs/highlight.js/blob/760dcd6080ba4ef7763ff1deff250530efa1dc94/src/languages/swift.js#L26

No idea what that `c` is doing there. There is no variable or function with that name in the Standard Library.

joshgoebel:
I asked @mportiz08 if they know anything about this and whether it could be merged into their PR.  If no one knows what it's for then does seem like something we could just yank.

joshgoebel:
I'd be ok with a PR to remove this it's definitely not a general purpose keyword.

svanimpe:
@joshgoebel When I finish the project I'm working on, I'll see if I can fix some of the Swift issues I've reported myself.

svanimpe:
@joshgoebel What do you think about removing the list of built-ins all together?

I looked into updating them, however, I now feel like we should just remove them. In Swift, functions can be overloaded not only on the number and types of their parameters, but also on the parameter names. Using only a function's base name to highlight it will lead to many false matches. Proper highlighting would require a type checker.

joshgoebel:
> I now feel like we should just remove them. In Swift, functions can be overloaded not only on the number and types of their parameters, but also on the return type and on the parameter names. Using only a function's base name to highlight it will lead to many false matches.

Removing a large swath of keywords makes auto-detection worse for the _whole library_, so generally it's frowned upon.

Can you give a few specific examples of the type of issues you're imagining?  Or is this `c` issue truly representative?  Typically for something like this we'd add a rule to eliminate false positives.  If these are built-in functions could we match then as `built_in(...)` so there would be no false positives unless the built-in was immediately proceeded by a `(`, etc?

svanimpe:
I'm only talking about the global functions in the Standard Library here, not keywords.

The problem is that Swift has now moved most functions into types as methods, and the few functions that remain global often have general names such as `min`, `max`, `zip`, `print`, `stride`, `sequence` etc. These aren't reserved names, so you can declare variables with that name, or even other global functions, as long as they are valid overloads.

[...]

PR #2908, opened 2020-12-04
(swift) Revamped keywords and added support for operators

### LINKED ISSUE #2856: (Swift) left and right keywords and precedencegroup highligthing
Highlight.js highlights `left` and `right` as keywords. It's true that they have some special meaning (see https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#grammar_precedence-group-declaration), however, they aren't reserved keywords, so 99% of the time, they will just be variables.

What is the policy regarding keywords? Should only reserved words be highlighted as keywords?

svanimpe:
I'm working on some significant improvements to the grammar, and this felt like something I could submit already.

* I've redone the keywords as that list was very outdated. The new list clearly separates the actual (reserved) keywords from contextual keywords (that are keywords only in specific contexts). In the future, I hope to move some of these contextual keywords into specific modes, which will avoid false matches.
* I've also fixed #2856 by leaving out some rarely used keywords that will result in more false than actual matches. These keywords relate to precedence groups and can be highlighted in a future PRECEDENCEGROUP mode.
* I've implemented the grammar for operators so they can be highlighted. This will also enable me to identify operator function declarations in the future.
* Note that the grammar for operators isn't 100% correct as I don't know how to match Unicode characters above `\uffff`.


svanimpe:
@joshgoebel I also have a new grammar for attributes ready (so we can get rid of the hardcoded list), but that depends on this PR.

svanimpe:
Hopefully, this will also allow me to fix #2895 eventually.

svanimpe:
I've detected an issue here: keywords like `private(set)` are matches as separate keywords `private` and `set`, even though I've changed the `$pattern` to look for parentheses as well. At first I thought this was due to the order of the keywords (`private` was listed before `private(set)`), but even after reordering them, it doesn't work.

Also, keywords that start with a number sign also don't work anymore and I just realized these keywords may also have parentheses with arguments after them, which may trip up the matching.

Do you have any ideas what's going on here?
I'm thinking I should just remove the parentheses again and match the keywords separately, but again, that will result in false matches.

svanimpe:
I did detect an issue here: some keywords, such as `#available` can be followed by parentheses with arguments, in which case the keyword matcher seems to add the arguments and doesn't match the keyword itself. For example

The text `#available` matches the keyword `#available`
The text `#available(test)` looks for a keyword `#available(test)` and doesn't match `#available`.

So I have more work to do here.

@joshgoebel How do you recommend I handle keywords with parentheses? The old grammar didn't match them and just happened to highlight keywords such as `private(set)` correctly as two separate keywords, because `set` is also listed as a contextual keyword (although that's referring to another context). However, a keyword like `unowned(unsafe)` wasn't matched, and also will not match as two keywords, because `unsafe` is not a keyword.

joshgoebel:
> The text #available(test) looks for a keyword #available(test) and doesn't match #available.

It would only do that if () are part of `keywords.$pattern`... and if so that is the correct behavior... `$pattern` is used to scan the current buffer and if there are any matches they are checked against the list of verbatim keywords.

You **probably** do NOT want `()` included in $pattern.  More "complex" keywords like `unowned(unsafe)` might better be handled with their own mode.

src/languages/swift.js
```diff
@@ -7,39 +7,52 @@ Website: https://swift.org
...
-    };
+  // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID413
+  // https://docs.swift.org/swift-book/ReferenceManual/zzSummaryOfTheGrammar.html
+  const SWIFT_KEYWORDS = {
+    // Override the pattern since the default of /\w+/ is not sufficient to capture
+    // the keywords that start with a number sign (#) or that have parentheses.
+    $pattern: /#?\w+(\(\w+\))?/,
+    keyword:
+      // Reserved keywords. These require backticks to use as an identifier:
+      '_ Any as associatedtype break case catch class continue default defer deinit do else enum extension ' +
```
joshgoebel, line 21: Since we're reworking all this... please switch to using an array with one keyword per line and then `KEYWORD.join(" ")` that array as necessary... see `ecmascript.js` for examples, etc...  this is for ease of maintenance in the future.

src/languages/swift.js
```diff
@@ -98,13 +111,72 @@ export default function(hljs) {
...
+  const operatorCharacter = either(
+    operatorHead,
+    /[\u0300–\u036F]/,
+    /[\u1DC0–\u1DFF]/,
+    /[\u20D0–\u20FF]/,
+    /[\uFE00–\uFE0F]/,
+    /[\uFE20–\uFE2F]/,
+    // TODO: The following characters are also allowed, but the regex doesn't work as intended.
+    // For example, it also matches -10 as an operator.
+    // /[\u{E0100}–\u{E01EF}]/u
```
joshgoebel, line 144: This doesn't seem right.  From what I'm reading this is the full range of what MIGHT be used for defining a custom operator... just because something _could potentially_ be an operator doesn't mean it most assuredly IS an operator and should be highlighted everywhere as an operator... correct?

This "wideness" might make sense in the context of something like `func [x] (a,b)` where [x] is an operator being defined (and we know that by context), but I don't think it makes sense in general.  Pretty sure in the standard case we should only highlight well-known built-in operators that Swift supports, like `>`, `<`, etc...

src/languages/swift.js
```diff
@@ -98,13 +111,72 @@ export default function(hljs) {
...
+  );
+  const OPERATOR = {
+    className: 'operator',
+    variants: [
+      { // TODO: Replace with negative look-behind when available.
+        className: 'keyword',
+        begin: /\s(?=as[?!]?\s)/,
+        excludeBegin: true,
```
joshgoebel, line 152: Can't the whole begin be a look-ahead avoiding `excludeBegin`?

src/languages/swift.js
```diff
@@ -98,13 +111,72 @@ export default function(hljs) {
...
+        end: /as[?!]?(?=\s)/
+      },
+      { // TODO: Replace with negative look-behind when available.
+        className: 'keyword',
+        begin: /\s(?=is\s)/,
+        excludeBegin: true,
+        end: /is(?=\s)/
```
joshgoebel, line 159: Is this look-ahead necessary?  Hasn't it already been proven by the begin?

src/languages/swift.js
joshgoebel, line 144: Pretty sure you can't use `\u{}` without being in UTF-8 regex mode...

src/languages/swift.js
```diff
@@ -98,13 +111,72 @@ export default function(hljs) {
...
+    /[\u1DC0–\u1DFF]/,
+    /[\u20D0–\u20FF]/,
+    /[\uFE00–\uFE0F]/,
+    /[\uFE20–\uFE2F]/,
+    // TODO: The following characters are also allowed, but the regex doesn't work as intended.
+    // For example, it also matches -10 as an operator.
+    // /[\u{E0100}–\u{E01EF}]/u
+  );
+  const OPERATOR = {
+    className: 'operator',
```
joshgoebel, line 147: Are these operators or keywords?

src/languages/swift.js
svanimpe: Will do!

svanimpe:
> You **probably** do NOT want `()` included in $pattern. More "complex" keywords like `unowned(unsafe)` might better be handled with their own mode.

I was thinking the same thing. And also to match the `#` keywords with the same rules as `@` attributes, as they have the same grammar (w.r.t. the arguments that can follow). For example:

```
#available(iOS 14, *)
@available(iOS 14, *)
```

The first is a keyword, whereas the second is an attribute 🤷‍♂️

I can just assign them a different class (`keyword` instead of `meta`). That would reduce the list of keywords to only the basic, simple keywords.

Or should I use `meta-keyword` here. Will that highlight with the keyword color?

src/languages/swift.js
svanimpe: Some of these characters are used elsewhere in the grammar, such as `<GenericParam>`, or `OptionalType?`, but other than that, they're always operators.

I assume the set of "head" character for operators and other identifiers is disjunct. For example, if you look at the range A1-FF, the ones that are missing from the list here are listed explicitly as valid head characters for regular (non-operator) identifiers:

\u00A8\u00AA\u00AD\u00AF\u00B2–\u00B5\u00B7–\u00BA
\u00BC–\u00BE\u00C0–\u00D6\u00D8–\u00F6\u00F8–\u00FF

So I think this should work just fine, as long as the cases where operators are used as punctuation are handled by specific modes. In the case of generics and optional types, that's already the case.

src/languages/swift.js
svanimpe: I'll give that a try.

src/languages/swift.js
svanimpe: I'll check.

src/languages/swift.js
svanimpe: Is it OK if I just leave this as a TODO and wait for #2756? 

Swift has full Unicode support, even for identifiers/titles, so there will be other cases where we can't detect a match because it uses characters above 0xFFFF

src/languages/swift.js
svanimpe: Is this question about `as? as! try? try!`?

If so, the grammar specifies them as operators and keywords at the same time, so I'm highlighting them as keywords, which is what other grammars do too, although most don't bother highlighting the question or exlamation mark. I preferred to include that in the match, otherwise it would be matched as a postfix operator.

joshgoebel: If they are keywords then you should remove "operator", it's not serving any purpose.

svanimpe: OK, I'll add them to list of "special" keywords (like the ones that have parentheses) instead, so I don't have to change the keywords $pattern to allow question marks or exclamation points.

My goal here was to group these keywords under the "operator" umbrella as well, in case a future grammar says "any operator is allowed here", which would then also include these keywords.

joshgoebel:
> I can just assign them a different class (keyword instead of meta). That would reduce the list of keywords to only the basic, simple keywords. Or should I use meta-keyword here? Will that highlight with the keyword color?

You should assign them the most semantically correct class.  You can never 100% assume anything about color - that is 100% up to themes and sometimes themes defy all reasoning.  `meta-keyword` would seem a reasonable usage if they are truly special kind of keywords.  Often we use that for compiler directives, special statements, etc... you can always look thru other grammars (to see how they use a classname) though this can always sometimes lead one astray. :)

svanimpe:
@joshgoebel I'm running into a lot of issues with how keywords are handled in HLJS.

For example:

- `x.var()` highlights `var` as a keyword, whereas it clearly is a method. This is allowed in Swift, because you can declare properties and methods with keyword names by wrapping them in backticks (`` func `var`() { }``)
- `Any` and `Self` highlight as normal types, whereas they should be keywords. This is due to the rule that keywords are only matched as leftovers. In this case, the `TYPE` mode already matches these.
- Swift as lots of keywords with special characters. For example `private(set)`, `#if` and `@objc(name)` are generally all highlighted as keywords and require special modes.

I'm starting to think I should just stop using the `keywords:` attribute and just match *all* keywords in a KEYWORD mode. That would remove a lot of "dancing around the keyword rules" that I currently have to do. What do you think?

svanimpe:
@joshgoebel I've made a lot of changes here:

- I've moved all keywords to a KEYWORDS mode as that was the only way to highlight them properly. The new grammar handles all sorts of keywords now, even those with special characters, and those that require (or may have) a leading dot. Fixes #2856
- Likewise, built-ins now also highlight properly. The new grammar should remove 99% of false positives. The remaining 1% are global functions that are overloads of a built-in, but those should be extremely rare (and acceptable to highlight as a built-in). Fixes #2847 
- I've added support for quoted identifiers, as those shouldn't highlight as keywords.
- I've also added support for implicit parameters and property wrapper projections (`$0`, `$name`, etc). Fixes #2871
- I've written a grammar for attributes and adopted the convention of highlighting built-in attributes as keywords, and user-defined attributes as metadata.
- I've extended the number of modes supported in string interpolation, including expressions that contain parentheses.

I've also changed the author to myself, as there's hardly any original code left, so I should be the one the contact/blame in case there's an issue with the grammar 😅 Is that OK?

joshgoebel:
> x.var() highlights var as a keyword, whereas it clearly is a method. This is allowed in Swift, because you can declare properties and methods with keyword names by wrapping them in backticks ( func `var`() { })

Yes, the "typical" solution here is to add a guard/rule for dispatch (to cover these cases), as I believe I've already mentioned several places.  That is the direction the library is headed as well since we'd like to highlight dispatch in the future as a thing.

> Any and Self highlight as normal types, whereas they should be keywords. This is due to the rule that keywords are only matched as leftovers. In this case, the TYPE mode already matches these.

I have less advice here.  This is sort of unusual as most grammars do not match just any identifier like this.  What I'd do though is add a special rule for the *exceptions* (Any, Self)...

> Swift as lots of keywords with special characters. For example private(set), #if and @objc(name) are generally all highlighted as keywords and require special modes.

`#` isn't necessarily hard but pretty sure I already provided guidance that the others should indeed by handled with custom rules.

> I'm starting to think I should just stop using the keywords: attribute and just match all keywords in a KEYWORD mode. That would remove a lot of "dancing around the keyword rules" that I currently have to do. What do you think?

Generally this should be done as a last resort.  All three of the things you mention here should not be difficult to work-around and is already common in other grammars.

joshgoebel:
> I've also changed the author to myself, as there's hardly any original code left, so I should be the one the contact/blame in case there's an issue with the grammar 😅 Is that OK?

No objection.  I'll need some time to review the current state of things though.

svanimpe:
Well, I did go for that last resort, as that lets me match all keywords without too many exception modes. However, I think removing the `keywords` attribute generated some issues with language detection 😕 I have no idea how that works, so I'll just wait for your feedback.

src/languages/lib/swift.js
```diff
@@ -0,0 +1,285 @@
...
+
+// Keywords that may have a leading dot.
+export const optionalDotKeywords = [
+  /init\b/,
+  /self\b/
+];
+
+// Regular keywords and literals.
+export const keywords = [
+  /\bAny\b/,
```
joshgoebel, line 17: This \b \B repetition would be nice to abstract and avoid repeating on every line.

src/languages/lib/swift.js
```diff
@@ -0,0 +1,285 @@
...
+  /\bSelf\b/,
+  /\b_\b/, // pattern
```
joshgoebel, line 19: Too common, should be removed or will need a custom rule to be scored 0.  This is one nice thing about `keywords` it's very easy to alter their relevance individually.

src/languages/lib/swift.js
```diff
@@ -0,0 +1,285 @@
...
+  /\binit!\B/,
+  /\binout\b/,
+  /\binternal\(set\)\B/,
+  /\binternal\b/,
+  /\bin\b/,
+  /\bis\b/, // operator
+  /\blazy\b/, // contextual
+  /\blet\b/,
+  /\bmutating\b/, // contextual
+  /\bnil\b/, // literal
```
joshgoebel, line 62: How do we make sure this is properly tagged as a "literal"?  Ditto for true, false, else... is a separately list not needed?

src/languages/lib/swift.js
```diff
@@ -0,0 +1,285 @@
...
+// NOTE: Contextual keywords are reserved only in specific contexts.
+// Ideally, these should be matched using modes to avoid false positives.
+
+// TODO: Create a PRECEDENCE_GROUP mode to match the remaining contextual keywords:
+// assignment associativity higherThan left lowerThan none right
+// These aren't included in the list because they result in mostly false positives.
+
+// Keywords that start with a number character (#).
+// #available is handled separately.
+export const numberKeywords = [
```
joshgoebel, line 114:
```suggestion
export const numberSignKeywords = [
```

Otherwise it's just ambiguous and confusing.

src/languages/lib/swift.js
```diff
@@ -0,0 +1,285 @@
...
+// @available is handled separately.
+export const keywordAttributes = [
+  /autoclosure/,
+  concat(/convention\(/, either('swift', 'block', 'c'), /\)/),
+  /discardableResult/,
+  /dynamicCallable/,
+  /dynamicMemberLookup/,
+  /escaping/,
+  /frozen/,
+  /GKInspectable/,
```
joshgoebel, line 249: If they (in the context of the list) are literal strings, then prefer normal quoting not `//`.

src/languages/swift.js
```diff
@@ -1,65 +1,66 @@
...
-        'underestimateCount unsafeAddressOf unsafeBitCast unsafeDowncast unsafeUnwrap ' +
-        'unsafeReflect withExtendedLifetime withObjectAtPlusZero withUnsafePointer ' +
-        'withUnsafePointerToObject withUnsafeMutablePointer withUnsafeMutablePointers ' +
-        'withUnsafePointer withUnsafePointers withVaList zip'
-    };
+  // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID413
+  // https://docs.swift.org/swift-book/ReferenceManual/zzSummaryOfTheGrammar.html
+  const DOT_KEYWORD = {
+    begin: concat(/\./, either(...Swift.dotKeywords, ...Swift.optionalDotKeywords)),
+    returnBegin: true,
```
joshgoebel, line 19: Prefer `regex.lookahead` to returnBegin whenever you can, we'd like to deprecate returnBegin in the  future.

src/languages/swift.js
```diff
@@ -1,65 +1,66 @@
...
+    contains: [
+      {
+        begin: /\./
+      },
+      {
+        className: 'keyword',
+        begin: either(...Swift.dotKeywords, ...Swift.optionalDotKeywords)
+      }
+    ]
+  };
```
joshgoebel, line 29:
```suggestion
  const DOT_KEYWORD = {
    begin: concat(/\./, either(...Swift.dotKeywords, ...Swift.optionalDotKeywords)),
    returnBegin: true,
    relevance: 0,
    contains: [
      {
        begin: /\./,
        relevance: 0
      },
      {
        className: 'keyword',
        begin: either(...Swift.dotKeywords, ...Swift.optionalDotKeywords)
      }
    ]
  };
```

The is a double/triple relevance issue... when building complex rules that really only match a "single" item you need to tune their reelvance such that it is only 1, not 2 or 3.

src/languages/swift.js
```diff
@@ -1,65 +1,66 @@
...
+  const BUILT_IN = {
+    className: 'built_in',
+    begin: concat(/\b/, either(...Swift.builtIns), /(?=\()/)
+  };
+  const BUILT_INS = [BUILT_IN_GUARD, BUILT_IN];
+
+  // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID418
+  const OPERATOR = {
+    className: 'operator',
+    variants: [
```
joshgoebel, line 59:
```suggestion
    className: 'operator',
    relevance: 0,
    variants: [
```

Operators are not given relevance, too common and not in enough grammars yet.

joshgoebel: Many of your auto-detect issues are probably right here + plus the double/triple relevance issues.

src/languages/swift.js
```diff
@@ -126,22 +127,98 @@ export default function(hljs) {
...
+    returnBegin: true,
+    contains: [
+      {
+        className: 'keyword',
+        begin: /(@|#)available/
+      },
+      {
+        begin: /\(/,
+        end: /\)/,
+        endsParent: true,
```
joshgoebel, line 157: Necessary?

src/languages/swift.js
```diff
@@ -126,22 +127,98 @@ export default function(hljs) {
...
+    begin: /\$\d+/
+  };
+  const PROPERTY_WRAPPER_PROJECTION = {
+    className: 'variable',
+    begin: `\\$${Swift.identifierCharacter}+`
+  };
+  const IDENTIFIERS = [QUOTED_IDENTIFIER, IMPLICIT_PARAMETER, PROPERTY_WRAPPER_PROJECTION];
+
+  // https://docs.swift.org/swift-book/ReferenceManual/Attributes.html
+  const AVAILABLE = {
```
joshgoebel, line 146: AVAILABLE_ATTRIBUTE?

src/languages/swift.js
```diff
@@ -126,22 +127,98 @@ export default function(hljs) {
...
+
+  const TYPE = {
+    className: 'type',
+    begin: '\\b[A-Z][\\w\u00C0-\u02B8\']*',
+    relevance: 0
+  };
+  // slightly more special to swift
+  const OPTIONAL_USING_TYPE = {
+    className: 'type',
+    begin: '\\b[A-Z][\\w\u00C0-\u02B8\']*[!?]'
```
joshgoebel, line 189: Doesn't TypeScript also have this same thing?

joshgoebel:
Could you please use a linter plugin with your editor to lint based on our eslint files?  The linting it all over the place.

src/languages/swift.js
joshgoebel: Why do we need to match the `.` here at all though?  Are you just doing that to avoid the parent mode ending?

joshgoebel: Actually here I think I'd use the top rule to match `.lookahead(blah)` and then just use the inner contains to match the literal `blah`... that should simplify the whole thing.

test/markup/swift/functions.expect.txt
```diff
@@ -5,6 +5,6 @@
 
 <span class="hljs-class"><span class="hljs-keyword">class</span> <span class="hljs-title">MyClass</span> </span>{
   <span class="hljs-function"><span class="hljs-keyword">func</span> <span class="hljs-title">f</span><span class="hljs-params">()</span></span> {
-    <span class="hljs-keyword">return</span> <span class="hljs-literal">true</span>
+    <span class="hljs-keyword">return</span> <span class="hljs-keyword">true</span>
```
joshgoebel, line 8: **This should not have changed.**

src/languages/swift.js
```diff
@@ -126,22 +127,98 @@ export default function(hljs) {
...
+  };
+  const PROPERTY_WRAPPER_PROJECTION = {
+    className: 'variable',
+    begin: `\\$${Swift.identifierCharacter}+`
+  };
+  const IDENTIFIERS = [QUOTED_IDENTIFIER, IMPLICIT_PARAMETER, PROPERTY_WRAPPER_PROJECTION];
+
+  // https://docs.swift.org/swift-book/ReferenceManual/Attributes.html
+  const AVAILABLE = {
+    begin: /(@|#)available/,
```
joshgoebel, line 147: One of these matchers needs relevance, 0, etc...

joshgoebel (changes requested):
See comments.

src/languages/swift.js
svanimpe: What is `regex.lookahead`?
I don't see that property on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp

svanimpe: Is this even possible with a lookahead? I'm doing this workaround because lookbehind isn't supported. I want to highlight only the keyword, if it's preceded by a dot (which I don't want to highlight). The rule should just be:

```
begin: concat(/(?<=\.)/, either(...Swift.dotKeywords, ...Swift.optionalDotKeywords))
```

But that doesn't work yet.

Lookbehind will let me remove a lot of workarounds from the code, so I'm hoping it'll be supported soon. I also have some experience writing TextMate grammars, and lookbehind was quite essential there too.

src/languages/swift.js
svanimpe: The issue here was that I got a "0 width" error if I didn't provide a rule to match the dot. Is that because I'm using `returnBegin`? My assumption was that whatever I returned also had to be matched by submodes.

W.r.t the relevance, just so I understand it correctly: Is the issue here that both the outer rule **and** the inner rule add to the relevance, and because they're nested, that relevance can be boosted too much?

src/languages/swift.js
svanimpe: Do you mean the `endsParent`? I wasn't sure.

Will the parent rule automatically end if there's no more matching `contains` rule?
Will `endsParent` speed up the matching? (so it doesn't have to look for another subrule)

src/languages/swift.js
svanimpe: Maybe, optional types are becoming more common now. 
But I didn't change this code, it just got moved around a bit.

test/markup/swift/functions.expect.txt
svanimpe: Yeah, Swift lists `true`, `false`, and `nil` as both keywords and literals, and Xcode highlights them as keyword, which is why I've included them here.
But I'll change them back to the `literal` class.

src/languages/lib/swift.js
svanimpe: Can I mix strings and regexes in the same array, will that work?

svanimpe:
W.r.t. the linter: I've installed the ESLint extension, but that also complains about code I didn't write and that has the same format as other language grammars. Should I take every warning as an error that needs to be fixed?

And what about this one?

Screen Shot 2020-12-11 at 11 45 55

Sometimes the linter complains about these unicode character ranges, sometimes it doesn't 🤷‍♂️

src/languages/lib/swift.js
svanimpe: Done.

src/languages/lib/swift.js
svanimpe: Done.

src/languages/lib/swift.js
svanimpe: Fixed!

src/languages/lib/swift.js
svanimpe: Done.

test/markup/swift/functions.expect.txt
svanimpe: Done.

joshgoebel:
> W.r.t. the linter: I've installed the ESLint extension, but that also complains about code I didn't write and that has the same format as other language grammars. 

It should be safe to run `eslint --fix` directly on files in `languages/*`.  Though not all languages have yet been so linted (many have).

> Should I take every warning as an error that needs to be fixed?

Most of them should be pretty easily fixable... you can run `--fix` to fix all the spacing and simple things.

**no-misleading-character-class**

I don't undertstand that one, we may want to just disable it.

> It also seems to behave differently for my lib/swift.js file than for my language grammar.

That may be because of my ruleset... I'll take a look at that...

src/languages/swift.js
joshgoebel: > But I didn't change this code, it just got moved around a bit.

No blame; you're pretty much rewritten the whole file (as you pointed out before) so now that means your also shining the light on some other broken things we should fix while we see them... that's just what happens in huge language reboot PRs.

So I don't think we need two rules here anymore... just a single rule with an optional `[?!]`, but `relevance: 0` for all cases.

joshgoebel:
You can ignore some of the subtle type issues, but I think those are coming from TypeScript, not ESLint.  I just fetched the latest PR and the grammar file looks very clean now.

src/languages/lib/swift.js
```diff
@@ -0,0 +1,285 @@
...
+);
+
+// Valid first characters for identifiers.
+export const identifierHead = either(
+  /[a-zA-Z_]/,
+  /[\u00A8\u00AA\u00AD\u00AF\u00B2–\u00B5\u00B7–\u00BA]/,
+  /[\u00BC–\u00BE\u00C0–\u00D6\u00D8–\u00F6\u00F8–\u00FF]/,
+  /[\u0100–\u02FF\u0370–\u167F\u1681–\u180D\u180F–\u1DBF]/,
+  /[\u1E00–\u1FFF]/,
+  /[\u200B–\u200D\u202A–\u202E\u203F–\u2040\u2054\u2060–\u206F]/,
```
joshgoebel, line 228: Is \u2054 just supposed to be a one off?

joshgoebel:
```
> /[\u2020–\u2027]/.exec("\u2024")
null
> /[\u0300–\u036F]/.exec("\u0301")
null
>
```

The second case the linter flagged as problematic... the first it did not, yet it also seems to not work.  In my quick testing only the outer characters are matched.  Am I reading this wrong or misunderstanding the goal of these matchers?

joshgoebel:
@svanimpe Did you just copy and paste a bunch of stuff from the Swift docs?  That "-" you're using is _not_ "-" it's something that looks like it.  I think that's a lot of the problem here.

```js
> /[\u2020–\u2027]/u.exec("–")
[ '–', index: 0, input: '–', groups: undefined ]
```

‼

svanimpe:
> Did you just copy and paste a bunch of stuff from the Swift docs?

Yes, and then used regex find/replace to reformat it. I figured that was the best way to avoid mistakes, but clearly, it wasn't foolproof 😅 Good catch!

But I assume that didn't fix the issues with the 5-hex-digit unicode characters, and that those are still unsupported?

> Is \u2054 just supposed to be a one off?

Yes.

joshgoebel:
> and that those are still unsupported?

Yes, unrelated.

src/languages/swift.js
joshgoebel: > Lookbehind will let me remove a lot of workarounds from the code, so I'm hoping it'll be supported soon. 

Don't hold your breath.  Even after Safari supports it we have to decide when we want to adopt it... right now we green field browsers going back 3-4 years... so if we waited that long it'll be a LONG time coming.

For a simple "pair" `[".", something]` you don't even need this muck at all just use begin/end:

```js
  const DOT_KEYWORD = {
    begin: concat(/\./, lookahead(either(...Swift.dotKeywords, ...Swift.optionalDotKeywords))),
    end: either(...Swift.dotKeywords, ...Swift.optionalDotKeywords),
    skipBegin: true
```

If you want you could rebase onto https://github.com/highlightjs/highlight.js/pull/2834 and just use `beforeMatch` sugar.  It's implemented differently still, but it does the same thing.

src/languages/swift.js
joshgoebel: > W.r.t the relevance, just so I understand it correctly: Is the issue here that both the outer rule and the inner rule add to the relevance, and because they're nested, that relevance can be boosted too much?

Yes.  Sometimes one might want this but for matching "simple" things where you're only using rules to jump thru hoops, you do not.  Typically identifiable "things" in a grammar should always score 1, not more than 1.

---

See my latest thoughts above on how to do this without all these hoops.

src/languages/swift.js
joshgoebel: > My goal here was to group these keywords under the "operator" umbrella as well, in case a future grammar says "any operator is allowed here", which would then also include these keywords.

I think you may be over thinking that. :)

src/languages/swift.js
joshgoebel: Well this reads weird later:

```
  const ATTRIBUTES = [
    AVAILABLE,
    KEYWORD_ATTRIBUTE,
    USER_DEFINED_ATTRIBUTE
  ];
```

Is it an attribute or not?  Lets clear up the confusion.

joshgoebel: Oh I see now.  Still, lets find some way to remove the confusion. :)

src/languages/swift.js
joshgoebel: > Will the parent rule automatically end if there's no more matching contains rule?

Yes, if there is no `end` matcher to "hold the mode open".

> Will endsParent speed up the matching? (so it doesn't have to look for another subrule)

*Technically*, but not to the extent it should matter and we don't use this type of thinking in any other grammars.  If it's clear (gramatically) that a rule will short-circuit itself, we usually just let it.

src/languages/swift.js
svanimpe: Done (in the next commit). But what about this one:

```js
  const AVAILABLE = {
    begin: /(@|#)available/,
    returnBegin: true,
    contains: [
      {
        className: 'keyword',
        begin: /(@|#)available/
      },
      {
        begin: /\(/,
        end: /\)/,
        endsParent: true,
        keywords: Swift.availabilityKeywords.join(' '),
        contains: [
          OPERATOR,
          NUMBER,
          STRING
        ]
      }
    ]
  };
```

Here, I only want to highlight the keyword itself, but the match should also consume the argument list.

src/languages/swift.js
svanimpe: Done (in the next commit).

src/languages/swift.js
svanimpe: Done (in the next commit).

src/languages/swift.js
svanimpe: I just named it AVAILABLE_ATTRIBUTE since it's matched as part of the attributes. I guess that should work.

src/languages/swift.js
svanimpe: I've set the relevance to 0.

src/languages/swift.js
svanimpe: Done.

svanimpe:
After setting some relevances to 0, the language detection issues seem resolved. All tests are back to green now :)

src/languages/lib/swift.js
joshgoebel: Mixing isn't a problem.  The underlying utility stuff and compiler usually handles strings and regex interchangeably without any issue.  It's only an issue when doing something like `a+b` or string concatenation, which mostly we're trying to get away from.

src/languages/swift.js
joshgoebel: > The rule still seems to work when I replace 

Cause it's pretty much just two ways of doing the same thing.  Just one rewinds the cursor and one avoids moving it forward.

You could collapse this to a single rule I think, couldn't you?

```ts
  const AVAILABLE = {
        begin: /(@|#)available\(/,
        keywords: "@available #available"
        // ...
```

src/languages/lib/swift.js
```diff
@@ -0,0 +1,301 @@
+/* eslint-disable no-misleading-character-class */
```
joshgoebel, line 1: Please don't add these, if there are issues remaining we need to resolve them.

src/languages/lib/swift.js
```diff
@@ -0,0 +1,301 @@
...
+  // /[\u{50000}-\u{5FFFD}\u{60000-\u{6FFFD}\u{70000}-\u{7FFFD}\u{80000}-\u{8FFFD}]/u,
+  // /[\u{90000}-\u{9FFFD}\u{A0000-\u{AFFFD}\u{B0000}-\u{BFFFD}\u{C0000}-\u{CFFFD}]/u,
+  // /[\u{D0000}-\u{DFFFD}\u{E0000-\u{EFFFD}]/u
+);
+
+// Valid characters for identifiers.
+export const identifierCharacter = either(
+  identifierHead,
+  /\d/,
+  /[\u0300-\u036F\u1DC0-\u1DFF\u20D0-\u20FF\uFE20-\uFE2F]/
```
joshgoebel, line 249: I'm showing `\uFE20-\uFE2F` (the whole range) as problematic because it's a combining character.  Unless you want to add a specific test that proves this works as it should we should remove it/comment it out.

src/languages/swift.js
```diff
@@ -1,96 +1,127 @@
...
+  // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID418
+  const OPERATOR = {
+    className: 'operator',
+    relevance: 0,
+    variants: [
+      {
+        begin: Swift.operator
+      },
+      {
+        begin: `\\.(\\.|${Swift.operatorCharacter})+`
```
joshgoebel, line 83: What are both these cases accomplishing? I'm not sure I understand the second.

src/languages/swift.js
```diff
@@ -123,45 +161,139 @@ export default function(hljs) {
...
+  };
+  const ATTRIBUTES = [
+    AVAILABLE_ATTRIBUTE,
+    KEYWORD_ATTRIBUTE,
+    USER_DEFINED_ATTRIBUTE
+  ];
+
+  const TYPE = {
+    className: 'type',
+    begin: '\\b[A-Z][\\w\u00C0-\u02B8\']*',
```
joshgoebel, line 223: Please convert to regex.

src/languages/swift.js
```diff
@@ -123,45 +161,139 @@ export default function(hljs) {
...
+    relevance: 0
+  };
+  // slightly more special to swift
+  const OPTIONAL_USING_TYPE = {
+    className: 'type',
+    begin: '\\b[A-Z][\\w\u00C0-\u02B8\']*[!?]',
```
joshgoebel, line 229: Please convert to regex.

joshgoebel:
> the language detection issues seem resolved. All tests are back to green now :)

Indeed, looking pretty good overall I think.  I plan to spend a little time on this myself and see if the keywords can be simplified any.

src/languages/lib/swift.js
svanimpe: Yes this works and is valid Swift code:

```swift
infix operator ~︦ // ~ 0xFE26

func ~︦ (lhs: Int, rhs: Int) -> Int {
    lhs + rhs
}

1 ~︦ 2
```

Screen Shot 2020-12-13 at 13 07 35

Should I add this as a unit test?

I assume this isn't a problem here because there are different rules for the first character of an operator. I'm not familiar with these ranges, but I assume the rules for `operatorHead` don't include combining characters?

src/languages/swift.js
svanimpe: The first case is a regular operator: a `head` followed by zero or more valid characters.

The second is a dot-operator: only operators that start with a dot are allowed to use dots as characters (`...`, `...<`, `.*`, etc). So there rule here is: a dot followed by one or more characters that may also include dots.

src/languages/swift.js
svanimpe: I've already done that in the code for the next PR, as I'll have to change this rule anyway to fix https://github.com/highlightjs/highlight.js/issues/2895

src/languages/swift.js
svanimpe: Same here :)

src/languages/swift.js
joshgoebel, line 83:
```suggestion
        // dot-operator: only operators that start with a dot are allowed to use dots as 
        // characters (..., ...<, .*, etc). So there rule here is: a dot followed by one or more 
        // characters that may also include dots.
        begin: `\\.(\\.|${Swift.operatorCharacter})+`
```

src/languages/lib/swift.js
joshgoebel: It wouldn't hurt.  Do you have any idea what that warning is trying to tell us? Can the character also be used to make a compound character and then that doesn't work?

joshgoebel: >  but I assume the rules for operatorHead don't include combining characters?

This one regex is the only one still being flagged and just that portion of it.  So it's unique somehow.

joshgoebel:
I cleaned this up a little:

- Simplified `@|#available` as previous suggested.
- Moved the simpler keywords (most of keywords, literals, numericKeywords) back into the standard `keywords` framework

Any questions before we merge?

svanimpe:
I just wonder whether I'll run into issues again with the regular keyword system, as keywords (such as `Any` and `Self`) take priority over modes (such as `TYPE`), but regular keywords are matched *after* modes. However, the tests include `Any` and `Self`, so if those are still green, everything should be fine.

joshgoebel:
> I just wonder whether I'll run into issues again with the regular keyword system, as keywords (such as Any and Self) take priority over modes (such as TYPE)

That's why Any and Self are still separate rules (to prevent type from grabbing them).  The keyword system should be preferred when it works, and then use modes to fill in the gaps.  We'd like to extend the keyword system long-term where it makes sense not have every grammar build their own custom system.  So forcing you to get used to working with it is part of the long-term plan. :-)

That said, cases like TYPE are probably not going away anytime soon because keywords are processed "in the gaps" and probably always will be (that's their distinction from modes). So the solution is just to handle Any and Self with their own rule that fires before type.

Long-term multi-patterns and look-behind would resolve a lot of your other keyword issues I think.  Though you can really already do multi-patterns using `either` with `$pattern` for many cases. (as I do here)

joshgoebel:
@svanimpe Thanks for all the effort on this!

joshgoebel (approved):

svanimpe:
@joshgoebel I ran into a problem with the new keyword regex: `/\b\w+(\(\w+\))?/`

I'm trying to match tuples like `(name: Type, name: Type)` to prevent `name` from being highlighted as a keyword, in case it matches any (which is allowed). However, because keywords with parentheses are now handled by the regular keyword system again, the tuple rule takes precedence and screws up the keyword matching.

I still feel like matching keywords last is only going to keep producing problems like this, which is why I moved them all into modes 😕 To solve my current problem, I'd have to revert some of your changes and change keywords with parentheses back to regexes, so they're included in the KEYWORD_MODES. Is that OK for you?

joshgoebel:
> However, because keywords with parentheses are now handled by the regular keyword system again, the tuple rule takes precedence and screws up the keyword matching. ... revert some of your changes and change keywords with parentheses back to regexes

Please push the code that isn't working with a failing test and let me look at it.  Often there are easy work-arounds for many of these kinds of problems.  Off the top I wouldn't expect this to be hard to do.

Isn't a rule to match `[a-z]+:` sufficient... grabs all tuples... and keywords match the remainder?

PR #2918, opened 2020-12-15
enh(cpp) support pack expansion in function declarations

### LINKED ISSUE #2917: (CPP) The highlighter doesn't support variadic templates/parameter packs
**Describe the issue**

This is valid C++ syntax:

``` C++
template<typename... T>
void foo(const T &... args) {}
```

Yet, when highlighting with `highlight`, the returned object's `illegalBy.msg` is `Illegal lexeme "." for mode "function"`.

`cpp`

`highlight`

**Sample Code to Reproduce**

Here's a modified version of the example JSFiddle: https://jsfiddle.net/dueh84nq/1/ - I changed it to display the illegal/illegalBy values.

**Expected behavior**

Since the syntax is legal, I would expect it to be highlighted somehow, not rejected as illegal.

**Additional context**

Here's some info on C++'s parameter packs: https://en.cppreference.com/w/cpp/language/parameter_pack

mortie:
Make dots legal in function declarations, which is necessary for variadic functions with parameter packs in C++: https://en.cppreference.com/w/cpp/language/parameter_pack

Resolves #2917 

### Checklist
- [x] Added markup tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ New Languages:
 
 Language improvements:
 
+- enh(cpp): Support C++ pack expansion in function arguments [Martin Dørum][]
 - enh(makefile): Add `make` as an alias (#2883) [tripleee][]
 - fix(http) avoid recursive sublanguage and tighten rules (#2893) [Josh Goebel][]
 - enh(swift) Improved grammar for strings (#2819) [Steven Van Impe][]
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -6,6 +6,7 @@ New Languages:
 
 Language improvements:
 
+- enh(cpp): Support C++ pack expansion in function arguments
```
joshgoebel, line 9: Please add your name at the end here and link around line ~31 or so.

joshgoebel (approved):

CHANGES.md
mortie: Done.

joshgoebel (approved):

joshgoebel:
@mortie  Thanks so much!

Issue #2758, opened 2020-10-16
(swift) does not recognize core Apple foundational libraries


AustinConlon:
I found this while writing [this](https://stackoverflow.com/questions/64380964/how-do-you-add-a-cagradientlayer-to-a-uibackgroundconfiguration-in-the-new-colle) on Stack Overflow. I'm not sure what the root of the problem is, but I'm guessing it starts in the function signature. Here's the code:
```
func configuredGridCell() -> UICollectionView.CellRegistration<UICollectionViewCell, Emoji> {
    return UICollectionView.CellRegistration<UICollectionViewCell, Emoji> { (cell, indexPath, emoji) in
        var content = UIListContentConfiguration.cell()
        content.text = emoji.text
        content.textProperties.font = .boldSystemFont(ofSize: 38)
        content.textProperties.alignment = .center
        content.directionalLayoutMargins = .zero
        cell.contentConfiguration = content
        var background = UIBackgroundConfiguration.listPlainCell()
        background.cornerRadius = 8
        background.strokeColor = .systemGray3
        background.strokeWidth = 1.0 / cell.traitCollection.displayScale
        cell.backgroundConfiguration = background
    }
}
```

joshgoebel:
<img width="1178" alt="Screen Shot 2020-10-15 at 9 10 28 PM" src="https://user-images.githubusercontent.com/6473/96201554-de722e00-0f2a-11eb-81ce-85079b21740c.png">

Looks good to me... you need to tell it your code is Swift... otherwise it's auto-detected as ObjectiveC and that's obviously not doing what you wanted.

joshgoebel:
Are all the same built-in classes equally valid in Swift as in ObjC?

```
  var API_CLASS = {
    className: 'built_in',
    begin: '\\b(AV|CA|CF|CG|CI|CL|CM|CN|CT|MK|MP|MTK|MTL|NS|SCN|SK|UI|WK|XC)\\w+',
  };

```

AustinConlon:
> you need to tell it your code is Swift

Since I encountered this as a Stack Overflow user, does this mean the question has to have the Swift tag?

AustinConlon:
> Are all the same built-in classes equally valid in Swift as in ObjC?
> 
> ```js
>   var API_CLASS = {
>     className: 'built_in',
[...]

Issue #2857, opened 2020-11-13
(Swift) `params` does not handled nested parens properly


svanimpe:
In the following example:

```swift
func sizeOfRectangle(from corner1: (x: Double, y: Double), to corner2: (x: Double, y: Double)) -> (width: Double, height: Double) {
  let width = abs(corner2.x - corner1.x)
  let height = abs(corner2.y - corner1.y)
  return (width, height)
}
```

Only the part `from corner1: (x: Double, y: Double)` is highlighted as a parameter list. I assume the grammar thinks the closing parenthesis of the tuple is the closing parenthesis of the parameter list?

The grammar also doesn't support line breaks in parameter lists, so only the first line is highlighted.

I'm also wondering why the parameter list is highlighted at all? It can be really complex in Swift, so it doesn't make sense to me that it's just one color.

joshgoebel:
> I assume the grammar thinks the closing parenthesis of the tuple is the closing parenthesis of the parameter list?

No, it's because `params` has `endsParent` so the nested params (any inside `()` group is considered more params [though it should really be unnamed] terminates as soon as the first `)` is hit because endParent fires for both and walks back up the mode tree.

To do this right you typically need a new mode.  If we had sugar for it it might look like this:

```js
          {  className: 'params',
            begin: /\(/, end: /\)/, endsParent: true,
            contains: [
              {'self', endsParent: false }, // this is magic sugar that doesn't exist plus being invalid JS, lol
              NUMBER,
```

So one has to have two modes OR remove the endsParent... I'm not sure what it's intended to do.  I mean I know what it's intended to do (end the `function`) but not why it's needed since the `{` will end the function anyways... unless perhaps you can have function declarations without bodies?  Many languages do this same trick but don't require `endsParent`.  First time I've seen it actually.

> The grammar also doesn't support line breaks in parameter lists, so only the first line is highlighted.

It most certainly does, I'm not sure what you're seeing.  There is no special handling of line breaks anywhere though.  The function is either terminated by `{` or by the closing `)` of any `params` group (or a few illegal characters).

> I'm also wondering why the parameter list is highlighted at all? It can be really complex in Swift, so it doesn't make sense to me that it's just one color.

[...]

PR #2920, opened 2020-12-16
enh(swift) Improved highlighting for types and generic arguments (#2920)

svanimpe:
Here's the next step for the Swift grammar rewrite:

- New grammar for types to better highlight type identifiers, including those used as generic arguments, as well as any keywords used within generic arguments.
- Includes rules to prevent punctuation used within types (`<`, `>`, `?`, `!`, `...`, and `&`) from being highlighted as operators.
- Fixes https://github.com/highlightjs/highlight.js/issues/2758

With these improvements, I should be able to write a new grammar for function declarations, to fix the remaining issues https://github.com/highlightjs/highlight.js/issues/2895 and https://github.com/highlightjs/highlight.js/issues/2857

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ Language improvements:
   - New grammar for attributes
   - Added support for quoted identifiers, implicit parameters, and property wrapper projections
   - Support for more complex expressions in string interpolation
+- enh(swift) New grammar for types (#2920) [Steven Van Impe][]
 - fix(http) avoid recursive sublanguage and tighten rules (#2893) [Josh Goebel][]
 - fix(asciidoc): Handle section titles level 5 (#2868) [Vaibhav Chanana][]
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,7 +19,7 @@ Language improvements:
   - New grammar for attributes
   - Added support for quoted identifiers, implicit parameters, and property wrapper projections
   - Support for more complex expressions in string interpolation
-- enh(swift) New grammar for types (#2920) [Steven Van Impe][]
+- enh(swift) Improved highlighting for types and generic arguments (#2920) [Steven Van Impe][]
 - fix(http) avoid recursive sublanguage and tighten rules (#2893) [Josh Goebel][]
 - fix(asciidoc): Handle section titles level 5 (#2868) [Vaibhav Chanana][]
 
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -270,8 +270,7 @@ export default function(hljs) {
       }
     ]
   };
-  // Generic arguments
-  TYPE.contains.push({
+  const GENERIC_ARGUMENTS = {
     begin: /</,
     end: />/,
     keywords: KEYWORDS,
@@ -281,7 +280,8 @@ export default function(hljs) {
       OPERATOR_GUARD,
       TYPE
     ]
-  });
+  };
+  TYPE.contains.push(GENERIC_ARGUMENTS);
 
   // Add supported submodes to string interpolation.
   for (const variant of STRING.variants) {
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -16,6 +16,15 @@ import {
 
 /** @type LanguageFn */
 export default function(hljs) {
+  // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID411
+  const BLOCK_COMMENT = hljs.COMMENT(
+    '/\\*',
+    '\\*/',
+    {
+      contains: [ 'self' ]
+    }
+  );
+
   // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID413
   // https://docs.swift.org/swift-book/ReferenceManual/zzSummaryOfTheGrammar.html
   const DOT_KEYWORD = {
@@ -77,6 +86,11 @@ export default function(hljs) {
   ];
 
   // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID418
+  const OPERATOR_GUARD = {
+    // Prevent -> from being highlighting as an operator.
+    begin: /->/,
+    relevance: 0
+  };
   const OPERATOR = {
     className: 'operator',
     relevance: 0,
@@ -92,6 +106,10 @@ export default function(hljs) {
       }
     ]
   };
+  const OPERATORS = [
+    OPERATOR_GUARD,
+    OPERATOR
+  ];
 
   // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#grammar_numeric-literal
   // TODO: Update for leading `-` after lookbehind is supported everywhere
@@ -205,7 +223,7 @@ export default function(hljs) {
         .join(' ')
     },
     contains: [
-      OPERATOR,
+      ...OPERATORS,
       NUMBER,
       STRING
     ]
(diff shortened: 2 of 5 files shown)


src/languages/swift.js
```diff
@@ -16,6 +16,15 @@ import {
...
 export default function(hljs) {
+  // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID411
+  const BLOCK_COMMENT = hljs.COMMENT(
+    '/\\*',
+    '\\*/',
+    {
+      contains: [ 'self' ]
+    }
+  );
+
```
svanimpe, line 27: This code was just moved around, for consistency.

src/languages/swift.js
```diff
@@ -224,24 +242,46 @@ export default function(hljs) {
...
-  const OPTIONAL_USING_TYPE = {
-    className: 'type',
-    begin: '\\b[A-Z][\\w\u00C0-\u02B8\']*[!?]',
-    relevance: 0
+    begin: lookahead(/\b[A-Z]/),
+    relevance: 0,
+    contains: [
+      { // Common Apple frameworks, for relevance boost
+        className: 'type',
+        begin: concat(/(AV|CA|CF|CG|CI|CL|CM|CN|CT|MK|MP|MTK|MTL|NS|SCN|SK|UI|WK|XC)/, Swift.identifierCharacter, '+')
```
joshgoebel, line 252: Nice!

joshgoebel: Add `\b` to make sure this matches only the start?

src/languages/swift.js
```diff
@@ -224,24 +242,46 @@ export default function(hljs) {
...
+      },
+      { // Optional type
+        begin: /[?!]+/,
+        relevance: 0
+      },
+      { // Variadic parameter
+        begin: /\.\.\./,
+        relevance: 0
+      },
+      { // Protocol composition
```
joshgoebel, line 267: Why do we need these vs just letting the mode end?  Is it all just in service of protocol composition so that this would work:

```
Booger... & Cat
```

joshgoebel: Could Cat or (& Cat) just be matched separately as it's own type?  This smells a bit like parsing to me vs highlighting.

src/languages/swift.js
```diff
@@ -224,24 +242,46 @@ export default function(hljs) {
...
+  // Generic arguments
+  TYPE.contains.push({
+    begin: /</,
+    end: />/,
+    keywords: KEYWORDS,
+    contains: [
+      ...KEYWORD_MODES,
+      ...ATTRIBUTES,
+      OPERATOR_GUARD,
+      TYPE
```
joshgoebel, line 282: 'self'?

joshgoebel: Nevermind, it's nested.

src/languages/swift.js
```diff
@@ -77,6 +86,11 @@ export default function(hljs) {
   ];
 
   // https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID418
+  const OPERATOR_GUARD = {
+    // Prevent -> from being highlighting as an operator.
```
joshgoebel, line 90: So the default operator stuff is a bit "wide". :-)

joshgoebel:
Awesome, more PRs this size would be great. :)

src/languages/swift.js
joshgoebel: Do you not want to highlight `?` and `!` as part of the type itself?

joshgoebel: Ah, this is because of operator conflicts...? I see.  That make sense now...

src/languages/swift.js
svanimpe: Would that make a difference?
Is this not already handled by the `begin: lookahead(/\b[A-Z]/)` rule?

src/languages/swift.js
svanimpe: Yes, it's just to parse certain operator characters that are used as punctuation here.

> Do you not want to highlight ? and ! as part of the type itself?

This isn't done by other highlighters. I assume that's because it's only syntactic sugar, and not really part of the type name.

src/languages/swift.js
svanimpe: Yeah, the nesting can get pretty complex.

src/languages/swift.js
svanimpe: Yes. `->` is regarded as syntax/punctuation. The Swift grammar also states that this token is reserved and cannot be declared as an operator. However, the characters themselves are valid operator characters, so we need a rule to prevent it from being highlighted as an operator.

svanimpe:
Any comments about the relevance? Should any be set to 0 (or to 1, in case I set them to 0)?

src/languages/swift.js
joshgoebel: > This isn't done by other highlighters.

src/languages/swift.js
joshgoebel: I guess in this case it may not actually matter... just originally you say it should start with \b but then your inner rule is much wider... which given different child rules could become be a problem... but given the children we actually have `\b` might actually not change anything since I think the other things you allow are all word boundaries to being with.  So I think I flagged it as more of a future thing, but that might be inconsistent with how I've flagged other things, LOL.  We can just resolve this and leave it unchanged.

joshgoebel:
> Any comments about the relevance? Should any be set to 0 (or to 1, in case I set them to 0)?

I didn't see any issues.  We want relevance for the "Apple like" libraries (for parity with Obj-C)... but nothing else is worthy of relevance... they are mostly too common "!?" or things that we don't typically give relevance for (random identifiers).

CHANGES.md
```diff
@@ -19,6 +19,7 @@ Language improvements:
   - New grammar for attributes
   - Added support for quoted identifiers, implicit parameters, and property wrapper projections
   - Support for more complex expressions in string interpolation
+- enh(swift) New grammar for types (#2920) [Steven Van Impe][]
```
joshgoebel, line 22:
```suggestion
- enh(swift) Better highlight type identifiers (#2920) [Steven Van Impe][]
```

joshgoebel: Or something, I have no idea at all what "new grammar for types" means.. it's not a new grammar, it's just some new rules, so we should try and describe the actual change that users will see.  So I looked at the notes from your PR to whip up that suggestion.

src/languages/swift.js
```diff
@@ -224,24 +242,46 @@ export default function(hljs) {
...
   };
-  const BLOCK_COMMENT = hljs.COMMENT(
-    '/\\*',
-    '\\*/',
-    {
-      contains: [ 'self' ]
-    }
-  );
+  // Generic arguments
+  TYPE.contains.push({
```
joshgoebel, line 274:
```suggestion
  TYPE.contains.push(TEMPLATE_WHATEVER_THINGY)
```

Should we perhaps give this rule a name with a constant vs just inlining it?

joshgoebel (approved):
Lets tweak the release notes slightly but otherwise this looks good... perhaps consider if we should have a constant for that one thing or not, but I'm ok either way.

src/languages/swift.js
svanimpe: Neither Xcode nor VS Code (TextMate grammar) highlight this. Types can get pretty complex in Swift (see the tests), so it makes sense to only highlight the identifiers themselves, not the punctuation.

src/languages/swift.js
svanimpe: Done

CHANGES.md
svanimpe: Done

svanimpe:
One thing I should note here is that the following will still result in false matches:

```swift
(var: Int, let: Double)
```

Here, `var` and `let` are highlighted as keywords, even though they're used as identifiers for tuple elements (which is allowed). This should rarely happen, but still...

I will look into this with my next PR, as this is very similar to function parameters, which also need improving.

joshgoebel:
**Doesn't need to happen here**, but I wonder if we should just expand `(AV|CA|CF|CG|CI|CL|CM|CN|CT|MK|MP|MTK|MTL|NS|SCN|SK|UI|WK|XC)` to be a full list of the full names of all the Apple frameworks then use that in both ObjC and Swift?  Perhaps not necessarily if we're not having issues, but it would be slightly more precise... someone else with a codebase of `AV*` objects is going to be sorely disappointed when their code is always detected as ObjC or Swift.

svanimpe:
>  I wonder if we should just expand (AV|CA|CF|CG|CI|CL|CM|CN|CT|MK|MP|MTK|MTL|NS|SCN|SK|UI|WK|XC) to be a full list of the full names of all the Apple frameworks

I don't think such a list even exists, and if it did, it would change every year. Apple just reserves all two-letter prefixes for its own use, even though it now also uses some three-letter prefixes.

Personally, I don't think the current list even belongs here, because Swift doesn't use prefixes. You only see these prefixes when interacting with Objective-C frameworks. I've only included it here to avoid misdetection :)

joshgoebel:
> because Swift doesn't use prefixes. You only see these prefixes when interacting with Objective-C frameworks. I've only included it here to avoid misdetection :)

Right, but some of these frameworks are super-common frameworks that a lot of swift code might interact with because they are platform libraries, no?

svanimpe:
Yes, Swift development on Apple platform still involves talking to a lot of Objective-C frameworks. So the prefixes are still relevant in that aspect, but they should become less relevant over the coming years, as people switch to SwiftUI.

PR #2923, opened 2020-12-17
enh(docs) Make demo more static

Hirse:
This PR updates the [`demo.html`](https://highlightjs.org/static/demo/) page by moving as much of the initialization as possible to the build step.

### Changes
- Remove dependency on jQuery
  - Move jQuery and Vue files to `tools/vendor` as they are still required for `developer.html`
- Static generation
  - Generate list of categories during build
  - Generate list of themes during build
  - Set language categories (`common`, `all`) during build
- Use css-`flex` instead of JS to set the height of the side bar lists
- Use WebWorker for highlighting where possible
- Use favicon from website
- Lazily load theme files on selection by setting `disabled` attribute on `link`

### Checklist
- [x] ~Added markup tests~ NA


demo/demo.js
```diff
@@ -1,117 +1,64 @@
...
 
-      $styles.removeClass('current');
-      $style.addClass('current');
-      selectStyle($style.text());
-    });
-    $styles.first().click();
-  }
+    if (currentStyle !== nextStyle) {
+      document.querySelector(`link[title="${nextStyle}"]`).removeAttribute("disabled");
+      document.querySelector(`link[title="${currentStyle}"]`).setAttribute("disabled", "disabled");
```
joshgoebel, line 58: Would `alternative` be better than disabled?  I don't really know but it's what I've used for this kind of thing in the past.

joshgoebel: Don't we still have to toggle `alternative` on/off?

joshgoebel:
So I think I agree with the overall goals/ambition here (more static rendering, throw out jQuery, etc)... BUT:

_Why all the Worker complexity and not just render super-simpler HTML and call our default init function to do all the heaving lifting?_ 

What do we gain with all that custom JS you've written vs almost no JS and just leaning on the libraries existing "just do it" auto support?

demo/demo.js
Hirse: I can have a look, but here I just followed what the previous script did:
https://github.com/highlightjs/highlight.js/blob/7e108baaa3be0a7ad402d60259d060fcefa44021/demo/demo.js#L86-L90

Hirse:
I added the worker mostly because I saw it on the [`/usage`](https://highlightjs.org/usage) page, thought it looked cool, and wanted to try it out.

Switched to a simple `initHighlightingOnLoad` for now.

>     * Why not just considering rendering everything server-side... why run Highlight.js on the client at all for the demo?

I had that thought too, but was wondering whether the demo should also demo the client-side usage or only the results?

demo/demo.js
Hirse: The current usage (or `alternative`/`disabled`) works fine in Firefox and Chromium. If you are on Mac, could you try it for Safari?

joshgoebel: Ah, I was looking more closely at the new code than the old... 

From MDN (learning something new):

> Alternate (rel="alternate stylesheet", title="..." must be specified): disabled by default, can be selected.

So alternates are simply disabled by default... I've always toggled their "alternate" status (making them active by default), but seems you can just as easily toggle their "disabled" status, thereby also making them active. :-)

So this seems a fine enough strategy.

demo/style.css
```diff
@@ -29,93 +29,100 @@ pre {
...
+
+aside ul,
+aside li {
   list-style: none;
   margin: 0;
   padding: 0;
 }
 
-#sidebar ul {
+aside ul {
```
joshgoebel, line 70: Nice!

joshgoebel:
> I added the worker mostly because I saw it on the /usage page, thought it looked cool, and wanted to try it out.

Trying new things is fun.  What did you think?  I think it's a bit annoying and wonder if long-term we shouldn't just make it much easier to use workers "out of the box".  No one is really raising any issues though so I haven't considered it a high priority... and seems less relevant than ever since all the recent catastrophic regex fixes... or maybe one benefit of workers is that the implementor writes all the code and the library is *entirely* isolated inside the worker process.

Any thoughts on big picture Worker stuff?

> Switched to a simple initHighlightingOnLoad for now.

👍🏼  You should really consider joining our Slack.

> wondering whether the demo should also demo the client-side usage or only the results?

*shrugs*  If that is the argument though then I think it's also an argument for using the simpler `initHighlightingOnLoad`, so I think it's good that we simplified that.  Personally I think the demo is intended to show *visually* what our library does, not be an example of usage (we have plenty of that in the README).

No need to hold up this PR - and I don't think we HAVE to server-side render by any means... but I don't think there is any strong reason that the demo highlighting portion MUST be rendered on the client.

demo/demo.js
```diff
@@ -1,117 +1,40 @@
...
-
-  function initStyles() {
-    var $styles;
-
-    $linkTitle.each(function(i, link) {
-      $styleContainer.append('<li>' + link.title + '</li>');
-    });
-
-    $styles = $styleContainer.find('li');
+hljs.debugMode();
```
joshgoebel, line 1: Why debug mode?

demo/index.html
```diff
@@ -2,56 +2,46 @@
 <html lang="en">
 <!-- This file is intended to be built using the instructions here: http://highlightjs.readthedocs.io/en/latest/building-testing.html#building -->
 <head>
-  <meta charset="utf-8">
+  <meta charset="utf-8" />
```
joshgoebel, line 5: Why? HTML5 does not need to be valid XHTML nor is there a good reason to force it to be off the top of my head.  Is there a big reason for these changes?

demo/demo.js
Hirse: Because it was before, but no preference either way.
https://github.com/highlightjs/highlight.js/blob/4d9d9e1832e3a8ce48fc116e2aee1a5225351bdc/demo/demo.js#L4

demo/index.html
Hirse: No, mostly personal preference.

Hirse:
> 👍🏼 You should really consider joining our Slack.

The button in the Readme doesn't seem to work anymore.
[slack](https://join.slack.com/t/highlightjs/shared_invite/zt-jatdlkw4-h3LdjU5rC23t7aQ6zqoxzw)

demo/demo.js
joshgoebel: Oh, hmmm... seems like it shouldn't really be needed... but if the page every blew up it would be easier to diagnose...  I guess if it was there before we can just defer to that then.

demo/index.html
joshgoebel: I like this [answer from StackOverflow](https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5):

> **In HTML5, the meaning of <foo /> depends on the type of element.**

> On HTML elements that are designated as void elements (essentially "An element that existed before HTML5 and which was forbidden to have any content"), end tags are simply forbidden. The slash at the end of the start tag is allowed, but has no meaning. It is just syntactic sugar for people (and syntax highlighters) that are addicted to XML.

Since we're using HTML5 not XHTML I think we should be clear and just write clean and simple HTML5, not XHTML.  If we really wanted to write XHTML we should change our DTD to say so, but I don't see any benefit to this.

joshgoebel:
Try again, updated the link.  It expires every 30 days. :(

demo/index.html
Hirse: Removed.

demo/index.html
```diff
@@ -4,54 +4,44 @@
 <head>
   <meta charset="utf-8">
   <title>highlight.js demo</title>
+  <link rel="shortcut icon" type="image/png" href="https://highlightjs.org/static/favicon.png">
```
joshgoebel, line 7: Long-term should we considering pulling this into our repo itself vs linking to the website? Not sure i have a strong opinion.

tools/build_browser.js
```diff
@@ -75,11 +75,39 @@ async function installDemo(languages, { minify }) {
     && lang.sample
   );
 
-  const css = await glob("styles/*.css", {cwd:"./src"})
-  const styles = css.map((el) => (
-    { "name": _.startCase(path.basename(el,".css")), "path": el }
-  ));
-  renderTemplate("./demo/index.html", "./demo/index.html", { styles, languages, minify });
+  languages.forEach((language) => {
```
joshgoebel, line 78: Can all this be pulled out into a `renderIndex()` function?  installDemo is getting a bit long.

joshgoebel (changes requested):
Looks good other than my one comment about factoring out `renderIndex()`... I'll circle back later actually test it locally before merging.

demo/index.html
Hirse: I don't think it's necessary to pull it into the repo as it's only relevant once the demo is part of the website. We could use a relative path (`../favicon.png`), but that's not really better (and doesn't show up locally).

joshgoebel: Then we'll leave it as is for now.

tools/build_browser.js
Hirse: Done.

joshgoebel:
In the future perhaps please just add new commits as you go along... GitHub makes it much easier to review the changes (ie, just the new commits)... when you rewrite the whole history the review screen forces me to start over and look at *everything single change*.

We squash almost all commit history anyways in PRs so there is no need for the history to be "clean".

joshgoebel:
Screen Shot 2020-12-18 at 1 09 13 AM

misc (undefined)

Did we break something?

Hirse:
common, misc, and all are hard-coded because they have a specific place in the order. Added another filtering to only show them if there are languages for the category:
- `node ./tools/build.js vbnet`
  image
- `node ./tools/build.js :common`
image

joshgoebel:
Yea. Clean and simple fix.

demo/style.css
```diff
@@ -29,93 +29,100 @@ pre {
...
 
 code {
+  display: block;
   font-family: Consolas, Monaco, monospace;
   font-size: 10pt;
+  overflow-x: auto;
+  padding: 0.5em;
   scrollbar-width: thin;
+  -moz-tab-size: 4;
+  tab-size: 4;
```
joshgoebel, line 39: Where are we using tabs that these needed to be added?

demo/style.css
```diff
@@ -29,93 +29,100 @@ pre {
 }
 
 code {
+  display: block;
   font-family: Consolas, Monaco, monospace;
   font-size: 10pt;
+  overflow-x: auto;
+  padding: 0.5em;
```
joshgoebel, line 36: Isn't this already coming from the themes?

demo/style.css
```diff
@@ -29,93 +29,100 @@ pre {
...
   scrollbar-width: thin;
+  -moz-tab-size: 4;
+  tab-size: 4;
 }
 
-body pre code.hljs {
-  padding-left: 1.5em;
-  /* compatible with school-book */
-  padding-top:15px;
+aside {
```
joshgoebel: I'm not sure if we're fighting over styling here or if this is accidental.  I _very specifically_ tweaked this last time to add more padding... and almost didn't notice that you removed this... I've added it back.

Hirse: The top padding is compatible with school-book, but the left padding is not:
image

Having school-book style padding on other themes feels a bit odd (especially since it's only on the top and left)
image
image

I would suggest letting the theme set the padding.

demo/style.css
Hirse: Ah, yes. I added it here while I was highlighting asynchronously with the worker so the code blocks don't jump around.

demo/style.css
Hirse: 1c, autoit, crmsh, dts, gml, hsp, makefile, roboconf, sqf, tp

For example:
https://github.com/highlightjs/highlight.js/blob/3a89dfb51006751b34501ea4f78d744b6d98e91a/test/detect/1c/default.txt#L3

demo/style.css
joshgoebel: > Having school-book style padding on other themes feels a bit odd  

I don't find it problematic.  As a general rules themes should not enforce pixel-precise padding (and no other themes do).  I may even remove school book in v11 (or remove it's BG at least).  If a modern theme were submitted that did something similar I would reject it.

> I would suggest letting the theme set the padding.

No, sorry, the default is pretty awful IMHO.  I think I'll add it to the list of things to change in v11.

demo/style.css
joshgoebel: Ok, I was not aware of this.  And how did we decide on 4 as the proper tab width?  What is the default?

demo/style.css
joshgoebel: But the CSS should have nothing to do with the worker...?

joshgoebel:
If there is some way to override all the themes padding (other than schoolbook) without custom JS (which is overkill) I'd be open to that... ie, some sort of CSS that I'm not aware of off the top of my head... but I don't think that's easily doable?

demo/style.css
joshgoebel: I wonder if those languages prefer/use tabs over spaces or if this is just a legacy oddity... This is a spaces codebase and my first urge was to go and change them all to spaces. :)

Hirse: I just picked 4 as it tends to look better than 8 and seems to usually be the default in most editors.

Hirse:
> If there is some way to override all the themes padding (other than schoolbook) without custom JS (which is overkill) I'd be open to that... ie, some sort of CSS that I'm not aware of off the top of my head... but I don't think that's easily doable?

We could set the name of the current theme as class on the `body` (or `main`) which would allow more specific selectors, but that does need a tiny bit of JS and I don't think it's really worth it.

joshgoebel (approved):

PR #2926, opened 2020-12-18
(chore) fix contributor parsing in AUTHORS.txt

Hirse:
When building the list of contributors for the `package.json`, contributors that added their link instead of their email are parsed incorrectly:
image

### Changes
- Update format of the list of authors 
- Updates parsing logic
- Format `package.json` contributor list to string format
  image

npm documentation:
> image
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#people-fields-author-contributors

### Checklist
- [x] ~Added markup tests~ NA

### DIFF
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -41,23 +41,18 @@ async function buildNodeHighlightJS() {
 }
 
 async function buildPackageJSON() {
-  const CONTRIBUTOR = /^- (.*) <(.*)>$/
+  const CONTRIBUTOR = /^- ([^<(]+( ?<.+>)?( ?\(http.+\))?)/;
 
-  let authors = await fs.readFile("AUTHORS.txt", {encoding: "utf8"})
-  let lines = authors.split(/\r?\n/)
-  let json = require("../package")
-  json.contributors = lines.reduce((acc, line) => {
-    let matches = line.match(CONTRIBUTOR)
+  const authors = await fs.readFile("AUTHORS.txt", { encoding: "utf8" });
+  const lines = authors.split(/\r?\n/);
+  const packageJson = require("../package");
 
-    if (matches) {
-      acc.push({
-        name: matches[1],
-        email: matches[2]
-      })
-    }
-    return acc;
-  }, []);
-  await fs.writeFile(`${process.env.BUILD_DIR}/package.json`, JSON.stringify(json, null, '   '));
+  packageJson.contributors = lines
+    .map(line => line.match(CONTRIBUTOR))
+    .filter((matches) => matches)
+    .map((matches) => matches[1]);
+
+  await fs.writeFile(`${process.env.BUILD_DIR}/package.json`, JSON.stringify(packageJson, null, 2));
 }
 
 async function buildLanguages(languages) {
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -27,7 +27,7 @@ Former Core Team members:
 Contributors:
 
 - Peter Leonov <gojpeg@gmail.com>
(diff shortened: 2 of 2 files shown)


tools/build_node.js
```diff
@@ -41,23 +41,18 @@ async function buildNodeHighlightJS() {
...
-    if (matches) {
-      acc.push({
-        name: matches[1],
-        email: matches[2]
-      })
-    }
-    return acc;
-  }, []);
-  await fs.writeFile(`${process.env.BUILD_DIR}/package.json`, JSON.stringify(json, null, '   '));
+  const CONTRIBUTOR = /^- ([^<(]+( ?<.+>)?( ?\(http.+\))?)/;
```
joshgoebel, line 44: Would \s+ be inadequate here at the beginning?

tools/build_node.js
```diff
@@ -41,23 +41,18 @@ async function buildNodeHighlightJS() {
...
+
+  const authors = await fs.readFile("AUTHORS.txt", { encoding: "utf8" });
+  const lines = authors.split(/\r?\n/);
+  const packageJson = require("../package");
+
+  packageJson.contributors = lines
+    .map(line => line.match(CONTRIBUTOR))
+    .filter((matches) => matches)
+    .map((matches) => matches[1]);
```
joshgoebel, line 53: Should we call `trim()`?

tools/build_node.js
```diff
@@ -41,23 +41,18 @@ async function buildNodeHighlightJS() {
...
+
+  await fs.writeFile(`${process.env.BUILD_DIR}/package.json`, JSON.stringify(packageJson, null, 2));
```
joshgoebel, line 55: Does this mean 2 spaces?  Was the change from 3 spaces to 2 here intentional?

Hirse: Yes. 
npm itself formats the `package.json` to two spaces when modifying, so was aligning with that.

tools/build_node.js
joshgoebel: N/m, but hmmmmm... will `.*` or `.*?` not get the job done equally well?

Hirse: Where exactly?
The format in the authors file is stable, so don't have to be too flexible.

tools/build_node.js
Hirse: With the current regex that should not be necessary.

tools/build_node.js
joshgoebel, line 44:
```suggestion
  const CONTRIBUTOR = /^- (.*(<.+>)?\s*(\(http.+\))?)/;
```

Simpler, no?

tools/build_node.js
joshgoebel: Ah, I see we're not actually splitting the string into any of the components, just matching the whole thing.

tools/build_node.js
joshgoebel: Maybe I'm just being pedantic...  the character group is bugging me if a simpler .* does the job.

```
/^- (.*?(<.+>)?\s*(\(http.+\))?)$/
```

Cleaner, no?

Hirse: The problem is that the authors file contains lines like this
```
- Josh Goebel <hello@joshgoebel.com> (maintainer)
```
where we don't want the part in parens.

See https://regexr.com/5ipfo for testcases

joshgoebel: Yeah... then .* can be too greedy in some cases.  I will resolve. :)

joshgoebel (approved):

PR #2930, opened 2020-12-18
(swift) Improved highlighting for functions, initializers, and subscripts

svanimpe:
This PR improves highlighting for functions, initializers, and subscripts. This includes detailed highlighting for generic parameters as well as function parameters.

To differentiate between tuple elements and function params, I've also added a mode to match tuples, whose element names should not be highlighted as `params`.

Fixes https://github.com/highlightjs/highlight.js/issues/2895
Fixes https://github.com/highlightjs/highlight.js/issues/2857

One open question: Do we still need the `illegal` attributes?


svanimpe:
As explained on https://github.com/highlightjs/highlight.js/pull/2908, I had to revert some of your keyword changes and change keywords with parentheses back to regexes, so they're included in the KEYWORD_MODES and not matched by TUPLE first.

src/languages/lib/kws_swift.js
```diff
@@ -240,7 +240,7 @@ export const identifierHead = either(
   /[\u2C00-\u2DFF\u2E80-\u2FFF]/,
   /[\u3004-\u3007\u3021-\u302F\u3031-\u303F\u3040-\uD7FF]/,
   /[\uF900-\uFD3D\uFD40-\uFDCF\uFDF0-\uFE1F\uFE30-\uFE44]/,
-  /[\uFE47-\uFFFD]/
+  /[\uFE47-\uFEFE\uFF00-\uFFFD]/ // Should be /[\uFE47-\uFFFD]/, but we have to exclude FEFF.
```
joshgoebel, line 243: I assume you are concating this with something else in one of your rules.  You're probably doing something like:

```
\uFEFF+[group that includes \uFEFF*]
```

This is the same problem as `\s*\s*` In a worse case the engine will chase it's tail trying to figure out whether which portion of \uFEFF are matched by the left side rather than the right.  The ambiguity must be removed.

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
   };
   TYPE.contains.push(GENERIC_ARGUMENTS);
 
+  // https://docs.swift.org/swift-book/ReferenceManual/Expressions.html#ID552
+  // Prevents element names from being highlighted as keywords.
+  const TUPLE_ELEMENT_NAME = {
+    begin: lookahead(concat(Swift.identifier, /\s*:/)),
+    end: lookahead(/:/),
+    keywords: "_",
```
joshgoebel, line 296:
```suggestion
    keywords: "_|0",
```

And what is end trying to accomplish here?  If you just watch to match the identifier portion why not do it with begin with a lookahead for `:` inside begin?

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+    relevance: 0
+  };
+  // Matches tuples as well as the parameter list of a function type.
+  const TUPLE = {
```
joshgoebel, line 300: This is a lot of parsing, why would not just matching `blah:` work without trying to find `()` pairs?

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+  // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID362
+  // Matches both the keyword func and the function title.
+  // Grouping these lets us differentiate between the operator function <
+  // and the start of the generic parameter clause (also <).
+  const FUNCTION_TITLE = {
+    begin: lookahead(/\bfunc\b/),
+    relevance: 0,
+    contains: [
+      {
+        className: 'keyword',
```
joshgoebel, line 329: We should use `keywords` on the parent rule  instead and you don't need a look ahead.

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+        begin: /\bfunc\b/
```
joshgoebel, line 330:
```suggestion
  const FUNCTION_TITLE = {
    begin: /\bfunc\b/,
    keywords: "func",
    relevance: 0,
    contains: [
```

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+      TYPE,
+      TUPLE
+    ],
+    illegal: /["']/
+  };
+  const FUNCTION = {
+    className: 'function',
+    begin: lookahead(/\bfunc\b/),
+    contains: [
+      FUNCTION_TITLE,
```
joshgoebel, line 393: Same notes here, this can be greatly simplified I think... just match `function` (no lookahead) use keywords to classify it... then your contains only needs to focus on what comes AFTER, not try to match the same thing repeatedly.

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+        begin: either(QUOTED_IDENTIFIER.begin, Swift.identifier, Swift.operator),
+        // Required to make sure the opening < of the generic parameter clause
+        // isn't parsed as a second title.
+        endsParent: true,
+        relevance: 0
+      }
+    ]
+  };
+  const GENERIC_PARAMETERS = {
+    begin: /\s*</,
```
joshgoebel, line 347: This \s* looks out of place and should probably be handled elsewhere.

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+        begin: /\b_\b/
+      },
+      {
+        className: 'params',
+        begin: Swift.identifier
+      }
+    ]
+  };
+  const FUNCTION_PARAMETERS = {
+    begin: /\s*\(/,
```
joshgoebel, line 373: Ditto.  This rule probably belongs inside FUNCTION contains if the goal is to allow space.

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+    begin: either(
+      lookahead(concat(Swift.identifier, /\s*:/)),
+      lookahead(concat(Swift.identifier, /\s+/, Swift.identifier, /\s*:/))
+    ),
+    end: lookahead(/:/),
+    relevance: 0,
+    contains: [
+      {
+        className: 'keyword',
+        begin: /\b_\b/
```
joshgoebel, line 364: Is `_` truly a keyword in Swift?

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+      FUNCTION_TITLE,
+      GENERIC_PARAMETERS,
+      FUNCTION_PARAMETERS
+    ],
+    illegal: /\[|%/
+  };
+
+  // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID375
+  // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID379
+  const INIT_SUBSCRIPT_TITLE = {
```
joshgoebel, line 402: Same comment here, you don't need lookahead and you can use `keywords` for the keyword matching.

joshgoebel: That should collapse the rule and avoid the need for `contains` at all.

joshgoebel: Which would probably make this simple enough to be moved directly to `INIT_SUBSCRIPT`.

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+      },
+      {
+        begin: /\s+/,
+        relevance: 0
+      }
+    ]
+  };
+  const INIT_SUBSCRIPT = {
+    className: 'function',
+    begin: lookahead(/\b(subscript|init[?!]?)\s*[<(]/),
```
joshgoebel, line 419:
```suggestion
  const INIT_SUBSCRIPT = {
    className: 'function',
    begin: /\b(subscript|init[?!]?)\s*[<(]/,
    keywords: { // ... 
```

joshgoebel: You will need $pattern here of course to deal with the `?!`...

joshgoebel: You should really look at `beginKeywords` also as that could also be used in MANY of these cases... and is even simpler than `keywords` since it uses avoids the need for $pattern for more complex cases.

src/languages/swift.js
```diff
@@ -56,7 +60,7 @@ export default function(hljs) {
   // find all the regular keywords
   const KEYWORDS = {
     $pattern: either(
-      /\b\w+(\(\w+\))?/, // kw or kw(arg)
```
joshgoebel, line 59: Ah your problem with this was your tuple rule eating the `()`, yes?  That's why we try to avoid parsing and do simpler rules... if your rule only matched `blah:` and avoided trying to figure out for *certain* that something looked like a tuple you'd avoid this problem completely.  Since you're only using it as a guard and not highlighting it's not the end of the world if it matches a bit wide (as long as it's not stealing things we need to highlight).

Are there other syntactic things in Swift that use `ident:`?

joshgoebel (changes requested):
Feels a bit like you're still asking yourself "how do i parse the language" vs "what's the simplest rule that could solve this problem"... that tends to result in very large and complex rulesets... so a lot of my thoughts are on that and simplifying.

src/languages/swift.js
joshgoebel: I'm also guessing it's technically wrong to assume all `()` combos would be tuples... so approach is 100% "correct" in the Swift language sense.

src/languages/lib/kws_swift.js
svanimpe: The new ranges should be fine as they exclude FEFF, no?

src/languages/swift.js
svanimpe: I'm trying to detect the element names in a tuple and consume them so they don't get highlighted as other things (keywords, function params, etc).

src/languages/swift.js
svanimpe: Because there is a lot of `blah:` in Swift.

Tuple elements, function parameters, dictionary key/value pairs, type/supertype relationship, enum/raw type relationship, switch case labels, labeled control flow statements, ...

The current rule is already pretty naive in assuming that every `blah:` between parenthesis is a tuple element name, so I should add more tests to find cases where it breaks down. It should work relatively well, though, as most of the cases I've listed should not appear in parentheses.

src/languages/swift.js
svanimpe: This is probably me incorrectly thinking `begin` consumes the token. It doesn't then?
Will `begin/keyword` or `beginKeyword` work when there's a constains rule (`title`) in this case that also matches?

Again, I probably did this to make sure `func` is matched as a keyword, not as an identifier (`title`).

src/languages/swift.js
svanimpe: I'll look into this.

src/languages/swift.js
svanimpe: I'll look into this. Just want to make sure `func` isn't matched as the `title`.

src/languages/swift.js
svanimpe: This is because there's no `end` rule anymore, so the spaces have to be matched in order for the `FUNCTION` mode to continue. That's why I added optional leading spaces in a few places.

src/languages/swift.js
svanimpe: Yes, but only in a few places. I thought it was neater to allow them only in specific places, and not everywhere (which would be the result of a contained rule for spaces, as would trailing spaces.

src/languages/swift.js
svanimpe: Yes, it's used as a wildcard pattern.
Which reminds me that I should also add support for other patterns in TUPLE, namely `(let name, var otherName)` etc.

src/languages/swift.js
svanimpe: Does `beginKeyword` support matching keywords like `init!`? Or is it tied to the default $pattern?

src/languages/swift.js
svanimpe: See above, there's just a looooot of things that use `identifier:` in Swift. Within parentheses, it's almost always a tuple element/function parameter name. I'll look into it more and see what cases I missed that may cause issues.

Of course, parentheses can also be just parentheses in expressions, but because tuple elements are also expressions, that should match nicely.

svanimpe:
I totally understand, but in this case I don't think we can simplify it much. My first attempt didn't include a rule for tuples, and as a result, every `blah:` in a function declaration would match as `param`, even if it was part of a tuple or function type, and not a parameter name. I added the tuple rule to differentiate between `params` in the function declaration, and every other use of tuples/function types.

src/languages/lib/kws_swift.js
joshgoebel: Yes, that's one way to fix it if it's not that big a deal.  I was just explaining it to perhaps improve your understanding since you asked about it in the main thread.

src/languages/swift.js
joshgoebel: > See above, there's just a looooot of things that use identifier: in Swift. 

_But if we're not highlighting those things, it's irrelevant how many variant there are..._ and if we are it still should mostly be irrelevant since we'd likely be highlighting within a child scope and the parent scope (which had the simpler `blah:` rule) would no longer be in effect.  And if we were highlighting them ALL in the parent scope then this problem goes away entirely. 

But _right now_ I don't see any need for the tuple rule.  And simplifying it also simplifies keywords.

Is all this just to prevent keywords being highlighted in tuples?  Wouldn't a negative lookahead for `:` on keywords `$pattern` solve this?

src/languages/swift.js
joshgoebel: Sample code?  And see my note above about just using a negative look-ahead for `:`... that solves keywords I think.

src/languages/swift.js
joshgoebel: Please add the test cases you think are needed and then if the tests fails with a simpler `blah:` rule you can throw it back to me to take a closer look at the failing test case.  I'm remain convinced this can be much simplified.

src/languages/swift.js
joshgoebel: `begin` consumes whatever it matches unless you use `excludeBegin`.  But that's not a problem.   You really shoudl try:

```
beginKeywords: `func`
```

This does the match and handles the fact that it's a keyword with single line of code.

src/languages/swift.js
joshgoebel: See my note above on `beginKeywords` which is sugar for this pattern.

src/languages/swift.js
joshgoebel: Yes, I understand why.  They should be inside the FUNCTION mode though not appended to the child modes.

src/languages/swift.js
joshgoebel: Repeating yourself vs doing it in a single place is usually the wrong approach.  Instead of every child including `\s*` it should be in a single place.  And it's not "everywhere". There are 3 rules in FUNCTION and 2 of them include this... and it won't harm the 3rd in any case.

src/languages/swift.js
joshgoebel: Again, `begin` consumes... so contains can never match something in begin unless you use `lookahead` or returnBegin...

src/languages/swift.js
```diff
@@ -283,6 +288,143 @@ export default function(hljs) {
...
+    illegal: /["']/
+  };
+  const FUNCTION = {
+    className: 'function',
+    begin: lookahead(/\bfunc\b/),
+    contains: [
+      FUNCTION_TITLE,
+      GENERIC_PARAMETERS,
+      FUNCTION_PARAMETERS
+    ],
```
joshgoebel, line 396:
```suggestion
    contains: [
      FUNCTION_TITLE,
      { begin: /\s*/ }, // allow whitespace
      GENERIC_PARAMETERS,
      FUNCTION_PARAMETERS
    ],
```

Or you could make a constant of course, but single place.  Done. Much simpler.

src/languages/swift.js
joshgoebel: It's just sugar around `either`:

```
      if (mode.beginKeywords) {
        mode.begin = '\\b(' + mode.beginKeywords.split(' ').join('|') + ')(?!\\.)(?=\\b|\\s)';
```

joshgoebel: Oh wait, yeah sorry you might still need to adjust `$pattern`... beginKeywords uses an either for the match but then uses the keywords engine for the highlighting.

joshgoebel: This:

```
    begin: lookahead(/\b(subscript|init[?!]?)\s*[<(]/),
    contains: [
      {
        className: 'keyword',
        begin: /subscript|init[?!]?/
      },
```

Becomes (off the top of my head):

```js
    begin: /\b(subscript|init[?!]?)\s*[<(]/,
    keywords: { 
      keyword: "subscript init init? init!",
      $pattern: /\w+[?!]?/
```

And now there is no potential sequencing randomness doe to `contains` since everything is grouped in a single cluster.

src/languages/swift.js
joshgoebel: Why is an explicit rule needed for that?  `let` and `var` are already keywords, no?

src/languages/lib/kws_swift.js
joshgoebel: Fixing poly cases properly can definitely be super annoying. :)  Evidently it's a much larger problem in Prism (they have 100s of cases)... I was able to fix ALL our backtracking issues (~30 or so I think) in a day.

src/languages/swift.js
joshgoebel: There are also ways to force end to hold open but then you'd need a child rule with `endsParent`... either of those those options would be preferable to appending `\s*` multiple places.  I think it's very explicit (and hence easier to understand) listing it right inside the contains though.

If I did it implicitly (holding end open) I'd want to add a comment explaining it was for whitespace anyways, so seems simpler to just have the rule right there.

joshgoebel:
>  I added the tuple rule to differentiate between params in the function declaration, and every other use of tuples/function types.

I might be missing something here... isn't that what FUNCTION_PARAMETERS is for?  You already have a scope for treating params differently if need be, no?

joshgoebel:
Lets try to simplify the other things first and then we can circle back to tuple at the very end. If we insist on p being treated differently than the tuple stuff then we may indeed need a separate rule.

joshgoebel:
I did have an idea the other day about throwing errors if keywords don't match `$pattern` (to help grammar creators) but you could extend the idea to simply auto-handle common situations... like:

- default to \w+
- add any OTHER characters found in keywords to the group

So given: 

```
keywords: "booger? what!"
```

You'd end up with `$pattern` as `/[\w?!]+/` auto-generated without needing any other effort.  And then you'd only need $pattern if you needed to REDUCE the scope (if \w was too wide).

Would that help simplify some of your issues with keywords?

joshgoebel:
I just realized a lot of the complexity you are adding is your not embracing the keywords system or not fully understanding how it works.  When you force keywords ALWAYS to always be a mode then they must consume content (vs being found inside the current buffer). This leads to uncommon/non-typical patterns like:

```js
{
  begin: lookahead("func"),
  contains: [ FUNC_KEYWORD, ...
```

Where you force begin not to consume because you need FUNC_KEYWORDS to consume... If begin consumed then FUNC_KEYWORD would have nothing to match, etc... `keywords` and `beginKeywords` avoids all this complexity because it highlights _within the match_.  There are times that works against you, but many times it can work for you.

src/languages/swift.js
svanimpe: Fixed the relevance in the next commit.

src/languages/swift.js
svanimpe: Changed `FUNCTION_TITLE` to use `beginKeywords`. However, in `FUNCTION` itself, I'm still using the lookahead as I can't consume the keyword there because I need it for `FUNCTION_TITLE`.

src/languages/swift.js
svanimpe: I've added a few more tests for `TUPLE`. So far, I didn't detect anything that required changes to the `TUPLE` rule.

src/languages/swift.js
svanimpe: Changed it to use `beginKeywords`.

src/languages/swift.js
svanimpe: See the note above. I can't consume `func` here because it need it for `FUNCTION_TITLE`.

The `FUNCTION_TITLE` rule requires a "lookbehind" to check that it follows the `func` keyword. Without that check, it's not possible to differentiate an operator function `<` from the start of the generic parameter clause (also `<`).

src/languages/swift.js
svanimpe: I think those leading spaces are just a leftover:

- The original version of this grammar combined a recursive `params` rule with `endsParent`, which caused https://github.com/highlightjs/highlight.js/issues/2857.
- My first fix was to keep the recursion, but remove `endsParent`.
- However, I then no longer had an end condition, so I had to make sure the rules kept matching up until the end of the parameter list.
- This meant I had to allowed leading spaces in certain places, but not just whitespace anywhere. Otherwise, a function could have two parameter lists for example.
- However, thanks to the `TUPLE` rule, I no longer need `contains: 'self'` for function parameters, so I can use `endsParent` again and remove the spaces.

So be warned, removing the `TUPLE` rule will break all of this again, as you will have to go back to recursive `FUNCTION_PARAMETERS`, which means you can't use `endsParent` anymore. You also can't use the `end: '{'` rule from the original version, as that was one of the causes for https://github.com/highlightjs/highlight.js/issues/2895

So long story short, the main motivation for `TUPLE` is to allow nested **parentheses** in `FUNCTION_PARAMETERS`, without nesting `FUNCTION_PARAMETERS` itself.

src/languages/swift.js
svanimpe: Fixed (see above).

src/languages/swift.js
svanimpe: Yes, correct. I didn't need a rule for it because I was only matching something with a trailing colon.

My brain was confusing this with the rule for function parameters, which allows two identifiers.

src/languages/swift.js
svanimpe: Fixed!

src/languages/swift.js
svanimpe: Fixed. See explanation above.

svanimpe:
I pushed a new commit with some simplifications regarding keywords, and the spaces in the function declaration.

> `subscript<X: A>(_ p: @attribute inout (x: Int, var: Int) = (0, 0))  { }`
> Ok I think I see now, you're trying to differentiate between:
> function params names (p)
> tuples named items (x, var)
> So first... what are the x and var here exactly (is this whole thing a type spec)?

`x` and `var` are tuple element names here, and `(x: Int, var: Int)` is a tuple type.

> Let's try to simplify the other things first and then we can circle back to tuple at the very end. If we insist on function params being treated differently than the tuple stuff then we may indeed need a separate rule. Though then I think I might at least consider something like (roughly):
> 
> ```js
>   const TUPLE = {
>     // make sure it's really a tuple
>     begin: /\(/ ... lookahead(/\s*/, concat(Swift.identifier, /\s*:/)),
>     end: /\)/,
> ```
> 
> This prevents false positives on `blah(blah)` style keywords and also prevents TUPLE from firing on non-tuple like expressions.

This would make sense in the global scope, but would cause problems in function declarations. There, I need TUPLE to match wide, so it consumes all nested parentheses: both regular tuples, and the parameter list of a function type, both of which aren't required to use identifiers (but may do so).

> You'd end up with $pattern as /[\w?!]+/ auto-generated without needing any other effort. And then you'd only need to manually adjust $pattern if you needed to REDUCE the scope (if \w was too wide) or target the pattern more precisely.

This could cause many other problems. A keyword is a very specific list of characters in a specific order. Only allowing additional characters would match too wide. For example, `open` is currently a contextual keyword, but we can't match `open!` or `!open` as keywords, because those are valid expressions using operators. So if `init!` is a keyword, we can only match `init!`, and nothing else that happens to contain an exclamation point.

> It goes without saying that I think there is much value in specifying keywords in a single place/constant and having them "just work" rather than writing a ton of individual keyword rules. This simplicity is amazing and elegant in the many simpler languages where it already "just works".

The problem here is that Swift is no longer a "simple" language like those from earlier generations. There is no such thing as a list of old-school globally-reserverd keywords in Swift. A lot of keywords are contextual, but can be valid identifiers in other contexts. For example, the keyword `open` is contextual because it only applies to class declarations. So in `open class X { }` it's a keyword, but in `var open = true` it's an identifier.

The only way to highlight these properly is do as much parsing as possible. That's how you end up with a grammar like TextMate 😕 Right now, I'm just trying to get HLJS to a point where it's usable for me, but long-term, I may have to move to a semantic highlighter that isn't based on regular expressions.

src/languages/swift.js
joshgoebel: I finally see what you are doing... you're using "func" to serve as an alternative "look-behind" for TITLE  otherwise you don't have a great way to avoid it double-matching non-titles.  This isn't a problem for many other languages because there are simply no other title like things.  Swift makes everything complicated when `<` can be a title or a template.

I'll give this some more thought.

src/languages/swift.js
joshgoebel: > Otherwise, a function could have two parameter lists for example.

If this never happens in real Swift code (why would a function have two param lists) then it's not a concern we need to worry about.  Perhaps I misunderstand what you're saying though. 

> So long story short, the main motivation for TUPLE is to allow nested parentheses in FUNCTION_PARAMETERS, without nesting FUNCTION_PARAMETERS itself.

Ah... or else `)` will potentially end everything prematurely... often a placeholder recursive rule is used for this sort of them, but then if it needed all the same children might as well just use `FUNCTION_PARAMETERS` as you're thinking.  And TUPLE might make more sense than that since it provides a bit of addl value.

src/languages/swift.js
joshgoebel: This is how simple function would be if not for the `<` craziness. :)

joshgoebel:
> The only way to highlight these properly is do as much parsing as possible. That's how you end up with a grammar like TextMate

Yes, that is my concern.  Why we don't have any strict "size requirements" per se Swift has rapidly gone from just 5kb to 20kb... and it's included in our common bundle, so that makes the size slightly relevant. It's surpassed JS/TS which is one of our other more complex/larger grammars that isn't just pure keywords - which is only 14kb (if you count the ECMAScript keywords also).

I'm starting to wonder what the upper limit on fidelity is... because I have a feeling you could easily end up with a very large grammar if you just kept cramming things in... lets take just one example:

```
  const AVAILABLE_ATTRIBUTE = {
    begin: /(@|#)available\(/,
    end: /\)/,
    keywords: {
      $pattern: /[@#]?\w+/,
      keyword: Swift.availabilityKeywords
        .concat([
          "@available",
          "#available"
        ])
        .join(' ')
    },
    contains: [
      ...OPERATORS,
      NUMBER,
      STRING
    ]
  };
```

I think all this does is provide scope for the "unique" availabilityKeywords (which arguably could be moved to the global scope, though perhaps it's less" correct")... this could otherwise be simplified to two lines of code:

```
          "@available",
          "#available"
```

Just added to the appropriate list of keywords already matched by existing rules.

@allejo @egor-rogov Any thoughts on where to draw the limits on raw size, or should their simply be no limits?  Does it matter at all that Swift is in the "common" set... I know we have some other huge grammars (mostly keywords) but we don't ship them in the common set.  

Any thoughts on complexity vs "perfection"? Is it worth an extra 18 lines of code and additional rules just to avoid highlighting a few keywords that might be used as identifiers?

@svanimpe That's what we're talking about in this case, right?  All of this is to avoid mis-highlighting those things as keywords in other contexts, yes?

joshgoebel:
Another example, operators:

```
// Valid first characters for operators.
export const operatorHead = either(
  /[/=\-+!*%<>&|^~?]/,
  /[\u00A1-\u00A7]/,
  /[\u00A9\u00AB]/,
  /[\u00AC\u00AE]/,
  /[\u00B0\u00B1]/,
  /[\u00B6\u00BB\u00BF\u00D7\u00F7]/,
  /[\u2016-\u2017]/,
  /[\u2020-\u2027]/,
  /[\u2030-\u203E]/,
  /[\u2041-\u2053]/,
  /[\u2055-\u205E]/,
  /[\u2190-\u23FF]/,
  /[\u2500-\u2775]/,
  /[\u2794-\u2BFF]/,
  /[\u2E00-\u2E7F]/,
  /[\u3001-\u3003]/,
  /[\u3008-\u3020]/,
  /[\u3030]/
);

// Valid characters for operators.
export const operatorCharacter = either(
  operatorHead,
  /[\u0300-\u036F]/,
  /[\u1DC0-\u1DFF]/,
  /[\u20D0-\u20FF]/,
  /[\uFE00-\uFE0F]/,
  /[\uFE20-\uFE2F]/
  // TODO: The following characters are also allowed, but the regex isn't supported yet.
  // /[\u{E0100}-\u{E01EF}]/u
);
```

This is just the keyword lists, not counting the rules in the grammar (another 25 lines)... if I had been writing the grammar I probably would have supported only the common operators:

```
  /[/=\-+!*%<>&|^~?]/,
```

**And said "we simply don't support every possibly operator under the sun"** - or resisted the topic if it came up later.  I'm not suggesting we undo this work now that it's done, but I think we should have some big picture idea for the next time this comes up... I feel like in the past the goal has been 'minimum ruleset to provide good highlighting' experience and that this is perhaps pushed far past that.

I feel like here we're trying to build a grammar to cover the WHOLE spec (again speaking strictly of operators) rather than just "what might be most helpful in a highlighting context"... and those are VERY different goals which lead to very different outcomes.

src/languages/swift.js
```diff
@@ -291,8 +295,7 @@ export default function(hljs) {
   // https://docs.swift.org/swift-book/ReferenceManual/Expressions.html#ID552
   // Prevents element names from being highlighted as keywords.
   const TUPLE_ELEMENT_NAME = {
-    begin: lookahead(concat(Swift.identifier, /\s*:/)),
-    end: lookahead(/:/),
+    begin: concat(Swift.identifier, /\s*:/),
```
joshgoebel, line 298: I didn't understand why we needed lookahead at all here.

src/languages/swift.js
```diff
@@ -351,7 +351,7 @@ export default function(hljs) {
       lookahead(concat(Swift.identifier, /\s*:/)),
       lookahead(concat(Swift.identifier, /\s+/, Swift.identifier, /\s*:/))
     ),
-    end: lookahead(/:/),
+    end: /:/,
```
joshgoebel, line 354: No good reason not to consume this that I can see, so we just do that, since it's simpler.

src/languages/swift.js
```diff
@@ -215,22 +219,22 @@ export default function(hljs) {
...
+          end: /\)/,
+          keywords: Swift.availabilityKeywords.join(' '),
+          contains: [
+            ...OPERATORS,
+            NUMBER,
+            STRING
+          ]
+        }
+      ]
+    }
```
joshgoebel, line 237: I think this is slightly simpler and easier to follow.

joshgoebel:
@svanimpe I think we're closing in on this.  I simplified a few things.  I see why we need TUPLE since it's serving dual-purposes (and also if we really want tuples highlighted differently than param names)... and I now understand why you're writing the function rules as you are.  I added a few comments on my commit.

Any questions about what I changed?

joshgoebel:
And yes if illegal isn't hurting anything we should keep - it's a HUGE help for auto-detect because of how fast it can rule languages out.

src/languages/swift.js
```diff
@@ -275,6 +283,7 @@ export default function(hljs) {
     end: />/,
     keywords: KEYWORDS,
     contains: [
+      ...COMMENTS,
       ...KEYWORD_MODES,
       ...ATTRIBUTES,
       OPERATOR_GUARD,
```
joshgoebel, line 289: Can you think of any good reason the compiler shouldn't just flatten `contains` for you avoiding keeping track of which items and rules vs arrays of rules?

src/languages/swift.js
svanimpe: Yeah, nesting `FUNCTION_PARAMETERS` within itself will also highlights identifiers in tuples and function types as `params`, which is what I wanted to avoid, as those aren't highlighted elsewhere, and aren't `params` anyway.

I also wanted to narrow down the scope of `params` to just the identifiers, and not the entire parameter list, as that causes overwhelmingly large highlights, and also highlights all the punctuation in the parameter list (and everything else we don't normally highlight).

src/languages/swift.js
svanimpe: Fixed

svanimpe:
> @svanimpe That's what we're talking about in this case, right? All of this is to avoid mis-highlighting those things as keywords in other contexts, yes?

In the case of `@/#available`, my goal was to properly highlight the arguments for that keyword. Those arguments are very specific and not like regular code. They also contain some specific keywords. Those keywords should never be added as global keywords, as they aren't used anywhere else, and will only cause confusion if we mis-highlight them.

Additionally, `@available` is the only attribute that has an argument list, so by handling it separately, we don't have to account for argument lists in the general `ATTRIBUTE` rules.

> And then said "we simply don't support every possibly operator under the sun, too complex" 

If code size is becoming an issue, then reducing the list of identifier/operator characters would make sense. If we reduce them to only the ones that are easy to type, we can probably remove all the Unicode characters and shrink the code size.

Note that we still have to support custom operators using these characters, so we don't break functions (with operator titles) again.

Also, if we narrow down the list of characters we support, we may be excluding users who don't write their code in English, and for whom the characters we remove may be normal everyday characters ...

> Is the goal to cover 90% of cases reasonably well or to cover 100%?

The goal should be correctness without completeness. I'm using HLJS to highlight teaching examples, and I cannot afford to have a single mis-highlight in my course. Those are really confusing to new learners.

I'm only doing this work because it's easy for me to integrate HLJS into my publishing setup, and fixing it seemed easier than integrating a different solution (which will likely be the long-term plan).

src/languages/swift.js
svanimpe: This is probably a leftover from when I used `contains`. The goal here was to check that the upcoming code is indeed an identifier, but leave matching to the contained rules.

Will the `keywords: "_|0"` rule still work with this change?

src/languages/swift.js
svanimpe: Yes, that should be fine. I didn't consume it just it case I need it to detect a type annotation (which always starts with `:`), but I managed to get proper highlighting without having to do that.

src/languages/swift.js
svanimpe: But this highlights the entire thing (including identifiers and punctuation) as a keyword I think? This is what I want to avoid. Generally only the attribute title itself is highlighted as the keyword. For example:

Screen Shot 2020-12-26 at 11 07 55

As we don't support highlighting identifiers and argument labels, I also didn't support them in `AVAILABLE_ATTRIBUTE`. However, setting a general `className: "keyword"` will highlight the entire thing as a keyword I think?

src/languages/swift.js
svanimpe: I'm not sure what you mean here. Am I not already flattening the nested arrays with the leading `...` operator?

src/languages/swift.js
joshgoebel: Yeah, when i tried to fix it I quickly hit what you were doing with it... this is just not a common pattern so I didn't see it at a glance. :)

src/languages/swift.js
joshgoebel: _Yes, keywords match the buffer, they do not consume the input._

src/languages/swift.js
joshgoebel: >  I didn't consume it just it case I need it to detect a type annotation

We really need look-behind. :(  Lots of complexity dealing with these kind of things without it.

src/languages/swift.js
joshgoebel: Yes, but it's error prone and you have to constantly remember which is a mode and which is a list of modes... vs just not requiring the `...` and letting the compiler do the aggregation.

src/languages/swift.js
joshgoebel: No.  The first mode terminates immediately after `begin` since there is no end... and then `starts` immediately starts a new mode... All tests are green.  Feel free add a failing test if I'm missing something but pretty sure I go tthis one right.

joshgoebel:
> If code size is becoming an issue, then reducing the list of identifier/operator characters would make sense. If we reduce them to only the ones that are easy to type, we can probably remove all the Unicode characters and shrink the code size.

Well, as I said the problem is we have no hard and fast rules, we're just playing it by ear. :-) And complexity also plays a role, though your code tends to be very clean and clear, which does lower the fear of the complexity. :) 

> Note that we still have to support custom operators using these characters, so we don't break functions (with operator titles) again.

Well there is always a side chance we could make that work with `\S` (not space)... or some such rather than an explicit rule.

> Also, if we narrow down the list of characters we support, we may be excluding users who don't write their code in English, and for whom the characters we remove may be normal everyday characters ...

I think that 98% of the library already skews English, all our docs are English (though there is talk of changing that, but it requires contributors who know those languages), etc... so that seems a somewhat natural limitation for the time being (even if suboptimal).  Proper multi-lingual support (and related docs) is a pretty big undertaking... Plus I'm not really even sure we can honestly have great multi-lingual support until someone digs into https://github.com/highlightjs/highlight.js/issues/2756

> The goal should be correctness without completeness. I'm using HLJS to highlight teaching examples, and I cannot afford to have a single mis-highlight in my course. Those are really confusing to new learners.

Our goals may be slightly misaligned.  While I generally agree I also don't think 100% correctness is possible (see our regex support in like every language). Sometimes it's impossible (because we're not a full parser).  Other times it's simply too complex/hard to achieve (lack of look-behind, engine limitations, etc).  So we try to highlight 98% of reasonable code well - even if 2% of edge cases might be broken and unfixable.

If a strange edge case is found of course we'd consider fixing it - but if it proved to be hard or impossible to fix that does not necessary mean we would cripple the highlighter by refusing to highlight a thing that works fine for 99% of people in 99% of circumstances.

> I'm only doing this work because it's easy for me to integrate HLJS into my publishing setup, and fixing it seemed easier than integrating a different solution (which will likely be the long-term plan).

At some point you may want to consider forking the internal Swift grammar into your own 3rd party grammar.  Full control without any review - and you're limited only by what the engine makes possible - not what we think is a good idea or good for the core library or not. :)

src/languages/swift.js
joshgoebel: Obviously `...` is more explicit, but is there value in that explicitness or does it just get in the way?

src/languages/swift.js
joshgoebel: I think this question is the only blocker?  So let me know if you find a testable issue here. :)  This missed the 10.5 window but will still be great to get it merged. :)

svanimpe: Ah, never mind. I just didn't see the `starts` in there.

Do we still need the `contains` in the middle? There's only one rule after all, and there can only be one occurrence of the argument list.

src/languages/swift.js
svanimpe: What would the alternative code be?

This may be just a case of me not being a JS developer.

svanimpe:
> So we try to get "close enough" - to highlight 98% of reasonable code well - even if 2% of edge cases might be broken and unfixable.

The text I'm currently highlighting is a programming fundamentals course, so definitely no edge cases in there, only basics :D

> At some point you may want to consider forking the internal Swift grammar into your own 3rd party grammar.

Long term, I may have to move to pre-rendering, as that isn't limited by speed or code size. The Swift compiler actually includes a library that does most of the hard word: https://github.com/apple/swift-syntax
That should be able to categorize all the tokens in the source code, without having to worry about regular expressions anymore. But I still have to look into exactly how much information it provides, how well it can handle incomplete code snippets (that don't compile on their own), how I can hook it into my setup, etc... For now, HLJS is still easier to use.

src/languages/swift.js
joshgoebel:
```js
    contains: [
      ...COMMENTS,
      ...KEYWORD_MODES,
      ...ATTRIBUTES,
      OPERATOR_GUARD,
```

vs

The compiler figure it out for you.

```
    contains: [
      COMMENTS,
      KEYWORD_MODES,
      ATTRIBUTES,
      OPERATOR_GUARD,
```

The question is does forcing the reader/writer to keep track of what needs ... and what doesn't add any value... I'm not sure it does.

joshgoebel:
> The text I'm currently highlighting is a programming fundamentals course, so definitely no edge cases in there, only basics :D

That doesn't mean there are no edge cases (from a highlighting perspective) though it prolly helps. :)  Often if one can rewrite the source easily they can avoid many of these kind of pitfalls though.

> The Swift compiler actually includes a library that does most of the hard work: https://github.com/apple/swift-syntax

I was wondering if a Swift solution for parsing Swift might not be a better fit for perfect parsing.

> For now, HLJS is still easier to use.

That is one advantage for sure. :)  

---

I'm also (slightly) interested in 3rd party parsing engines... (allowing someone to replace our parsing engine entirely) so if you wanted to write a full JS Swift parser that spit out tokens I'd be open to a discussion about what kind of plug-in architecture we'd need to support something like that.

src/languages/swift.js
joshgoebel: Technically PERHAPS you could get away with just `starts: { end: /\)/, ... }` but I'm not sure it would be as obvious what is happening at a glance so that doesn't predispose me to simplify further. :)

src/languages/swift.js
joshgoebel: I'm not interested in an array here ever having any special type of semantic meaning... so that means technically it would be easy to move this type of thing into the compiler and make the type passed (array or item) irrelevant.

sa-orlov (approved):

sa-orlov (approved):

joshgoebel:
@svanimpe I will take your merging as a tacit agreement that all is well and this is good to go. :)

joshgoebel (approved):

src/languages/swift.js
svanimpe: By "compiler", do you mean some part of HLJS that processes these language files? Or is this a JS "feature" ? In the first case, I'd still prefer the explicit version I think. It's less surprising/confusing.

For example, in https://github.com/highlightjs/highlight.js/pull/2938, I have the following code:

```js
keywords: [
  ...Swift.precedencegroupKeywords,
  ...Swift.literals
].join(' '),
```

I tried leaving off the leading `...` here, but that didn't work. Of course, this being JS and all, I didn't get any errors. It just didn't work 🤷‍♂️

joshgoebel:
@svanimpe Thanks for all the hard work on this. :)  Will ship in Jan with 10.6.

PR #2937, opened 2020-12-24
continue CSS Consistency

joshgoebel:
### Changes

Continues work towards CSS consistency across all our CSS grammars
that allow raw CSS (ie, they are supersets).  Cleans up all grammars a lot.

Less is still a hot mess, but the other 3 are getting a lot cleaner.

### Checklist


joshgoebel:
Should be easy to review this one one commit at a time if the overall view is too messy.  I tried to keep them clean.

sa-orlov (approved):

sa-orlov (approved):

src/languages/lib/css-shared.js
```diff
@@ -1,3 +1,137 @@
+export const MODES = (hljs) => {
```
egor-rogov, line 1: Doesn't `MODES` sound too broad for these 3 modes? You'll have to guess what `  const modes = css.MODES(hljs);` really means.

src/languages/css.js
```diff
@@ -5,141 +5,144 @@ Website: https://developer.mozilla.org/en-US/docs/Web/CSS
...
-          hljs.APOS_STRING_MODE,
-          hljs.QUOTE_STRING_MODE,
+        className: 'selector-pseudo',
+        variants: [
+          {
+            begin: ':(' + css.PSEUDO_CLASSES.join('|') + ')'
+          },
+          {
+            begin: '::(' + css.PSEUDO_ELEMENTS.join('|') + ')'
+          }
```
egor-rogov, line 66: Are two variants better that one with `::?`?

joshgoebel: CLASSES and ELEMENTS are entirely separate lists.  This is more precise (and clearer).

src/languages/lib/css-shared.js
joshgoebel: Seems clear to me?   MODES just for CSS vs MODES in general.

Not sure I follow.  Did you have an alternative suggestion that you think is clearer?

I wrapped them so I could pass in `hljs`... seems silly to have a 3+ separate functions and have to pass them all hljs.

egor-rogov: No, I don't have anything else but 3 separate functions. OK if single function seems clear enough to you.

egor-rogov (approved):
Looks good to me.

PR #2938, opened 2020-12-27
(swift) Support operator and precedencegroup declarations.

svanimpe:
This was the final thing on my list :)

- Adds support for `title` highlighting in an operator declaration.
- Same for a `precedencegroup` declaration.
- Highlights the contextual keywords in a `precedencegroup`. These keywords were removed from the global list because they're very rarely used and would cause mostly incorrect highlights as global keywords.

### DIFF
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -416,11 +416,8 @@ export default function(hljs) {
   // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID380
   const OPERATOR_DECLARATION = {
     beginKeywords: 'operator',
+    end: hljs.MATCH_NOTHING_RE,
     contains: [
-      {
-        match: /\s+/,
-        relevance: 0
-      },
       {
         className: 'title',
         match: Swift.operator,
@@ -433,11 +430,8 @@ export default function(hljs) {
   // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID550
   const PRECEDENCEGROUP = {
     beginKeywords: 'precedencegroup',
+    end: hljs.MATCH_NOTHING_RE,
     contains: [
-      {
-        match: /\s+/,
-        relevance: 0
-      },
       {
         className: 'title',
         match: Swift.typeIdentifier,
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -413,6 +413,49 @@ export default function(hljs) {
     ],
     illegal: /\[|%/
   };
+  // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID380
+  const OPERATOR_DECLARATION = {
+    beginKeywords: 'operator',
+    contains: [
+      {
+        match: /\s+/,
+        relevance: 0
+      },
+      {
+        className: 'title',
+        match: Swift.operator,
+        endsParent: true,
+        relevance: 0
+      }
+    ]
+  };
+
+  // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID550
+  const PRECEDENCEGROUP = {
+    beginKeywords: 'precedencegroup',
+    contains: [
+      {
+        match: /\s+/,
+        relevance: 0
+      },
+      {
+        className: 'title',
+        match: Swift.typeIdentifier,
+        relevance: 0
+      },
+      {
+        begin: /{/,
+        end: /}/,
+        relevance: 0,
+        endsParent: true,
+        keywords: [
+          ...Swift.precedencegroupKeywords,
+          ...Swift.literals
+        ].join(' '),
+        contains: [ TYPE ]
+      }
+    ]
+  };
 
   // Add supported submodes to string interpolation.
   for (const variant of STRING.variants) {
@@ -460,6 +503,8 @@ export default function(hljs) {
           ...KEYWORD_MODES
         ]
       },
+      OPERATOR_DECLARATION,
+      PRECEDENCEGROUP,
       {
         beginKeywords: 'import',
         end: /$/,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,15 @@ Language grammar improvements:
 
 - enh(swift) Improved highlighting for operator and precedencegroup declarations. (#2938) [Steven Van Impe][]
 
+Parser:
+
+- add `modes.MATCH_NOTHING_RE` that will never match
+  - This can be used with `end` to hold a mode open (it must then be ended with
+    `endsParent` in one of it's children modes) [Josh Goebel][]
+
 [Michael Newton]: https://github.com/miken32
 [Steven Van Impe]: https://github.com/svanimpe/
+[Josh Goebel]: https://github.com/joshgoebel
 
 
 ## Version 10.5.0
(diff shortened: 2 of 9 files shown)


src/languages/swift.js
```diff
@@ -283,6 +283,49 @@ export default function(hljs) {
...
   TYPE.contains.push(GENERIC_ARGUMENTS);
 
+  // https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID380
+  const OPERATOR_DECL = {
+    beginKeywords: 'operator',
+    contains: [
+      {
+        match: /\s+/,
+        relevance: 0
+      },
```
joshgoebel, line 293:
```suggestion
  const OPERATOR_DECL = {
    beginKeywords: 'operator',
    ends: hljs.NEVER,
    contains: [
      {
```

I'm not sure what to name it (never isn't quite right) but for things like this rather than all that muck to eat spaces it should be easier to just write "leave the mode open until you find a token"...

src/languages/swift.js
```diff
@@ -283,6 +283,49 @@ export default function(hljs) {
...
+      {
+        begin: /{/,
+        end: /}/,
+        endsParent: true,
+        keywords: [
+          ...Swift.precedencegroupKeywords,
+          ...Swift.literals
+        ].join(' '),
+        contains: [ TYPE ]
+      }
```
joshgoebel, line 325:
```suggestion
      {
        begin: /{/,
        end: /}/,
        relevance: 0,
        endsParent: true,
        keywords: [
          ...Swift.precedencegroupKeywords,
          ...Swift.literals
        ].join(' '),
        contains: [ TYPE ]
      }
```

The match group isn't what's relevant, the keywords inside are.

src/languages/swift.js
```diff
@@ -362,6 +405,8 @@ export default function(hljs) {
           ...KEYWORD_MODES
         ]
       },
+      OPERATOR_DECL,
```
joshgoebel, line 408: Let's not abbreviate?  This is all gziped and minified later anyways.

joshgoebel:
Fairly straight-forward. :)

src/languages/swift.js
joshgoebel: I'd also be open to a new attribute to suggest this if it was clear and easy to read.  But it would really just be sugar for the above.

joshgoebel:
Please add issue link to the post so the issue can auto-close when we merge. :)  Pretty sure we had an issue on this?

sa-orlov (approved):

sa-orlov (approved):

sa-orlov:
and how to keep it?

sa-orlov (approved):

sa-orlov (approved):

sa-orlov (approved):

sa-orlov (commented):
You Best

sa-orlov (commented):
You Best

sa-orlov (approved):

src/languages/swift.js
svanimpe: Is `MATCH_NEVER` just a regex that never matches, such as `\b\B`? I guess it's a bit confusing because it makes it seems as though the mode never ends. But I assume you'd have to combine this with a child mode that ends its parent?

src/languages/swift.js
svanimpe: Fixed.

src/languages/swift.js
svanimpe: Fixed.

svanimpe:
I've made the suggested changes/fixes.

I'll leave the whitespace handling up to you. Perhaps you can already merge this PR and decide on that issue later, as it probably applies to lots of code, not just this PR?

src/languages/swift.js
joshgoebel: > Is MATCH_NEVER just a regex that never matches, such as \b\B? 

Well, it doesn't exist yet, but we'd add it - and yes, exactly.

> I guess it's a bit confusing because it makes it seems as though the mode never ends. 

Exactly, *unless manually ended by a child*... obviously the mode ends upon input termination.  My concern is that confusion between "never ends, literally" and "never ends, but by child".

> But I assume you'd have to combine this with a child mode that ends its parent?

Yes, as you already have... this would just avoid that extraneous parsing of whitespace.

I very much thing the approach is correct, and obviously we could do it with just `\b\B` and a comment, but it seems nice to bless the pattern with a proper name. @allejo Any ideas?

joshgoebel:
> I'll leave the whitespace handling up to you. Perhaps you can already merge this PR and decide on that issue later, as it probably applies to lots of code, not just this PR?

Yes, I'm happy to fix.

I don't see a rush (no pending release).  Often I use PRs as pressure on myself to get new things added.  I wouldn't leave this open for weeks/months for such a reason, but if it remains open a few extra days over the holidays and encourages me to add support for never match, then that seems a win win.  I'm curious if allejo has any better thoughts for the naming.

joshgoebel:
Added the code, we only need the name.  :)

joshgoebel:
@allejo Any thoughts on naming?  I'm leaning towards just `NEVER_RE` and being done with it.

allejo:
Would this ever be used anywhere else besides `end` in a mode? If not, I'd lean towards something like `ENDS_NEVER` so it doesn't end up being used in other places.

```js
const OPERATOR_DECL = {
    beginKeywords: 'operator',
    ends: hljs.ENDS_NEVER,
    // ...
}
```

Otherwise, if this can be used elsewhere, `NEVER_RE` sounds good to me

joshgoebel:
> Would this ever be used anywhere else besides end in a mode? If not, I'd lean towards something like ENDS_NEVER so it doesn't end up being used in other places.

OMG, at first glance I like it a lot.  I'll give it a noodle but I'm not sure where else you'd use it. `begin: NEVER` doesn't make any sense...  If I can't think of anything else we'll go with ENDS_NEVER and I'll get this merged.

joshgoebel:
@allejo 

Oh, but how do you feel about the fact that technically it's wrong?  Usually you'd be pairing this with `endsParent` in a child:

```
const OPERATOR_DECL = {
    beginKeywords: 'operator',
    ends: hljs.ENDS_NEVER,
    contains: [
      { 
        begin: ...,
        endsParent: true
    ]
}
```

Does ENDS_NEVER cause confusion?  Technically it's a could be seen as a lie.

allejo:
I think reading `ends: NEVER` would cause the same confusion as `ENDS_NEVER`, no? If we want to go down a more explicit/verbose route, what about `ENDED_BY_CHILD`?

joshgoebel:
> I think reading ends: NEVER would cause the same confusion as ENDS_NEVER, no?

This is a good point... LOL.  Maybe a more literal:

- `MATCH_NOTHING_RE`.
- `MATCH_NEVER_RE`.

>  If we want to go down a more explicit/verbose route, what about ENDED_BY_CHILD?

Because there are sometimes good reasons to end a mode and never end it at all... like HTTP response body...

allejo:
You have my support for `ends: MATCH_NOTHING_RE`. It's the most accurate since it's purposely not matching anything for the end of a mode but doesn't imply it can never be ended.

joshgoebel:
@svanimpe Thanks for all the hard work.

joshgoebel (approved):

svanimpe:
@joshgoebel What is the release schedule for 10.6?

joshgoebel:
I just updated the dates... this next week is going to be busy so I'd say next weekend sounds about right.

PR #2941, opened 2020-12-28
Use Array.from(NodeList).forEach() for IE11

### LINKED ISSUE #2940: IE11 Object doesn't support property or method 'forEach'
Since https://github.com/highlightjs/highlight.js/commit/3198e92629c8c09f9372e279dd833884e4edb839 the highlighting in IE11 is broken:

> SCRIPT438: Object doesn't support property or method 'forEach'

The reason:

https://github.com/highlightjs/highlight.js/blame/82f7a68cef989d2de3aa2b02faf2f80f6ce7ee60/src/highlight.js#L746-L747

The solution: 

```js
Array.from(NodeList).forEach(...)
```

limonte:
Fixes #2940

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -744,7 +744,7 @@ const HLJS = function(hljs) {
     initHighlighting.called = true;
 
     const blocks = document.querySelectorAll('pre code');
-    blocks.forEach(highlightBlock);
+    Array.from(blocks).forEach(highlightBlock);
   };
 
   // Higlights all when DOMContentLoaded fires


sa-orlov (approved):

sa-orlov (approved):

sa-orlov (approved):

sa-orlov (approved):

joshgoebel:
Closing per comment on issue.  Our codebase purposely uses these newer features and syntax *because* we no longer support older browsers, like IE11.

If you truly need IE11 support I'd recommend Prism or perhaps the older release of Highlight.js (version 9), although it has known security issues - but they may not matter if you are only using the library to highlight your own code.

sa-orlov (commented):
He ok

sa-orlov (commented):
Ok

limonte:
Thanks for the feedback @joshgoebel!

I'll stick to v9 then. IE11 is still a must for a lot of projects.

joshgoebel:
>  IE11 is still a must for a lot of projects.

Sure, but less than 1% total browser share according to caniuse.com... so we focus on the 99% not the 1%.

> I'll stick to v9 then.

If you're processing input from *users* know that there are many ways they could easily potentially lockup/crash your site because of bad regexes... for dealing with user input v9 simply isn't secure anymore though you could always review the v10 security release and try to backport the fixes or only use languages that didn't contain vulnerabilities.

Also see my suggestion of using Prism (though they are discussing dropping support for IE11 also, just a matter of time).

PR #2951, opened 2021-01-07
(fix) Improve plugin-api types and docs

Hirse:
### Changes
- Fix types of `after:highlightBlock` to reflect actual parameter
- Clean up plugin-api docs

### Checklist
- [x] ~Added markup tests~, or they don't apply here because no code change

### DIFF
--- a/docs/plugin-api.rst
+++ b/docs/plugin-api.rst
@@ -13,15 +13,15 @@ You can add a plugin via the ``addPlugin`` API.
 ::
 
   // a plugin can be a class
-  addPlugin(new SimplePlugin())
-  addPlugin(new MoreComplexPlugin(options))
+  addPlugin(new SimplePlugin());
+  addPlugin(new MoreComplexPlugin(options));
 
   // or simply a keyed object of functions
   addPlugin({
-    'after:highlightBlock': (args) => {
-      ...
+    'after:highlightBlock': ({ block, result, text }) => {
+      // ...
     }
-  })
+  });
 
 Class based plugins
 ^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


docs/plugin-api.rst
```diff
@@ -13,15 +13,15 @@ You can add a plugin via the ``addPlugin`` API.
...
-  addPlugin(new SimplePlugin())
-  addPlugin(new MoreComplexPlugin(options))
+  addPlugin(new SimplePlugin());
+  addPlugin(new MoreComplexPlugin(options));
 
   // or simply a keyed object of functions
   addPlugin({
-    'after:highlightBlock': (args) => {
-      ...
+    'after:highlightBlock': ({ block, result, text }) => {
```
joshgoebel, line 21: Ok I see why you did this [consistency]... the second "high level example" (line 41) has arguments... but I wonder if at this high level we should just use `args` and save the details for the API docs below vs creating yet another place we're going to forget to update in the future when this changes?

Mainly: _The exact arguments don't matter for these high-level examples as none of them are used._

Any thoughts?

Hirse: Yes, that was the intention.

I'm fine removing the exact object properties, but we should probably change it to `data` then to be consistent with the types.

Specifically for this case I wouldn't worry too much about updating since it's in the same file.

joshgoebel: Well I already got it wrong once. :)  But it also doesn't change that often... My preference is `data` but it's only 60/40 opinion... so if you're feeling lazy I can be persuaded to just merge as-is.

Hirse: Alright, let's leave it then. I'll update it next time. ;)

joshgoebel: Does your linter flag this?

joshgoebel (approved):

PR #2953, opened 2021-01-07
enh(parser) beginKeyword no longer bestows double relevance

joshgoebel:
Previously something like:

```
{ 
  beginKeywords: "def"
  // which also implies:
  // keywords: "def"
}
```

Would bestow 2 relevance. 1 for the rule match and a second for the
`keywords` match.  This is not correct behavior. A single "token" should
default to a single relevance point, not 2.

### Changes

This changes the default relevance of a mode containing `beginKeywords`
to 0 forcing it to only gain relevance from keyword matches alone.

This of course revealed a few auto-detect balance issues (because some 
grammars had been benefiting from this extra boost) which this PR
also resolves.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Language grammar improvements:
 
 Parser:
 
+- `beginKeyword` no longer bestows double relevance (#2953) [Josh Goebel][]
 - add `modes.MATCH_NOTHING_RE` that will never match
   - This can be used with `end` to hold a mode open (it must then be ended with
     `endsParent` in one of it's children modes) [Josh Goebel][]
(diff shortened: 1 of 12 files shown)


joshgoebel:
Commit history here is clean if one wanted to review per commit.

egor-rogov (approved):
Looks good, but do we really need floating-point relevance (1.5)? I guess 2 didn't work, but still...

joshgoebel:
> Looks good, but do we really need floating-point relevance (1.5)? I guess 2 didn't work, but still...

**Well, need is a strong word.**  I don't want to encourage it (if not necessary), but it was helpful here;  and honestly DOUBLE relevance (not to mention 10x) is really a pretty huge multiple to be the only choice.  And many of the really "smart" things we could do with relevance will ultimately result in fractional relevance values.

`Math.floor` is just be being ridiculously careful with compatibility.  I wager it wouldn't matter, but was easy enough to add and then remove with v11 (allowing fractional relevance values).

PR #2954, opened 2021-01-07
(chore) deprecate C-like

joshgoebel:
Deprecates `c-like` grammar (though still allows it's use) and copies
the contents fully into `c` and `cpp` grammars which will now diverge
as time goes on.

`c` still has auto-detect disabled at this point as it's just a clone
of `cpp`, but when someone is ready to simplify the `c` grammar tests
may be added and auto-detect may be turned back on.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,12 @@ Language grammar improvements:
 
 - enh(swift) Improved highlighting for operator and precedencegroup declarations. (#2938) [Steven Van Impe][]
 
+Grammar changes:
+
+- Deprecate `c-like`, though you should not be using it directly anyways.
+  - will be removed in v11.
+- `c` and `cpp` are now wholly unique grammars that will diverge over time
+
 Parser:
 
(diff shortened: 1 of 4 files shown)


joshgoebel:
This diff is huge but essentially it's just a few trivial things:

- copy `c-like` into `c.js` (tweaking `name` and `aliases`)
- copy `c-like` into `cpp.js` (tweaking `name` and `aliases`)
- `c.js` has `autoDetect: false` still (see comment in PR main text)
- `c-like` attempt to see if `c` and `cpp` are loaded yet and if not will claim all C and C++ aliases for itself (to support users using it directly even though they should not be)
- update `c-like` comments to explain the current situation

joshgoebel:
This work is to force the grammars to begin to diverge even if no one wants to put in the FULL time yet to make them perfect. `c-like` paved the way for this to happen (in v10) but did not actually encourage this.

egor-rogov (approved):

PR #2955, opened 2021-01-07
enh(sql): added `ilike` keyword

0xflotus:
### Changes

I added `ilike` to keyword list in sql

### Checklist

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -312,3 +312,4 @@ Contributors:
 - Guillaume Grossetie <ggrossetie@yuzutech.fr>
 - Steven Van Impe <steven.vanimpe@icloud.com>
 - Martin Dørum <martid0311@gmail.com>
+- 0xflotus (https://github.com/0xflotus)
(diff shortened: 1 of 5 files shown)


joshgoebel:
Please provide a list of which major databases suppose ILIKE and which do not.  We're quite selective about what we add to the main SQL grammar as there are individual grammars for the different DBs.

0xflotus:
> Please provide a list of which major databases support ILIKE (do they all?). We're quite selective about what we add to the main SQL grammar as there are individual grammars for the different DBs.

PostgreSQL supports it. That's the major DB I use.

joshgoebel:
This appears to be a non-standard Postgres feature.  Please use the `pgsql` grammar instead.  We won't be adding this to `sql`.

0xflotus:
> Thanks for the PR, but this appears to be a non-standard Postgres feature. Please use the `pgsql` grammar instead. We won't be adding this to `sql`.

Thanks for the hint.

Issue #2277, opened 2019-11-12
Propose: Improve `highlight` API


joshgoebel:
The old API:

```js
highlight(language, code, ignore_illegals, continuation) {
```

The new API take an options object literal:

```js
highlight(code, { language, ignoreIllegals}) {}
```

So to upgrade your code:

```js
// before
html = highlight("vbscript","your code goes here", true)
// after
html = highlight("your code goes here", {language: "vbscript", ignoreIllegals: true })
```

Advantages:

- target of the verb [highlight] is now the first argument, which makes more sense, such as `format(drive)`, `highlight(code)`
- no need to remember order of the arguments
- we can still support the old API for legacy (ie, non-breaking change)
- fewer arguments are generally better (easier to understand, now at the call site you always see the names of the optional arguments, etc)

Checklist:

- [x] new `highlight` API as shown above
- [ ] <del>`highlightAuto` -> `highlight(code)`</del>
- [ ] <del>`highlightAuto` is deprecated</del>
- [x] `highlightBlock` -> `highlightElement` #3003
- [x] `highlightBlock` -> deprecated #3003
- [x] `fixMarkup` -> deprecated, will be removed with v11
- [x] `initHighlightingOnLoad` -> `highlightAll()`
- [x] `initHighlighting` -> `highlightAll()`

egor-rogov:
Yes, the suggested API looks better to my taste. But I can't see enough value in it to justify API breaking change.
Making it non-breaking - well, why not.

yyyc514:
Any objection to:

- `highlightBlock` -> `highlightElement`
- `initHighlighting` -> `start()` or `highlightAll()`
- `initHighlightingOnLoad` -> `startOnLoad()`

Both for clarity and simplicity.

egor-rogov:
`start` doesn't looks clear to me, `highlightAll` is better to my taste.

joshgoebel:
@allejo Any thoughts on these API naming changes? I think this was a topic before we added you to the team.

allejo:
I agree with Egor where "start" isn't clear. I'd lean more towards:

- `highlightDOMElement` or `highlightDOMNode` (I don't immediately think a DOM element if it's just `highlightElement`)
- `highlightAll`
- `highlightAllOnLoad`

joshgoebel:
I don't like this highlightAllOnLoad vs highlightAll nonsense and forcing users to think about WHEN.  I think instead:

```
[...]

PR #2962, opened 2021-01-14
(parser) new highlightAll() API

joshgoebel:
Progress towards #2277.

### Changes

Added new `highlightAll()` API which deprecates:

- initHighlighting()
- initHighlightingOnLoad()

These are slated to be removed with version 12. (with v11 they will likely just because aliases to `highlightAll()`)

---

The new function can be called at any time and will "just work".

- If called after DOM has loaded, it will immediately highlight.
- If called before, it will wait until the DOM finishes loading, then
highlight.

If there are any cases where `highlightAll()` is needed prior to the DOM `onload` firing, we can discuss those cases between now and v12.  That would seem to be an edge case though I'd guess and could easily be handled with a custom unload handler.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,10 +11,19 @@ Language grammar improvements:
 
 Parser:
 
+- new simpler `highlightAll()` API (#2962) [Josh Goebel][]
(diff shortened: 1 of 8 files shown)


egor-rogov (approved):
Good thing!

PR #2968, opened 2021-01-20
Resolve npm audit

StoneCypher:
Two minor changes to get `npm audit` to hush its face.

I'm not sure it makes sense to put this in `CHANGES.md`?  It's near-trivial

### Changes

1. Bumped `ini` from 1.3.5 to 1.3.8 (prototype pollution)
1. Removed `eslint-plugin-standard` as it's now redundant

### Checklist
- [x] Added markup tests, or they don't apply here because... `it's just package-lock`


joshgoebel:
> Removed eslint-plugin-standard as it's now redundant

How so?

> I'm not sure it makes sense to put this in CHANGES.md? It's near-trivial

Correct. No need to add a line item.

---

Please undo the version bump, the version will be auto-bumped with the next release.

joshgoebel:
FYI this is all stuff I usually handle at release time, so no need to make PRs for this in the future.

StoneCypher:
> > Removed eslint-plugin-standard as it's now redundant
>
> How so?

These are instructions given by the plugin itself, on install.  [The plugin was deprecated in July of 2019](https://github.com/standard/standard/issues/1316).

image


.

> Please undo the npm version bump, the version will be auto-bumped with the next release.

Auto-bumps are dangerous, and there's a security problem ***right now***.  Also, most auto-bumpers are getting this wrong; they're upgrading to 1.3.7 instead of 1.3.8 because of cross-pinning, which looks like it resolves the problem, but actually does not.

You sure?  The thing that gets fixed here is exploitable today.

StoneCypher:
The security fix I hope to merge here is from December 11 of 2020

I really hope you consider letting this version upgrade go in sooner, rather than later.  This is a non-trivial problem.  Several significant DOSes can be built with this.

joshgoebel:
> Auto-bumps are dangerous

Before a new release I typically review the dependencies (for outdated releases) and bump those that seem necessary/relevant (also paying attention to security flags).  It's a part of the release process, but there is nothing truly automatic about it.

> and there's a security problem right now. 

With what? `ini`?  **All our released packages do not depend upon `ini`**, it is merely a development dependency. A new "release" including this bump would include *exactly the same code* as 10.5.

---

In ANY case the main package version bump would happen when I decide to do a release (whether it's this afternoon or a month from now) it would not be included with this PR.  So please revert that and then the rest of the PR could be merged.

StoneCypher:
okay.

joshgoebel:
I tracked this to ground zero:

- `ini` is a dependency of `js-beautify`
-  we use `js-beautify` in one dev util script that we never actually run (should probably remove)
- `js-beautify` only uses `env` from `config-chain`, **so it's not vulnerable** (`env` doesn't use `ini`)

It's good to be aware of security bumps but it's also important to realize that just because a library pulls in a package (as a dependency of a dependency) does not automatically mean it is vulnerable.  

Always worth asking the question though, so thanks for bring it up.

joshgoebel:
Since we use `eslint` now for periodic formatting I'll probably remove this beautification script and it's dependencies from 10.6 entirely... fewer deps is always better than more.

package.json
```diff
@@ -6,7 +6,7 @@
     "syntax"
   ],
   "homepage": "https://highlightjs.org/",
-  "version": "10.5.0",
+  "version": "10.5.1",
```
joshgoebel, line 9: This still needs to be reverted to 10.5.0.

StoneCypher: you don't bump patch for patches?  

ok sec

StoneCypher: my urge to call un-bumping "denting," though

joshgoebel: When we do another release we'll bump the version, not until then.

joshgoebel: It's likely to be 10.6 though, not 10.5.1... which is why I don't bump until I do an actual release and find out what the version number will be. :)

PR #2977, opened 2021-01-27
Addeded Arduino Dark theme

### LINKED ISSUE #2975: [Request] Arduino Dark theme
I would like to highlight _Arduino_ source code with dark theme.
It's really nice theme for highlighting Arduino code.

I would like to request to add _Arduino dark theme_ based on _One Dark Pro_ – https://github.com/konrad91/OneDarkArduino.

**Additional context...**
Here is sample image of code highlighted by that theme:
Arduino Dark Theme

This feature request is related to [that one](https://github.com/yusanshi/embed-like-gist/issues/11)...

BelKed:
Resolves #2975.

### Changes
Addeded Arduino Dark theme.

### Checklist
- [x] Addeded Arduino Dark theme


src/styles/arduino-dark.css
```diff
@@ -0,0 +1,86 @@
...
+.hljs-selector-pseudo,
+.hljs-type,
+.hljs-string,
+.hljs-selector-id,
+.hljs-selector-class,
+.hljs-quote,
+.hljs-template-tag,
+.hljs-deletion {
+  color: #61afef;
+}
```
joshgoebel, line 56: This is a bit unexpected... can't we follow the same patterns used over in the light font to make sure that this also works equally well highlighting HTML, CSS, etc, not just source code?

joshgoebel (changes requested):

src/styles/arduino-dark.css
BelKed: OK, I edited it in https://github.com/highlightjs/highlight.js/pull/2977/commits/3452ccfa61a2109c4043a7338452d3b412fba3ab.

joshgoebel:
The problem isn't just how they are grouped, it's using the same color for all those *very different* classes... HTML/CSS/markup highlighting is not going to look good.  The original Arduino theme uses 3 different colors (for those 4 groups)... so the dark theme probably also should.

Any themes we add need to have reasonable styling for ALL the grammars we support, so they can be used with whatever language a user decides to highlight.  A theme that only highlights code well but not HTML or CSS is not something we'd want to merge.

src/styles/arduino-dark.css
```diff
@@ -67,6 +68,7 @@
 .hljs-title,
 .hljs-section {
   color: #c678dd;
+  font-weight: normal;
```
joshgoebel, line 71: What is this trying to do? The default would already be normal...

src/styles/arduino-dark.css
```diff
@@ -9,6 +9,7 @@
   overflow-x: auto;
   padding: 0.5em;
   background: #282c34;
+  font-family: Hack;
```
joshgoebel, line 12: Sorry, themes may not specify custom fonts.

BelKed:
I edited colors as you want. Here is HTML sample:
HTML


And here is CSS sample:
CSS

src/styles/arduino-dark.css
BelKed: Oh, I was trying my font and I forget to remove...

src/styles/arduino-dark.css
BelKed: Yes, I know, I forget to delete...

joshgoebel:
I thought Arduino was perhaps now shipping with a dark theme.  Please confirm that this is just a theme of your own design? (or is it based on something else?)

BelKed:
This theme is based on One Dark theme for Arduino IDE – https://github.com/konrad91/OneDarkArduino...

joshgoebel:
Don't we already support this via "Atom One Dark"?  It seems that is the intention of that theme, no?

BelKed:
Oh, I see. I close this pull request.

PR #2978, opened 2021-01-27
Allow for browser-testing of 'extra' language.

dbadb:
Trivial change to search the extra directory for detect/${lang}/default.txt
for inclusion into index.html for the browser-based testing.  Without
this change, my custom language wouldn't appear in the browser.
No problem for me if you don't wish to integrate this change.

### Changes

one liner self explanatory.

### Checklist
- [x ] Added markup tests, or they don't apply here because...

### DIFF
--- a/tools/lib/language.js
+++ b/tools/lib/language.js
@@ -40,10 +40,6 @@ class Language {
     this._sample = "";
     if (fs.existsSync(this.samplePath))
       this._sample = fs.readFileSync(this.samplePath, {encoding: "utf8"});
-    else
-    {
-      console.log("language.js no samplePath for " + this.name);
-    }
     return this._sample;
   }
 
--- a/tools/lib/language.js
+++ b/tools/lib/language.js
@@ -44,11 +44,13 @@ class Language {
   }
 
   get samplePath() {
-    if (!this.moduleDir) {
-      return `./test/detect/${this.name}/default.txt`;
+    if (this.moduleDir) {
+      // this is the 'extras' case.
+      return `${this.moduleDir}/test/detect/${this.name}/default.txt`;
     }
     else {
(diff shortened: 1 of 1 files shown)


tools/lib/language.js
```diff
@@ -44,7 +44,12 @@ class Language {
   }
 
   get samplePath() {
-    return `./test/detect/${this.name}/default.txt`
+    if (!this.moduleDir) {
+      return `./test/detect/${this.name}/default.txt`;
+    }
+    else {
+      return `${this.moduleDir}/test/detect/${this.name}/default.txt`;
+    }
```
joshgoebel, line 52:
```suggestion
    if (this.moduleDir) {
      return `${this.moduleDir}/test/detect/${this.name}/default.txt`;
    } else {
      return `./test/detect/${this.name}/default.txt`;    
    }
```

Avoid unnecessary negation.  Does this still look good?

dbadb: Looks fine - my usual practice is to place the common case first.
Want me to make that change?

joshgoebel: Yes, please.

IMHO: An if/else to me expressed equal weight on both branches... so I merely tried to simplify.  If I'd thought harder I might have turned the exception case into a guard and then left the "happy path" as the fall thru.

Hirse: Is it not enough to do the following for both cases?
```js
return `${this.moduleDir || "."}/test/detect/${this.name}/default.txt`; 
```
or slightly more verbose
```js
const prefix = this.moduleDir || ".";
return `${prefix}/test/detect/${this.name}/default.txt`; 
```

joshgoebel: Yeah, nice, but we already went one way. :)

joshgoebel (approved):

PR #2979, opened 2021-01-28
Fix build missed opts

contributorpw:
I added direct reading of options from the commander, since the previous challenge did not work.

I haven't noticed any problems on this topic yet.

### Changes

The previous

```js
$> commander.minify;
undefined
```

If we call `node build.js` with `-n` or `--no-minify`

```js
$> commander.opts().minify;
false
```

If we call `node build.js` without `-n` or `--no-minify`

```js
$> commander.opts().minify;
true
```

### Checklist
- [x] Markup tests don't apply here because it's a code misprint.


joshgoebel:
Good catch!

PR #2983, opened 2021-02-01
enh(autoit) add missing directives and built-ins

mlipok:
no issue just review

### Changes
added WinWaitActive WinWaitClose WinWaitNotActive

### Checklist


joshgoebel:
Please explain and justify reasoning for "other fixes". We typically like to understand the reasoning and context when things are changed or removed from grammars.

mlipok:
Ok. In next 24h. Currently I'm at my work.

mlipok:
Default is moved as in AutoIt HelpFile it is described the same as Null and this is the same kind of Keyword like "True False And Null Not Or" they are **LITERAL**. Typically they could be used as an operator (And Not Or) as a value in function call, as a part of expression. 
For comaprition "Dim Do Else ElseIf EndFunc EndIf EndSelect" are used to declare variables (Dim) or as a conditional statements.

In relation to meta-keyword **comments** there is no something like comments in AutoIt. There are:
```lang-autoit
#comments-start
...
...
#comments-end
```
From the AutoIt HelpFile Remarks:
```
The #comments-start and #comments-end directives can be nested.
You can also use the abbreviated keywords #cs and #ce.
Additionally, the directives themselves can be commented out!
```

For this reason meta-keyword **comments** are not proper as this is only a part of **#comments-start** and **#comments-end**  directives.

mlipok:
btw. comments are defined in this part:
```lang-js
  const COMMENT = {
    variants: [
      hljs.COMMENT(';', '$', {
        relevance: 0
      }),
      hljs.COMMENT('#cs', '#ce'),
      hljs.COMMENT('#comments-start', '#comments-end')
    ]
  };
```

joshgoebel:
> btw. comments as Far as I understand are handled in this part:

Indeed... but it seems to me `#comments` is still a directive and could be used for single line comments, no? [ref](https://github.com/Fmk0/work/edit/master/AutoIT/README.md)

 If we wanted to be consistent might we also support that with `COMMENT` as well?

I worry about treating comments differently than other directives but perhaps given their status as comments it make sense?

mlipok:
As to your statement "but it seems to me #comments is still a directive "

"#comments" no

but:
"#

> comments-start"
> "#comments-end"
> "#cs"
> "#ce"

Yes.

image

btw. As you can see there is many other directives not supported yet, like:
> "#region"
> "#endregion"
> "#forceref"
> "#ignorefunc"
> "#Au3Stripper**********"

Maybe also other which I must deeply review later.

mlipok:
"Indeed... but it seems to me #comments is still a directive and could be used for single line comments, no?"
Anything that in line starts from hash is treated as a commented line and is ommited at runtime.

image

joshgoebel:
> Anything that in line starts from hash is treated as a commented line and is ommited at runtime.

OK... Then I'm pretty sure we'd also need to add:

```js
      hljs.COMMENT(/^#/, /$/, {
        relevance: 0
      }),
```

To the list of variants... would we not?

mlipok:
Yes....

one Remark: proper name of the language is "AutoIt"   a never "AutoIT"

mlipok:
Ah... there is also a list of directives that was not enlisted yet.

joshgoebel:
Do you plan to add them to this PR?  If so please upgrade to an array for easier maintenance in the future:

```js
const DIRECTIVES = [
  "include",
  "NoTrayIcon",
  ...
]
// ...
    keywords: {
      'meta-keyword': DIRECTIVES
    },
```

mlipok:
will try next few days

mlipok:
Take a look at my recent update, if this looks fine I will suplement them more, as this is not all.

mlipok:
can I use wildcard in an array declaration ?
#AutoIt3Wrapper_*
#Au3Stripper_*

Or do I have to list them separately?

src/languages/autoit.js
```diff
@@ -1,25 +1,46 @@
...
 
   const COMMENT = {
     variants: [
+      hljs.COMMENT(/^#/, /$/, {
+        relevance: 0
+      }),
```
joshgoebel, line 43: This isn't needed if we're happy with the current behavior.

`
#something
`

Is this a comment or a meta?  Your own highlighting makes it look "meta"... which is the current behavior... changing it to a comment would actually be a bit more complex than just adding this rule.

src/languages/autoit.js
```diff
@@ -1,25 +1,46 @@
...
+    "include",
+    "include-once",
+    "NoTrayIcon",
+    "OnAutoItStartRegister",
+    "pragma",
+    "Region",
+    "RequireAdmin",
+    "#Tidy_Off",
+    "#Tidy_On",
+    "#Tidy_Parameters"
```
joshgoebel, line 31: These will never work unless you add `$pattern` to your `keywords` hash.  The default keyword matcher is `\w+` which will NEVER match the #.  I think `/#?\w+/` might work.  See `powershell.js` for an example of doing this.

mlipok:
```suggestion
    "Tidy_Off",
    "Tidy_On",
    "Tidy_Parameters"
```
?

src/languages/autoit.js
mlipok:
```suggestion
```

src/languages/autoit.js
joshgoebel: wWell why did you have the # in front of them in the first case? How are they used in code? Is the  doubled?   I assumed you'd see:

```
##Tidy_Off
```

Otherwise I wasn't sure why you added the extra `#` at all...

And you'd still need $pattern to deal with the `-`...

joshgoebel:
> can I use wildcard in an array declaration ?

No, you'd need a regular mode rule if you needed to user regex or something more complex.  Keyword lists can only be literal strings.

> Or do I have to list them separately?

Yes, but if there are more than just a few variants then we really should likely use a mode instead to avoid listing out a long list. 

> Could you take a look and point me to the issue ? because I screw it up.

Not sure what you screwed up?

src/languages/autoit.js
joshgoebel: This is what broke the tests...

mlipok:
> Not sure what you screwed up?
I mean that 2 checks was canceled and 3 was failed

src/languages/autoit.js
mlipok: by mistake.

mlipok:
> > can I use wildcard in an array declaration ?
> No, ......
> > Or do I have to list them separately?
> Yes, but if there are more than just a few variants then we really should likely use a mode instead to avoid listing out a long list.

this is a source list which I would like tho use as a reference:

```lang-autoit
#AutoIt3Wrapper_Add_Constants?1
#AutoIt3Wrapper_Au3Check_Parameters?1
#AutoIt3Wrapper_Au3Check_Stop_OnWarning?1
#AutoIt3Wrapper_Au3stripper_Stop_OnError?1
#AutoIt3Wrapper_AutoIt3Dir?1
#AutoIt3Wrapper_Aut2Exe?1
#AutoIt3Wrapper_AutoIt3?1
#AutoIt3Wrapper_Change2CUI?1
#AutoIt3Wrapper_Compile_Both?1
#AutoIt3Wrapper_Compression?1
#AutoIt3wrapper_EndIf?1
#AutoIt3Wrapper_Icon?1
#AutoIt3wrapper_If_Compile?1
#AutoIt3wrapper_If_Run?1
#Autoit3wrapper_Jump_to_First_Error?1
#AutoIt3Wrapper_OutFile?1
#AutoIt3Wrapper_OutFile_Type?1
#AutoIt3Wrapper_OutFile_X64?1
#AutoIt3Wrapper_PlugIn_Funcs?1
#AutoIt3Wrapper_Res_Comment?1
#AutoIt3Wrapper_Res_CommpanyName?1
#Autoit3Wrapper_Res_Compatibility?1
#AutoIt3Wrapper_Res_Cursor_Add?1
#AutoIt3Wrapper_Res_Description?1
#AutoIt3Wrapper_Res_Field?1
#AutoIt3Wrapper_Res_File_Add?1
#AutoIt3Wrapper_Res_FileVersion?1
#AutoIt3Wrapper_Res_FileVersion_AutoIncrement?1
#AutoIt3Wrapper_Res_Fileversion_First_Increment?1
#AutoIt3Wrapper_Res_HiDpi?1
#AutoIt3Wrapper_Res_Icon_Add?1
#AutoIt3Wrapper_Res_Language?1
#AutoIt3Wrapper_Res_LegalCopyright?1
#AutoIt3Wrapper_Res_LegalTrademarks?1
#AutoIt3Wrapper_Res_ProductName?1
#AutoIt3Wrapper_Res_ProductVersion?1
#AutoIt3Wrapper_Res_requestedExecutionLevel?1
#AutoIt3Wrapper_Res_Remove?1
#AutoIt3Wrapper_Res_SaveSource?1
#AutoIt3Wrapper_Run_Stop_OnError?1
#AutoIt3Wrapper_Run_After?1
#AutoIt3Wrapper_Run_After_Admin?1
#AutoIt3Wrapper_Run_Au3Check?1
#AutoIt3Wrapper_Run_Au3Stripper?1
#AutoIt3Wrapper_Run_Before?1
#AutoIt3Wrapper_Run_Before_Admin?1
#AutoIt3Wrapper_Run_Debug?1
#AutoIt3Wrapper_Run_Debug_Mode?1
#AutoIt3Wrapper_Run_SciTE_Minimized?1
#AutoIt3Wrapper_Run_SciTE_OutputPane_Minimized?1
#AutoIt3Wrapper_Run_Tidy?1
#AutoIt3Wrapper_ShowGUI?1
#AutoIt3Wrapper_ShowProgress?1
#AutoIt3Wrapper_Testing?1
#AutoIt3Wrapper_Tidy_Stop_OnError?1
#AutoIt3Wrapper_UPX_Parameters?1
#AutoIt3Wrapper_UseUPX?1
#AutoIt3Wrapper_UseX64?1
#AutoIt3Wrapper_Version?1
#AutoIt3Wrapper_Versioning?1
#AutoIt3Wrapper_Versioning_Parameters?1
#Au3Stripper_Ignore_Funcs?1
#Au3Stripper_Ignore_Variables?1
#Au3Stripper_Off?1
#Au3Stripper_On?1
#Au3Stripper_AlwaysStrip_Off?1
#Au3Stripper_AlwaysStrip_On?1
#Au3Stripper_Parameters?1
#Tidy_Off?1
#Tidy_On?1
#Tidy_Parameters?1
#Tidy_ILC_Pos?1
```

joshgoebel:
> this is a source list which I would like tho use as a reference:

Yeah that's far too long a list to hand-code... You'd probably want to add a separate mode to catch those... ie something like:

```
{
  start: /#AutoIt3Wrapper_\S+/,
  end: /$/,
  className: // ...
```

Perhaps using `keywords` also if you just want the keyword portion highlighted, etc...

And make sure it comes before PREPROCESSOR in the list.

joshgoebel:
If that's too much then you can just finish up this PR with a changelog entry and we can get this piece of it merged as-is and you could circle back later to try your hand at the more complex stuff?

mlipok:
> If that's too much then you can just finish up this PR with a changelog entry and we can get this piece of it merged as-is and you could circle back later to try your hand at the more complex stuff?

As for now it is fine.
Will back to this later.

changelog entry ?

joshgoebel:
"Updated the changelog at CHANGES.md"

An entry for this change need to be added and your name added as an author footnote.

joshgoebel:
You need to pull the latest commits into this branch.  From console in a clean checkout:

```
git checkout -b patch-1
git cherry-pick ef344822386ef2af4e147e596630c355ea40e907
git cherry-pick 9c2f19b70b81ede4ca5fb09e43b71bb889b1510b
git push
```

src/languages/autoit.js
```diff
@@ -1,22 +1,40 @@
 /*
 Language: AutoIt
 Author: Manh Tuan <junookyo@gmail.com>
+Modified: Michał Lipok
```
joshgoebel, line 4:
```suggestion
```

We don't generally track this per-file like this other than main author and HUGE rewrites.

joshgoebel:
Thanks, I merged the two PRs together for you!

joshgoebel (approved):

PR #2987, opened 2021-02-05
enh(scala) fix triple quoted string

### LINKED ISSUE #2986: (Scala) Strings in triple quotes are highlighted incorrectly.
**Describe the issue**
For example, the code:

```
val s = """ this is a string "this is still a string" another quote: " after the quote """
```

create a String with the content:

```
 this is a string "this is still a string" another quote: " after the quote 
```
This code is highlighted incottectly. `this is still a string`, and `after the quote` is not formatted as a string, although they should be. The rule is that single quotes are just part of the string when they are surrounded by triple quotes.

For more context please see [this](https://meta.stackoverflow.com/q/404946/2359227) Meta Stack Overflow question.

Scala

I am not sure. The one that Stack Overflow is using.

**Sample Code to Reproduce**
```
val s = """ this is a string "this is still a string" another quote: " after the quote """
```

**Expected behavior**
`s` to be string, and not to open code segments in the middle of the string.

**Additional context**
Nothing I can think of :)

joshgoebel:
Resolves #2986

### Checklist

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -30,22 +30,23 @@ export default function(hljs) {
     className: 'string',
     variants: [
       {
+        begin: '"""',
+        end: '"""'
+      },
+      {
         begin: '"',
         end: '"',
         illegal: '\\n',
         contains: [ hljs.BACKSLASH_ESCAPE ]
       },
       {
-        begin: '"""',
-        end: '"""',
-        relevance: 10
-      },
-      {
         begin: '[a-z]+"',
         end: '"',
         illegal: '\\n',
-        contains: [ hljs.BACKSLASH_ESCAPE,
-          SUBST ]
+        contains: [
+          hljs.BACKSLASH_ESCAPE,
+          SUBST
+        ]
       },
       {
         className: 'string',
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ New Languages:
 
 Language grammar improvements:
 
-- enh(scala) fix triple quoted strings [Josh Goebel][]
+- enh(scala) fix triple quoted strings (#2987) [Josh Goebel][]
 - enh(perl) Much improved regex detection (#2960) [Josh Goebel][]
 - enh(swift) Improved highlighting for operator and precedencegroup declarations. (#2938) [Steven Van Impe][]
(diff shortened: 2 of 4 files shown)


src/languages/scala.js
```diff
@@ -29,23 +29,24 @@ export default function(hljs) {
...
       {
         begin: '"',
         end: '"',
         illegal: '\\n',
         contains: [ hljs.BACKSLASH_ESCAPE ]
       },
-      {
-        begin: '"""',
-        end: '"""',
-        relevance: 10
```
toshetah, line 41: Basically the main change I see here is that you removed the `relevance`, and that you changed the order. Can you please explain what do those changes do? Next time I might just contribute it myself :)

toshetah (approved):

src/languages/scala.js
joshgoebel: Well if `"` matches first that breaks everything, so we have to match `"""` first... and the relevance is a much longer explication... find the thread on huge relevance and read it for a start for more thoughts... 10 is just too high for something that is shared by several languages now, so I removed the huge relevance bump.

_The sequencing is what fixes the issue._

toshetah (approved):

toshetah:
I'd add another test of multi line string:

```
val multiLineString =
  """aaa
    |ddd
    |sdfa""".stripMargin
```

should produce:

```
aaa
ddd
sdfa
```

and:

```
```
val multiLineString =
  """aaa
    |ddd
    |sdfa"""
```

should produce:

```
aaa
    |ddd
    |sdfa
```

Code run at [Scastie](https://scastie.scala-lang.org/N2V8tPHcQguXehJyhxD4pA).

joshgoebel:
If you wanted to contribute additional tests to the project (expanding on `strings.txt`) that'd be great... long-term it'd be nice to test all the different types of strings, escaping of special chars, etc...

joshgoebel:
> I'd add another test of multi line string:

Of course our parsing engine doesn't care about the how the string is interpreted, we just highlight it. So your last two examples are irrelevant in our context - as far as I can tell. The highlighted is the same in both cases, regardless of what the end result it inside the runtime.

Issue #2996, opened 2021-02-10
Discuss: (PHP) built-ins are currently purposely not highlighted.


xCrimin4L:
**Describe the issue**
> When PHP is put into the pre and code tags, everything renders correctly except certain functions. They are missing <span class="hljs-keyword">and just left as regular text while other functions get the <span>.

**Which language seems to have the issue?**
> Seems to be PHP but maybe other languages also that need the class="hljs-keyword".

**Are you using `highlight` or `highlightAuto`?**
> I've tried both and both are giving me the same exact problem.

**Sample Code to Reproduce**
> https://jsfiddle.net/b2dj79qh/

**Expected behavior**
> You should see "json_decode" & "file_get_contents" functions white as the other function "isset" is syntaxed.

**Additional context**
> I'm pretty sure they are all considered the same (isset, file_get_contents, json_decode, etc...), so I feel like they should all be the same color no?

joshgoebel:
There  (`isset`, etc.) are reserved words. https://www.php.net/manual/en/reserved.keywords.php

_PHP does not highlight built-ins_ such as `file_get_contents` or `json_decode`.

If there is a full list of built-ins available I'd be curious to see it, but I worry it might be too large (since PHP is also included in our `common` build). I'd prefer to perhaps see the `blahBlah()` pattern detected instead and flagged as a generic function call. See #2500

joshgoebel:
We need to get the details here on how long a full list would be and see if anyone is willing to work on generic function dispatch detection for PHP.  If neither of these seems to be viable then this will be auto-closed at some point as "expected behavior" - with #2500 tracking the overall issue (of improved fidelity across all grammars).

Ayesh:
Here is a list: https://www.php.net/manual/en/indexes.functions.php

I don't think highlighting built-in functions at the cost of storing this many functions is worth it. I think even the built-in interfaces list is a far fetch to be honest.

joshgoebel:
> I don't think highlighting built-in functions at the cost of storing this many functions is worth it. 

Yikes. Fully agree.

> I think even the built-in interfaces list is a far fetch to be honest.

You mean the `built_in` list we already have?  I might agree with you (with only a quick glance). Are these things not likely to show up in actual PHP code or snippets?  These lists don't exist solely for highlighting purposes but also to allow us to "autodetect" if a given snippet of code is PHP or not, etc...

But a list of pure interfaces is no help at all if interfaces aren't commonly named in actual PHP code...

[...]

PR #2997, opened 2021-02-11
[PHP] Updates to keep up with PHP 8

Ayesh:
At the moment, highlight.js is a bit behind the major changes in PHP 8.0 and upcoming syntax in PHP 8.1. 
 - PHP 8.0 adds a new [`WeakMap` built-in class](https://php.watch/versions/8.0/weakmap) class.
 - Add `trait` to list of class-like naming patterns. This was supported since PHP 5.4.
 - PHP 8.1 adds support for Enums with [`enum` syntax](https://php.watch/versions/8.1/enums#enum-syntax). Adding `enum` keyword.
 - PHP 8.0 adds a new built-in interface called [`Stringable`](https://php.watch/versions/8.0/stringable).
 - PHP 8.0 match expression throws an [`UnhandledMatchError` exception](https://php.watch/versions/8.0/match-expression#UnhandledMatchError)

This PR updates the PHP keywords list to with these changes. Please note that PHP 8.1 is not yet released, and the Enum syntax is not merged to PHP source yet. The voting is in favor, and it will likely be merged on after the [vote](https://wiki.php.net/rfc/enumerations) ends next week. 

### Changes
 - Add `WeakMap`, `Stringable`, `UnhandledMatchError` to built-in class and interface list. 
 - Update keywords list to include `mixed`. PHP also supports `static` keyword as return/paremter types, but `static` is already in the keywords list. 
 - Add `trait` and `enum` to class-like name matching list.

### Checklist

### Screenshots

image


test/detect/php/default.txt
```diff
@@ -48,6 +48,16 @@ enum Foo: string {
     case Test = 'test';
 }
 
+function read(mixed $key): string {
```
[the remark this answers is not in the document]
joshgoebel: IE: See the failing tests "AssertionError: default.txt should be detected as php, but was zephir".

This is likely because you added code here that "looks like" Zephir (which is very close to PHP).

I'd suggest reverting this part of your changes and seeing if that makes tests green.

joshgoebel (commented):
> Please note that PHP 8.1 is not yet released, and the Enum syntax is not merged to PHP source yet. The voting is in favor, and it will likely be merged on after the vote ends next week.

So then we probably hold this PR at least until the end of the voting.

---

Looks like you still have a problem with the markup tests failing... is it possible you've added code that hits an "illegal"? (which short circuits all the highlighting?)  That's what it looked like just at a glance...

Are your tests passing locally?

joshgoebel:
Thanks for the great write-up and clean commit history.  Very nice.

SUPPORTED_LANGUAGES.md
```diff
@@ -136,7 +136,7 @@ Languages that listed a **Package** below are 3rd party languages and are not bu
 | Oracle Rules Language   | ruleslanguage          |         |
 | Oxygene                 | oxygene                |         |
 | PF                      | pf, pf.conf            |         |
-| PHP                     | php, php3, php4, php5, php6, php7 |    |
+| PHP                     | php, php3, php4, php5, php6, php7, php8 |    |
```
joshgoebel, line 139: Curious: Does PHP use `php8` or `php5` as file extensions?   I'm not sure why we need adding version numbers here... (I'm not questing THIS particular add, just the pattern)  

We don't do anything different for different versions of PHP. I'm thinking with version 11 all these need to disappear and just have the single "php"? Do you know anything that would suggest otherwise?

Ayesh: I'd be hugely in favor to get rid of all version-specific extensions.

As far as I remember, the `php5` extension/mime-type was used in cPanel default configuration options. I don't think anybody continues this pattern anymore. There is no PHP 6 at all either. I think this list was just kept updated by incrementing just this number, but not dropping anything in the name of backwards compatibility?

test/detect/php/default.txt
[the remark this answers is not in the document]
Ayesh: I will try this, thank you. I used that particular snippet because it has the new `mixed` type in it, but I think at least for the `default.txt` file, the simpler, the better.

Ayesh:
Thanks a lot for the review @joshgoebel - I'm so grateful for these insights. I will work on the failing tests. They fail locally (node 15.2) for me as well. I'm not 100% sure what keeps them from failing, but I will take a thorough look.

SUPPORTED_LANGUAGES.md
joshgoebel: Before my time.  This is a potential breaking change though, so we're stuck with it for now.  But I will add it to the list for v11 changes.

joshgoebel:
Try commenting out all the `illegal` keys and see if it starts working... if so that's a clue.

test/detect/php/default.txt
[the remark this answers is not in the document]
joshgoebel: `default.txt` should in practice be a "clear, representative sample" of a language... no more, no less.  So perhaps they should change indeed slowly over time, but the current system makes it frustrating to do so.

The current auto-detect system is all relative to other languages vs being more yes/no... if it was a more isolated and binary decision such as "this is PHP" or "this is not" these tests would make a lot more sense.

Our sample size is really too small for what we're attempting to do with it anyways - so it's become a very artificial thing.

Ayesh:
Thank you, tinkering with `illegal` key did the trick! There was a `[:($"]` regex there, and it was ruling out the potential `enum Foo: string` syntax because of the `:` sign. I relaxed the illegal regex to allow `:` characters. Triggering CI because it passed locally. 

I will update with a negative look behind to tune the regex there.

joshgoebel:
> I will update with a negative look behind to tune the regex there.

Can't use negative look-behinds, they aren't supported by all browsers yet.

Ayesh:
I understand, thank you.

I can't think about a way to make the `illegal` rule more strict. I think allowing `:` character is worth the relaxed `illegal` rule to allow `enum Foo: string {}` syntax. 

> So then we probably hold this PR at least until the end of the voting.
Yes. I will remove the draft status from this PR on Feb 17, that's when the vote should end and code gets officially merged. 

Thanks a lot for shedding the light throughout this PR @joshgoebel.

src/languages/php.js
```diff
@@ -156,11 +156,11 @@ export default function(hljs) {
       },
       {
         className: 'class',
-        beginKeywords: 'class interface',
+        beginKeywords: 'class interface trait enum',
```
joshgoebel, line 159: Try using `variants` here, something like:

```
variants: [
  { beginKeywords: "enum", illegal: [reduced set] },
  { beginKeywordS: "class interface trait", illegal: [full set] }
]
```

It's a shortcut to writing all new rules (which is what the mode compiler does internally).. All the other keys (from the top) are inherited.

Ayesh:
Brilliant, that was great idea. Tests should pass now too. Thank you.

joshgoebel:
You seem to know your PHP, perhaps you could add your thoughts to: https://github.com/highlightjs/highlight.js/issues/2996 if you have time.

Ayesh:
@joshgoebel Spot on with the `beginKeywordS`. I force-pushed with it fixed.

joshgoebel (approved):
Pending making sure PHP 8.1 goes as planned.

Ayesh:
I added a new commit that expands our number detection. It now supports all numerals supported up to PHP 8.1.
 - Decimal numbers
 - Floats
 - Binary numbers (`0b1100` and `0B1100`).
 - Hex number support (`0xAFAF` and `0XAFAF`).
 - Octal number support (`0777`)
 - [New `0o` and `0O` prefix support for Octals](https://php.watch/versions/8.1/explicit-octal-notation) (`0o777` and `0O777`) (Already accepted and merged to upcoming PHP 8.1)
 - Scientific notation (`7E-10` and `1.2e3`)
 - [Underscore number separator](https://php.watch/versions/7.4/underscore_numeric_separator) (Already implemented in PHP 7.4)


image

Thank you.

src/languages/php.js
```diff
@@ -60,7 +60,13 @@ export default function(hljs) {
       HEREDOC
     ]
   };
-  const NUMBER = {variants: [hljs.BINARY_NUMBER_MODE, hljs.C_NUMBER_MODE]};
+  const NUMBER = {
+    className: 'number',
+    variants: [
+      { begin: `\\b0b[01]+(?:_[01]+)*\\b|\\b0o[0-7]+(?:_[0-7]+)*\\b|\\b0x[\\da-f]+(?:_[\\da-f]+)*\\b|(?:\\b\\d+(?:_\\d+)*(\\.(?:\\d+(?:_\\d+)*))?|\\B\\.\\d+)(?:e[+-]?\\d+)?` },
```
joshgoebel, line 66: Nice.

Please expand this into multiple variants for each variant of number vs a huge regex that tends to be impossible to read and maintain over time.

Ayesh: Thank you. It's now expanded into four variants - binary, octal, hex, and decimal.

Ayesh:
I think this PR will creep up with more changes, so I will wait for the Enum vote to finish on Feb 17, and then take out this PR from the draft state. Once these changes are made, I will see to open another with the only two missing syntax changes:

So far, the two other features we could improve are:
 - Attributes (`#[Foo(Bar)]`): They are currently matched as a comment because PHP also follows hash-comment syntax. 
 - Named parameters: (`call_foo(foo: 1, bar: 2)`). I could get attributes to work, but I will do some more testing with named parameters.

joshgoebel:
> I think this PR will creep up with more changes

Personally I'd rather see this frozen and then another PR opened (it can be on top of this commit history, and rebased later) - but I'm also trying to be flexible since you're one of the cleanest committers I've ever seen here. :-)  Perhaps the real issue is my rush to review things, lol.  Don't get a lot of draft PRs.

> Attributes (#[Foo(Bar)])

That would be a great and easy add I think... would these be `meta` then?

> Named parameters: (call_foo(foo: 1, bar: 2)).

This involves detecting `functionDispatch(` style patterns, most likely, yes?  If so we should highlight the dispatch also (function name), though there is some question on what class we should see. See #2500, etc.  But if you just slap something on it I can sort out the proper class afterwards.

Ayesh:
Hi @joshgoebel - thanks for the help with attributes and named parameters. You are right, I think it's good to go with "meta" for Attributes. Rust seems to get same treatment too.

---

So far, the only change that is not yet landed on PHP is Enums. All other features are already released or merged. I rebased off #3000, and applied all other patches except a couple related to Enums. I will send a separate PRs for Enums, Attributes, and Named parameters. 

Thank you.

Ayesh:
Taking this PR out of draft state, because all changes are already landed in PHP, and all changes except octal numerals are working in the already released PHP 8.0. 
Fingers crossed :)

joshgoebel (approved):

joshgoebel:
@Ayesh Thanks so much!

Ayesh:
Yay, thank you.

PR #3001, opened 2021-02-14
fix(cpp) constructor support for initializers

### LINKED ISSUE #1980: (cpp) constructor does not support initializers
```js
highlight('c++', 'void Do() {}').value
// ''<span class="hljs-function"><span class="hljs-keyword">void</span> <span class="hljs-title">Do</span><span class="hljs-params">()</span> </span>{}', highlight OK

highlight('c++', 'A(): a(10) {}').value
// 'A(): a(<span class="hljs-number">10</span>) {}', only highlight number

highlight('c++', 'explicit A(): a(10) {}').value
// 'explicit A(): a(10) {}', not highlight
```

joshgoebel:
Resolves #1980.

### Changes

Detects the `:` for an initializer sequence and highlights strings and
numbers that appear in the sequence.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@ New Languages:
 Language grammar improvements:
 
 - enh(gml) Add additional GML 2.3 keywords (#2984) [xDGameStudios][]
-- fix(cpp) constructor support for initializers () [Josh Goebel][]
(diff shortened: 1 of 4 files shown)


egor-rogov (approved):
Looks good

PR #3003, opened 2021-02-14
enh(parser) add highlightElement, deprecate highlightBlock

joshgoebel:
More work towards #2277.

### Changes

- adds `highlightElement(el)` API and new plugin callbacks
- deprecates old `highlightBlock(el)` API and old callbacks
- both sets of plugin callbacks will be supported until v12 

Supporting both sets of plugin callbacks should allow plugins to be written that easily work with both v10 and v11 (for a smooth transition for all).

Old API will continue to be supported until v12.

- [x] Make sure API isn't double called if a plugin defines both for compatibility

### Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -90,6 +90,9 @@ module.exports = {
     },
     {
       files: ["test/**/*.js"],
+      globals: {
+        should: "readonly"
+      },
       env: {
         mocha: true
       },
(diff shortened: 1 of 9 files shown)


allejo (approved):

driesvints:
@joshgoebel could it be that this isn't reflected in the docs yet? `highlightBlock` isn't deprecated there yet and there's no docs on `highlightElement`. I'm seeing the below warnings but have no idea how to use the new `highlightElement`.

Screenshot 2021-05-26 at 11 46 50

joshgoebel:
The docs weren't building correctly.  I fixed.

https://highlightjs.readthedocs.io/

Thanks.  The API is the same for that method, only the name changed.

driesvints:
@joshgoebel thanks for checking in on that! :)

Issue #2973, opened 2021-01-24
Inconsistent syntax highlighting for C funcitons in the same stdio.h header


drankinatty:
**Inconsistent Syntax Highlight for C functions from stdio.h**

On StackOverflow when answering questions, highlightjs fails to apply consistent syntax highlighting to many functions, but this is particularly acute when it applies (or fails to apply) consistent syntax highlighting to functions from the same standard header `stdio.h` For example:

The functions `fputs()` and `puts()` are highlighted, but `fopen()`, `fgetc()` and `fclose()` all have no highlight applied at all. The is confusing to new programmers wondering why some functions are one color and some are not.

**This applies to C, but is applicable to C++ as well**

Here is the link to the answer: [Strncat causes exception](https://stackoverflow.com/a/65865974/3422102)

**Are you using `highlight` or `highlightAuto`?**

**Sample Code to Reproduce**
```c
#include <stdio.h>

#define MAXC 100        /* if you need a constant, #defien one (or more) */

int main (int argc, char* argv[]) {
    
    if (argc < 2) { /* validate one argument give for filename */
        fputs ("error: too few arguments.\n", stderr);
        return 1;
    }
    char buf[MAXC] = "", c;
    size_t n = 0;
    FILE* fp = fopen (argv[1], "r");

    if (fp == NULL) {                   /* validate file open for reading */
        perror ("fopen-argv[1]");       /* on failure, perror() tells why */
        return 1;
    }

    /* while buf not full (saving 1-char for \0), read char */
    while (n + 1 < MAXC && (c = fgetc(fp)) != EOF)
        buf[n++] = c;                   /* assign char to next element in buf */

    buf[n] = 0;                         /* nul-terminate buf */
    
    puts (buf);                         /* output result */
    
    fclose(fp);
}

```

**Expected behavior**

All functions, especially from the same standard header `stdio.h` should have consistent highlight applied.

drankinatty:
Why does the code posted here have consistent highlight of `stdio.h` functions while that on StackOverflow does not?

drankinatty:
See also bug filed on [Meta-StackOverflow - Why is our highlightjs syntax highlight inconsistent while Code filed with bugs to highlightjs on GitHub is consistent?](https://meta.stackoverflow.com/questions/404664/why-is-out-highlightjs-syntax-highlight-inconsistent-while-code-filed-with-bugs)

joshgoebel:
[...]

PR #3005, opened 2021-02-14
enh(cpp) Highlight all function dispatches

joshgoebel:
Work towards resolving #2973.

Related: #2500.

### Changes

- Highlights all function dispatch as `function.dispatch`, which aliases
to `built_in` for now

### Checklist


allejo:
Should `_` be a keyword for an "only used for relevancy" list (or was this just a temporary name)? And for the Arduino grammar, was your split of `built_in` keywords arbitrary was there some meaning to how you split them?

joshgoebel:
> Should _ be a keyword for an "only used for relevancy" list (or was this just a temporary name)? 

It's a convention, though I also considered `_*` as in `_builtins`, `_something`, `_whatever` to allow for better naming.

> And for the Arduino grammar, was your split of built_in keywords arbitrary was there some meaning to how you split them?

It's split cleanly between built-in classes (still highlighted as keywords) and built-in functions (highlighted via the `blahBlah(` rule now.

joshgoebel:
Ping.

allejo:
> It's a convention now (hence the patch to avoid emitting HTML for them), though I also considered `_*` as in `_builtins`, `_something`, `_whatever` to allow for better naming.

Yea I noticed the patch in the core changes. I'm not convinced that `_` is the best name for this. I'd lean towards using something like `_relevancy`. Then in the future, `_*` could be used for future internal or special behavior stuff.

joshgoebel:
>  Then in the future, _* could be used for future internal or special behavior stuff.

I like free naming too much so I codified `_*` and added it to docs. I do think it's an improvement from the magic `_` key.  We always have `__`  or `$` or all sort of options for adding future functionality including passing an object:

```js
keywords: {
  delicious: { extraSpecialHighlighting: true, list: "for if def when" }
}
```

I don't want to worry too much about saving the namespace for the future.

src/languages/cpp.js
```diff
@@ -120,19 +236,27 @@ export default function(hljs) {
...
-      'auto_ptr deque list queue stack vector map set pair bitset multiset multimap unordered_set ' +
-      'unordered_map unordered_multiset unordered_multimap priority_queue make_pair array shared_ptr abort terminate abs acos ' +
-      'asin atan2 atan calloc ceil cosh cos exit exp fabs floor fmod fprintf fputs free frexp ' +
-      'fscanf future isalnum isalpha iscntrl isdigit isgraph islower isprint ispunct isspace isupper ' +
-      'isxdigit tolower toupper labs ldexp log10 log malloc realloc memchr memcmp memcpy memset modf pow ' +
-      'printf putchar puts scanf sinh sin snprintf sprintf sqrt sscanf strcat strchr strcmp ' +
-      'strcpy strcspn strlen strncat strncmp strncpy strpbrk strrchr strspn strstr tanh tan ' +
-      'vfprintf vprintf vsprintf endl initializer_list unique_ptr _Bool complex _Complex imaginary _Imaginary',
+    built_in: '_Bool _Complex _Imaginary',
+    _relevance_hints: COMMON_CPP_HINTS,
```
allejo, line 240:
```suggestion
    _relevance_only: COMMON_CPP_HINTS,
```

I know it doesn't matter but just to match the docs

allejo (approved):

src/languages/cpp.js
joshgoebel: I prefere hints (of the two), but it's not a magic phrase - they don't have to match.

PR #3007, opened 2021-02-15
(chore) Builds ES and CJS modules side by side with conditional exports

joshgoebel:
Researching #2600.

### Changes

- Builds ES and CJS modules side by wide
- Uses conditional exports

### Checklist

### DIFF
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -117,6 +117,16 @@ async function buildLanguages(languages) {
   log("");
 }
 
+const CORE_FILES = [
+  "LICENSE",
+  "README.md",
+  "VERSION_10_UPGRADE.md",
+  "SUPPORTED_LANGUAGES.md",
+  "SECURITY.md",
+  "CHANGES.md",
+  "types/index.d.ts"
+]
+
 async function buildNode(options) {
   mkdir("lib/languages");
   mkdir("es/languages");
@@ -124,10 +134,14 @@ async function buildNode(options) {
   mkdir("styles");
   mkdir("types");
 
-  install("./LICENSE", "LICENSE");
-  install("./README.md", "README.md");
-  install("./types/index.d.ts", "types/index.d.ts");
-  install("./src/core.d.ts","lib/core.d.ts");
+
+  CORE_FILES.forEach(file => {
+    install(`./${file}`, file);
+  });
+  install("./src/core.d.ts", "lib/core.d.ts");
+  install("./src/core.d.ts", "es/core.d.ts");
+  install("./src/core.d.ts", "lib/common.d.ts");
+  install("./src/core.d.ts", "es/common.d.ts");
 
   log("Writing styles.");
   const styles = await fs.readdir("./src/styles/");
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -10,9 +10,15 @@ async function buildESMIndex(name, languages) {
   const header = `import hljs from './core.mjs';`;
   const footer = "export default hljs;";
 
+  const safeImportName = (s) => {
+    s = s.replace(/-/g, "_");
+    if (/^\d/.test(s)) s = `L_${s}`;
+    return s;
+  };
+
   const registration = languages.map((lang) => {
     let out = '';
-    const importName = "L_" + lang.name.replace("-","_")
+    const importName = safeImportName(lang.name);
     let require = `import ${importName} from './languages/${lang.name}.js';`;
     // TODO: break this with v11? All modules must export default?
     if (lang.loader) {
@@ -125,7 +131,7 @@ const CORE_FILES = [
   "SECURITY.md",
   "CHANGES.md",
   "types/index.d.ts"
-]
+];
 
 async function buildNode(options) {
   mkdir("lib/languages");
@@ -156,11 +162,12 @@ async function buildNode(options) {
   // filter languages for inclusion in the highlight.js bundle
   languages = filter(languages, options.languages);
 
+  const common = languages.filter(l => l.categories.includes("common"));
   await buildESMIndex("index", languages);
-  await buildESMIndex("common", languages.filter(l => l.categories.includes("common")));
+  await buildESMIndex("common", common);
   await buildESMUtils();
   await buildCJSIndex("index", languages);
-  await buildCJSIndex("common", languages.filter(l => l.categories.includes("common")));
+  await buildCJSIndex("common", common);
   await buildLanguages(languages);
 
   log("Writing highlight.js");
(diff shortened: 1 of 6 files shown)


aduh95:
You need to change 
https://github.com/highlightjs/highlight.js/blob/6af784fe6b7786db69df3c20848e178b75f712da/package.json#L23
to use the `.mjs` instead (or keep using the `.js` extension and add `"type":"module"` to the `package.json`).

tools/build_node.js
```diff
@@ -6,7 +6,35 @@ const { filter } = require("./lib/dependencies");
...
+    return s;
+  };
+
+  const registration = languages.map((lang) => {
+    let out = '';
+    const importName = safeImportName(lang.name);
+    let require = `import ${importName} from './languages/${lang.name}.js';`;
+    // TODO: break this with v11? All modules must export default?
+    if (lang.loader) {
+      require = require += `.${lang.loader}`;
```
aduh95, line 25: I'm not sure to understand what this is doing? If you are not updating `out`, it is simply a no-op right?
```suggestion
      require += `.${lang.loader}`;
```

tools/build_node.js
```diff
@@ -18,30 +46,69 @@ async function buildNodeIndex(languages) {
...
 async function buildPackageJSON() {
   const packageJson = require("../package");
+
+  const exports = {
+    ".": dual("./lib/index.js"),
+    "./lib/common": dual("./lib/common.js"),
+    "./lib/core": dual("./lib/core.js"),
+    "./lib/languages/*": dual("./lib/languages/*.js"),
+  };
+  packageJson.exports = exports;
```
aduh95, line 110: From the Node.js documentation

> Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. require('your-package/package.json'). **This will likely be a breaking change.**

Should we add `package.json` to the export map? I'm not using it myself, but I wanted to make sure this was considered before landing in case it is useful for someone else.

joshgoebel: It would just need a single entry to itself, not dual, right?

aduh95: Yes, there is only one JSON format thankfully 😅

tools/build_node.js
joshgoebel: Ugh yeah, this didn't come across right - but I'm just going to remove those bits entirely anyways.

tools/build_node.js
```diff
@@ -6,42 +6,102 @@ const { filter } = require("./lib/dependencies");
...
+  };
+
+  const registration = languages.map((lang) => {
+    let out = '';
+    const importName = safeImportName(lang.name);
+    const require = `import ${importName} from './languages/${lang.name}.js';`;
+    out += require;
+    out += `hljs.registerLanguage('${lang.name}', ${importName});`;
+    return out;
+  });
```
aduh95, line 26: nit: I think it's actually more readable without the variables
```suggestion
  const registration = languages.map((lang) => 
    `import ${safeImportName(lang.name)} from './languages/${lang.name}.js';` +
    `hljs.registerLanguage('${lang.name}', ${importName});`);
```

joshgoebel:
Good?

tools/build_node.js
```diff
@@ -41,10 +41,28 @@ async function buildCJSIndex(name, languages) {
...
 
 async function buildNodeLanguage(language, options) {
+  const EMIT = `function emitWarning() {
+    if (!emitWarning.warned) {
+      emitWarning.warned = true;
+      process.emitWarning(
+        'Using file extension in specifier is deprecated, use "highlight.js/lib/languages/%%%%" instead'
+      );
+    }
+  }
```
aduh95, line 51: Alternatively you could put this in a separate module and `import` it in the `.js.js` files: that would reduce the size of the npm package, and only one warning would be printed out.

tools/build_node.js
```diff
@@ -41,10 +41,28 @@ async function buildCJSIndex(name, languages) {
 }
 
 async function buildNodeLanguage(language, options) {
+  const EMIT = `function emitWarning() {
+    if (!emitWarning.warned) {
+      emitWarning.warned = true;
+      process.emitWarning(
+        'Using file extension in specifier is deprecated, use "highlight.js/lib/languages/%%%%" instead'
+      );
```
aduh95, line 49: We might want to add a `type` parameter:

> The following additional handling is implemented if the warning `type` is `'DeprecationWarning'`:
>
> - If the `--throw-deprecation` command-line flag is used, the deprecation warning is thrown as an exception rather than being emitted as an event.
> - If the `--no-deprecation` command-line flag is used, the deprecation warning is suppressed.
> - If the `--trace-deprecation` command-line flag is used, the deprecation warning is printed to stderr along with the full stack trace.

```suggestion
      process.emitWarning(
        'Using file extension in specifier is deprecated, use "highlight.js/lib/languages/%%%%" instead',
        'DeprecationWarning'
      );
```

tools/build_node.js
joshgoebel: I like the warning per file.  If someone has 5 broken requires, they should see 5 warnings. (ie, per call site)

The size difference isn't huge.

tools/build_node.js
```diff
@@ -6,85 +6,192 @@ const { filter } = require("./lib/dependencies");
...
+  await fs.writeFile(`${process.env.BUILD_DIR}/lib/${name}.js`, index);
 }
 
-async function buildNodeLanguage(language) {
+async function buildNodeLanguage(language, options) {
+  const EMIT = `function emitWarning() {
+    if (!emitWarning.warned) {
+      emitWarning.warned = true;
+      process.emitWarning(
+        'Using file extension in specifier is deprecated, use "highlight.js/lib/languages/%%%%" instead',
```
aduh95, line 48:
```suggestion
        'Using file extension in specifier is deprecated, use "highlight.js/lib/languages/%%%%" instead of "highlight.js/lib/languages/%%%%.js"',
```

tools/build_node.js
```diff
@@ -6,85 +6,192 @@ const { filter } = require("./lib/dependencies");
...
+        'DeprecationWarning'
+      );
+    }
+  }
+  emitWarning();`;
+  const CJS_STUB = `${EMIT}
+    module.exports = require('./%%%%.js');`;
+  const ES_STUB = `${EMIT}
+    import lang from './%%%%.js';
+    export default lang;`;
```
aduh95, line 58: nit: you could use a more concise form here.
```suggestion
    export { default } from './%%%%.js';`;
```

joshgoebel: That looks less clear to me.  Possibly just my being unfamiliar with the syntax. Looks like it would export `{ default ` instead of `default`.

PR #3009, opened 2021-02-15
ehn(api): add unregisterLanguage method

aduh95:
### Changes

Currently, there is no way of un-do a `registerLanguage` operation except re-loading the whole library. Adds a `unregisterLanguage` method to the public API.

### Checklist
- [x] No markup tests, this shouldn't change anything to the markup

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -821,6 +821,11 @@ const HLJS = function(hljs) {
    */
   function unregisterLanguage(languageName) {
     delete languages[languageName];
+    for(const alias of Object.keys(aliases)) {
+      if(aliases[alias] === languageName) {
+        delete aliases[alias];
+      }
+    }
   }
 
   /**
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -815,6 +815,15 @@ const HLJS = function(hljs) {
   }
 
   /**
+   * Remove a language grammar module
+   *
+   * @param {string} languageName
+   */
+  function unregisterLanguage(languageName) {
+    delete languages[languageName];
+  }
+
+  /**
    * @returns {string[]} List of language internal names
    */
   function listLanguages() {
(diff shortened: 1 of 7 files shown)


joshgoebel:
Curious: What is your use case for needing to unregister a language?

joshgoebel:
This would also need to be documented in `api.rst`.

src/highlight.js
```diff
@@ -814,6 +814,15 @@ const HLJS = function(hljs) {
     }
   }
 
+  /**
+   * Remove a language grammar module
+   *
+   * @param {string} languageName
+   */
+  function unregisterLanguage(languageName) {
+    delete languages[languageName];
```
joshgoebel, line 823: Technically do we need to clean up the `alias` table also?

joshgoebel (changes requested):
As this would be a core API method we'd also probably want at least one test to show that it indeed does what it's expected to do.

aduh95:
> Curious: What is your use case for needing to unregister a language?

Useful for testing, in the test suite we have to remove the module from the cache to remove the test languages added by the other tests:

https://github.com/highlightjs/highlight.js/blob/742e87b4c512faaf8118323e3192aa94628e262f/test/detect/index.js#L3-L4

This could be avoided if the other tests were cleaning up their test languages. This is needed to move the test suite to ESM syntax (https://github.com/joshgoebel/highlight.js/pull/1).

src/highlight.js
```diff
@@ -821,6 +821,11 @@ const HLJS = function(hljs) {
    */
   function unregisterLanguage(languageName) {
     delete languages[languageName];
+    for(const alias of Object.keys(aliases)) {
+      if(aliases[alias] === languageName) {
+        delete aliases[alias];
```
joshgoebel, line 826: You'll probably want to install an ESLint helper in your editor of choice.

joshgoebel:
> This is needed to move the test suite to ESM syntax

Ah, that makes sense since we can't just delete the require cache anymore. :)

joshgoebel (approved):

joshgoebel:
@aduh95 Thanks!

PR #3026, opened 2021-02-28
fix(parser) Make language alias registration case insensitive

davido:
### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,7 +29,7 @@ Deprecations:
 API:
 
 - enh(api) add `unregisterLanguage` method (#3009) [Antoine du Hamel][]
-- Make language registration case insensitive [David Ostrovsky][]
+- enh: Make alias registration case insensitive (#3026) [David Ostrovsky][]
 
 [Stef Levesque]: https://github.com/stef-levesque
 [Josh Goebel]: https://github.com/joshgoebel
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,7 @@ Deprecations:
 API:
 
 - enh(api) add `unregisterLanguage` method (#3009) [Antoine du Hamel][]
+- Make language registration case insensitive [David Ostrovsky][]
 
 [Stef Levesque]: https://github.com/stef-levesque
 [Josh Goebel]: https://github.com/joshgoebel
@@ -38,6 +39,7 @@ API:
 [Vyron Vasileiadis]: https://github.com/fedonman
 [Antoine du Hamel]: https://github.com/aduh95
 [Vaibhav Chanana]: https://github.com/il3ven
+[David Ostrovsky]: https://github.com/davido
 
 ## Version 10.6.0
 
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -875,7 +875,7 @@ const HLJS = function(hljs) {
     if (typeof aliasList === 'string') {
       aliasList = [aliasList];
     }
-    aliasList.forEach(alias => { aliases[alias] = languageName; });
+    aliasList.forEach(alias => { aliases[alias.toLowerCase()] = languageName; });
   }
 
   /**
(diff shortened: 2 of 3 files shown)


davido:
Upstream discussion is here: [1].

[1] https://crbug.com/gerrit/14091

joshgoebel:
This will not fix: `Could not find the language 'latex', did you forget to load/include a language module?`

The problem is that your installation of Highlight.js is not loading the `latex` language module (as the error is telling you).  Adding all the aliases in the world will not fix the issue.  Latex is not including in the "common" distributable so you'll need to build a custom build or add additional JS files (see our README for directions).

It's possible you think you've fixed it locally because you're using a different build of the library than is being used in production?  But that error would not be shown if latex was loaded and registered.

---

This might still be a reasonable PR but I'm going to briefly noodle over whether we simply shouldn't make aliases case-insensitive rather than having this type of issue continue to come up.

davido:
> This will not fix: Could not find the language 'latex', did you forget to load/include a language module?

I have not experienced this issue.

I debugged gerrit@HEAD and turned out, that this function

```javascript
      /**
       * @param {string} name - name of the language to retrieve
       * @returns {Language | undefined}
       */
      function getLanguage(name) {
        name = (name || '').toLowerCase();
        return languages[name] || languages[aliases[name]];
      }
```

returned undefined for `tex` language name.

> The problem is that your installation of Highlight.js is not loading the latex language module (as the error is telling you). Adding all the aliases in the world will not fix the issue. Latex is not including in the "common" distributable so you'll need to build a custom build or add additional JS files (see our README for directions).

Gerrit build instruction for hljs is [simple](https://gerrit.googlesource.com/gerrit/+/refs/heads/master/lib/highlightjs/building.md#steps-to-create-the-pack-file):

```
    $>  # start in some temp directory
    $>  git clone https://github.com/highlightjs/highlight.js
    $>  cd highlight.js
    $>  patch -p1 < ../0001-Add-tex-alias-to-latex-mode.patch
    $>  git clone https://github.com/highlightjs/highlightjs-closure-templates
    $>  ln -s ../../highlightjs-closure-templates/soy.js src/languages/soy.js
    $>  mkdir test/detect/soy && ln -s ../../../highlightjs-closure-templates/test/detect/soy/default.txt test/detect/soy/default.txt
    $>  npm install
    $>  node tools/build.js
```

Now, grepping `latex` produce these matches:

```
  $ grep -i -latex build/highlight.min.js
  hljs.registerLanguage("latex",(()=>{"use strict";return e=>{const n=[{
  });return{name:"LaTeX",aliases:["tex","TeX"],
```

And the `LaTeX` sources with `.tex` extension just work.

What am I missing?

joshgoebel:
> I have not experienced this issue.

Ah, ok we're talking about different issues then.  

> [`getLanguage`] returned undefined for tex language name.

Ah yes, thank for running this down. :-) 

> `name = (name || '').toLowerCase();`

Ugh. So we're already normalizing aliases at request/runtime, but not at registration time.  I think the fix we truly need here is on line ~878 instead:

```js
aliasList.forEach(alias => { aliases[alias.toLowerCase()] = languageName; });
```

If the lookup is downcase, then the registration should be also.  Could you confirm if this also resolves your issue and perhaps update the PR if so?

davido:
> Could you confirm if this also resolves your issue

yes

> [...] and perhaps update the PR if so.

Done.

joshgoebel:
Please give yourself some credit in CHANGES.md. :)  Otherwise looks good.

davido:
> Please give yourself some credit in CHANGES.md.

Done.

CHANGES.md
```diff
@@ -29,6 +29,7 @@ Deprecations:
 API:
 
 - enh(api) add `unregisterLanguage` method (#3009) [Antoine du Hamel][]
+- Make language registration case insensitive [David Ostrovsky][]
```
joshgoebel, line 32:
```suggestion
- enh: Make alias registration case insensitive (#3026) [David Ostrovsky][]
```

joshgoebel (approved):

joshgoebel:
@davido Thanks!

PR #3030, opened 2021-03-02
(chore) Fix Authors format

Hirse:
Follow up from #3024, fyi @olcayusta 

### Changes
The entries in the AUTHORS.txt should follow the npm people field short-format as they get included in the package.json.
image

See https://docs.npmjs.com/cli/v7/configuring-npm/package-json#people-fields-author-contributors

### Checklist
- [x] ~Added markup tests~

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -317,4 +317,4 @@ Contributors:
 - davidhcefx <davidhu0903ex3@gmail.com>
 - xDGameStudios <xDGameStudios@gmail.com>
 - Ayesh Karunaratne <ayesh@aye.sh>
-- Olcay Usta <https://github.com/olcayusta>
+- Olcay Usta (https://github.com/olcayusta)


joshgoebel:
How do we prevent people from continuing to do this wrong - or else perhaps the format should not be so strict at all?  Just taking PRs to fix this constantly seems annoying...

olcayusta:
Yes, I missed it. Obviously I added it as it is in the task list. I think this task can be removed. I don't think it will cause a problem.

joshgoebel:
@olcayusta No worries, you aren't in trouble... you're not the first person to get this tiny detail wrong... I just want to figure out how to resolve it more permanently... perhaps we need to update the issue template todo checklist to provide guidance?

allejo:
> How do we prevent people from continuing to do this wrong - or else perhaps the format should not be so strict at all? Just taking PRs to fix this constantly seems annoying...

Here you go, https://github.com/highlightjs/highlight.js/pull/3036

Issue #2890, opened 2020-11-23
Scheme -  inconsistent highlighting


johbra:
The highlighting of the Scheme function
`(define add-point
    (lambda [p1 p2]
      (make-point
        (+ (point-x p1) (point-x p2))
        (+ (point-y p1) (point-y p2)))))`
using the following HTML document
`<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8"/>
        <link rel="stylesheet"
              href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.4.0/styles/default.min.css">
        <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.4.0/highlight.min.js"></script>
        <script charset="UTF-8" src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.4.0/languages/scheme.min.js"></script>
        <script>hljs.initHighlightingOnLoad();</script>
    </head>
    <body>
        <h2>Code example</h2>
        <pre><code class="scheme">(define add-point
  (lambda [p1 p2]
    (make-point
      (+ (point-x p1) (point-x p2))
      (+ (point-y p1) (point-y p2)))))
        </code></pre>
    </body>
</html>
`
yields inconsistent highlighting: The fourth line of the function had to be highlighted exactly as the fifth line. But that's not the case.

joshgoebel:
<img width="667" alt="Screen Shot 2020-11-23 at 9 31 53 PM" src="https://user-images.githubusercontent.com/6473/100039477-5279de80-2dd3-11eb-8573-b7521c55fd88.png">

Strange indeed.

joshgoebel:
May need a language champion for Scheme to sort this out.  This type of syntax just give me a headache. :)

```
  var LAMBDA = {
    className:"lambda",
    begin: /lambda/,
    endsWithParent: true, returnBegin: true,
    contains: [
      NAME,
      {
        begin: /\(/, end: /\)/, endsParent: true,
        contains: [IDENT],
      }
    ]
  };
```

Our grammar expects a lambda to have a "name" and then parens and a list of "idents" (which we don't color).

So it seems that first ( and then does nothing until it finds the first closing ), which snaps it back out of lambda and normal highlighting resumes.

joshgoebel:
Is there another form where you pass the lamba arguments with `()` vs `[]`?

johbra:
[...]

PR #3035, opened 2021-03-03
(scheme) Add 'racket' alias

elihunter173:
### Changes

Add `racket` alias to the scheme language since scheme highlighting is also used for racket. https://github.com/highlightjs/highlight.js/issues/2890

### Checklist
- [x] Added markup tests, or they don't apply here because... Single alias added

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -318,3 +318,4 @@ Contributors:
 - xDGameStudios <xDGameStudios@gmail.com>
 - Ayesh Karunaratne <ayesh@aye.sh>
 - Olcay Usta <https://github.com/olcayusta>
+- Eli W. Hunter <elihunter173@gmail.com>
--- a/SUPPORTED_LANGUAGES.md
+++ b/SUPPORTED_LANGUAGES.md
@@ -172,7 +172,7 @@ Languages that listed a **Package** below are 3rd party languages and are not bu
 | SQL                     | sql                    |         |
 | STEP Part 21            | p21, step, stp         |         |
 | Scala                   | scala                  |         |
-| Scheme                  | scheme                 |         |
+| Scheme                  | scheme, racket         |         |
 | Scilab                  | scilab, sci            |         |
 | Shape Expressions       | shexc                  | [highlightjs-shexc](https://github.com/highlightjs/highlightjs-shexc) |
 | Shell                   | shell, console         |         |
--- a/src/languages/scheme.js
+++ b/src/languages/scheme.js
@@ -198,6 +198,7 @@ export default function(hljs) {
 
   return {
     name: 'Scheme',
+    aliases: [ 'racket' ],
     illegal: /\S/,
     contains: [
       hljs.SHEBANG(),


joshgoebel:
I'm not actually sure this would be good.  From their site:

> Racket is ... a dialect of Lisp and a descendant of Scheme

_But aren't their many descendants of Scheme and dialects of Lisp?_  I'm not sure we want to start adding aliases every time a new variant of scheme pops up (even if our highlighting is compatible).  In a way I see this as scope creep because now it  forces us in a way to *continue* to be Racket compatible, even if that might be an onerous burden in the future. And even if we aren't truly compatible now - but just aren't aware of it yet.

I'd perhaps rather continue to say "we don't support Racket officially, but Scheme works pretty well".  If one is using a scheme-like language it's easy enough to just specify the language as `scheme` OR if that's not enough to add this alias manually with a single line of JS using `registerAlias`.

I'm not sure we want to start codifying this into the core library.

@allejo Thoughts?

joshgoebel:
@elihunter173 Are you aware of us doing something similar with any of our other aliases (perhaps we do)?  If so then perhaps it's time we reviewed the entire policy around this (with v11 around the corner) vs dealing with this as a one-off.

joshgoebel:
Trying to make this less nebulous: As a general rule _I feel an aliases should NOT expand the assumed scope of a grammar._  Aliases should literally be an alias or an equivalency, ie:

- JS is an alias of JavaScript.
- But TypeScript (assume we had no built-in support) is NOT an alias for JavaScript, it's a very different superset.
- Atom is an alias of XML.  Atom is literally XML code.
- Delphi is Object Pascal code, making Pascal a valid alias.
- FreePascal and Lazarus are not aliases of Delphi.  They are not Delphi code, they are other Pascal implementations.
- Zsh is not a bash alias.  Zsh is shell code, but NOT bash code.
- Racket is not a Scheme alias.  Racket is a descendant of Scheme with it's own unique differences.

This reasoning feels pretty consistent to me.

---

- `scheme` may work well, for racket, but that doesn't imply it should be an alias.
- `js` may work well for TypeScript (in a pinch),but that doesn't imply it should be an alias.
- `delphi` may work well for FreePascal, but that doesn't imply it should be an alias.

elihunter173:
That seems totally reasonable to me, after all I won't be the one maintaining this. Feel free to close the PR.

My path to this PR was I wanted to add racket support to my discord bot. Since discord uses highlight.js for syntax highlighting, I match my language codes with the classes of highlight.js. I saw racket wasn't a class but the scheme class was used/modified to support racket and I knew zsh was an alias of bash, so I thought that was the pattern this package followed. I don't have much stake in this PR to be honest, just thought this would be a nice quality of life improvement.

joshgoebel:
> Since discord uses highlight.js for syntax highlighting, I match my language codes with the classes of highlight.js. 

I'd suggest building this mapping into your bot so it knows if it's pasting racket to use scheme.  Probably only a few lines of code (if that).

> I knew zsh was an alias of bash, so I thought that was the pattern this package followed

Indeed, a reasonable deduction.  I'll leave this open a bit longer and it'll likely result in a new issue to fix some of those more ambiguous aliases when we release version 11.

PR #3036, opened 2021-03-03
Add linting for AUTHORS.txt

allejo:
Related to #3030

### Changes

Add a linting option to our CI

### Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,11 +11,17 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        command:
+          - lint
+          - lint-authors
+          - lint-languages
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
         node-version: "*" # latest
     - run: npm ci
-    - run: npm run lint
-    - run: npm run lint-languages
+    - run: npm run ${{ matrix.command }}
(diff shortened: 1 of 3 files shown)


joshgoebel:
Wow.  This is nice, but it makes me wonder if we're going to a bunch of trouble here for no real reason... is it perhaps sufficient to have a list in authors?  Do we really need to add every single person to contributors in package.json that is published to npm?  Could it be that idea was overthought?  Do other libraries do this?  Is anyone going to be offended if we stopped?

Does the linter allow for simple things like:

```
- Josh Goebel <hello@joshgoebel.com> (maintainer)
- Ivan Sagalaev <maniac@softwaremaniacs.org> (original author)
```

joshgoebel:
Seems to.  Can it be enhanced to actually SHOW the content of the line that is causing the validation error, not just its number?  If so I think we can merge these and see how it goes. :)

allejo:
> Do other libraries do this? Is anyone going to be offended if we stopped?

I see a lot of other projects simply point to: [github.com/highlightjs/highlight.js/graphs/contributors](https://github.com/highlightjs/highlight.js/graphs/contributors) and an `AUTHORS.txt` would be reserved for collaborators to the project and not just one-time PR contributors.

joshgoebel:
> I see a lot of other projects simply point to: github.com/highlightjs/highlight.js/graphs/contributors and an AUTHORS.txt would be reserved for collaborators to the project and not just one-time PR contributors.

I could easily to talked into this. :)   Thoughts?

allejo:
I don't really think our AUTHORS file does much other than burden us with manual maintenance. I would argue that being listed in the top 100 contributors (via a native GH feature) to a project would feel better than being on a list that can be updated just for a quick PR.

joshgoebel:
You gonna feel bad about us not needing this PR then?

allejo:
Not one bit. It's one 1 less thing to maintain

PR #3039, opened 2021-03-03
fix(parser) `highlightAll()` now works if the library is lazy loaded

### LINKED ISSUE #3020: Syntax highlighting fails in React applications when using the new highlightAll() method
**Describe the issue**
I am trying to use highlightjs on a React project, which uses Next.js (although I don't think the issue is related). I have added highlight.js using npm (v106.0) and loaded in a style sheet for a theme.

I have a simple function component set up like this:

```
import hljs from "highlight.js";
function Post(props) {
  useEffect(() => {
    hljs.highlightAll();
  }, []);

  return (
    <>
      <pre>
        <code> /* code blocks that i want to highlight */ </code>        
      </pre>
    </>
  );
}
```

This is using the new `highlightAll()` method call, but when the component is loaded the highlighting fails to take effect, and the `<pre><code>` block is left unstyled.

**Expected behavior**
I would expect the code block to be styled and the language highlighted using highlightjs and the stylesheet I have selected.

The error is not language specific, but all code blocks are failing to get instantiated and highlighted, and any styling applied what so ever.

**What I think the issue is**
I think the issue is that the new `highlightAll()` is not compatible with React and the way that React shadow loads components. [Line #772](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L772) makes a check if the dom has loaded, else it breaks out:

```
  function highlightAll() {
    // if we are called too early in the loading process
    if (!domLoaded) { wantsHighlight = true; return; }

    const blocks = document.querySelectorAll('pre code');
    blocks.forEach(highlightElement);
  }
```

The only place that `domLoaded` is set to true is in the [`boot()` method](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L778):
```
  function boot() {
    domLoaded = true;
    // if a highlight was requested before DOM was loaded, do now
    if (wantsHighlight) highlightAll();
  }
```

The problem is that the only place `boot()` is called from is an [event listener](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L785):
```
  if (typeof window !== 'undefined' && window.addEventListener) {
    window.addEventListener('DOMContentLoaded', boot, false);
  }
```

Since React will not cause the fire of the `DomContentLoaded` event when a user navigates to a new page, `boot()` is never called to set this variable to true, and when I am calling `highlightAll()` from my code it's is exiting too early. The `boot()` is also [not exposed](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L960) to be able to call directly from my code.

**Workaround**
For now, the solution is to call `hljs.initHighlighting()` ([Line 747](https://github.com/highlightjs/highlight.js/blob/master/src/highlight.js#L747)) since that method does not make the domLoaded check.

But, this method is marked as deprecated, so a fix needs to be put in place to allow this highlighter to continue to work with React applications. Perhaps allow a `true` parameter to be passed to force it to run the syntax highter code.

joshgoebel:
Previously we hooked `DOMContentLoaded` to keep track of whether the
page had loaded or not, but that does not work if we are lazy loaded
FAR into the future.  In that case we need to check `document.readyState`
and if it's complete then we know that the page has previously been
loaded and we're now good to go.

Should resolve #3020.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## Version 10.7.0 (in progress)
 
+Bugs:
+
+- fix(parser) `highlightAll()` now works if the library is lazy loaded [Josh Goebel][]
+
 New Languages:
 
 - Added 3rd party RiScript grammar to SUPPORTED_LANGUAGES (#2988) [John C][]
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -762,30 +762,29 @@ const HLJS = function(hljs) {
   }
 
   let wantsHighlight = false;
-  let domLoaded = false;
 
   /**
    * auto-highlights all pre>code elements on the page
    */
   function highlightAll() {
     // if we are called too early in the loading process
-    if (!domLoaded) { wantsHighlight = true; return; }
+    if (document.readyState === "loading") {
+      wantsHighlight = true;
+      return;
+    }
 
     const blocks = document.querySelectorAll('pre code');
     blocks.forEach(highlightElement);
   }
 
   function boot() {
-    domLoaded = true;
     // if a highlight was requested before DOM was loaded, do now
     if (wantsHighlight) highlightAll();
   }
 
-  const LOADED_STATES = ["interactive", "complete"];
   // make sure we are in the browser environment
   if (typeof window !== 'undefined' && window.addEventListener) {
     window.addEventListener('DOMContentLoaded', boot, false);
-    if (LOADED_STATES.includes(document.readyState)) { domLoaded = true; }
   }
 
   /**
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -781,9 +781,11 @@ const HLJS = function(hljs) {
     if (wantsHighlight) highlightAll();
   }
 
+  const LOADED_STATES = ["interactive", "complete"];
   // make sure we are in the browser environment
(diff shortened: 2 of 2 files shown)


trm217:
As stated in #3020 I suggest getting rid of the domLoaded variable and instead directly check against the document readystate if the highlighting can be done.

Here is the code with my suggested change:
```js
/**
 * auto-highlights all pre>code elements on the page
 */
function highlightAll() {
  // if we are called too early in the loading process
  if (document.readyState === "loading") { wantsHighlight = true; return; }
  const blocks = document.querySelectorAll('pre code');
  blocks.forEach(highlightElement);
}
function boot() {
  // if a highlight was requested before DOM was loaded, do now
  if (wantsHighlight) highlightAll();
}

// make sure we are in the browser environment
if (typeof window !== 'undefined' && window.addEventListener) {
  window.addEventListener('DOMContentLoaded', boot, false);
}
```

joshgoebel:
@trm217 Agreed, much nicer!

allejo (approved):

trm217:
When can we expect this change to be available on NPM?

joshgoebel:
In the next month or so.  There is no urgency since the immediate (and trivial) fix is just to use the older API until 10.7.

PR #3042, opened 2021-03-05
(chore) Improve 3rd party language quick start guide

jf990:
Proposing some changes to https://github.com/highlightjs/highlight.js/blob/master/extra/3RD_PARTY_QUICK_START.md

I worked out the process while working on https://github.com/highlightjs/highlightjs-curl, so that repo and this doc are in alignment, as well as the proposed template repo https://github.com/jf990/highlightjs-language-template

based on a review here there are a few final changes to complete, but wanted to get some feedback before finishing.

### Changes

Update https://github.com/highlightjs/highlight.js/blob/master/extra/3RD_PARTY_QUICK_START.md to improve the doc.

### Checklist

- [ ] ~~Added markup tests, or they don't apply here because...~~ (doc update)
- [ ] Updated the changelog at `CHANGES.md` TBD


extra/3RD_PARTY_QUICK_START.md
```diff
@@ -9,48 +9,55 @@ Take a look at some of the real-life examples first:
...
+
+- Put your language file in `src/languages/language.js`.
+- Add detect tests in `test/detect/language` (rename `language` to match your language name).
+- Add markup tests in `test/markup/language`.
+- Add a `package.json` file.
+- Include a license.
 
 ## Testing
 
+Going back to your clone of `highlight-js` core repo, create an `extra/language` folder for your language. Copy your repo `src` and `test` folders here.
```
joshgoebel, line 34: Is that what people are doing?  It's much easier to just have/move your own repo's `git checkout` directly inside the extra folder... no need to copy or link anything... I wouldn't recommend copying.

If symlinking the folder into extra works we might recommend that (I haven't tested).

extra/3RD_PARTY_QUICK_START.md
```diff
@@ -9,48 +9,55 @@ Take a look at some of the real-life examples first:
...
 
+Set your repo directory structure to follow exactly the example(s) above.
+
+For example, if you have a `language` language, create your repo directory structure as follows:
+
+- Put your language file in `src/languages/language.js`.
+- Add detect tests in `test/detect/language` (rename `language` to match your language name).
+- Add markup tests in `test/markup/language`.
+- Add a `package.json` file.
+- Include a license.
```
joshgoebel, line 30:
```suggestion
For example, if you your grammar is named `pascal` you would create your repo directory structure as follows:

- Put your grammar file in `src/languages/pascal.js`.
- Add detect tests in `test/detect/pascal/`. 
- Add markup tests in `test/markup/pascal/`.
- Add a `package.json` file.
- Add a LICENSE so people know how your work is licensed.
```

extra/3RD_PARTY_QUICK_START.md
```diff
@@ -9,48 +9,55 @@ Take a look at some of the real-life examples first:
...
+- Checkout a clone of the core repo [highlight-js](https://github.com/highlightjs/highlightjs) from GitHub.
+- Read [Language contributors checklist](https://highlightjs.readthedocs.io/en/latest/language-contribution.html).
 
-- Checkout highlight-js from github...
-- 3rd party languages are placed into the `extra` directory
+## Create a repo
 
-So if you had a `xyz` language you'd create an `extra/xyz` folder, and that would be your language module. All paths below are relative to that.
+Each language is developed in its own repo. This helps keep language definitions and maintainers independent of the core work.
+Determine if you will host the repository yourself or you want it to be part of the [highlightjs organization on GitHub](https://github.com/highlightjs).
```
joshgoebel, line 18: This should be reworded to encourage people to start with their own repo - we can migrate it to our organization later if the user requests.  What happens when we let them decide first is we end up having a few people who never finish their grammar and then we have dead repos in our organization for a long time - and it's hard to decide when they should be removed.  Just awkward.

Would be much nicer to have someone send a link to their repo all shiney and ready to go and then if everything looks good it can be transferred or forked, etc... I think that's the sequencing we'd like to head towards.

CC @allejo

extra/3RD_PARTY_QUICK_START.md
```diff
@@ -59,12 +66,11 @@ Publish it to NPM also if you'd like. This will make it much easier for anyone u
...
 When your language module is ready create a PR that adds it to the `README.md` file in the list of languages.
 
-
 ## The Future
 
 More work could be done on:
 
 - Allowing you to ONLY run your own tests, not the whole suite.
-- Allowing you to maintain a 3rd party module WITHOUT it being inside of a `highlight-js` checkout (this requires discussion though)
+- ~~Allowing you to maintain a 3rd party module WITHOUT it being inside of a `highlight-js` checkout (this requires discussion though)~~
```
joshgoebel, line 74: Did you cross this out because your feel the discussion is over, or?

extra/3RD_PARTY_QUICK_START.md
jf990: I don't know about the git checkout process inside an existing repo, I need to learn how to do that. If that works it solves a major issue I have with this process and how I approached the set up for the template project.

extra/3RD_PARTY_QUICK_START.md
jf990: I agree with your suggestion, but as you also mentioned i think we should use a non-language file here because I wanted it to be the exact file used in the template project, and then very obvious all the files you need to rename. Should we go back to `xyz`?

extra/3RD_PARTY_QUICK_START.md
jf990: related to your comment above, my initial approach was start with a highlightjs-xyz repo firs. but I agree, the better process is start with your own repo first, and transfer to highlightjs after approval. I'll update the docs to promote this process instead of what I had.

extra/3RD_PARTY_QUICK_START.md
joshgoebel: Well git checkout or symlink (for any Unix people)... I never tried symlinking but can't imagine why it wouldn't work.

extra/3RD_PARTY_QUICK_START.md
joshgoebel: `xyz` is ok I think, though a placeholder is easy enough to search/replace...

extra/3RD_PARTY_QUICK_START.md
allejo: I agree with Josh. Having a complete and functional grammar would at least allow hljs maintainers to approve PRs should it be abandoned. But if it's incomplete, then it'd be up to the maintainers to complete it.

jf990: including a reference to the template should make this easier

jf990:
updated and tested inside `extra` as a cloned repo and updated assuming we are going to move forward with the template repo. @joshgoebel @allejo suggestions applied.

extra/3RD_PARTY_QUICK_START.md
```diff
@@ -9,56 +9,69 @@ Take a look at some of the real-life examples first:
...
+```bash
 node ./tools/build.js -t cdn
 
 ...
-Building extra/highlightjs-xyz/dist/xyz.min.js.
+Building extra/highlightjs-your-language/dist/your-language.min.js.
 ...
 ```
 
+Copy `extra/highlightjs-your-language/dist/your-language.min.js` to your repo's dist folder `{your-language-repo}/dist/your-language.min.js`, commit and push it.
```
joshgoebel, line 66: Why?  There should be nothing to copy.  If their repo is under extra then the file is already in the dist folder exactly where it belongs, it only needs to be committed.

joshgoebel:
Under testing we should mention `ONLY_EXTRA` env which will more tightly scope the test to only running against languages inside `extra` rather than all the core languages.  This only works for markup tests, but those are the most common that need to be run over and over.

```
ONLY_EXTRA=true npm run test-markup
```

jf990:
@joshgoebel suggestions are up. thanks!

joshgoebel (approved):

PR #3045, opened 2021-03-12
(python) highlight print and exec as builtins

### LINKED ISSUE #1468: (python) exec and print should be highlighted as built-ins
If I'm using just `print` then this is highlight.
But if I'm using `print()` then this isn't highlight.
\<code class="python highlight">\</code>

samuelcolvin:
fix #1468 

### Changes

Highlight `print` and `exec` as a builtin in python.

### Checklist


joshgoebel (approved):

joshgoebel:
Done and done.  Thanks so much!

PR #3051, opened 2021-03-17
enh(http) support headers without preamble

### LINKED ISSUE #3019: (http) highlight http headers without HTTP preamble
**Describe the issue**
It should be possible to declare http key/value pairs without requiring HTTP preamble.  

For example, just:

```HTTP
Some-header: value
```

HTTP

highlightAuto

...

**Sample Code to Reproduce**

```HTTP
Referer: value
```

**Expected behavior**

Should output:

```HTML
<span class="hljs-attribute">Referer</span>
```

**Additional context**
Appears to be a regression from 10.5. It used to work. Maybe #2893?

joshgoebel:
Resolves #3019.

### Checklist

### DIFF
--- a/test/markup/cpp/function-declarations.expect.txt
+++ b/test/markup/cpp/function-declarations.expect.txt
@@ -21,4 +21,3 @@
 
 <span class="hljs-function"><span class="hljs-keyword">void</span> <span class="hljs-title">A</span><span class="hljs-params">()</span>: a(<span class="hljs-number">10</span>) {</span>}
 <span class="hljs-function"><span class="hljs-keyword">explicit</span> <span class="hljs-title">A</span><span class="hljs-params">()</span>: a(<span class="hljs-number">10</span>) {</span>}
-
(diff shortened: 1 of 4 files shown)


joshgoebel:
@marcoscaceres

test/markup/cpp/function-declarations.expect.txt
```diff
@@ -21,4 +21,3 @@
 
 <span class="hljs-function"><span class="hljs-keyword">void</span> <span class="hljs-title">A</span><span class="hljs-params">()</span>: a(<span class="hljs-number">10</span>) {</span>}
 <span class="hljs-function"><span class="hljs-keyword">explicit</span> <span class="hljs-title">A</span><span class="hljs-params">()</span>: a(<span class="hljs-number">10</span>) {</span>}
```
allejo, line 23:
```suggestion
<span class="hljs-function"><span class="hljs-keyword">explicit</span> <span class="hljs-title">A</span><span class="hljs-params">()</span>: a(<span class="hljs-number">10</span>) {</span>}

```

The C++ markup expected file is being touched in this PR by accident it seems. Not sure if you'd want to squash/clean it up somehow.

allejo (approved):

test/markup/cpp/function-declarations.expect.txt
joshgoebel: This is just fixing a line break issue, so I don't really care.

marcoscaceres:
Thanks for working on this, @joshgoebel!

Issue #1263, opened 2016-07-29
Rethink sublanguages


isagalaev:
There are many subtly different ways of embedding one syntax into another:

- The host language clearly defines boundaries, between which we can use a completely independent highlighting process. It's the case with `<script>..</script>` in HTML.

- The host language defines boundaries, but the closing boundary can only happen at the top level of the sublanguage. This the case of PHP in HTML, where `?>` inside a PHP comment wouldn't end the sublanguage.

- The sublanguage ends when it parses out a single top-level mode. This is the case of JSX.

Current implementation of `subLanguage` only works for the first case. And the `skip` attribute only fixes a few common cases.

Current idea is to implement sublanguages as sub-parsers embedded in the host parser tree.

yyyc514:
> Current idea is to implement sublanguages as sub-parsers embedded in the host parser tree.

This is on my radar for the future... obviously you have some thoughts on this, but I'm not getting much of a feel for your idea with just that one-liner.

And chance when you have some time you might be able to expand that one liner to 2-3 short paragraphs kind of laying out at least your big-picture thoughts (if you can still remember them).

Right now to fix JSX I'm thinking it really needs to not be a sublanguage at all and just an actual part of the JS language.  Because you can do things like each line can either be HTML or a JS comment... so you can't just simply switch to HTML mode for a block... because:

```
<div>
// this is Javascript code, a comment that I can use to annotate the HTML with JS comments, fun fun
</div>
```

So how `html` would need to know about JS comments... I wonder if your ideas solve any of this (it's ok if they don't).  This really isn't one language embedded into another exactly it's like a big mixing bowl where they just randomly throw JS and HTML together in strange ways. :)

PR #3053, opened 2021-03-17
enh(parser) new highlight() API

### LINKED ISSUE #2277: Propose: Improve `highlight` API

joshgoebel:
Resolves #2277.

### Changes

- Implements the new `highlight(code, options = {})` API
- `language` and `ignoreIllegals` are options now (vs params)
- `continuation` is for internal use only and hence not supported any longer

Compatibility:

- v10.7 supports BOTH old and new APIs fully.
- v11 will support both as well (to ease migrations) but drop `continuation` support from the old API.
- v12 will drop old API support entirely.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,11 @@ Language grammar improvements:
 
 Deprecations:
 
+- `highlight(languageName, code, ignoreIllegals, continuation)` deprecated as of 10.7
+  - Please use the newer API which takes `code` and then accepts options as an object
+  - IE: `highlight(code, {language, ignoreIllegals})`
+  - `continuation` is for internal use only and no longer supported
+
 - `highlightBlock(el)` deprecated as of 10.7.
   - Please use `highlightElement(el)` instead.
   - Plugin callbacks renamed `before/after:highlightBlock` => `before/after:highlightElement`
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ const hljs = require('highlight.js/lib/core');  // require only the core library
 // separately require languages
 hljs.registerLanguage('xml', require('highlight.js/lib/languages/xml'));
 
-const highlightedCode = hljs.highlight('xml', '<span>Hello World!</span>').value
+const highlightedCode = hljs.highlight('<span>Hello World!</span>', {language: 'xml'}).value
 ```
 
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4,15 +4,12 @@ Library API
 Highlight.js exports a few functions as methods of the ``hljs`` object.
 
 
-``highlight(languageName, code, ignore_illegals, continuation)``
-----------------------------------------------------------------
+``highlight(languageName, code, ignoreIllegals, continuation)`` (deprecated with 10.7)
+--------------------------------------------------------------------------------------
 
-Core highlighting function.
-Accepts a language name, or an alias, and a string with the code to highlight.
-The ``ignore_illegals`` parameter, when present and evaluates to a true value,
-forces highlighting to finish even in case of detecting illegal syntax for the
-language instead of throwing an exception.
-The ``continuation`` is an optional mode stack representing unfinished parsing.
+**This is the old API, please see the new API immediately below.**
+
+``continuation`` is an optional mode stack representing unfinished parsing.
 When present, the function will restart parsing from this state instead of
 initializing a new one.  This is used internally for `sublanguage` support.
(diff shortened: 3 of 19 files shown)


allejo (approved):
Do we want to output a deprecation message for this?

paladox:
I'm wondering if we could add support for continuation again? We use it for the gerrit project specifically here https://github.com/GerritCodeReview/gerrit/blob/master/polygerrit-ui/app/elements/diff/gr-syntax-layer/gr-syntax-layer.ts#L445.

joshgoebel:
This has been removed for many different reasons.  

- Technically it's leaking parser internals. 
- It's not intended to "drive the parser" at all (line by line or otherwise).  
- It merely is a detail of our internal sublanguage implementation - one that's likely to change in the future.
- **Line by line driving is fundamentally broken.**  Grammars rules are not guaranteed to break cleanly at line boundaries - and some most definitely do not.
- It was added to our documentation in 2019 that this is not intended to be used to for line-by-line highlighting (93badf79c) [as soon as I found out people were doing so]. 

Many of the things we'd like to investigate in v11 may require deep internal changes.  We might switch from an iterative parser to a completely recursive parser. There is rethinking sublanguages (#1263), a potential new back-tracking parser (#1140), etc.   Just because a line has been processed may not even mean it's processed.  We might hit a rule mismatch a few lines later and then rewind and try again with an entirely different matching strategy.

To allow us to iterate as freely as possible without worrying about breaking changes that means our internals need to be fully internal.  As such `continuation` is gone... and `_top` becomes private API. No state in, no state out.

---

To do line-by-line content _properly and 100% correctly_  one should post-process the final output (which can be done with 10 lines of JS or so). There are other options with the **private** Emitter API, but use at your own risk.  Post-processing the final text output is the most stable choice and what most implementors should do.

joshgoebel:
Someone should really write a `line_by_line` plugin using `after:highlight` that wraps this up nicely and then we could just point everyone to that.

paladox:
@joshgoebel I guess it should be:

‘’’html

result = this.hljs.highlight(
        baseLine,
        {
            language: this.baseLanguage,
            ignoreIllegals: true
        }
        state.baseContext
      );

result.value = result.value.split('\n').map(line => {
     const prepend = tokenStack.map(token => `<span class="${token}">`).join('');
     const matches = line.matchAll(/(<span class="(.*?)">|<\/span>)/g);
     for (const match of matches) {
       if (match[0] === '</span>') tokenStack.shift();
       else tokenStack.unshift(match[2]);
     }
     const append = '</span>'.repeat(tokenStack.length);
     return prepend + line + append;
   }).join('\n');
this.baseRanges.push(this._rangesFromString(result.value, rangesCache));
‘’’

?

paladox:
So this https://gerrit-review.googlesource.com/c/gerrit/+/302602 ?

joshgoebel:
You tell me...?   I don't completely follow larger context of what you are doing here or what baseLine and revisionLine are...

joshgoebel:
@paladox On an unrelated note: https://github.com/gerrit-review/gerrit/blob/master/polygerrit-ui/app/elements/diff/gr-syntax-layer/gr-syntax-layer.js#L349

If these are still actual issues today I'd love to see issues filed and perhaps we can fix them so you don't have to deal with them downstream.

PR #3055, opened 2021-03-18
enh(parser) Detect comments based on english like text, rather than specific list of keywords

### LINKED ISSUE #2827: Idea: Find text that looks like plain english, vs hunting for certain words

The very limited set of words that could appear inside comments to indicate they are comments (vs code) is far too limited and really look like it's been designed with a particular test in mind rather than even which words are likely to be most common in comments.

```
export const PHRASAL_WORDS_MODE = {
  begin: /\b(a|an|the|are|I'm|isn't|don't|doesn't|won't|but|just|should|pretty|simply|enough|gonna|going|wtf|so|such|will|you|your|they|like|more)\b/
};
```

Instead, what if we tried to find whole sequences of text that look like regular sentences/explanation, etc... rather than code... and then some relevance was granted for these sequences vs individual words.

Roughly: (needs work)

```
([()"]?([A-Za-z'-]{3,}|is|a|I|so|us|to|at|if|in)[.]?[()":]?([,.]\s|\s|$|\))){3}
```

See context for an example of this in action on regex101.

Nope.

**Additional context...**

Screen Shot 2020-11-03 at 3 42 21 PM

joshgoebel:
Resolves #2827.

### Changes

Before we boosted comments relevance by:

```
begin: /\b(a|an|the|are|I'm|isn't|don't|doesn't|won't|but|just|should|pretty|simply|enough|gonna|going|wtf|so|such|will|you|your|they|like|more)\b/
```

Now we look for "english like phrases" instead:

```
begin: /[ ]+(([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[:]?([.][ ]|[ ]|\))){3}/
```

### Checklist

### DIFF
--- a/src/lib/modes.js
+++ b/src/lib/modes.js
@@ -73,7 +73,6 @@ export const COMMENT = function(begin, end, modeOptions = {}) {
     },
     modeOptions
   );
-  // mode.contains.push(PHRASAL_WORDS_MODE);
   mode.contains.push({
     className: 'doctag',
     // hack to avoid the space from being included. the space is necessary to
--- a/src/lib/modes.js
+++ b/src/lib/modes.js
@@ -83,13 +83,48 @@ export const COMMENT = function(begin, end, modeOptions = {}) {
     excludeBegin: true,
     relevance: 0
   });
+  const ENGLISH_WORD = regex.either(
+    // list of common 1 and 2 letter words in English
+    "I",
+    "a",
+    "is",
+    "so",
+    "us",
+    "to",
+    "at",
+    "if",
+    "in",
+    "it",
+    "on",
+    // note: this is not an exhaustive list of contractions, just popular ones
+    /[A-Za-z]+['](d|ve|re|ll|t|s|n)/, // contractions - can't we'd they're let's, etc
+    /[A-Za-z]+[-][a-z]+/, // `no-way`, etc.
+    /[A-Za-z][a-z]{2,}/ // allow capitalized words at beginning of sentences
+  );
   // looking like plain text, more likely to be a comment
   mode.contains.push(
     {
       // TODO: how to include ", (, ) without breaking grammars that use these for
       // comment delimiters?
       // begin: /[ ]+([()"]?([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[()":]?([.][ ]|[ ]|\))){3}/
-      begin: /[ ]+(([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[:]?([.][ ]|[ ]|\))){3}/
+      // ---
+
+      // this tries to find sequences of 3 english words in a row (without any
+      // "programming" type syntax) this gives us a strong signal that we've
+      // TRULY found a comment - vs perhaps scanning with the wrong language.
+      // It's possible to find something that LOOKS like the start of the
+      // comment - but then if there is no readable text - good chance it is a
+      // false match and not a comment.
+      //
+      // for a visual example please see:
+      // https://github.com/highlightjs/highlight.js/issues/2827
+
+      begin: regex.concat(
+        /[ ]+/, // necessary to prevent us gobbling up doctags like /* @author Bob Mcgill */
+        '(',
+        ENGLISH_WORD,
+        /[.]?[:]?([.][ ]|[ ])/,
+        '){3}') // look for 3 words in a row
     }
   );
   return mode;
--- a/src/lib/modes.js
+++ b/src/lib/modes.js
@@ -73,18 +73,25 @@ export const COMMENT = function(begin, end, modeOptions = {}) {
     },
     modeOptions
   );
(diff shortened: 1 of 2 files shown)


allejo:
> Now we look for "english like phrases" instead:
> 
> ```regex
> begin: /[ ]+(([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[:]?([.][ ]|[ ]|\))){3}/
> ```

How did you come up with this regex?

joshgoebel:
See #2827.  Testing with a big lump of comments.  It's just attempting to match 3 english looking words in a row...

src/lib/modes.js
```diff
@@ -73,12 +73,25 @@ export const COMMENT = function(begin, end, modeOptions = {}) {
...
+    excludeBegin: true,
     relevance: 0
   });
+  // looking like plain text, more likely to be a comment
+  mode.contains.push(
+    {
+      // TODO: how to include ", (, ) without breaking grammars that use these for
+      // comment delimiters?
+      // begin: /[ ]+([()"]?([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[()":]?([.][ ]|[ ]|\))){3}/
+      begin: /[ ]+(([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[:]?([.][ ]|[ ]|\))){3}/
```
allejo, line 92:
```suggestion
      // TODO: how to include ", (, ) without breaking grammars that use these for
      // comment delimiters?
      // begin: /[ ]+([()"]?([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[()":]?([.][ ]|[ ]|\))){3}/
      //
      // see https://github.com/highlightjs/highlight.js/issues/2827
      begin: /[ ]+(([A-Za-z'-]{3,}|is|a|I|so|us|[tT][oO]|at|if|in|it|on)[.]?[:]?([.][ ]|[ ]|\))){3}/
```

I would add a link to the issue explaining what this regex works on. OR explaining how you reached this seemingly arbitrary list of words. The regex makes sense in the context of that URL and with the regex101 screenshot but not just from reading code.

src/lib/modes.js
```diff
@@ -73,12 +73,25 @@ export const COMMENT = function(begin, end, modeOptions = {}) {
...
   );
-  mode.contains.push(PHRASAL_WORDS_MODE);
+  // mode.contains.push(PHRASAL_WORDS_MODE);
   mode.contains.push({
     className: 'doctag',
-    begin: '(?:TODO|FIXME|NOTE|BUG|OPTIMIZE|HACK|XXX):',
+    // hack to avoid the space from being included. the space is necessary to
+    // match here to prevent the plain text rule below from gobbling up doctags
+    begin: '[ ]*(?=(TODO|FIXME|NOTE|BUG|OPTIMIZE|HACK|XXX):)',
+    end: /(TODO|FIXME|NOTE|BUG|OPTIMIZE|HACK|XXX):/,
```
allejo, line 82:
```suggestion
    begin: '[ ]*(?=(TODO|FIXME|NOTE|BUG|OPTIMIZE|HACK|XXX):?)',
    end: /(TODO|FIXME|NOTE|BUG|OPTIMIZE|HACK|XXX):?/,
```

I've seen code snippets that don't have the `:`, so unless this causes problems, I'd prefer the `:` be optional.

allejo (approved):
I don't see any downsides to this, it's better than the list of arbitrary words. Just some minor suggestions but LGTM and it's something that can be improved later on.

src/lib/modes.js
joshgoebel: Please provide some examples?... opening this up makes this a LOT fuzzier... perhaps we need a word boundary then... or else you match NOTEABLE, XXXXXXXX, BUGGY, BUGGER, etc... what does it look like when someone doesn't use a `:`?  I kind of like the precision of it as it stands now... that `:` is the "one true way". LOL.

src/lib/modes.js
joshgoebel: I will explain this further with comments then and we can link to the issue that's a good idea.  Basically it's just common 1 and 2 letters words since just matching `\w{1,2}` would be FAR too broad... I'm sure it's also missing a few words that we'll add over time (or not, since this is mostly an invisible helper for auto-detect).

src/lib/modes.js
allejo: You're right, a word boundary would be better in this case. Definitely not as common to not have a `:`, but code linters often support it, e.g.

- https://github.com/JohnPostlethwait/fixme/blob/master/bin/fixme.js
- https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Tests/Commenting/FixmeUnitTest.js
- https://github.com/matteosan1/workspace/blob/master/qt4/tdpalio2online/tdpalioserver/corsa.cpp

However, I don't feel strongly about this

PR #3057, opened 2021-03-18
enh(parser) Warn if unescaped HTML is present

### LINKED ISSUE #2886: Detect HTML/JS injection attacks and warn users pro-actively

https://github.com/highlightjs/highlight.js/issues/2884
https://github.com/ccampbell/rainbow/issues/249

This comes up again and again (though thankfully not TOO often).  Beginners are VERY confused about this whole HTML escaping thing.  

While we can't do anything smart about this yet (because we unfortunately allow HTML inside code blocks for "clever" users) I'd like this to change with v11.  With v11 we should drop this HTML pass-thru behavior and move it to a plugin (making it very much opt-in).  The default behavior should be that HTML is silently dropped and I'd even consider adding some sort of error:

```
[code block]
WARNING.
Are you missing a bunch of HTML code you expected to see here? 
Your HTML wasn't properly escaped and that can lead to serious
security issues. _Learn More_

Properly escape your code and the highlighting you expect will kick in.
[/code block]
```

This would of course be a breaking change so we'd need to wait until v11.  For 95% of users I can't see the downside to this and it seems we could potentially educate and prevent a lot of harm.  Someone wanting the HTML to pass thru would install a plug-in and thus change the behavior, get the old behavior back, etc.

Silent dropping but no error... but that just leads to support issues... I suppose we could log the error to the console vs actually showing it on the webpage.

joshgoebel:
Closes #2886.

Note: This cannot be merged until V11.

### Changes

If unescaped HTML is detected inside the element that has been requested
to highlight then it will be stripped and a warning will be logged to
the console.

This warning can be supressed with the option `ignoreUnescapedHTML`.

This is done AFTER the `before` plugin hooks fire, allowing for a plugin
to potentially grab the HTML before hand, replace it with text, and
later do something special - say perhaps merging it back it.

Currently this can happen without touching the `ignoreUnescapedHTML`
option as the element is passed to the plugin BEFORE the security checks
are performed.

An alternative would be to always do the security checks first - and if
a user wishes to use a plugin that allows HTML then they would have to
both install that plugin AND turn on `ignoreUnescapedHTML`.

Hard to say if one might be preferrable over the other?

### Checklist


joshgoebel:
Screen Shot 2021-03-18 at 7 29 53 PM

allejo:
> An alternative would be to always do the security checks first - and if
> a user wishes to use a plugin that allows HTML then they would have to
> both install that plugin AND turn on `ignoreUnescapedHTML`.

I would personally prefer this option. Making it tedious to opt-in to this behavior should make people think, "why is this so tedious to enable?" For example, in React, if you want to bypass the React sanitization and such, you have to use the [convoluted `dangerouslySetInnerHTML` prop](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml). It's convoluted by design to actively discourage this and only if you actually know what you're doing, then you should be doing it.

joshgoebel:
> I would personally prefer this option. Making it tedious to opt-in to this behavior should make people think, "why is this so tedious to enable?"

Isn't it kind of double annoying though if the line right before it is?

```
hljs.options({ignoreUnescapedHTML: true});
hljs.registerPlugin(MergeUnescapedHTML);
```

A little redundant, no?

allejo:
> ```
> hljs.options({ignoreUnescapedHTML: true});
> hljs.registerPlugin(MergeUnescapedHTML);
> ```
> 
> 
> A little redundant, no?

Redundant? Yes. But the way I see it is the first line is to disable the warning emitted by hljs. And the second line is to actually handle the behavior.

Is it be possible for a plugin to change the hljs options? i.e. have the `MergeUnescapedHTML` plugin set `ignoreUnescapedHTML` to true.

> So then when the security check runs AFTER all plugins run it would see that the plugin already handled the HTML and hence no need for a security warning... does that make sense?

But I would agree that doing the security check after the plugin runs would be a good idea since it'd allow plugins to fix/silence security warnings.

joshgoebel:
> But I would agree that doing the security check after the plugin runs would be a good idea since it'd allow plugins to fix/silence security warnings.

Then I will leave it as it is for now - the plugin runs first.  So a properly structured plugin can fix the security issue - making adding HTML merging back a one-liner.

> Is it be possible for a plugin to change the hljs options? 

No, and I'm not sure that'd be a good idea. :) It did cross my mind.  Hopefully we'd have so few options in core that this type of thing wouldn't be necessary (plugins reconfiguring the core engine).

joshgoebel:
Can this be approved then (the functionality)?  The tests won't go green until this is rebased on top of V11... so it'll probably be held for a bit, but I'd like to know if it's good to go or not.

src/highlight.js
```diff
@@ -702,15 +703,19 @@ const HLJS = function(hljs) {
 
     if (shouldNotHighlight(language)) return;
 
-    // support for v10 API
     fire("before:highlightElement",
       { el: element, language: language });
 
+    if (!options.ignoreUnescapedHTML && element.innerHTML !== element.textContent) {
```
allejo, line 709:
```html
<div id="foo">
  I'm just a &lt;code&gt;sample&lt;/code&gt; block.
</div>

<div id="innerHTML"></div>
<div id="textContent"></div>
<div id="result"></div>

<script>
document.getElementById("innerHTML").innerText = document.getElementById("foo").innerHTML;
document.getElementById("textContent").innerText = document.getElementById("foo").textContent;
document.getElementById("result").innerText = document.getElementById("foo").innerHTML == document.getElementById("foo").textContent;
</script>
```

https://jsfiddle.net/uj5qzn2e/

Am I doing something wrong? Or does `innerHTML` need to be have HTML entities transformed before they can be compared with `textContent`

joshgoebel: Oh, my test cases locally didn't include escaped HTML... that certainly breaks the comparison approach.  I think instead perhaps we just need to check node.children...

allejo (approved):
`element.children.length > 0` works much better :+1:

PR #3062, opened 2021-03-21
(chore) Improve Kotlin language demo

bluefireoly:
The previous demo did not really show how the themes will look like with kotlin, this is why I added much more language features of Kotlin to the example.

The new example code includes:
- interfaces, classes, inheritance
- import statements
- properties declared in the constructor
- a variable
- default parameters and named parameters
- Strings
- numbers
- a custom property getter
- String templates
- field keyword
- function with parameters
- a lamda expression with a named input parameter
- function invocation without round brackets
- an Integer range
- an annotation
- an extension function
- nullable properties
- the ?: operator
- an if statement

### Checklist

### DIFF
--- a/test/detect/kotlin/default.txt
+++ b/test/detect/kotlin/default.txt
@@ -1,12 +1,27 @@
-import kotlin.lang.test
+import kotlinx.serialization.Serializable
+import kotlin.random.Random
 
-interface A {
-    fun x()
+interface Building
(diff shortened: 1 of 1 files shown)


joshgoebel (approved):

joshgoebel:
@bluefireoly  Thanks!

PR #3065, opened 2021-03-22
enh(shell) add alias ShellSession

ryantm:
### Changes

ShellSession is the canonical name for this "language" used by Linguist which is used by
GitHub.

https://github.com/github/linguist/blob/2a6f9e29e198afbaa721164680bbc3b292dad260/lib/linguist/languages.yml#L5340

### Checklist
- [x] Added markup tests, or they don't apply here because... this just adds a language alias

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,13 @@
+## Version next
+
+- enh(shell) add alias ShellSession [Ryan Mulligan][]
+
+[Ryan Mulligan]: https://github.com/ryantm
+
+
 ## Version 10.7.1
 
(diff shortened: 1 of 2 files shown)


joshgoebel:
> ShellSession is the canonical name for this "language" used by Linguist which is used by
GitHub.

I don't necessarily find the canonical-ality of GitHub or Linguist to be a great reason for us to add aliases, but since we also already have the name of this grammar as "shell session" this probably makes sense based on that alone.

joshgoebel (approved):

joshgoebel:
Thanks!

PR #3066, opened 2021-03-22
enh(shell) consider one space after prompt as part of prompt

ryantm:
### Changes
By having the space after the prompt be part of the prompt, you can
have css like

```
code.language-ShellSession .hljs-meta,
code.language-shellsession .hljs-meta,
code.language-Shell .hljs-meta,
code.language-shell .hljs-meta,
code.language-Console .hljs-meta,
code.language-console .hljs-meta {
  user-select: none;
}
```

which allows you to avoid selecting the prompt, where the selection
does not include the space after the prompt.

### Checklist
- [X] Added markup tests, or they don't apply here because... tests updated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## Version next
 
 - enh(shell) add alias ShellSession [Ryan Mulligan][]
+- enh(shell) consider one space after prompt as part of prompt [Ryan Mulligan][]
 
 [Ryan Mulligan]: https://github.com/ryantm
 
@@ -69,7 +70,6 @@ Deprecations:
 [David Ostrovsky]: https://github.com/davido
 [AndyKIron]: https://github.com/AndyKIron
 [Samuel Colvin]: https://github.com/samuelcolvin
-[Ryan Mulligan]: https://github.com/ryantm
 
 ## Version 10.6.0
 
--- a/src/languages/shell.js
+++ b/src/languages/shell.js
@@ -17,7 +17,7 @@ export default function(hljs) {
         // We cannot add \s (spaces) in the regular expression otherwise it will be too broad and produce unexpected result.
         // For instance, in the following example, it would match "echo /path/to/home >" as a prompt:
         // echo /path/to/home > t.exe
-        begin: /^\s{0,3}[/~\w\d[\]()@-]*[>%$#]/,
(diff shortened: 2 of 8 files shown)


joshgoebel:
I have mixed thoughts on this one.  This runs contrary to our long-standing policy of NOT over-matching (ie, generally not including spaces in matched elements like this), _even when_ the visual effect would be exactly the same.  But then one might argue this is actually a more precise match in some ways - that the space truly is part of the prompt.  I'll noodle on it and see if my thoughts firm up any with a bit more time.

ryantm:
> one might argue this is actually a more precise match in some ways - that the space truly is part of the prompt.

Yes, I'm pretty sure you can define your prompt without the space, though I think most people would hate that. When you start typing a command you certainly never type a space first!

ryantm:
I fixed up CHANGES.md, resolving the merge conflict.

joshgoebel:
> I fixed up CHANGES.md, resolving the merge conflict.

Please don't continue to bother with this. :)  I'm used to this by now - it's no bother for me to do it at the very end.

CHANGES.md
```diff
@@ -69,7 +70,6 @@ Deprecations:
 [David Ostrovsky]: https://github.com/davido
 [AndyKIron]: https://github.com/AndyKIron
 [Samuel Colvin]: https://github.com/samuelcolvin
-[Ryan Mulligan]: https://github.com/ryantm
```
joshgoebel, line 72: This is a nice catch but wouldn't have hurt anything either. ;)

src/languages/shell.js
```diff
@@ -17,7 +17,7 @@ export default function(hljs) {
         // We cannot add \s (spaces) in the regular expression otherwise it will be too broad and produce unexpected result.
         // For instance, in the following example, it would match "echo /path/to/home >" as a prompt:
         // echo /path/to/home > t.exe
-        begin: /^\s{0,3}[/~\w\d[\]()@-]*[>%$#]/,
+        begin: /^\s{0,3}[/~\w\d[\]()@-]*[>%$#]\s?/,
```
joshgoebel, line 20: Are there cases where this is too board?  I think this will also match tabs and newlines... I think perhaps `[ ]?` would be better?

ryantm: Yeah, good call. I changed it to that.

joshgoebel (approved):

PR #3068, opened 2021-03-23
Version 11 backlog

joshgoebel:
The best way to review this is one commit at a time, they all shoudl be
clean.

- Don't bother to review the VERSION_11.md as you go, it's just a rough list - there is a commit at the end that cleans it up proper.

### Changes

Please see #2558.

### Checklist


joshgoebel (commented):
Self-approving so this can be merged.

PR #3073, opened 2021-03-25
(chore) Fix typescript type exports

### LINKED ISSUE #2945: [Request] Expose TypeScript types.
Sometime I want to use some type from Highlight.js internal types. e.g. `Language`, `LanguageFn`.  
But I can't use it now.

Expose all internal TypeScript types from `highlight.js`.  
So I can use it for more typing.

No.

**Additional context...**
Types should be able to import like this.

```ts
import { LanguageFn, Language } from 'highlight.js';
```

joshgoebel:
Resolved #2945.

### Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,7 @@ module.exports = {
     ecmaVersion: 2015,
     sourceType: "module"
   },
-  // parser: '@typescript-eslint/parser',
+  parser: '@typescript-eslint/parser',
   plugins: [
     "@typescript-eslint"
   ],
--- a/src/core.d.ts
+++ b/src/core.d.ts
@@ -1,0 +1,6 @@
+// commented out
+// <reference path="../types/index.d.ts" />
+
+import hljs from "highlight.js";
+export default hljs;
+
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -65,6 +65,7 @@ async function buildNode(options) {
   install("./LICENSE", "LICENSE");
   install("./README.md", "README.md");
   install("./types/index.d.ts", "types/index.d.ts");
+  install("./src/core.d.ts","lib/core.d.ts");
 
   log("Writing styles.");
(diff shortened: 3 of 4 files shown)


Hypnosphi:
@joshgoebel can you please backport this into v10? I'd be happy to migrate to v11, but unfortunately I can't, because I use dynamic imports, see https://github.com/webpack/webpack/issues/13865

joshgoebel:
_V10.0 is about to drop off the supported list entirely._   The core team really does not have time to support multiple versions.  Even a PR here isn't super trivial as it would have to pull over the new changes but then also back out the changes to types that occurred between 10 -> 11.

> but unfortunately I can't, because I use dynamic imports, see webpack/webpack#13865

`import` isn't the only way to use the library.  You could use the stand-alone JS builds (as a separate file even) or even append our web build at the very end of your build process (I'm sure Webpack has a way of doing this).  So there are ways to use v11 with Webpack despite the bug.

joshgoebel:
If the exports was a _huge_ issue we might perhaps drop them entirely, but it seems only a few broken tools are having issues and that mostly everything else works - unless others would like to chime in here.

Hypnosphi:
More context: we're developing a component library, and our `Code` component loads languages on demand in runtime depending on the passed `language` prop value. How would I do that without using dynamic imports or migrating away from webpack?

Source: https://github.com/JetBrains/ring-ui/blob/typescript/components/code/code.tsx#L15-L21

joshgoebel:
Well, I'm not a webpack person I have no idea what the magic comment is trying to do but I'd imagine for dynamic run-time dependencies (via import) you would just use the new ESM CDN builds that we publish (as of 11.2) unmodified... the grammars are all package up in individual files for dynamic usage... and the "main" Highlight.js bundle could be:

- loaded separately 
- concated to the end of your web pack build
- imported dynamically as well (via ESM CDN build)

PR #3077, opened 2021-03-26
enh(c) Update keyword list for C11/C18

### LINKED ISSUE #3010: enh(c) Update keyword list for C11/C18
Pulling this out because we should eventually apply this to our `C` grammar (removing the C++ keywords and types, etc).

---

The full list of *keywords* for C11/C18 is

```
                auto                          if                             unsigned
                break                         inline                         void
                case                          int                            volatile
                char                          long                           while
                const                         register                       _Alignas
                continue                      restrict                       _Alignof
                default                       return                         _Atomic
                do                            short                          _Bool
                double                        signed                         _Complex
                else                          sizeof                         _Generic
                enum                          static                         _Imaginary
                extern                        struct                         _Noreturn
                float                         switch                         _Static_assert
                for                           typedef                        _Thread_local
                goto                          union
```

_Originally posted by @ztane in https://github.com/highlightjs/highlight.js/issues/2973#issuecomment-779284055_

joshgoebel:
Resolves #3010.

### Changes

- Applies the keyword list found in #3010
- Splits out `type` keywords

### Checklist
- [x] Added markup tests, or they don't apply here because we're still
reworking the whole grammar

### DIFF
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -27,7 +27,7 @@ export default function(hljs) {
     '[a-zA-Z_]\\w*' + regex.optional(TEMPLATE_ARGUMENT_RE) +
   ')';
   const CPP_PRIMITIVE_TYPES = {
-    className: 'keyword',
+    className: 'type',
     begin: '\\b[a-z\\d_]*_t\\b'
   };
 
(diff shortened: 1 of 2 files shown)


joshgoebel:
@klmr Any chance you could review the one?

src/languages/c.js
```diff
@@ -105,19 +105,61 @@ export default function(hljs) {
...
+    "_Bool",
+    "_Complex",
+    "_Generic",
+    "_Imaginary",
+    "_Noreturn",
+    "_Static_assert",
+    "_Thread_local"
+  ];
+
+  const C_TYPES = [
```
klmr, line 146: `_Bool`, `_Complex` and `_Imaginary` are *types*. So are (starting with the next version, C23) `_Decimal32`, `_Decimal64` and `_Decimal128`.

src/languages/c.js
```diff
@@ -105,19 +105,61 @@ export default function(hljs) {
...
+    "short",
+    "long",
+    "char",
+    "void"
+  ];
+
+  const KEYWORDS = {
+    keyword: C_KEYWORDS,
+    type: C_TYPES,
+    literal: 'true false nullptr NULL',
```
klmr, line 161: `nullptr` doesn’t exist in C, it’s only a keyword literal in C++.

src/languages/c.js
```diff
@@ -245,15 +286,15 @@ export default function(hljs) {
         { // containers: ie, `vector <int> rooms (9);`
           begin: '\\b(deque|list|queue|priority_queue|pair|stack|vector|map|set|bitset|multiset|multimap|unordered_map|unordered_set|unordered_multiset|unordered_multimap|array)\\s*<',
```
klmr, line 287: I guess this is beyond the scope of this PR but these are all C++ types (and the entire rule related to a C++ only feature, templates).

src/languages/c.js
```diff
@@ -105,19 +105,61 @@ export default function(hljs) {
...
+    "char",
+    "void"
+  ];
+
+  const KEYWORDS = {
+    keyword: C_KEYWORDS,
+    type: C_TYPES,
+    literal: 'true false nullptr NULL',
+    // TODO: apply hinting work similar to what was done in cpp.js
     built_in: 'std string wstring cin cout cerr clog stdin stdout stderr stringstream istringstream ostringstream ' +
```
klmr, line 163: The list of built-ins still contains *lots* of C++-isms that need to be removed. There are also some duplications with the list of keywords now (`_Bool`, `_Complex` …).

By contrast, I would consider adding the list of [convenience macros aliasing underscore-keywords](https://en.cppreference.com/w/c/keyword):

```
alignas
alignof
complex
bool
imaginary
noreturn
static_assert
thread_local
```

… and potentially the `_Atomic` aliases from here: https://en.cppreference.com/w/c/atomic#Types_2

I would also consider adding [`_Pragma`](https://en.cppreference.com/w/c/preprocessor/impl) which is not a C keyword but is, for all intents and purposes, treated exactly like one.

The rest of the list is still pretty arbitrary: it contains some (but not all) names functions from the C standard library.

src/languages/c.js
```diff
@@ -105,19 +105,61 @@ export default function(hljs) {
...
+    "_Alignof",
+    "_Atomic",
+    "_Bool",
+    "_Complex",
+    "_Generic",
+    "_Imaginary",
+    "_Noreturn",
+    "_Static_assert",
+    "_Thread_local"
+  ];
```
klmr, line 144: Two keywords are missing: `asm` and `fortran`. These are only “conditionally supported” but, where they are, they’re keywords. At least `asm` is supported by all major compilers.

src/languages/c.js
joshgoebel: > The rest of the list is still pretty arbitrary:

That's fine once we more it to hinting as long as they are "common" functions in general C code.

joshgoebel: > and potentially the _Atomic aliases from here:

You mean like? `atomic_ulong`

klmr: Yes, exactly.

Re the hinting: sure, that makes sense.

joshgoebel: >  But the C++-only ones should be removed, otherwise they bias the detection unduly towards C and away from C++.

No argument there of course. :)

joshgoebel:
@klmr Thanks, i think this gets us a BIT closer. :)

PR #3078, opened 2021-03-26
enh(cpp) simpler struct matching

### LINKED ISSUE #2520: Add some type of look-behind support

### LINKED ISSUE #2736: (c) Declaration of struct variable triggers class title mode
**Describe the issue**

When declaring a variable of type `struct foobar`, this triggers the rule for a struct/class title, which leads to the whole declaration being highlighted as a title. See https://meta.stackexchange.com/a/354753/1968.

`c`/`cpp`

`highlight`

**Sample Code to Reproduce**

```c
struct List *newnode = (struct List *)malloc(size * sizeof(struct List));
```

This is rendered as

```xml
<class><keyword>struct</keyword> <title>List</title> *<title>newnode</title> = (<title>struct</title> <title>List</span> *)<span class="hljs-title">malloc</title>(<title>size</title> * <title>sizeof</title>(<title>struct</title> <title>List</title>));</class>
```

**Expected behavior**

Should render as

```xml
<keyword>struct</keyword> List *newnode = (<keyword>struct</keyword> List *)<built_in>malloc</built_in>( * <keyword>sizeof</keyword>(<keyword>struct</keyword> List));
```

**Additional context**

Without the initial `struct` the highlighter doesn’t trigger class/title mode. It still doesn’t correctly render the `struct` keywords in the rest of the initialisation but that’s already addressed (incidentally) by #2728.

### LINKED ISSUE #3112: Highlight identifier after 'impl' keyword in Rust language definition
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `highlight.js@9.18.3` for the project I'm working on.

It looks as if the identifier after the `impl` keyword is not being highlighted. I've added an extra element to the class definition to include `impl`.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/highlight.js/lib/languages/rust.js b/node_modules/highlight.js/lib/languages/rust.js
index 5e2af44..590dc15 100644
--- a/node_modules/highlight.js/lib/languages/rust.js
+++ b/node_modules/highlight.js/lib/languages/rust.js
@@ -89,7 +89,7 @@ module.exports = function(hljs) {
       },
       {
         className: 'class',
-        beginKeywords: 'trait enum struct union', end: '{',
+        beginKeywords: 'trait enum struct union impl', end: '{',
         contains: [
           hljs.inherit(hljs.UNDERSCORE_TITLE_MODE, {endsParent: true})
         ],
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

joshgoebel:
Resolves #2736.   Resolves #2520. Resolves #3112.

### Changes

- Adds new `beforeMatch` compiler extension
- (cpp) Simplified struct/class/enum/union matcher
- (java) Simplified class interface enum matcher
- adds `title.class` sub-scope support. Related #2521.

Seems to do about as well as GitHub now.

Screen Shot 2021-03-26 at 12 39 58 PM

### Checklist
- [x] docs for `beforeMatch`

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -452,7 +452,7 @@ const HLJS = function(hljs) {
         modeBuffer += codeToHighlight.slice(match.index, match.index + 1);
         if (!SAFE_MODE) {
           /** @type {AnnotatedError} */
-          const err = new Error('0 width match regex');
+          const err = new Error(`0 width match regex (${languageName})`);
           err.languageName = languageName;
           err.badRule = lastMatch.rule;
           throw err;
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -252,7 +252,8 @@ const HLJS = function(hljs) {
      */
     function emitMultiClass(mode, match) {
       let i = 1;
-      while (match[i]) {
+      // eslint-disable-next-line no-undefined
+      while (match[i] !== undefined) {
         const klass = language.classNameAliases[mode.className[i]] || mode.className[i];
         const text = match[i];
         if (klass) { emitter.addKeyword(text, klass); } else {
--- a/src/languages/actionscript.js
+++ b/src/languages/actionscript.js
@@ -10,6 +10,10 @@ import * as regex from '../lib/regex.js';
 /** @type LanguageFn */
 export default function(hljs) {
   const IDENT_RE = /[a-zA-Z_$][a-zA-Z0-9_$]*/;
+  const PKG_NAME_RE = regex.concat(
+    IDENT_RE,
+    regex.concat("(\\.", IDENT_RE, ")*")
+  );
   const IDENT_FUNC_RETURN_TYPE_RE = /([*]|[a-zA-Z_$][a-zA-Z0-9_$]*)/;
 
   const AS3_REST_ARG_MODE = {
@@ -92,20 +96,16 @@ export default function(hljs) {
       hljs.C_BLOCK_COMMENT_MODE,
       hljs.C_NUMBER_MODE,
       {
-        className: 'class',
-        beginKeywords: 'package',
-        end: /\{/,
-        contains: [ hljs.TITLE_MODE ]
+        beforeMatch: /\b(package)\s+/,
+        keywords: "package",
+        match: PKG_NAME_RE,
+        className: "title.class"
       },
       {
-        className: 'class',
-        beginKeywords: 'class interface',
-        end: /\{/,
-        excludeEnd: true,
-        contains: [
-          { beginKeywords: 'extends implements' },
-          hljs.TITLE_MODE
-        ]
+        beforeMatch: /\b(class|interface|extends|implements)\s+/,
+        keywords: "class interface extends implements",
+        match: IDENT_RE,
+        className: "title.class"
       },
       {
         className: 'meta',
--- a/src/languages/actionscript.js
+++ b/src/languages/actionscript.js
@@ -114,13 +114,12 @@ export default function(hljs) {
         keywords: { 'meta-keyword': 'import include' }
       },
       {
-        className: 'function',
         beginKeywords: 'function',
         end: /[{;]/,
         excludeEnd: true,
         illegal: /\S/,
         contains: [
-          hljs.TITLE_MODE,
+          hljs.inherit(hljs.TITLE_MODE, { className: "title.function" }),
           {
             className: 'params',
             begin: /\(/,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,9 +4,20 @@ Parser:
 
 - enh(parser) support multi-class matchers (#3081) [Josh Goebel][]
 - enh(parser) Detect comments based on english like text, rather than keyword list [Josh Goebel][]
+- adds `title.class` sub-scope support (#3078) [Josh Goebel][]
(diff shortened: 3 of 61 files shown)


joshgoebel:
Right now I'm using .title-class... but I wonder if perhaps we shouldn't a bit grander:

.title--class
.title__class

joshgoebel:
This is currently rebased on the multi-class stuff (so I could finish it), but in review that can be ignored.

joshgoebel (commented):
Self-approving.  @allejo If you have time to look at this later and see anything terrible please let me know.

.eslintrc.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     ecmaVersion: 2015,
     sourceType: "module"
   },
-  parser: '@typescript-eslint/parser',
+  // parser: '@typescript-eslint/parser',
```
NullVoxPopuli, line 15: why this change?

joshgoebel: It goes back and forth. :)  The uncomment never should have gotten committed.

joshgoebel: Do you know if you can use `parser` only for specific paths?

src/highlight.js
```diff
@@ -452,7 +453,7 @@ const HLJS = function(hljs) {
         modeBuffer += codeToHighlight.slice(match.index, match.index + 1);
         if (!SAFE_MODE) {
           /** @type {AnnotatedError} */
-          const err = new Error('0 width match regex');
+          const err = new Error(`0 width match regex (${languageName})`);
```
NullVoxPopuli, line 456: I appreciate the specification!

src/languages/actionscript.js
```diff
@@ -92,20 +96,16 @@ export default function(hljs) {
       hljs.C_BLOCK_COMMENT_MODE,
       hljs.C_NUMBER_MODE,
       {
-        className: 'class',
-        beginKeywords: 'package',
-        end: /\{/,
-        contains: [ hljs.TITLE_MODE ]
+        beforeMatch: /\b(package)\s+/,
```
NullVoxPopuli, line 99: this is super slick!

NullVoxPopuli (approved):
This is exciting! It'll make my own grammar cleaner in a few places :D

src/languages/actionscript.js
joshgoebel: It may still be replaced by multi-match though since that can do the same thing... not sure.

.eslintrc.js
NullVoxPopuli: you can! via the overrides options (which I recommend moving to _only_ overrides, as I've done here: https://github.com/NullVoxPopuli/eslint-configs)

joshgoebel: Nice!  For some reason I was thinking that something as high level as the parser couldn't be configured in overrides. Seems to work just fine though.

PR #3081, opened 2021-03-27
enh(parser) multi-class in a single mode

### LINKED ISSUE #2838: Proposal: Support multi-class highlighting inside a single mode/rule

Yes, very often you have language constructs like `[delim]content[delim]`... a string being a perfect example.  Many time you may want to highlight the delimiters differently than the content.  We provide no simple way to do that now without resorting to complex rule chains or ambiguous `contents`.  For example, lets match a simple single quoted string and highlight the whole thing as a string.

```js
{ 
  className: "string",
  begin: /'.*'/,
}
```

Easy, but now lets try to color it separately:

```js
{
  className: "string.delim",
  begin: /'/, end: /'/,
  contains: [ 
     { begin : /[^']*, className: "string" /
  ]
}
```

This is the shortest variant, and does get the job done, but we have string nested inside `string.delim`, which is strange.  If we were going to do nesting at all (which I'm not sure we should) here surely you'd want `string.delim` inside `string`. And of course it wouldn't help us at all if we wanted to classify the begin and end matcher with different classes.

Lets try again (and fail):

```js
{
  begin: /(?=')/,  // looks like a string looking ahead
  contains: [ 
     { begin: /'/, className: "string.delim" },
     { begin : /[^']*(?=')/, className: "string" },
  ]
}
```

Many modes/grammars get away using/abusing contains like this for a sequence and it works because the rules are distinct enough... that doesn't work for us here... I'm not sure how to make this work... after it finds the whole string it'll just keep matching, we have no way to know we've reached the "end".  It would work if the end delimiter were different:

```js
  // 'string'E // 'E terminates strings
  contains: [ 
     { begin: /'/, className: "string.delim" },
     { begin : /.*(?!'E)/, className: "string" },
     { begin: /'E/, className: "string.delim", endsParent: true },
  ]
```

Ok, so lets dig in use all the powers we have available:

```js
{
        begin: /(?='.*')/,  // perhaps look ahead just to see if we have a full string
        contains: [
           {
             begin: /'/, className: "string.delim",
             starts: {
               className: "string",
               end: /\b\B/, // hack to leave the mode open until a rule matches
               contains: [
                 {
                   begin: /(?=')/,
                   endsParent: true
                 }
               ],
               starts: {
                 contains: [
                   { begin: /'/, className: "string.delim", }
                 ]
               }
             }
             },
        ]
      },
```

Ok, that works, but man... find quote highlight it, start a new mode "string", use a magic `end` to prevent the mode from closing... keep matching things until we see a `'` (look ahead), then end the parent which starts ANOTHER new mode to match the final delimiter.  Ugh.  We could also try a pure chain:

```js
      {
        begin: /(?='.*')/,  // perhaps look ahead just to see if we have a full string
        contains: [
           {
             begin: /'/, className: "string.delim",
             starts: {
               className: "string",
               begin: /[^']*/,
               starts: {
                 contains: [
                   { begin: /'/, className: "string.delim", }
                 ]
               }
             }
          }
        ]
      },
```

Simple, one mode chains into the next into the next with `starts` until it hits the end and all rewinds... this of course requires us to manually match the middle of the expression, which is slightly annoying.

Of course these structures are a pain to write by hand, but we could of course use syntactic sugar (ie, build one of the above variants internally, without adding any features to the core of the parser).  Say we added some chain sugar:

```js
{
        className: "string",
        chain: [ 
          { match: /'/, className: "string.delim" },
          { match: /[^']*/, className: "string" },
          { match: /'/, className: "string.delim" }
        ]
}
```

Better.  This of course compiles into something much more complex... and we'd still left with having to specify the inside match of the string, which is kind of annoying.  This is also bad because single "modes" that secretly compile into massively complex chains make it much harder to build complex high-level structs based on composing those lower-level structures.  The interactions get very complex because what you think is a "simple low-level rule" is actually a massively complex rule that the compiler has just hidden all the complexity away from you.

This type of sugar works for many simpler things, but this "3 pair" (two delims, and an enclosure) is such a common pattern that I really think perhaps it should be added/supported by the parser at the lowest-level.  We already have the concept of `begin`, `end` and everything in-between.   We just don't provide an easy mechanism to assign separate CSS classes.

So I'd like to propose two low-level variations: 

One for modes with children, simply allows each "piece" to be individually classified. This is closest to what we already have and would be simplest to add I think:

```js
{
        className: {
          begin: 'string.delim',
          middle: 'string',
          end: 'string.delim'
        },
        begin: '"',
        end: '"',
        contains: // ...
      }
```

And the same thing for simpler regex matches when a single regex will get the job done:

```js
      {
        match: /(')(.*?)(')/,
        className: ['string.delim', 'string', 'string.delim'],
        // or possibly
        className: {
          0: 'string.delim',
          1: 'string',
          2: 'string.delim'
        },
      },
```

The latter format (keyed digits) would of be immediately recognizable to anyone whose worked on TextMate grammars before... and of course this style is not limited to 3 match components... you could easily have a complex regex that broke something down into 5 or 7 components, highlighting each of the pieces differently.

And since these are TRUE singular modes they can be composed easily (used anywhere modes can already be used) without any special caveats or complex interactions with `starts`, `endsParent`, `endsWithParent`, etc...

**Additional context...**

None.

joshgoebel:
Resolves #2838.

### Changes

Adds multi-class highlighting:

```js
{
        begin: [
          /function|function!/,
          /\s+/,
          hljs.IDENT_RE
        ],
        className: {
          1: "keyword",
          3: "title"
        },
}
```

### Checklist
- [x] needs unit tests
- [x ] docs


docs/mode-reference.rst
```diff
@@ -154,16 +154,40 @@ for one thing like string in single or double quotes.
...
+  },
+  }
+
+This would highlight ``function!`` as a ``keyword`` while highlighting the name
+of the function as ``title``. The space(s) between would be matched, but not
+highlighted.
+
+*Note: Internally this magic happens by wrapping each regex within it's own
+capture group.  If your regexes include any groups they must all be
+non-capturing, ie:* ``(?:regex)``.
```
allejo, line 185: Is there a way to detect if a RegExp object has capture groups inside of it? If not, I feel like we should specify that using capture groups with a traditional RegExp object will __not__ work since we're only checking if `begin` is an array. Add a link to #3095?

src/lib/ext/multi_class.js
```diff
@@ -0,0 +1,26 @@
...
+import * as regex from "../regex.js";
+
+const MultiClassError = new Error();
+
+/**
+ *
+ * @param {CompiledMode} mode
+ */
+export function MultiClass(mode) {
+  if (!Array.isArray(mode.begin)) return;
```
allejo, line 11:
```suggestion
  if (!Array.isArray(mode.begin)) {
    mode.isMultiClass = false;
    return;
  }
```

`isMultiClass` should be set to false since according to the type definition, it's a boolean that will exist in a CompiledMode.

allejo (changes requested):
Loving the simplicity needed to add this feature

docs/mode-reference.rst
joshgoebel: >  If not, I feel like we should specify that using capture groups with a traditional RegExp object will not work

I thought that's what it already says.  

> Add a link to #3095?

That's a great idea.

allejo: To me, this says that if I want to organize my regex and have groups, to always use non-capturing groups. Then it tells me that when an array is given, it will internally convert it to capturing groups for me. It doesn't tell me that I shouldn't/can't use capturing groups right now and expect this behavior.

src/lib/ext/multi_class.js
joshgoebel: I made the type optional instead.

docs/mode-reference.rst
joshgoebel: I thought about that (the necessity of array), but there are all sorts of reasons to use capture groups than multi-class... you can use capture groups as much as you want, just NOT with multi-match... 

The other problem with letting someone specify their own capture groups is you have to have ALL of them... you can't leave some content outside a group... so the array allows us to FORCE this behavior since we're in charge of the wrapping... and the more I use I think it's also kind of nice for people who like clean syntax more than pure regex.

I could take another look, but if you see the confusion could you just offer up what you think it should say? Since I plan to fix it soon anyways I'm not wanting to spend much time here.

src/highlight.js
```diff
@@ -247,10 +247,29 @@ const HLJS = function(hljs) {
     }
 
     /**
-     * @param {Mode} mode - new mode to start
+     * @param {Mode} mode
```
allejo, line 250:
```suggestion
     * @param {CompiledMode} mode
```

The only time this function is called, it'll always be a `CompiledMode`

allejo (approved):
This seems pretty sane to me

docs/mode-reference.rst
allejo: Oh wait. I forgot about this one. One sec!

allejo, line 185:
```suggestion
*Note: Internally, the array of regular expressions has each separate item converted
into capture groups. However, `manually adding capture groups is not yet supported <https://github.com/highlightjs/highlight.js/issues/3095>`_.
If your regexes use groups, they __must__ be non-capturing groups, i.e. ``(?:regex)``.
```

I'm bad at RST, so if there's a cleaner way of adding links, please fix it

types/index.d.ts
```diff
@@ -206,7 +206,8 @@ type CompiledMode = Omit<Mode, 'contains'> &
         endRe: RegExp
         illegalRe: RegExp
         matcher: any
-        isCompiled: true
+        isCompiled: true,
+        isMultiClass?: boolean,
```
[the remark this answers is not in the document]
joshgoebel: Perhaps technically, but that bugs me though i can't explain it.  This isn't the same as isCompiled IMHO.

PR #3092, opened 2021-03-29
(chore) throttle deprecation messages

### LINKED ISSUE #3069: (chore) Throttle logging of deprecation messages
Would it maybe make sense to throttle logging the same deprecation messages? We're using highlight.js as a dependency of another dependency so can't directly update the deprecated code. The console is simply drowned in the exactly same deprecation message. Once would suffice, IMO, so we could alert the dependency to change it.

_Originally posted by @mihkeleidast in https://github.com/highlightjs/highlight.js/issues/2277#issuecomment-805046833_

mihkeleidast:
Resolves #3069.

### Changes

I've added a simple object map to keep track of already logged deprecations so they would not be logged multiple times.

I've manually tested that the following code now only logs one deprecation message:
```js
function highlight(code) {
  return hljs.highlight("java", code, false);
}

highlight(code);
highlight(code);
highlight(code);
```

### Checklist
- [x] Added markup tests, or they don't apply here because... does not change markup. could not find any existing logger tests so did not add any either.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,10 +7,11 @@
 - enh(nginx) improving highlighting of some sections [Josh Goebel][]
 - fix(vim) variable names may not be zero length [Josh Goebel][]
 - enh(sqf) Updated keywords to Arma 3 v2.02 (#3084) [R3voA3][]
+- (chore) throttle deprecation messages (#3092) [Mihkel Eidast][]
 
 New Languages:
 
-- Added 3rd party Splunk search processing language grammar to SUPPORTED_LANGUAGES (#3090) [Wei Su][]  
+- Added 3rd party Splunk search processing language grammar to SUPPORTED_LANGUAGES (#3090) [Wei Su][]
 
 Theme Improvements:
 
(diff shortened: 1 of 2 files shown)


mihkeleidast:
Looks like `depd` has limited or no browser support - what would be recommended to use here instead? [util-deprecate](https://www.npmjs.com/package/util-deprecate) looks popular but that does not have per-callsite logging.

src/highlight.js
```diff
@@ -967,8 +964,7 @@ const HLJS = function(hljs) {
   @returns {string}
   */
   function deprecateFixMarkup(arg) {
-    logger.deprecated("10.2.0", "fixMarkup will be removed entirely in v11.0");
-    logger.deprecated("10.2.0", "Please see https://github.com/highlightjs/highlight.js/issues/2534");
+    logger.deprecated("10.2.0", "fixMarkup will be removed entirely in v11.0. Please see https://github.com/highlightjs/highlight.js/issues/2534");
```
joshgoebel, line 967: Please don't change the invokes, this is going to conflict with other work.  The only thing we should need to touch here if `logger.js`. Thanks.

mihkeleidast: Yeah no need for that if no need for depd.

joshgoebel (approved):

joshgoebel:
@mihkeleidast Thanks so much!

PR #3098, opened 2021-04-02
Markdown: don't allow inlines to break out of their containing block

polygitapp:
I was running into this issue where an underscore in a url would cause the rest of the markdown document to appear in italics. And double line breaks, headers, etc. would not automatically close it. The main goal of this PR is to fix that and a couple other minor things.

### Changes
* Split the CODE mode into separate modes for inline (CODE_SPAN) and blocks (FENCED_CODE, INDENTED_CODE).
* Enable `endsWithParent: true` on all inline elements (INLINE_HTML, LINK, CODE_SPAN, BOLD, ITALIC) and remove them from the top level mode.
     * However `endsWithParent` does not seem to work when the mode is used inside multiple other modes, so do a deepCopy of all the inline modes before using them. I'm totally open to other ways to workaround this. I'm not familiar with JS enough to know what the right approach here is.
     * Convert the inline modes from using regex literals to strings in order to do the copy.
* Add a new default `Paragraph` mode which contains the inline elements. The Paragraph mode as well as the List mode ends whenever it recognizes the beginning of any of the other top level modes.
* Update italics/bold so that underscore is not allowed within words https://spec.commonmark.org/0.29/#example-359.
* Added a ITALICS_CHILD and BOLD_CHILD mode that only appears inside of ITALICS and BOLD. This is to handle edge cases around nesting italics/bold.
* Updated the top level modes (BLANK_LINE, BLOCKQUOTE, HEADER, HORIZONTAL_RULE, FENCED_CODE, LINK_REFERENCE, LIST, INDENTED_CODE, PARAGRAPH) to begin/end at the beginning/ending of lines. Also updated them to accept unlimited white space in the front because they can all be nested inside of a list.
* Updated the horizontal rule to better match https://spec.commonmark.org/0.29/#thematic-break

There is a failing test in `/test/markup/dart/comment-markdown.txt`. The asterisks get read as italics instead of a bulleted list. Removing the `/// ` it parses correctly as markdown, but as part of a dart it doesn't work for some reason. I couldn't figure out how to fix it. I'm thinking it has to due with newlines? Would appreciate help as to how to fix that test.

```
/// text.
/// * bullet
///   * sub-bullet
```

Had to tweak a couple of the markdown tests. `**_` doesn't work anymore for bold + italics, because of the way it checks if the underscore is inside a word. Also the nested code blocks doesn't work... that seems like a small edge case though.  I think overall it is an improvement.

### Checklist

### DIFF
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -9,60 +9,58 @@ Category: common, markup
 import * as regex from '../lib/regex.js';
 
 export default function(hljs) {
+  let deepCopy = function(x) {
+    return JSON.parse(JSON.stringify(x));
+  };
   const INLINE_HTML = {
-    begin: /<\/?[A-Za-z_]/,
+    begin: '<\\/?[A-Za-z_]',
     end: '>',
     subLanguage: 'xml',
+    endsWithParent: true,
     relevance: 0
   };
   const HORIZONTAL_RULE = {
-    begin: '^[-\\*]{3,}',
-    end: '$'
+    className: 'section',
+    match: '^[ \\t]*([\\*-_])[ \\t]*\\1[ \\t]*\\1( |\\t|\\1)*$',
+    relevance: 0
   };
-  const CODE = {
+  const INDENTED_CODE = {
     className: 'code',
+    begin: '^( {4}|\\t)',
+    end: '^(?!( {4}|\\t))',
+    relevance: 0
+  };
+  const FENCED_CODE = {
     variants: [
-      // TODO: fix to allow these to work with sublanguage also
       {
-        begin: '(`{3,})[^`](.|\\n)*?\\1`*[ ]*'
-      },
-      {
-        begin: '(~{3,})[^~](.|\\n)*?\\1~*[ ]*'
-      },
-      // needed to allow markdown as a sublanguage to work
-      {
-        begin: '```',
-        end: '```+[ ]*$'
-      },
-      {
-        begin: '~~~',
-        end: '~~~+[ ]*$'
-      },
-      {
-        begin: '`.+?`'
-      },
-      {
-        begin: '(?=^( {4}|\\t))',
-        // use contains to gobble up multiple lines to allow the block to be whatever size
-        // but only have a single open/close tag vs one per line
+        begin: '^[ \t]*`{3,}',
+        end: '`{3,}[_\\t]*$',
+        returnBegin: true,
         contains: [
           {
-            begin: '^( {4}|\\t)',
-            end: '(\\n)$'
+            className: 'code',
+            begin: '`{3,}',
+            endsWithParent: true
           }
-        ],
-        relevance: 0
-      }
+        ]
+      },
+      {
+        begin: '^[ \t]*~{3,}',
+        end: '~{3,}[_\\t]*$',
+        returnBegin: true,
+        contains: [
+          {
+            className: 'code',
+            begin: '~{3,}',
+            endsWithParent: true
+          }
+        ]
+      },
     ]
-  };
-  const LIST = {
-    className: 'bullet',
-    begin: '^[ \t]*([*+-]|(\\d+\\.))(?=\\s+)',
-    end: '\\s+',
-    excludeEnd: true
   };
   const LINK_REFERENCE = {
     begin: /^\[[^\n]+\]:/,
+    end: '$',
     returnBegin: true,
     contains: [
       {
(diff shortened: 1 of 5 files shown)


joshgoebel:
Thanks for this! ...but this is very hard to review as-is.  It may need to be broken down into smaller chunks so that each item can be fully understood and discussed individually.  It also adds a lot of complexity and right now it's hard to see if that complexity is truly necessary or if these problems could perhaps be solved more simply.

If each of these items had been in a  separate commit that would be a lot easier to look thru.  Lets start with just one simple thing... HORIZONTAL_RULE. 

- Why is it now relevance 0?
- Why does it now have has a className that it did not have before?

These types of changes should not be made in a large commit without any explanation or justification. That makes me very scary to try and process this as a whole unit for fear of all sorts of "hidden" changes like this.

---

> so do a deepCopy of all the inline modes ... Convert the inline modes from using regex literals to strings 

This is not necessary. `endsWithParent` does indeed alter the child modes (pulling in the end matcher based on the parent), so it will not work when blindly placed inside multiple parents.  I'd first take a look at whether this endsWithParent change was truly necessary.  You don't really say what problem you were trying to solve here... but if so you'd want to use `hljs.inherit` to make copies at the object level.  That's why it exists.

And we definitely want no more strings... the whole library is moving the direction of converting all strings to proper regex... so the reverse is definitely not what we want (though I think you only did that to copy them).

---

Also our recent trend is trying to find ways to make grammars *simpler* when possible (and often it's possible).  Not more complex... so for something like this it'd be better to take one item at a time and first understand the issue with the existing grammar.  It's quite possible there are grammar features (or patterns) you may not even be aware of or better ways to approach many of these things (with far less complexity).  Possibly not, but right now with them all mixed together it's very hard to say.

Are you ok with rewinding this and working thru and helping us understand each of these items on their own merits?

joshgoebel:
If we're going to take things one at a time I think the simple things to tackle first would be:

- horizontal rule
- Split the CODE mode into separate modes for inline (CODE_SPAN) and blocks (FENCED_CODE, INDENTED_CODE). (and what problem this is solving, etc)
- Updated the top level modes (BLANK_LINE, BLOCKQUOTE, HEADER, HORIZONTAL_RULE, FENCED_CODE, LINK_REFERENCE, LIST, INDENTED_CODE, PARAGRAPH) to begin/end at the beginning/ending of lines. Also updated them to accept unlimited white space in the front because they can all be nested inside of a list.

Those three things alone would probably be a more manageable PR to review.

joshgoebel:
> I was running into this issue where an underscore in a url would cause the rest of the markdown document to appear in italics. And double line breaks, headers, etc. would not automatically close it. 

Off the top of my head fixing that might (possibly) only need to touch the URL rule.  Just because in Markdown double line break magically close things doesn't _necessarily_ mean we have to build that logic into every mode in the grammar.  It's certainly possible this is a real improvement, but it also simply may be unneeded complexity.  Again, difficult to say without looking at individual tests/cases.  It's quite common that the first someone someone proposes (including myself) is often overly complex and that further thought and discussion leads to much simpler and more elegant solutions.

polygitapp:
How about starting with disallowing underscores bolding/italicizing when inside words? The horizontal rule was just something I ran into while looking at the markdown spec and doesnt really matter. And its hard to explain the splitting and rearranging of the modes individually. It only works in conjunction with the addition of the paragraph mode etc.

joshgoebel:
> How about starting with disallowing underscores bolding/italicizing when inside words? 

Sure.

> The horizontal rule was just something I ran into while looking at the markdown spec and doesnt really matter. 

Your call - it'd probably be easy to sign off on - but we'd want to have a few markup tests to prove that the regex actually works on all the intended rule variants, etc... (which also serves to document the variants as well).

> And its hard to explain the splitting and rearranging of the modes individually. It only works in conjunction with the addition of the paragraph mode etc.

So the individual code modes have no value on their own - apart from the paragraph stuff?  If so, then we could start by what problem we're trying to solve by adding a paragraph mode.  Over time I've slowly been *removing* contextual modes from grammars, so before we add a new one I want to make sure there isn't some simpler solution to the problem.

joshgoebel:
> I was running into this issue where an underscore in a url would cause the rest of the markdown document to appear in italics. 

An exact code example of this original issue would also be super useful to have attached.

polygitapp:
Sure, here are some examples. Basically anything with an unescaped asterisk or underscore. Which is valid Markdown. Usually I run into it with urls www.example.com?test=foo_bar but it can happen in a lot of situations.

---
### Example 1
```
# Rule of Identity, A*I=A

In this chapter we will learn about...
```
CommonMark.org:
Screen Shot 2021-04-03 at 12 33 04 AM
GitHub:
Screen Shot 2021-04-03 at 12 35 47 AM
Highlight.js:
Screen Shot 2021-04-03 at 12 33 44 AM

---
### Example 2
```
Next Steps:
1. Email hillary_duff@gmail.com
2. ????
3. Profit!
```
CommonMark.org:
Screen Shot 2021-04-03 at 12 40 30 AM
GitHub:
Screen Shot 2021-04-03 at 12 40 57 AM
Highlight.js:
Screen Shot 2021-04-03 at 12 40 18 AM

---
### Example 3 https://gist.github.com/polygitapp/2f6d2e10e3de1b09c3f6530dde6a70e6
GitHub:
Screen Shot 2021-04-03 at 12 47 56 AM
Highlight.js
Screen Shot 2021-04-03 at 12 48 07 AM

joshgoebel:
Awesome... so I'd first take those 3 great test cases and add them all to `markup/markdown/bold_italics.txt`... 

So the general idea is that `_` should only work if proceeded by a space or something, right?  This is what look-behind is for, but of course we can't use that yet (browser support)... so at first glance I'd consider:

- `beforeMatch`
- matching all "word like things" with a separate rule to make sure no other rules trigger inside them (is there any Markdown actually allowed inside a word?)

We could also just include the space in the match, but that's dirty and I would consider it as a last resort.  All of these solutions should be <5 line patches.

That might still leave an edge case with nested bold/italic... but one thing/commit at a time.  Actually matching all word like things vs a beforeMatch might even leave you without an edge case... I'd first get the tests passing, make a small commit for that... then you could add an additional test for the nested bold/italics that *might* be an issue... and it will either pass or we'll have a second commit to tackle it.

joshgoebel:
What is the edge case with bold by the way, just something like:

```
**_bob went to the store_**
```

?

joshgoebel:
Any luck?

joshgoebel:
_Closing for inactivity._  We'd still be happy to review these changes in smaller chunks (as mentioned), but we tend not to leave stale issues open if it seems like someone has moved on to other things.  This can be reopened at any time when a new revision is ready for review - or if easier an entirely new PR can also be opened.

PR #3099, opened 2021-04-03
WIP: Update handlebars syntax for Glimmer and Ember syntaxes

NullVoxPopuli:
It seemed that the handlebars syntax here is pretty old, this PR aims to update it to all the features currently available to Glimmer and Ember projects

### Changes
- implement new syntax identification:
  - [ ] modifiers
  - [x] angle bracket components (previously, they were just XML elements)
  - [ ] named blocks
  - [ ] arguments differentiated from attributes
- new helpers:
  - [x] let
  - [x] fn
  - [ ] has-block
  - [ ] everything from emberjs/rfcs#560, emberjs/rfcs#561, emberjs/rfcs#532
  - [ ] ember-truth-helpers
- new modifiers:
  - [ ] on
- changed classification
  - for Helpers, (previously "builtins"), these have been changed from `name` to `function`, because helpers are functions.

### Checklist

### DIFF
--- a/src/languages/handlebars.js
+++ b/src/languages/handlebars.js
@@ -4,7 +4,7 @@ Requires: xml.js
 Author: Robin Ward <robin.ward@gmail.com>
 Description: Matcher for Handlebars as well as EmberJS additions.
 Website: https://handlebarsjs.com, https://emberjs.com
-Category: common, template
+Category: template
 */
 
 import * as regex from '../lib/regex.js';
--- a/src/languages/handlebars.js
+++ b/src/languages/handlebars.js
@@ -9,51 +9,102 @@ Category: template
 
 import * as regex from '../lib/regex.js';
 
-export default function(hljs) {
-  const BUILT_INS = {
-    $pattern: /[\w.\/]+/,
-    'builtin-name': [
-      'action',
-      'bindattr',
-      'collection',
-      'component',
-      'concat',
-      'debugger',
-      'each',
-      'each-in',
-      'get',
-      'hash',
-      'if',
-      'in',
-      'input',
-      'link-to',
-      'loc',
-      'log',
-      'lookup',
-      'mut',
-      'outlet',
-      'partial',
-      'query-params',
-      'render',
-      'template',
-      'textarea',
-      'unbound',
-      'unless',
-      'view',
-      'with',
-      'yield'
-    ]
-  };
 
-  const LITERALS = {
-    $pattern: /[\w.\/]+/,
-    literal: [
-      'true',
-      'false',
-      'undefined',
-      'null'
-    ]
-  };
+export default function(hljs) {
+// Merged: https://github.com/emberjs/rfcs/pull/560
+  const EQUALITY_HELPERS = [
+    'eq',
+    'neq'
+  ];
+  // Merged: https://github.com/emberjs/rfcs/pull/561
+  const NUMERIC_COMPARISON_HELPERS = [
+    'gt',
+    'gte',
+    'le',
+    'lte'
+  ];
+  // Merged: https://github.com/emberjs/rfcs/pull/562/files
+  const LOGICAL_OPERATOR_HELPERS = [
+    'and',
+    'or',
+    'not'
+  ];
+
+  const OTHER_OPERATORS = [
+    // ember-truth-helpers
+    'not-eq',
+    'xor',
+    'is-array',
+    'is-object',
+    'is-equal'
+  ];
+  const BLOCK_HELPERS = [
+    'let',
+    'each',
+    'each-in',
+    'if',
+    'unless'
+  ];
+  const INLINE_HELPERS = [
+    'log',
+    'debugger',
+    'has-block',
+    'concat',
+    'fn',
+    'component',
+    'get',
+    'hash',
+    'query-params'
+  ];
+
+  const MODIFIERS = [
+    'action',
+    'on'
+  ];
+
+  const LEGACY = [
+    'input',
+    'link-to',
+    'mut',
+    'bindattr',
+    'collection',
+    'in',
+    'loc',
+    'lookup',
+    'partial',
+    'render',
+    'template',
+    'textarea',
+    'unbound',
+    'view',
+    'with',
+    'yield'
+  ];
+
+  const SPECIAL = [
+    'outlet',
+    'yield',
+    'as'
+  ];
+
+  const LITERALS = [
+    'true',
+    'false',
+    'undefined',
+    'null'
+  ];
+
+  const KEYWORD_REGEX = /[\w.\/]+/;
+
+  const HELPERS = [
+    ...SPECIAL,
+    ...LEGACY,
+    ...INLINE_HELPERS,
+    ...OTHER_OPERATORS,
+    ...LOGICAL_OPERATOR_HELPERS,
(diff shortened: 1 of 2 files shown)


src/languages/handlebars.js
```diff
@@ -225,14 +289,28 @@ export default function(hljs) {
...
         end: /\}\}\}\}/,
-        contains: [CLOSING_BLOCK_MUSTACHE_CONTENTS]
+        contains: [ CLOSING_BLOCK_MUSTACHE_CONTENTS ]
+      },
+      // open angle-brocket component invocation
+      {
+        className: 'template-tag',
+        begin: /<:?/,
+        end: />/,
+        contains: [ OPENING_BLOCK_MUSTACHE_CONTENTS, BLOCK_PARAMS, KEYWORD_CONTENTS ]
```
joshgoebel, line 299: You can't put an empty block like KEYWORD_CONTENTS here, it'll do nothing because it has no `match` or `begin` tags.  If you ant to match keywords here you'd put keywords at the top-level.

```suggestion
        className: 'template-tag',
        keywords: // keywords go here
        begin: /<:?/,
        end: />/,
        contains: [ OPENING_BLOCK_MUSTACHE_CONTENTS, BLOCK_PARAMS ]
```

joshgoebel:
@nknapp Would you be around to help review this when it's ready?

joshgoebel:
If you get overwhelmed here I'd suggest to start with smaller chunks (ie, try to accomplish these things a few at a time) vs a huge PR that does it all.  For example you could perhaps clean up the keyword situation first, without trying to add any new tags, etc... so far the PR is "ok" (clean enough for review) but if it gets too complex then we'd ask you to break it into smaller pieces anyways just so we can see what you did and be able to properly review it, etc.

But since this is your first commit and "idk what I'm doing" I'd suggest taking it slow and working in small chunks. (which can be more easily reviewed and merged - which will build momentum and a sense of achievement).

Food for thought.

joshgoebel:
You'll likely need to add $pattern back at some point... may want to review the commit log to see why it was added and what problem it's solving.

joshgoebel:
_Also, is this official handlebars are are these Ember and Glimmer specific extensions?_

src/languages/handlebars.js
```diff
@@ -324,6 +333,12 @@ export default function(hljs) {
         end: /\}\}/,
         keywords: 'else if'
       },
+      {
+        className: 'bulit_in',
```
joshgoebel, line 337: typo :-)

src/languages/handlebars.js
```diff
@@ -3,8 +3,8 @@ Language: Handlebars
 Requires: xml.js
 Author: Robin Ward <robin.ward@gmail.com>
 Description: Matcher for Handlebars as well as EmberJS additions.
-Website: https://handlebarsjs.com
-Category: template
+Website: https://handlebarsjs.com, https://emberjs.com
+Category: common, template
```
joshgoebel, line 7: Please revert, this is not your decision to make - which languages are in our `common` set.  If you just want to add handlebars to the build your an do so on the console:

```
./tools/build -t node :common handlebars
```

Etc.

---

Also what is embers relationship to Handlebars?

src/languages/handlebars.js
```diff
@@ -205,7 +205,7 @@ export default function(hljs) {
   };
 
   const SUB_EXPRESSION_CONTENTS = hljs.inherit(HELPER_NAME_OR_PATH_EXPRESSION, {
-    className: 'name',
+    className: 'function',
```
joshgoebel, line 208: Please make sure that your PR explanation justifies (visual) changes like this and explains why they are necessary...

NullVoxPopuli: can do!

In this particular case, helpers _are_ functions (and take args), and not actually just a name, like you'd have with the name of a Component invocation.

NullVoxPopuli:
Thanks for the early comments! these are helpful!!

> You'll likely need to add $pattern back at some point... may want to review the commit log to see why it was added and what problem it's solving.

I'll check it out :+1:

> Also, is this official handlebars are are these Ember and Glimmer specific extensions?

handlebars doesn't have components, so about everything I'm doing is going to be on top of handlebars, but for ember and glimmer (they use the same hbs file extension though). 

I saw there was an HTMLBars syntax, but it looked like it was going to be removed.

I can move all of this into a new ember/glimmer specific syntax, if that's desired and try extending from handlebars.

src/languages/handlebars.js
NullVoxPopuli: > Also what is embers relationship to Handlebars?

ember has similar syntax, but has built a bunch of stuff on top, as a result, the handlebars syntax in highlight.js ~7 years out of date :(
(but that's why I'm making a PR!) :D

joshgoebel: > as a result, the handlebars syntax in highlight.js ~7 years out of date

I'm not sure that's a correct statement [in the general sense]. I know @nknapp put a lot of work into it just last year to improve it.

https://handlebarsjs.com  

AFAIK Handlebar exists as it's own independent project and there are surely those using it outside of Ember - _like our very own project, actually_.   So the question then becomes whether this is appropriate to be added to Handlebars at all _or whether a separate Ember grammar may be needed_.  I'm curious if nknapp has any thoughts on that. 

In any case I can certainly see how this would be helpful for Ember users, but if we decide it's not appropriate as part of Handlebars proper then it would need to be distributed as a 3rd party Ember grammar - rather than mixed with our existing support.

NullVoxPopuli:
I have a test template that I'm working through here in the developer.html: 
image
(taken from: https://github.com/joukevandermaas/vim-ember-hbs/pull/13)

My plan was to get the whole thing looking correct before taking the PR out of draft. 
Most of it should  "just" be adding component invocation support

src/languages/handlebars.js
NullVoxPopuli: Yeah that's totally fair! Most editors have an ember-variant to use for hbs highlighting

joshgoebel:
> I saw there was an HTMLBars syntax, but it looked like it was going to be removed.

Correct, also it's only an alias of Handlebars at this point.

> I can move all of this into a new ember/glimmer specific syntax, if that's desired and try extending from handlebars.

That is indeed the question.  Though at that point it becomes a 3rd party module so I wouldn't worry about "extending" anything (you could, but we don't really encourage or support run-time dependencies)... at that point you'd just be "forking" the grammar essentially and then carrying it forward as Ember/Glimmer specific.  But perhaps that's all you meant by "extending" in the first place. :-)

NullVoxPopuli:
I'm gonna move all this to a new file, called `glimmer.js`, and that way we don't need to worry about handlebars maintenance / maintainers also having to deal with ember/glimmer stuff. 
It also gives me a chance to start over, and make sure only what's needed is implemented -- handlebars has some features that ember/glimmer no longer has

joshgoebel:
> handlebars has some features that ember/glimmer no longer has

This definitely seems to since the deal that this is an entirely separate grammar, not Handlebars (though I was already leaning this way).

At this point you'd be creating a 3rd party grammar module (in your own repository) so I'd check out our documentation on that:

https://github.com/highlightjs/highlight.js/blob/main/extra/3RD_PARTY_QUICK_START.md

NullVoxPopuli:
Cool! thanks for having all this documentation written up!

PR #3114, opened 2021-04-07
add(nestedtext) NestedText support

joshgoebel:
- See https://nestedtext.org/

### Checklist


allejo (changes requested):
Taken from https://nestedtext.org/en/latest/examples.html, chunks of text are completely disappearing from the highlighted output.

```
debug: false
secret_key: t=)40**y&883y9gdpuw%aiig+wtc033(ui@^1ur72w#zhw3_ch

allowed_hosts:
  - www.example.com

database:
  engine: django.db.backends.mysql
  host: db.example.com
  port: 3306
  user: www

webmaster_email: admin@example.com
```

image

joshgoebel:
Bug in the parser that was already fixed in main.  I've rebased.

src/languages/nestedtext.js
```diff
@@ -0,0 +1,85 @@
...
+          /[ ]/,
+          /.*$/
+        ]
+      },
+      {
+        match: [
+          /^\s*/,
+          /-$/
+        ]
+      }
```
allejo, line 63:
```
list_of_dictionaries:
  -
    hello: world
    key: value

list_of_lists:
  -
    - green
    - blue
    - red
```

Would you mind adding this snippet to the markup text? This pattern wasn't immediately clear to me when reading through the NT spec. I can't help but feel this could be overlooked if not in our tests.

allejo (approved):
lgtm

src/languages/nestedtext.js
joshgoebel: Done.

PR #3116, opened 2021-04-08
(chore) consistency improvements with SAS

joshgoebel:
### Changes

- Fix some apparent bugs in grammar (including parens in matches when
they should not be, etc)
- More consistent application of class names
- Use new `title.function`
- Contextually detect function titles in a few additional places
- Explode keyword lists 

### Checklist

### DIFF
--- a/src/languages/sas.js
+++ b/src/languages/sas.js
@@ -84,7 +84,7 @@ export default function(hljs) {
       literal:
         'null missing _all_ _automatic_ _character_ _infile_ ' +
         '_n_ _name_ _null_ _numeric_ _user_ _webout_',
-      meta:
+      keyword:
         SAS_KEYWORDS
     },
     contains: [
@@ -99,10 +99,15 @@ export default function(hljs) {
         begin: /&[a-zA-Z_&][a-zA-Z0-9_]*\.?/
       },
       {
-        // Special emphasis for datalines|cards
-        className: 'emphasis',
-        begin: /^\s*datalines|cards.*;/,
-        end: /^\s*;\s*$/
+        begin: [
+          /^\s*datalines;|cards;/,
+          /(?:.*\n)+/,
+          /^\s*;\s*$/
+        ],
+        className: {
+          1: "keyword",
+          2: "string"
+        }
       },
       { // Built-in macro variables take precedence
         className: 'built_in',
@@ -110,12 +115,12 @@ export default function(hljs) {
       },
       {
         // User-defined macro functions highlighted after
-        className: 'name',
+        className: 'title.function',
         begin: /%[a-zA-Z_][a-zA-Z_0-9]*/
       },
       {
         className: 'meta',
-        begin: '[^%](' + SAS_FUN + ')[\(]'
+        begin: '[^%](' + SAS_FUN + ')(?=\\()'
       },
       {
         className: 'string',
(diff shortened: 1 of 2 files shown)


joshgoebel:
CC @mcaceresb  

I came across this when working on our themeing and CSS classes for v11.  `emphasis` is intended for markup (and this may become clearer in the future), not randomly applying to things you might wish to be italic.  Indeed SAS was the only grammar using `emphasis` outside of markup grammars (Markdown, etc).  So I've changed this to `string`.

You're also welcome to join the discussion on #2500 or #2521.  `meta` isn't the proper classification for a function, but I'm not sure what to replace it with so I left it.

mcaceresb:
@joshgoebel I think my idea behind using `emphasis` for datalines/cards was because I wanted it to be distinct from other highlighting (SAS' editor highlights datalines/cards with a light yellow, which it doesn't do anywhere else, so I figured I would pick something that might be distinct from other stuff that is highlighted).

joshgoebel:
>  I think my idea behind using emphasis for datalines/cards was because I wanted it to be distinct from other highlighting (SAS' editor highlights datalines/cards with a light yellow, which it doesn't do anywhere else

Yes, I certainly understand the motivation, but that's not how it should be thought about from a grammar perspective . The idea as a grammar writer is to describe what elements of code are, _not how they should be highlighted_.  You say "this is a keyword" not "this should be bold and green and underlined".  "These are datelines", not "these should be emphasized".  Perhaps one theme wishes to emphasize them, perhaps another does not.  That is a theming decision.

You might think keywords (or datelines) should be emphasized but some themes could decide to make them the most subtle color imaginable - _or even purposely not highlight them at all_...  So in this case we need to find the best *semantic* description of what these lines of code are.  We're rethinking this stuff [categorizations] (with the issues I linked to) but long term this might be more of a:

- string.unquoted.dataline

If you want a theme that makes all datelines bright yellow (or emphasizes them any other way you could imagine), you could certainly write one (and some people write canonical themes)... but for most themes not specifically designed with your grammar in mind then it's just an unquoted string.

If you want to read thru the TextMate scoping stuff and have a better idea of what a "dateline" might be than an unquoted string I'd be all ears.  Not that we use TextMate scopes, but they are something ever present on my mind as we rethinking our CSS classes.  If there is a VS Code extension for SAS you may want to check and see how it scopes datelines.

- https://macromates.com/manual/en/language_grammars
- https://www.sublimetext.com/docs/3/scope_naming.html

mcaceresb:
@joshgoebel I agree with you, and your explanation certainly makes more sense than my logic from back then. I think having it be an unquoted string makes _some_ sense, so to having it like this at the moment seems reasonable.

src/languages/sas.js
```diff
@@ -4,88 +4,492 @@ Author: Mauricio Caceres <mauricio.caceres.bravo@gmail.com>
...
 */
 
+import * as regex from '../lib/regex.js';
+
+/** @type LanguageFn */
 export default function(hljs) {
   // Data step and PROC SQL statements
-  const SAS_KEYWORDS =
-    'do if then else end until while ' +
-    '' +
```
joshgoebel, line 11: @mcaceresb  Is there a reason for these line breaks?  Are there implicit headings we should add and make explicit?

mcaceresb: @joshgoebel  I don't think so. I think it was just a visual aid when I was coding.

PR #3119, opened 2021-04-09
chore(themes) remove `builtin-name` CSS class

joshgoebel:
### Changes

Removed the `builtin-name` replacing it with `built_in` in the few (4-5) 
grammars it was used.  Also only 4 themes made any distinction between 
built_in and builtin-name, the rest choosing to style these exactly the
same - so it's never been clear what the difference is or there is none.

If more nuance is needed here it should be addressed via #2500 and #2521.

In only one grammar was both `built_in` and `builtin-name` used side by
side so other than that (in combo with very particular themes) this
should be a quiet change.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,7 @@ New Languages:
 
 Theme Improvements:
 
+- chore(themes) remove `builtin-name` CSS class (#3119) [Josh Goebel][]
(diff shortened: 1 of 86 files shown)


joshgoebel (commented):
Looks good.

PR #3123, opened 2021-04-10
Add Glimmer to SUPPORTED_LANGUAGES

NullVoxPopuli:
Adds Glimmer highlighting plugin to list of supported languages

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,6 +38,7 @@ Language Grammars:
 - Default CDN build drops support for several languages.
 - Some language grammar files have been removed.
 - Some redundant language aliases have been removed.
+- Added 3rd party Glimmer grammar to SUPPORTED_LANGUAGES(#3123) [NullVoxPopuli][]
 
 ### Other changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,7 +38,6 @@ Language Grammars:
 - Default CDN build drops support for several languages.
 - Some language grammar files have been removed.
 - Some redundant language aliases have been removed.
-- Added 3rd party Glimmer grammar to SUPPORTED_LANGUAGES(#3123) [NullVoxPopuli][]
 
 ### Other changes
 
@@ -79,6 +78,7 @@ Grammars:
 
 New Languages:
 
+- Added 3rd party Glimmer grammar to SUPPORTED_LANGUAGES(#3123) [NullVoxPopuli][]
 - Added 3rd party Splunk search processing language grammar to SUPPORTED_LANGUAGES (#3090) [Wei Su][]
 - Added 3rd party ZenScript grammar to SUPPORTED_LANGUAGES(#3106) [Jared Luboff][]
 - Added 3rd party Papyrus grammar to SUPPORTED_LANGUAGES(#3125) [Mike Watling][]
--- a/SUPPORTED_LANGUAGES.md
+++ b/SUPPORTED_LANGUAGES.md
@@ -78,6 +78,7 @@ Languages that listed a **Package** below are 3rd party languages and are not bu
 | GAUSS                   | gauss, gss             |         |
 | GDScript                | godot, gdscript        | [highlightjs-gdscript](https://github.com/highlightjs/highlightjs-gdscript) |
 | Gherkin                 | gherkin                |         |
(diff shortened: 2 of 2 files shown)


joshgoebel:
Now I'm trying to figure out what to do (if anything) about the fact that you take the same aliases we ship... so installing your plugin REPLACES our built-in functionality... I think that's a first.  Not sure if we need to disallow that (probably too drastic) or merely call it out, or what...

joshgoebel:
I'm thinking since Handlebars still exists outside of Glimmer/Ember ecosystem that the aliases (by default) might not be correct.  That someone should have to manually link them up if that's the behavior they desire.

NullVoxPopuli:
> I assume you think that is a feature not a bug? :)

yes, a feature :D 

The issue primarily comes from Glimmer files reusing the hbs extension.... but... 🤷 

However,

>  so installing your plugin REPLACES our built-in functionality

My plugin doesn't actually register itself, you need to do that manually via: 
```js
import hljs from 'highlight.js';
import { glimmer } from 'highlightjs-glimmer';

hljs.registerLanguage('glimmer', glimmer);
hljs.highlightAll();
```

which, idk.. I _could_ do in the plugin, but that's a side-effect, and side-effects are generally less flexible in broader ES modules (even though probably totally fine for how highlightjs is used)

joshgoebel:
> My plugin doesn't actually register itself, you need to do that manually via:

You should also ship a CDN distributable in `dist` and it should self register as your name, ie `glimmer`.  I'll create the same issue I filed for others recently.

_Aliases always self-register though whenever the language is registered with `registerLanguage`._

> The issue primarily comes from Glimmer files reusing the hbs extension

Which means that `hbs` is now ambiguous in meaning... which is problematic.  I'm thinking perhaps your alias list should instead be empty and your README should show a one-liner for how someone could hook up those aliases if that is their desire - to replace those aliases?

Trying to avoid surprising behavior.

NullVoxPopuli:
> You should also ship a CDN distributable in dist and 

workin on that now :+1:

> Trying to avoid surprising behavior.

I get that -- and I understand that this may be surprising behavior to _someone_, but Glimmer is _kind of_ a superset of handlebars, _buuuuut_, more importantly, I have a feeling the number of people in the world that would want _both_ syntaxes is < 10

joshgoebel:
>  _kind of_ a superset

kind of indeed.  And then the question is why not go all the way and also clobber `handlebars` itself - if installing glimmer truly means "I want glimmer everywhere"?  Or were you giving it a bit more lee way as it's the actual canonical name of our built-in grammar?

I think at a minimum your README should perhaps mention (as a "caveat") that it clobbers those aliases and someone would need to re-register them to point to `handlebars` if they did not desire that behavior.

@allejo Any thoughts on this?

NullVoxPopuli:
> actual canonical name of our built-in grammar?

yeah, moreso this. The term handlebars is rarely used in the Glimmer/Ember world, except when talking about the file extension itself.

> I think at a minimum your README should perhaps mention

I did add this (very recently): 

> NOTE highlightjs-glimmer cannot be used at the same time as highlightjs' handlebars syntax highlighting.

joshgoebel:
> NOTE highlightjs-glimmer cannot be used at the same time as highlightjs' handlebars syntax highlighting.

But that's wholly inaccurate (or at the very least imprecise) no? One can easily do:

```
<pre><code class="lang-glimmer">...</code></pre>
<pre><code class="lang-handlebars">...</code></pre>
```

So long as you're leaving `handlebars` alone (which I do think is correct), then this should work just fine.  What you're breaking is the aliases, not the whole grammar.

NullVoxPopuli:
> But that's wholly inaccurate (or at the very least imprecise) no? One can easily do:

:+1: I'll update the verbiage: https://github.com/NullVoxPopuli/highlightjs-glimmer/pull/10/commits/4a239f25cd967af0ad34a0e02ac1d288ceea62a8

NullVoxPopuli:
Success! 
image
https://github.com/NullVoxPopuli/highlightjs-glimmer/runs/2315197032

joshgoebel:
> 👍 I'll update the verbiage: NullVoxPopuli/highlightjs-glimmer@4a239f2

@allejo Any objections?  I'm "ok" after this README change I think.

joshgoebel:
Add to changes.md too please?

joshgoebel (approved):

joshgoebel:
@NullVoxPopuli Thanks!

PR #3126, opened 2021-04-12
chore(dev) improve match scope visualization

NullVoxPopuli:
this change trades vertical space to reclaim some horizontal space
that was previously lost when inspecting the class structure of the
emitted hljs highlight.

As a bonus, it's slightly easier to see hierarchy between parent/child
elements because the visible structure text is stacked / indented
like a tree, rather than inline, like it was previously.

Example here: https://hljs-glimmer.nullvoxpopuli.com/?debug=true

### Checklist
      no markup changes 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,11 +33,16 @@ Theme Improvements:
 - chore(themes) remove `builtin-name` CSS class (#3119) [Josh Goebel][]
 - chore(theme) Update GitHub theme css to match GitHub's current styling (#1616) [Jan Pilzer][]
 
+Dev Improvements:
+
+- Change the class structure inspector to utilize more vertical space
+
 [Josh Goebel]: https://github.com/joshgoebel
 [Ryan Mulligan]: https://github.com/ryantm
 [R3voA3]: https://github.com/R3voA3
 [Wei Su]: https://github.com/swsoyee
 [Jared Luboff]: https://github.com/jaredlll08
+[NullVoxPopuli]: https://github.com/NullVoxPopuli
 
 ## Version 10.7.1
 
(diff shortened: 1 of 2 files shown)


tools/developer.html
```diff
@@ -37,23 +37,37 @@
...
       border-radius: 5px;
       white-space: pre;
+
+      position: relative;
     }
 
     .visible-structure pre code span:before {
       display: inline-block;
-      content: attr(data-klass);
+      content: attr(class);
```
joshgoebel, line 51: Please revert this.

NullVoxPopuli: done :+1:

NullVoxPopuli:
oh holup, I can fix all the slight alignment issues

joshgoebel:
I'll take a closer look later, I wonder though if we can re-line up the content a bit better with an offset or negative margin on the blocks?
Screen Shot 2021-04-11 at 9 57 56 PM

IE, if this was offset better... not a big deal, just wondering.

NullVoxPopuli:
Yup, just remembered about align-items: baseline.

image

joshgoebel:
Damn.  That's sharp.

NullVoxPopuli:
Updated!

NullVoxPopuli:
Here is how it looks on my silly Glimmer stuff 
image

CHANGES.md
```diff
@@ -35,7 +35,7 @@ Theme Improvements:
 
 Dev Improvements:
 
-- Change the class structure inspector to utilize more vertical space
+- (chore) greatly improve match scope visualization in dev tool (#3126) [NullVoxPopuli][]
```
NullVoxPopuli, line 38: you're too kind ;)

joshgoebel (approved):

Issue #3118, opened 2021-04-09
Task: Create theme maintainers team and see who is still wanting to help maintain themes


joshgoebel:
Need to create a theme maintainers list so that theme maintainers who desire can keep their themes up-to-date.  

The overall idea being that themes (like grammars) are never "one and done" that over time they require a (small) amount of maintenance to keep up with changes in the core library, grammars, theming in general, etc.  Themes that have no owner/maintainer should possibly be moved into an `old_themes` repository/folder with a README that indicates they may not provide the best results compared to core themes.

_Of course we'd love it if all themes were maintained and provided a great experience._

What kind of things require upkeep?  See #2500, #2521, etc... plus the recent additions of `property`, `punctuation`, `operator`, and other prior and upcoming changes... but **this issue is just trying to get everyone onto a team so I can more easily contact you**.  After that I'll focus on typing up a list of things you may want to consider when refreshing your themes to take advantage of recent changes and improvements.

### Added by HLJS Team

- [x] default.css
- [ ] mono-blue.css
- [x] ascetic.css (Ivan)
- [x] dark.css (Ivan)
- [ ] googlecode.css

### Unmaintained

- [x] androidstudio.css
- [x] arduino-light.css (not with Arduino anymore, forwarded request along to Arduino peoples)
- [x] arta.css (pumbur@pumbur.net) (email bounced)
- [x] docco.css
- [x] ir-black.css (vaskas@programica.ru) (email bounced)
- [x] vs.css
- [x] foundation.css -  " Since Foundation has moved on substantially, I'm not even sure the existing theme makes much sense anymore. It would be best if someone recreated it anew."
- [ ] gml.css - (thumbs down on maintaining)

### Base16

These should be auto-build from Base16 sources... see #3132.

- [x] atelier-cave-dark.css
- [x] atelier-cave-light.css
- [x] atelier-dune-dark.css
- [x] atelier-dune-light.css
- [x] atelier-estuary-dark.css
- [x] atelier-estuary-light.css
- [x] atelier-forest-dark.css
- [x] atelier-forest-light.css
- [x] atelier-heath-dark.css
- [x] atelier-heath-light.css
- [x] atelier-lakeside-dark.css
- [x] atelier-lakeside-light.css
- [x] atelier-plateau-dark.css
- [x] atelier-plateau-light.css
- [x] atelier-savanna-dark.css
- [x] atelier-savanna-light.css
- [x] atelier-seaside-dark.css
[...]

PR #3130, opened 2021-04-13
upgrade srcery theme

redguardtoo:
upgrade srcery  theme

Update theme srcery with minor change, as I was enlisted in #3118

### Changes
Add code comment, adjust a few entity's color

### Checklist
- [x] they don't apply here because it's minor css change


joshgoebel:
Did you mean to exclude operator and punctuation?

src/styles/srcery.css
```diff
@@ -2,76 +2,92 @@
...
+/* Bright black */
+.hljs-link,
+.hljs-number,
 .hljs-comment,
-.hljs-deletion,
 .hljs-meta {
   color: #918175;
 }
+
+.hljs,
```
joshgoebel, line 76: This looks incorrect.  `hljs-subst` should perhaps get BG and FG applied, but not different layout attributes.  Should `subst` be split out with just the color and background?

redguardtoo: @joshgoebel , thanks for the review. I force pushed the latest code.

I didn't do any radical change. I follow the Emacs implementation of same theme and make some small adjustment.

joshgoebel:
I added some bits to the header and removed the "default" boilerplate (since the build process adds that now).

PR #3133, opened 2021-04-13
fix(devtool): use correct CSS to hide psuedo element used for hacking…

NullVoxPopuli:
… the width of the wrapping content to be *at least* the width of the green class tag / decoration

### Checklist


joshgoebel:
This was broken for me in that the padding on the bottom was negative or something weird.  How does this look for you?

NullVoxPopuli:
Just pushed this 
image
image
image
image

So far, it seems the best way to align things is to not touch the display property on the span element... which is obnoxious, because it makes aligning things to the baseline hard

NullVoxPopuli:
_but_, I did find a _really_ weird display option that allowed the border to expand... idk 🤔 Gonna keep looking

joshgoebel:
What was wrong with what I showed you again?

NullVoxPopuli:
I don't remember you showing me anything?

NullVoxPopuli:
also, I think this is in the best place _I_ can get it for now.
I am not happy with the various ways that blocks are broken up and separated from wrapping elements (because of how the code is wrapped with spans).

For example:

```
<style>
  code
</style>
```
all 3 of those lines end up being sibling elements, which makes any sort of `display: inline-*` not great. _but_, we need inline-flex for align-items: baseline. 
_but_, with inline-flex, _often_, entire sections of code are moved into one line, which is _worse_. :upside_down_face:

joshgoebel:
Grrrr, I really thought I pushed a commit... let me take a look at this latest version.

NullVoxPopuli:
@joshgoebel Rebased and fixed the text-alignment issue (I think)

NullVoxPopuli:
image


There is also a compressed version if we want 🤷 
image

joshgoebel:
Compressed please.

NullVoxPopuli:
Spacing reduced and PR rebased

PR #3135, opened 2021-04-14
enh(parser) className => scope, use "hybrid" pproach

### LINKED ISSUE #3085: Discuss: Change className to scope

Yes, once we add nested scope then technically className is no longer accurate, plus it's length has always bothered me.

I propose we rename it scope...

Before:

```
      {
        className: 'title.class',
        match: '^\\w+(?=:)'
      },
      {
        className: 'literal',
        match: '^-'
      },
```

After:

```js
      {
        scope: 'title.class',
        match: '^\\w+(?=:)'
      },
      {
        scope: 'literal',
        match: '^-'
      },
```

Nope.  TextMate grammars calls this `captures/name`:

```
 {  match = '(@selector\()(.*?)(\))';
    captures = {
       1 = { name = 'storage.type.objc'; };
       3 = { name = 'storage.type.objc'; };
    };
 };
```

**Additional context...**

None.

joshgoebel:
### Changes

Changes nested scopes to the hybrid TM approach suggested over
in the [Prism thread on the subject](https://github.com/PrismJS/prism/issues/2849).

Resolves #3085 

### Checklist

### DIFF
--- a/docs/css-classes-reference.rst
+++ b/docs/css-classes-reference.rst
@@ -1,22 +1,21 @@
-CSS classes reference
-=====================
+Scopes Reference
+================
 
 
-Stylable classes
+Stylable Scopes
 ----------------
 
-The general purpose classes are intended to be used for any language, but the
-other classes may also be used when they are semantically correct.  For example
-if you had a general purpose language that allowed inline URLs:
+The general purpose scopes are intended to be used for any language, but the
+other scopes may also be used when semantically correct.  For example if you had
+a general purpose language that allowed inline URLs:
 
 ::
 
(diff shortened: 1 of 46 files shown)


allejo (approved):
Shouldn't our [Mode Reference documentation](https://github.com/highlightjs/highlight.js/blob/main/docs/mode-reference.rst#classname) also be touched in this PR? i.e. specifying that `className` is deprecated in favor of `scope`?

Otherwise, this lgtm

joshgoebel:
I'll take another look, this may have grown from where it started and perhaps I still need to update a few docs.

PR #3138, opened 2021-04-14
enh(parser) allow multi-class capture groups

### LINKED ISSUE #3095: Todo: Multi-class should allow capturing match groups

https://github.com/highlightjs/highlight.js/pull/3081

Multi-class support does not currently allow capturing match groups.  _This should be allowed._

We should count the match groups, renumber them, and also renumber the keys passed to `className` (to account for any new "gaps" that the capture groups have introduced).  We have most of the code for this in `regex.js` already since we have to do a lot of this for our MultiRegexMatcher.

None.

**Additional context...**

None.

joshgoebel:
Resolves #3095

### Changes

Renumbers multi-class capture groups to allow for inner capture groups without breaking everything.

- Also rewrites backreferences within the individual regexes,
- Caveat: There is still no way for a back-reference from one regex to match a reference in a DIFFERENT regex (even though we know they are just getting joined).  We could allow this, but I'm not sure which way is more confusing.

I'm not sure if that caveat will become relevant or not?

### Explain

Lets say we 3 match scopes:

`{ 1 => ..., 2 => ..., 3 => ... }`

So what we need is a clean match like this:

`(a)(b)(c) => [ "a", "b", "c" ]`

But this falls apart with inner match groups:

`(a)(((b)))(c) => ["a", "b", "b", "b", "c" ]`

Our scopes are now "out of alignment" and we're repeating `b` 3 times.
What needs to happen is the numbers are remapped:

`{ 1 => ..., 2 => ..., 5 => ... }`

We also need to know that the ONLY groups that should be output
are 1, 2, and 5.  This PR handles this behavior correctly.
 

### Checklist

### DIFF
--- a/test/api/multiClassMatch.js
+++ b/test/api/multiClassMatch.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const hljs = require('../../build');
+hljs.debugMode();
 
 describe('multi-class matchers', () => {
   before(() => {
@@ -36,7 +37,7 @@ describe('multi-class matchers', () => {
           },
           {
             match: [
-              /func/,
+              /((func))/,
               /\(\)/,
               /{.*}/
             ],
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -249,11 +249,13 @@ const HLJS = function(hljs) {
       let i = 1;
       // eslint-disable-next-line no-undefined
       while (match[i] !== undefined) {
+        if (!mode._emit[i]) { i++; continue; }
         const klass = language.classNameAliases[mode.className[i]] || mode.className[i];
         const text = match[i];
-        if (klass) { emitter.addKeyword(text, klass); } else {
+        if (klass) {
+          emitter.addKeyword(text, klass);
+        } else {
           modeBuffer = text;
-          // emitter.addText(text);
           processKeywords();
         }
         i++;
--- a/src/lib/ext/multi_class.js
+++ b/src/lib/ext/multi_class.js
@@ -5,7 +5,52 @@ import * as regex from "../regex.js";
 const MultiClassError = new Error();
 
 /**
+ * Renumbers labeled scope names to account for additional inner match
+ * groups that otherwise would break everything.
  *
+ * Lets say we 3 match scopes:
+ *
+ *   { 1 => ..., 2 => ..., 3 => ... }
+ *
+ * So what we need is a clean match like this:
+ *
+ *   (a)(b)(c) => [ "a", "b", "c" ]
+ *
+ * But this falls apart with inner match groups:
+ *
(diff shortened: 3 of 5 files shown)


joshgoebel:
> There is still no way for a back-reference from one regex to match a reference in a DIFFERENT regex (even though we know they are just getting joined). We could allow this, but I'm not sure which way is more confusing.

I think the solution here (in the long-run) is to allow the full regex to be carefully written out by the user:

```
match: /(abc)((?:def))(\1)/,
scope: {
  1: "first", 2: "second", 3: "third"
}
```

And then add sanity checks to make sure the output QUANTITY is at least correct... we expect 9 characters, there are 9 characters total in all match groups. Of course this would again disallow inner capture groups... so I'm a fan of putting off this problem for another day and see what comes up in the real world.

test/api/multiClassMatch.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const hljs = require('../../build');
+hljs.debugMode();
```
allejo, line 4:
```suggestion
```

Probably want to remove this?

allejo (approved):
Ok this one took a while to wrap my head around but I like this implementation :+1:

allejo:
> I'm a fan of putting off this problem for another day and see what comes up in the real world.

+1 for waiting on this, I don't _think_ there are enough pros for warrant investing time for this

test/api/multiClassMatch.js
joshgoebel: Nope, tests should be running in debug mode to fully raise errors.

joshgoebel:
> Ok this one took a while to wrap my head around but I like this implementation 

I'm not sure how else to do this... all this is to work around the fact that in JS you can't get positional data on capture groups... so the only way to do it is make sure every character in the string is in a capture group.  This is why things like TextMate grammars can't run in the browser easily because they depend heavily on this ability to do their own multi-matching.

At least not unless this happens: https://github.com/tc39/proposal-regexp-match-indices

joshgoebel:
A nicer way MIGHT be to rewrite the regex programmatically to add the needed capture groups:

```
/(a)bcd(e)/ => /(a)(bcd)(e)/
```

Along with the same renumeration of course... that might be pretty straight forward, but for now I'll leave that in the "for another day" bucket also. :-)  I think the array syntax turned out pretty decent.

joshgoebel:
- Renumbers multi-class capture groups to allow for inner capture groups without breaking everything.
- Also rewrites back-references within those individual regexes to still be correct.

Caveat: There is still no way for a back-reference from one regex to match a reference in a DIFFERENT regex (even though we know they are just getting joined). We could allow this, but I'm not sure which way is more confusing.

PR #3139, opened 2021-04-14
enh(themes) Add all Base16 themes

### LINKED ISSUE #3132: Task: Build Base16 themes auto-magically
Regarding the other themes, I can't avoid noticing that many of them are from the Base16 project.

In the [Highlight] project, we simply automated the creations of themes using Mustache templates, which greatly simplifies any updates (e.g. when new syntax elements are introduced).

For the task, I've created some scripts, templates, and a slightly updated set of the Base16 YAML definitions, in order to correctly provide names and links of their original authors, and also split between Dark and Light themes (not every B16 theme supports both versions, although some do).

Here are my scripts and assets (feel free to use them, some are public domain and others are MIT licensed):

- https://gitlab.com/tajmone/highlight-extras/-/tree/master/base16
- https://github.com/tajmone/Base16-Sass

Via this automation system you might even end up with more HLJS Base16 themes than you started with, since I've added some extra themes to the list.

[Highlight]: http://www.andre-simon.de/

_Originally posted by @tajmone in https://github.com/highlightjs/highlight.js/issues/3118#issuecomment-817093587_

joshgoebel:
### Changes

Keep our base16 themes separate from the rest and add builds of other Base16 themes.

Resolves #3132.

### Checklist


NullVoxPopuli (approved):
I much prefer this separation :D

Hirse:
What happened to the sorting of the themes in `developer.html`?
Is this intentional?
Unsorted list of themes

joshgoebel:
Nope, accident.  Honestly though this shouldn't be manually updated at all.  We should probably make the whole thing a template and build it (with auto-populated CSS list) when we build demo.  This would also allow us to point people to the developer tool online in the future (for testing), which would be a big win.

PR #3141, opened 2021-04-16
(chore) simplify swift

joshgoebel:
### Changes

Take advantage of new features to simplify swift

### Checklist


joshgoebel:
@svanimpe Any chance you could review this?  Simplies swift using all the new stuff... all tests pass - the only changes to tests are to deal with the function => title.function updates...

src/languages/swift.js
```diff
@@ -399,11 +391,12 @@ export default function(hljs) {
...
-    match: /\b(subscript|init[?!]?)\s*(?=[<(])/,
-    keywords: {
-      keyword: "subscript init init? init!",
-      $pattern: /\w+[?!]?/
+    match: [
+      /\b(?:subscript|init[?!]?)/,
+      /\s*(?=[<(])/,
+    ],
+    className: {
+      1: "keyword"
```
joshgoebel, line 399: I wonder if this shouldn't be `title.function` instead of keyword?

NullVoxPopuli (approved):
I don't know swift, but this looks like a fairly straightforward change

src/languages/swift.js
svanimpe: If you mean `subscript` and `init`, those are just keywords. Subscripts and initializers are type members that don't have a title.

PR #3147, opened 2021-04-17
enh(themes)  devibeans - inspired by vim's jellybeans theme

terminaldweller:
Added a new theme inspired by vim's jellybeans theme. A live preview can be viewed [here](http://terminaldweller.com:9000).

### Checklist


joshgoebel:
Please update your header to include tagged fields:

```
/*!
   Theme: a11y-light
   Author: @ericwbailey
   Maintainer: @ericwbailey

   Based on the Tomorrow Night Eighties theme: https://github.com/isagalaev/highlight.js/blob/master/src/styles/tomorrow-night-eighties.css
 */
```

ie, `Theme:`, `Author:`, etc...

And can you please attempt a snap or two so someone can see it here for review without having to jump to an external website?  Also please show an HTML and Markup example as well so we can see how it looks highlighting the major different types of things we support.

src/styles/devibeans.css
```diff
@@ -1,6 +1,8 @@
-/*
-    devibeans (dark)
-    Created by Farzad Sadeghi (http://github.com/terminaldweller)
+/*!
+    Theme: devibeans(dark)
+    Author: @terminaldweller
```
joshgoebel, line 3: You can spell your name out here (if you'd like)... the author field is really where we want the Github IDs incase we need to ping you in the future regarding your theme.  This is ok though if it's what you prefer.

joshgoebel:
Awesome.  I think it just needs a changelog entry...

terminaldweller:
Added a new entry under New Themes in CHANGES.md. that should do it, right?

joshgoebel:
I almost forgot about our new theme checker.  Can you please run it and have a look?

```
./tools/checkTheme.js src/styles/devibeans.css
```

There are quite a few things your theme omits.  Is this intentional?

joshgoebel:
I might hold this a bit longer until I can circle back and figure out which of those things should be warnings vs errors... we'd *really* like themes to be "complete" as well as start using the new `punctuation`, `operator`, etc... of course it's your theme... so we can't mandate anything or tell you to make it "more colorful" if you like it just how it is... but we can require a theme to explicitly "opt-out'... so that we know for sure you *intended* to not highlight something rather than that you simply weren't aware of it, etc...  this is important for long-term theme maintenance as new scopes are added to the parser.

IE:

```css
/* we purposely do not highlight operators */
.hljs-operator {} 
```

So the question is which items do we want to require manually opting out of and which are purely 100% optional.  That's what we're still figuring out.

terminaldweller:
Added the few missing tags, explicitly opted out of the ones I don't want highlighting for.

joshgoebel:
@terminaldweller Thanks fo much!

PR #3150, opened 2021-04-19
enh(haskell) BinaryLiterals, NumericUnderscores, HexFloatLiterals

martijnbastiaan:
### Changes
Added support for the following Haskell language extensions:

* https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/binary_literals.html
* https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/numeric_underscores.html

It slightly over-approximates. For example, it would highlight `10_`. If that's an issue, I'll put more effort in the PR. IMO this isn't really a problem though, as it would be a syntax error anyway.

Edit: sorry for all the force pushes, should be done now :)

### Checklist
- [X] Added markup tests


src/languages/haskell.js
```diff
@@ -154,10 +154,16 @@ export default function(hljs) {
       PREPROCESSOR,
 
       // Literals and names.
+      hljs.inherit(hljs.BINARY_NUMBER_MODE, {
+        begin: '\\b(0b[01_]+)'
+      }),
+
+      hljs.inherit(hljs.C_NUMBER_MODE, {
+        begin: '(-?)(\\b0[xX][a-fA-F0-9_]+|(\\b(\\d|_)+(\\.(\\d|_)*)?|\\.(\\d|_)+)([eE][-+]?(\\d|_)+)?)'
```
joshgoebel, line 162: Generally we do not match `-` since it could often by unary or binary and we do not try and figure out such distinctions. (very hard to do also without look-behind)

Also please add additional tests for some of the `e` variants, etc.

joshgoebel: Are there two variants here? (hex and regular?)  It seems so.  Please split them out into `variants` for readability and maintenance rather than using a more complex single regex.  See many other grammars for examples.

joshgoebel (changes requested):
Looks pretty good, just needs a few changes!

src/languages/haskell.js
martijnbastiaan: Ah right, I just copied it from `C_NUMBER_MODE`. I'll make a few tests though :-)

joshgoebel: Well, if it's exactly the same it of course shouldn't need to be copied at all... but if it's now specific to this language then we should clean it up into variants for future use - which is just the general policy on such things.

I'd suggest we could perhaps clean up `C_NUMBER_MODE` as well but too many grammars are likely dependent on it being *exactly* how it is now - but that doesn't mean we can't have nicer and easy to read matches in the individual grammars. :)

joshgoebel: I'll add such thoughts to the v12 list.  v11 is already big enough.

martijnbastiaan: Ah let me clarify a bit: I copied it and then added `|_` to optionally match the underscores. 

I'm a bit fuzzy on what you want me to do; should I just make a clean version for Haskell numbers?

joshgoebel: See https://github.com/highlightjs/highlight.js/blob/main/src/languages/swift.js#L121 for an example.

When we start adding *custom* numeric rules to grammars then we always do it as nicely as possible (for future readability and maintenance).  That means separate `variants` for hex, binary, etc... (as you see in Swift) not just one large regex with multiple rules "hidden" inside.

martijnbastiaan: Excellent. I'll get to it tonight. Thanks for the quick responses :)

martijnbastiaan:
@joshgoebel I've rewritten Haskell's number parsing taking inspiration from Swift's implementation. I've included a whole bunch of test cases taken from the GHC documentation. As a happy coincidence, the highlighter now also supports the `HexFloatLiterals` language extension :). Let me know if there's anything more I can do.

joshgoebel:
Awesome.  This looks MUCH much closer to mergeable. :-)

Not sure what all the valid/invalid is trying to accomplish since we randomly highlight some invalids?  Is this sample code just copied and pasted from GHC?  If so, ok, but we should add a comment saying so with a link to the source to explain what's going on and that the invalid can mostly be ignored  - and that all the valid should be highlighted.

If it's not then I'm not a direct copy/paste I'm sure what valid having any `invalids` serves?

martijnbastiaan:
Yeah, you're right, leaving the invalids in is a bit silly. Removed!

src/languages/haskell.js
```diff
@@ -62,6 +62,42 @@ export default function(hljs) {
...
+  /* See:
+
+     - https://www.haskell.org/onlinereport/lexemes.html
+     - https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/binary_literals.html
+     - https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/numeric_underscores.html
+     - https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/hex_float_literals.html
+
+  */
+  const decimalDigits = '([0-9]_*)+';
+  const hexDigits = '([0-9a-gA-F]_*)+';
```
joshgoebel, line 74: a-g?

src/languages/haskell.js
```diff
@@ -62,6 +62,42 @@ export default function(hljs) {
...
+    className: 'number',
+    relevance: 0,
+    variants: [
+      // decimal floating-point-literal (subsumes decimal-literal)
+      {
+        match: `(-?)\\b(${decimalDigits})(\\.(${decimalDigits}))?` + `([eE][+-]?(${decimalDigits}))?\\b`
+      },
+      // hexadecimal floating-point-literal (subsumes hexadecimal-literal)
+      {
+        match: `(-?)\\b0x_*(${hexDigits})(\\.(${hexDigits}))?` + `([pP][+-]?(${decimalDigits}))?\\b`
```
joshgoebel, line 88: The `p` portions is expression in decimal, not hex?

src/languages/haskell.js
```diff
@@ -62,6 +62,42 @@ export default function(hljs) {
...
+      },
+      // octal-literal
+      {
+        match: `(-?)\\b0b(${octalDigits})\\b`
```
joshgoebel, line 92: Is this correct? From manual: 

> octal (prefixed by 0o or 0O)

joshgoebel: Do we have no octal tests that would catch this mistake?

src/languages/haskell.js
martijnbastiaan: Woops, leftover of a debugging attempt :facepalm:. Fixed!

src/languages/haskell.js
martijnbastiaan: You're right, that's a copy+paste error. 

I didn't know about the `[oO]`, but that seems to be a rule across all notations (`[xX]`, `[bB]`). I've updated the patterns accordingly.

src/languages/haskell.js
joshgoebel: I've seen this before, but was just double checking.

joshgoebel:
Please remove the `-?` for consistency with every other grammar.  We have no way to know if that `-` is unary or an operator, do we.  I assume Haskell has the `-` operator for subtraction?

IE:

```
5-3
```

The `-` is NOT part of the number 3, it's an operator, correct?

martijnbastiaan:
Good point! 

> The - is NOT part of the number 3, it's an operator, correct?

That's absolutely correct. I've updated the code to reflec this. I've also added a comment to the negative number tests, in case someone else stumbles upon it in the future. Could you check whether the comment is accurate?

joshgoebel (approved):
Looks great other than my last question!

src/languages/haskell.js
```diff
@@ -154,10 +190,10 @@ export default function(hljs) {
       PREPROCESSOR,
 
       // Literals and names.
+      NUMBER,
 
       // TODO: characters.
       hljs.QUOTE_STRING_MODE,
-      hljs.C_NUMBER_MODE,
```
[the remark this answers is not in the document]
martijnbastiaan: No, not really, just an artifact from trying different implementations. I've undone the reordering.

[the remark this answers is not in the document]
joshgoebel: Probably doesn't matter but order does change behavior so if there wasn't a good reason - best to leave it in it's original spot. :-)

joshgoebel:
@martijnbastiaan Thanks for all the effort!

joshgoebel (approved):

martijnbastiaan:
@joshgoebel Thanks for all the patience! This is what makes me love OSS :heart:

PR #3158, opened 2021-04-21
enh(markdown) various improvements

joshgoebel:
Do not lose improvements from #3152. 

### Changes

Prevents emphasis from false flagging in various circumstances and improves emphasis accuracy in some circumstances.

- `_` does not terminate until whitespace is found
  -  allows `_italic_word_with_underscore_`
- normal words are consumed so that characters inside do not flag as `_`
  - allows `noitalic_here_at_all`
- emphasis must conclude on the same line it starts
  - prevents false positives that can wreck the entire rest of document
  - this is the same behavior GitHub highlighting and VS Code seem to use 

I don't love consuming all the words, may need more thought.  

I'm also not 100% sure the guard of "same line" is needed, but it's a follow-up on discussion in #3152.  It could certainly be argues false negative that mess up the whole document might be worse than failing to highlighting emphasis across multiple lines.

For example:

```
Now I am *bold forever and ever

no matter what else comes
```

Currently we will continue to search forever for a `*` that we will never find... where-as technically the `*` should not count at all since there is no matching close `*`.  It gets very hard to prove this "no matching close" across multiple lines though without a full markdown parser since we really care about blocks (a markdown concept) not lines.

### Checklist

### DIFF
--- a/test/markup/markdown/bold_italics.expect.txt
+++ b/test/markup/markdown/bold_italics.expect.txt
@@ -16,3 +16,9 @@
 
 <span class="hljs-bullet">*</span> list with a <span class="hljs-emphasis">*italic item*</span>
 <span class="hljs-bullet">*</span> list with a <span class="hljs-strong">**bold item**</span>
+
+ word_with_underscores
+ <span class="hljs-emphasis">_italic_word_with_underscore_</span>
+
+ word<span class="hljs-emphasis">*with*</span>asterisks
+ <span class="hljs-emphasis">*italic*</span>wordwith<span class="hljs-emphasis">*asterisks*</span>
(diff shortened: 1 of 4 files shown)


test/markup/markdown/bold_italics.expect.txt
```diff
@@ -16,3 +16,27 @@
...
 <span class="hljs-bullet">*</span> list with a <span class="hljs-strong">**bold item**</span>
+
+ word_with_underscores
+ <span class="hljs-emphasis">_italic_word_with_underscore_</span>
+
+ word<span class="hljs-emphasis">*with*</span>asterisks
+ <span class="hljs-emphasis">*italic*</span>wordwith<span class="hljs-emphasis">*asterisks*</span>
+
+
+// Special handling
```
NullVoxPopuli, line 27: 💯

NullVoxPopuli (approved):

joshgoebel:
I'm still up in the air on some of this so I'm moving to draft for now.

PR #3159, opened 2021-04-22
beginScope and endScope

### LINKED ISSUE #3156: [Request] Multi-match for `end` matchers

Working on JSX I end up with the following mess:

```js
{
    end: regex.lookahead(/<\/[A-Za-z-]+>|\/>/),
    starts: {
      contains: [
        { match: /\/>/, className: "tag", endsParent: true },
        { match:
          [/<\//, /[A-Za-z-]+/,/>/],
          className: {1: "tag", 2: "name", 3:"tag"},
          endsParent: true },
      ]
}
```

Once we find the end (with a look-ahead) we have to jump into a new mode (with `starts`) and two rules using multi-class on `match/begin` to highlight the individual pieces of tag as well as using `endParent` to prevent those same modes from gobbling up anymore then a SINGLE end tag.

And of course the look-ahead also prevents END_SAME_AS_BEGIN since we have no capture.

Not sure.

```js
{
  begin: ...,
  end: [ /</, /[A-Za-z-]+/, />/],
  scope: { 
    1: "tag", 2: ... // scopes for `begin`
    end: {1: "tag.bracket", 2: "tag.name", 3:"tag.bracket" }}
}
```

This is problematic though because eventually I'd like to use `begin` and `end` to allow scoping the ENTIRE begin and end block individually, ie:

```js
{
  begin: /"/, end: /"/
  scope: { begin: "string.quote", end: "string.quote", middle: "string" }
```

Though I suppose a string vs an object is easy to differentiate at runtime.  

A dedicated key is another option:

```js
{
  end: [ /</, /[A-Za-z-]+/, />/ ],
  endScope: { 1: "tag.bracket", 2: "tag.name", 3:"tag.bracket" }
}
```

TextMate grammars use `captures`, `beginCaptures`, `endCaptures`. Making a case for a separate key. A fuller example:

```js
{
  begin: [ /</, /[A-Za-z-]+/ ],
  beginScope: {1: "tag.bracket", 2:"tag.name"},
  end: [ /</, /[A-Za-z-]+/, />/ ],
  endScope: { 1: "tag.bracket", 2: "tag.name", 3:"tag.bracket" }
}
```

With `scope` and `match` being sugar for `beginScope` and `begin`.

**Additional context...**

This is the next obvious improvement for multi-match.

joshgoebel:
Resolves #3156.

### Changes

Adds endScope and beginScope:

```
  {
    begin: /"/,
    end: /"/
    beginScope: "string.delim",
    endScope: "string.delim",
    scope: "string"
  }
```

To produce something like:

```
<string>
  <string.delim>"</>
    hello world
  <string.delim>"</>
</>
```

Multi-match may also be used with `end`, just as with `begin`.

### Checklist
- [x] Added tests, or they don't apply here because...

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,7 +43,11 @@ module.exports = {
   overrides: [
     {
       files: ["types/*.ts", "src/*.ts"],
-      parser: '@typescript-eslint/parser'
+      parser: '@typescript-eslint/parser',
+      rules: {
+        "import/no-duplicates": "off",
+        "import/extensions": "off"
+      }
     },
     {
       files: ["src/**/*.js"],
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,8 @@ Language Grammars:
 
 Parser:
 
+- enh(parser) add `beginScope` and `endScope` to allow separate scoping begin and end (#3159) [Josh Goebel][]
+- enh(parsed) `endScope` now supports multi-class matchers as well (#3159) [Josh Goebel][]
 - enh(parser) `highlightElement` now always tags blocks with a consistent `language-[name]` class [Josh Goebel][]
   - subLanguage `span` tags now also always have the `language-` prefix added
 - enh(parser) support multi-class matchers (#3081) [Josh Goebel][]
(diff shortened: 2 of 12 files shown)


joshgoebel:
Two commits are probably easier to review separately.

NullVoxPopuli (approved):
I love this! in certain templating languages dealing with nested stuff is tricky, and this is perfect!

PR #3166, opened 2021-04-23
chore(stan) more typical `meta` scope usage

joshgoebel:
### Changes

- Uses `meta` for the whole `#include` line
- Uses `string` for the filename being included

### Checklist
- [x] Updated the changelog at `CHANGES.md` (not needed, chore)


src/languages/stan.js
```diff
@@ -516,8 +524,7 @@ export default function(hljs) {
       },
       {
         className: 'keyword',
-        begin: /\btarget\s*\+=/,
-        relevance: 10
+        begin: /\btarget\s*\+=/
```
joshgoebel, line 527: This is not unique enough to be relevance 10.

NullVoxPopuli (approved):

PR #3167, opened 2021-04-23
(chore) remove meta-keyword and meta-string in favor of nesting

### LINKED ISSUE #3163: Review/fix meta, meta-string, and meta-keyword scopes
Question:  Do we really need `meta-string` and `meta-keyword` vs subscopes/parent scopes?

`meta-keyword` is used 28 times in core grammars and `meta-string` is use 20 times.  With only 3 exceptions these are used under a parent `meta` scope.  Stan looks like it probably should also be `meta/meta-keyword` (nested), which would bring the exceptions down to only 2.

The `meta-` prefix should either be dropped and the hierarchy used for styling or these should become `meta.keyword` and `meta.string` (sub scopes).  Given how hierarchal these already are I think the prefix should be dropped so we simply have `keyword` inside a larger `meta` scope, etc... ie, how it is now, but without the `meta-` prefix.

This is purely to simplify our scoping and remove the duplication.

Long-term I'd still like to possible to get to a more TextMate grammar meaning of `meta`, but that might take some time.

Exceptions:

- `dts` uses `meta-keyword` outside meta (fixed in #3164)
- `r` uses `meta-keyword` outside meta
- `stan` uses `meta-keyword` outside meta (fixed in #3166)

joshgoebel:
Resolves #3163.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,8 @@ Parser:
 
 Grammars:
 
+- enh(all) `.meta-keyword` => `.meta .keyword` (nested scopes) (#3167) [Josh Goebel][]
+- enh(all) `.meta-string` => `.meta .string` (nested scopes) (#3167) [Josh Goebel][]
 - enh(swift) add `actor` keyword (#3171) [Bradley Mackey][]
 - enh(crystal) highlight variables (#3154) [Josh Goebel][]
 - fix(ruby) Heredoc without interpolation (#3154) [Josh Goebel][]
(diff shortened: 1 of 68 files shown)


allejo (approved):
Hope you don't mind I pushed to your branch. Doing a "replace all" was easier than doing suggestions on here.

LGTM 👍

joshgoebel:
Oh it was because my search and replace didn't care about the quoting. I see now. :)

PR #3173, opened 2021-05-01
Fix highlighting for rust raw byte string literals

nico-abram:
Here's an example of what this should fix: https://doc.rust-lang.org/reference/tokens.html?highlight=raw%20string#raw-byte-string-literals

imagen

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -121,6 +121,7 @@ Dev Improvements:
 [Jared Luboff]: https://github.com/jaredlll08
 [NullVoxPopuli]: https://github.com/NullVoxPopuli
 [Mike Watling]: https://github.com/Pickysaurus
+[Nico Abram]:https://github.com/nico-abram
 
 
 ## Version 10.7.1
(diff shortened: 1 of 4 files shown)


bumbread (approved):

joshgoebel:
Thanks so much!

Please add markup tests to `test/markup/rust/strings`, etc to verify this new behavior.

nico-abram:
> 
> 
> Thanks so much!
> 
> Please add markup tests to `test/markup/rust/strings`, etc to verify this new behavior.

Done @joshgoebel

nico-abram:
Sorry, I forgot to update the expect file, just pushed again

joshgoebel:
Perfect. Can you add changlog entry I don't think I saw it.

nico-abram:
Added changelog entry

joshgoebel (commented):
@nico-abram Thanks so much for the contribution!

joshgoebel (approved):

PR #3175, opened 2021-05-02
Add highlighting support for Wren

joshgoebel:
**Wren is unbelievably amazing.**

Screen Shot 2021-05-02 at 2 49 59 AM

I worry the attributes may be a bit much and that perhaps we should highlight them as comments instead though.

### Checklist


joshgoebel (commented):
Self-reviewed.

PR #3177, opened 2021-05-05
(taggerscript) support Unicode escapes and fix function definitions

phw:
### Changes

This updates the taggerscript grammar to:

- Add support for Unicode escape sequences like `\u2014` (which is a new feature)
- Fix the function definition, functions starting with underscores or numbers are actually allowed

### Checklist

I have not yet updated the changelog, I was unsure how you handle this. Currently the changelog only shows actual releases, not ongoing changes. Happy to update this, though.


src/languages/taggerscript.js
```diff
@@ -22,7 +22,7 @@ export default function(hljs) {
 
   const FUNCTION = {
     className: 'keyword',
-    begin: /\$(?!noop)[a-zA-Z][_a-zA-Z0-9]*/,
+    begin: /\$(?!noop)[_a-zA-Z0-9]*/,
```
joshgoebel, line 25: Can a function be a single character, is `_` by itself a valid function name?

src/languages/taggerscript.js
```diff
@@ -35,7 +35,12 @@ export default function(hljs) {
 
   const ESCAPE_SEQUENCE = {
     className: 'symbol',
-    begin: /\\./
+    begin: /\\[^u]/
```
joshgoebel, line 38: This isn't needed, just change the order so that unicode comes first...

src/languages/taggerscript.js
joshgoebel: Class name looks wrong here (I know you didn't change it).  Probably should be `title.function`?  Would you mind updating?

Does the `(` always immediately follow the function name?

joshgoebel (changes requested):
Thanks so much! Added a few notes.

src/languages/taggerscript.js
phw: Yes to all of the above, strictly speaking it would be `/\$(?!noop)[_a-zA-Z0-9]+/` actually. But for syntax highlighting the above works better, because an unescaped $ always starts a function, so highighting `$` already works well for this purpose.

src/languages/taggerscript.js
phw: Ok, I'll test and update

src/languages/taggerscript.js
joshgoebel: >  it would be /\$(?!noop)[_a-zA-Z0-9]+/ actually

We should go with that then.  We need to be as strict as is reasonable as this is also used for auto-detection.

> But for syntax highlighting the above works better, because an unescaped $ always starts a functio

I don't follow.

If the `(` is always immediate then we should just change this to a matcher without the `end` to be more precise.

```
match: /\$(?!noop)[_a-zA-Z0-9]+(?\s*\(=)/
```

If whitespace isn't allowed the \s* could be removed.

phw: Regarding the name, that was changed here:

https://github.com/highlightjs/highlight.js/commit/0c8940069e69395036a48c51371b85699f04da75#diff-976da83ed3c74e03a20bb75c0361c839f1c917c3c426d7145781ec508dbab6eb

> Things like `$if`, `$ne`, `$eq` look more like keywords in a
  conventional language, certainly not as function names being
  defined by user

I tend to agree on this. The language does not allow you to define your own functions. But not sure really how this is all used or turns out for formatting. TextMate grammars have a class specifically for "functions provided by the framework/library", something like this would probably fit if that exists.

phw: Ok. Wouldn't `match: /\$(?!noop)[_a-zA-Z0-9]+` be sufficient? Why the `(?=\()` isn't that captured by the `end: /\(/` already?

About being explicit I will then also change the current generic definition of escape from `\.` to a list of the actual characters, because this is actually a very limited set.

joshgoebel: > Why the (?=\() isn't that captured by the end: /\(/ already?

Is this a function:

```
$lookslikefunction but... HEY HOW ARE YOU #!*#@#&@#& 
how is the weather
(
```

If there is no middle to speak of then we don't need begin/end, we only need match.

joshgoebel: Is there no distinction between what `$_lossless` and `$if` are semantically in the language?

phw: I have made the definitions more explicit as suggested, including removing the end parameter and explicitly giving the possible escape sequences. I also fixed some issues with nesting of $noop()

> Is there no distinction between what $_lossless and $if are semantically in the language?

No, both are just internally defined functions that run with no or some parameters and return something. In the end that even is true for $noop(), just that this is a specific builtin that is otherwise treated as kind of a comment (and $noop() is a rather strong indicator that the language is matching).

There is no definitive list of all possible functions, as the language can be extended. But not with the language itself.

src/languages/taggerscript.js
```diff
@@ -9,33 +9,36 @@ export default function(hljs) {
     className: 'comment',
     begin: /\$noop\(/,
     end: /\)/,
-    contains: [ {
+    contains: [{
+      begin: '/\\./'
```
joshgoebel, line 13: Is this meant to be different than ESCAPE_SEQUENCE?

joshgoebel: (other than the lack of highlighting)

src/languages/taggerscript.js
```diff
@@ -9,33 +9,36 @@ export default function(hljs) {
...
     } ],
     relevance: 10
   };
 
   const FUNCTION = {
     className: 'keyword',
-    begin: /\$(?!noop)[a-zA-Z][_a-zA-Z0-9]*/,
-    end: /\(/,
-    excludeEnd: true
+    begin: /\$(?!noop)[_a-zA-Z0-9]+(?=\()/
```
joshgoebel, line 26: Yes, nice!

src/languages/taggerscript.js
```diff
@@ -9,33 +9,36 @@ export default function(hljs) {
...
   };
 
   const VARIABLE = {
     className: 'variable',
-    begin: /%[_a-zA-Z0-9:]*/,
-    end: '%'
+    begin: /%[_a-zA-Z0-9:]+%/
```
joshgoebel, line 31: Another great improvement.

src/languages/taggerscript.js
```diff
@@ -9,33 +9,36 @@ export default function(hljs) {
...
+
+  const ESCAPE_SEQUENCE_UNICODE = {
+    className: 'symbol',
+    begin: /\\u[a-fA-F0-9]{4}/
   };
 
   const ESCAPE_SEQUENCE = {
     className: 'symbol',
-    begin: /\\./
+    begin: /\\[\\nt$%,()]/
```
joshgoebel, line 41: Ok, that's another way to fix it AND more explicit. :) Awesome.

src/languages/taggerscript.js
phw: Not really, but my main purpose here was to catch escaped parantheses. But having this a string clearly was a bug, this is of course supposed to be a regex. So I pushed an update and changed it to `/\\[()]/`  and also extended the test case to check for this.

This is supposed to handle things like:

```
$noop(\))
$noop(\()
```

Or more complext stuff with nesting:

```
$noop(\($noop(abc$func(\))\)xyz))
```

All of the above is just a single $noop, but the () inside need to be nested properly and eventually get escaped with a backslash.

phw:
Thanks a lot for the thorough review. I actually now understand how the parser rules are applied, something I obviously did not fully grasp when I originally wrote this grammar.

Please take a look at the latest commits with extended tests.

src/languages/taggerscript.js
```diff
@@ -23,7 +23,7 @@ export default function(hljs) {
 
   const FUNCTION = {
     className: 'keyword',
-    begin: /\$(?!noop)[_a-zA-Z0-9]+(?=\()/
```
joshgoebel: I was wondering earlier if we really needed this. Simpler is better. :)

joshgoebel (approved):

joshgoebel (approved):

Issue #3093, opened 2021-03-31
(cpp) std::string* not highlighted in 10.7.1


milutin:
**Describe the issue**
There is highlighted regression in 10.7.1 with following line
std::string* p = nullptr;

```cpp
std::string* p = nullptr;
```

**Which language seems to have the issue?**
cpp

**Are you using `highlight` or `highlightAuto`?**
highlightBlock

**Sample Code to Reproduce**
highlight10.7.1: http://jsfiddle.net/L10xf382/8/
highlighting10.6: http://jsfiddle.net/L10xf382/10/

**Expected behavior**
highlighting10.6: http://jsfiddle.net/L10xf382/10/

**Additional context**

joshgoebel:
See https://github.com/highlightjs/highlight.js/pull/3005.

The built-in list was removed in favor of a `something(` rule (using the previous built-in list for hinting only)... it seems some things on that list were not functions though (like std).  Could you pease review the list and see if anything else is NOT a function?

```
-    built_in: 'std string wstring cin cout cerr clog stdin stdout stderr string
stream istringstream ostringstream ' +
-      'auto_ptr deque list queue stack vector map set pair bitset multiset mult
imap unordered_set ' +
-      'unordered_map unordered_multiset unordered_multimap priority_queue make_
pair array shared_ptr abort terminate abs acos ' +
-      'asin atan2 atan calloc ceil cosh cos exit exp fabs floor fmod fprintf fp
uts free frexp ' +
-      'fscanf future isalnum isalpha iscntrl isdigit isgraph islower isprint is
punct isspace isupper ' +
-      'isxdigit tolower toupper labs ldexp log10 log malloc realloc memchr memc
mp memcpy memset modf pow ' +
-      'printf putchar puts scanf sinh sin snprintf sprintf sqrt sscanf strcat s
trchr strcmp ' +
-      'strcpy strcspn strlen strncat strncmp strncpy strpbrk strrchr strspn str
str tanh tan ' +
-      'vfprintf vprintf vsprintf endl initializer_list unique_ptr _Bool complex
 _Complex imaginary _Imaginary',
```

I see some types in there as well now that I'm looking... CC @klmr

joshgoebel:
Ah, ok, we'll wait and see if klmr can chime in then. :)  I could take a guess and get MOST of it I think, but a few I'm not sure about.

klmr:
[...]

PR #3178, opened 2021-05-05
Extend cpp keyword set

krisvanrens:
Added a set of commonly used C++ keywords/identifiers. We can't add them all of course, but things like `optional` and `variant` which are common vocabulary types since 2017 should be in there in my opinion.

I would also like to order the lists in the language file lexicographically, making extension easier (without having to search). Is this OK? I will update this PR if desired.

### Changes

* Extend set of common keywords/identifiers for C++.

### Checklist / `TODO`
- [x] Remove redundant primitive types from keyword list.
- [x] Cleanup keyword/function list.
- [x] Update/review tests where needed.
- [x] Explode dense lists and order lexicographically.
- [x] Update the changelog at `CHANGES.md`.


src/languages/cpp.js
```diff
@@ -218,11 +218,31 @@ export default function(hljs) {
...
+    'swap',
+    'exchange',
+    'forward',
+    'move',
+    'declval'
   ];
 
   const CPP_KEYWORDS = {
     keyword: 'int float while private char char8_t char16_t char32_t catch import module export virtual operator sizeof ' +
+      'int8_t int16_t int32_t int64_t int_fast8_t int_fast16_t int_fast32_t int_fast64_t ' +
```
joshgoebel, line 242: These should already be handled by `CPP_PRIMITIVE_TYPES`, no?

src/languages/cpp.js
```diff
@@ -218,11 +218,31 @@ export default function(hljs) {
     'stderr',
     'stringstream',
     'istringstream',
-    'ostringstream'
+    'ostringstream',
+    'optional',
```
joshgoebel, line 222: Are these keywords or functions?  This is a list of common functions, if this are keywords they don't belong in this list.  It's also possible the list itself needs cleanup if it includes other keywords.

joshgoebel:
> I would also like to order the lists in the language file lexicographically, 

We can do that when done in a separate commit at the end so it doesn't complicate the review.

We also should go ahead and explode `KEYWORDS` into a proper array itself (one word per line, etc).

src/languages/cpp.js
krisvanrens: Yes, you're right, forgive me. I will update the list.

I'm wondering though why the `charXX_t` (and other `_t` types) are there in the list as well then. They should all be handled by `CPP_PRIMITIVE_TYPES`. Perhaps they can be removed.

src/languages/cpp.js
krisvanrens: It's a mix, so is the already existing part of the list.

E.g. `cos` is a function, `stringstream` is a type etc. Most of the items are functions, but some of them are often-used types. Regarding the interpretation the line is not 100% clear though, because of operator overloading in C++, a static type can be used as though it looks like a function. E.g.:

```c++
std::cout << "Hi there!";
```

Here `std::cout` is a global object of type `ostream` with an overloaded `operator<<()`, making it's use look like a function.

If the list should be functions only, it should definitely be cleaned. I'm willing to pick this up if you'd like me to.

src/languages/cpp.js
joshgoebel: > I'm wondering though why the charXX_t (and other _t types) are there in the list as well then. 

Me too, I'd say they could safely be removed?

krisvanrens: Yes, I'll check this and extend the test for this. There currently is a test that only tries a single example.

src/languages/cpp.js
joshgoebel: Yes, it should be functions... and that list should be used in `FUNCTION_DISPATCH` instead of KEYWORDS (not sure why keywords got there). Any types should be moved out into `TYPES` and we should add them to keywords in addition to the other keys. Any keywords should be moved into KEYWORDS.

```
keywords: { type: TYPES } 
```

CPP_PRIMITIVE_TYPES should probably be `className: 'type'` also.

So we'll probably end up with something like:

```
  const CPP_KEYWORDS = {
    keyword: KEYWORDS,
    built_in: '_Bool _Complex _Imaginary',
    literal: 'true false nullptr NULL ...'
  };

```

src/languages/cpp.js
krisvanrens: This should be solved in the latest version now.

src/languages/cpp.js
krisvanrens: Hey Josh, I will look into this in the coming days -- I have been having quite a busy life lately ;-)

joshgoebel:
Would be nice to have this in v11. :)  Probably getting close to release.

krisvanrens:
Ha, I'm working on it as we speak ;-) Still figuring out some of the JS stuff going on in the language file. I will push a new version later this evening (evening for me, that is).

krisvanrens:
Hey!

I've cleaned up some parts according to Josh's comments above. However, three tests fail for some reasons not completely clear to me. The first one is an autodetection error from a C++ comment test that is detected as 'sqf' now, and another one where Arduino C++ is detected as C++ instead of Arduino. This latter one probably also triggers the last test to fail.

Furthermore I've got a few remarks/questions:

- Handling of function highlighting is not bug-free. E.g. when a C++ function is a function template that is called with explicit template arguments it's not being detected as a function (e.g.: `func()` (OK)m `func<int>()` (not OK)). I'm not sure how to fix this yet as I don't feel very confident with the HL.js language code yet.
- I'm not sure exactly what I changed when I moved the `COMMON_HINTS` array out of the `CPP_KEYWORDS` `_relevant_hints` field. The `CPP_KEYWORDS` object is used in numerous places. It feels kind of random/messy when I look at it -- but then again, I'm not really understanding the language parsing file yet.
- In the reorganization and cleanup of the types/keywords/function arrays, it was inevitable that I had to sort them.
- I'm not really sure how to beautify the über-long regex for containers.
- Handling of common objects that are used through their operators is not working correctly yet. For example: `std::cout << "blah";` should highlight `cout` because it is a very commonly used object. It is an object that is used through `operator<<()`, this kind-of is a function, but not really, and HL.js doesn't detect it as such because no parentheses are used. This and probably others as well need a specific `COMMON_OBJECTS` class or something to trigger the highlighting in the right context.
- How far should I go in the adding of keywords/common functions? Right now there's a set that probably covers about 75% of all common C++ code (this is really wild guess). But I can imagine someone who uses this for some more exotic code (some of the code that tends to be presented at conferences via the likes of reveal.js etc.) there may be inconsistencies because not all hundreds and hundreds of common C++ functions are in there. There are still quite a number missing. I'd like to extend the list some more, but I'm not sure where to stop.

At least I checked all the examples and have added a few TODOs for this language parsing file. I might be able to pick these up in a couple of weeks (I'm too busy now). It feels like this language parsing file may need a proper update/refresh anyway.

joshgoebel:
> The first one is an autodetection error from a C++ comment test that is detected as 'sqf' now, and another one where Arduino C++ is detected as C++ instead of Arduino. This latter one probably also triggers the last test to fail.

I can look into these, probably Arduino needs to be updated based on other changes you made here.

> Handling of function highlighting is not bug-free. E.g. when a C++ function is a function template that is called with explicit template arguments it's not being detected as a function (e.g.: `func()` (OK)m `func<int>()` (not OK)). I'm not sure how to fix this yet as I don't feel very confident with the HL.js language code yet.

We'd need to expand the regex to match the `<>` part as well.  Examples of this can be found in other grammars but we may want to use the new multi-match stuff here and try to simplify.


> I'm not sure exactly what I changed when I moved the `COMMON_HINTS` array out of the `CPP_KEYWORDS` `_relevant_hints` field. 

Before the list was mixed (having more than functions) so it needed to be here so the non-functions could still be found and counted (since FUNCTION_DISPATCH wouldn't cover them).  If it now only contains functions then it no longer needs to be in `CPP_KEYWORDS`, so it was right to remove it.

> The `CPP_KEYWORDS` object is used in numerous places. It feels kind of random/messy when I look at it -- but then again, I'm not really understanding the language parsing file yet.

That's because the grammar has too many nested modes... and keywords don't inherit, so if you want the keywords to be present in every sub-mode then you have to keep specifying them over and over.  Flattening the grammar to avoid all these nested modes would help with this.

> I'm not really sure how to beautify the über-long regex for containers.

Why does it need to be beautified?  I mean you could put the list in an array and then built it with the `regex` helpers (`regex.concat`, `regex.either`, etc)... that would be an improvement.

> Handling of common objects that are used through their operators is not working correctly yet. For example: `std::cout << "blah";` should highlight `cout` because it is a very commonly used object. It is an object that is used through `operator<<()`, this kind-of is a function, but not really, and HL.js doesn't detect it as such because no parentheses are used. This and probably others as well need a specific `COMMON_OBJECTS` class or something to trigger the highlighting in the right context.

Perhaps, though I thought I remembered @klmr saying something about not highlighting these, or am I thinking of something else?

- How far should I go in the adding of keywords/common functions? Right now there's a set that probably covers about 75% of all common C++ code

This is sufficient.  The list is only for heuristics - all functions are highlighted either way.  Someone using the library for serious work should be tagging languages in any case, NOT relying on auto-detect which is not always reliable.

---

This looks like it's in a fairly good spot if I can fix the test issues.  We'll see about getting this merged then we can discuss what's next for another PR.

joshgoebel:
It looks like the keywords list still includes some types... `int`, etc... those should be in TYPES.

test/detect/cpp/comment.txt
```diff
@@ -1,16 +0,0 @@
...
-(The explanation above describes two-word generation. If you type "-d 3",
-the program will find three-word groups, and so on. Greater depths make more
-sense, but they require more input text and take more time to process.)
-
-http://www.eblong.com/zarf/markov/
-*/
-
-
-/* make cpp win deterministically over others with C block comments */
-cout << endl;
```
joshgoebel, line 16: This is a terrible detect test being nothing but comments.. we shouldn't expect to necessarily get this right anyways, so removing.

krisvanrens: Yes I agree.

krisvanrens:
> It looks like the keywords list still includes some types... `int`, etc... those should be in TYPES.

Mmm. Yes, you're right. Let me fix that. I initially went for "this is the list of reserved keywords in C++" but it contains type keywords as well of course. In the context of a highlighting tool it doesn't matter if a keyword is reserved or not of course.

krisvanrens:
> We'd need to expand the regex to match the `<>` part as well. Examples of this can be found in other grammars but we may want to use the new multi-match stuff here and try to simplify.

This is fixed now.

> That's because the grammar has too many nested modes... and keywords don't inherit, so if you want the keywords to be present in every sub-mode then you have to keep specifying them over and over. Flattening the grammar to avoid all these nested modes would help with this.

This can be fixed in later versions I would say.

> > I'm not really sure how to beautify the über-long regex for containers.
> 
> Why does it need to be beautified? I mean you could put the list in an array and then built it with the `regex` helpers (`regex.concat`, `regex.either`, etc)... that would be an improvement.

Yeah, that's exactly what I meant. I left it in for now, this could be picked up later on.

> > Handling of common objects that are used through their operators is not working correctly yet. For example: `std::cout << "blah";` should highlight `cout` because it is a very commonly used object. It is an object that is used through `operator<<()`, this kind-of is a function, but not really, and HL.js doesn't detect it as such because no parentheses are used. This and probably others as well need a specific `COMMON_OBJECTS` class or something to trigger the highlighting in the right context.
> 
> Perhaps, though I thought I remembered @klmr saying something about not highlighting these, or am I thinking of something else?

Hmmm. OK, I can look up this conversation for a later version. For now they are in the functions list (which is not entirely correct).

> This is sufficient. The list is only for heuristics - all functions are highlighted either way. Someone using the library for serious work should be tagging languages in any case, NOT relying on auto-detect which is not always reliable.
> 
> This looks like it's in a fairly good spot if I can fix the test issues. We'll see about getting this merged then we can discuss what's next for another PR.

OK.

I think the highlighting is working quite well now. I've been evaluating the test examples and didn't see any weirdness. It turns out quite a bit of the things didn't work before (many of the keyword functions). Mostly because of the lack of matching for `<..>` before function parentheses.

In a next PR I would like to dig in a little deeper to simplify the grammar. It really feels like some parts of it are dead, wrong or not up-to-date with the latest C++ standard.

krisvanrens:
Oh, and by the way, the Arduino 'default' markup test fails again. In the developer markup tester it is auto-detected as JavaScript. I think it's not detected as Arduino code in the tests as well.

test/markup/cpp/function-declarations.expect.txt
```diff
@@ -1,25 +1,25 @@
...
-<span class="hljs-function"><span class="hljs-keyword">friend</span> <span class="hljs-keyword">void</span> <span class="hljs-title">showB</span><span class="hljs-params">(B::SomeType x)</span> </span>{}
-<span class="hljs-function"><span class="hljs-keyword">inline</span> <span class="hljs-keyword">int</span> <span class="hljs-title">add</span><span class="hljs-params">(<span class="hljs-keyword">int</span> a, <span class="hljs-keyword">int</span> b)</span> </span>{}
+<span class="hljs-function"><span class="hljs-keyword">decltype</span>(<span class="hljs-type">auto</span>) <span class="hljs-title">look_up_a_string_1</span><span class="hljs-params">()</span> </span>{ <span class="hljs-keyword">return</span> <span class="hljs-built_in">lookup1</span>(); }
+<span class="hljs-function"><span class="hljs-type">void</span> <span class="hljs-title">look_up_a_string_2</span><span class="hljs-params">()</span> </span>{ <span class="hljs-keyword">return</span> <span class="hljs-built_in">lookup2</span>(); }
```
joshgoebel, line 2: Is void really a type?  I've kind of considered it an anti-type, hence a keyword... thoughts?

test/markup/cpp/function-title.txt
```diff
@@ -1,5 +1,5 @@
-int main() {
-  A a = new A();
+int main(int argc, char** argv) {
+  auto a = std::make_unique<A>();
```
joshgoebel, line 2: Please add a second test (or just a second file) in this file vs *removing* the simple case to replace it with a complex case.

```suggestion
  A a = new A();
  auto a = std::make_unique<A>();
```

joshgoebel:
> Oh, and by the way, the Arduino 'default' markup test fails again.

Because Arduino is ALSO cpp and uses that same grammar... so you need to update those markup tests as well (keywords are now `type`, etc)...  There is a commented line in `markup/index.js` that can rewrite all the tests for you but if you use it you need to read the diff carefully to make sure it's doing what you expect.

Please don't rewrite history here but just add commits so I can more easily review.

> In the developer markup tester it is auto-detected as JavaScript.

No matter, markup tests have nothing to do with auto-detection.

>  I think it's not detected as Arduino code in the tests as well.

I don't see a problem... run `./tools/checkAutodetect.js`... all is good.

test/markup/cpp/function-declarations.expect.txt
krisvanrens: `void` is the empty type ;-) Indeed in C++ it is considered bad practice to use `void` pointers etc. but for a function that returns nothing, you choose the `void` return type. But it definitely is a type.

test/markup/cpp/function-title.txt
krisvanrens: Yes, will do.

I changed this, because I got autodetection issues with the old version; it was detected as PHP for some reason.

krisvanrens:
> > Oh, and by the way, the Arduino 'default' markup test fails again.
> 
> Because Arduino is ALSO cpp and uses that same grammar... so you need to update those markup tests as well (keywords are now `type`, etc)... There is a commented line in `markup/index.js` that can rewrite all the tests results for you but if you use it you need to read the diff carefully to make sure it's doing what you expect.

I see, yes. I now fixed it using the results from the test output of Node, because the 'Arduino' language highlighting is not in the developer markup checker.

> 
> Please don't rewrite history here but just add commits so I can more easily review.

Uhm, I'm not sure what I did wrong then -- this was not my intention. I pulled in your commits with `git pull -r` then committed my additions.

> 
> > In the developer markup tester it is auto-detected as JavaScript.
> 
> No matter, markup tests have nothing to do with auto-detection.

OK.

test/markup/cpp/function-title.txt
joshgoebel: Markup (`tests/markup`) tests have _nothing_ to do with auto-detect tests.

joshgoebel:
> because the 'Arduino' language highlighting is not in the developer markup checker.

It's whatever you choose to build... `built -t browser :common arduino` would add it, etc.

joshgoebel:
> Uhm, I'm not sure what I did wrong then -- this was not my intention. I

No worries could just be me reading it wrong... a lot of the more git-savvy contributors do that and often it makes reviews harder.

test/markup/cpp/template-complexity.expect.txt
```diff
@@ -1,21 +1,21 @@
...
 }
 
 <span class="hljs-comment">// Disable overload for already valid operands.</span>
-<span class="hljs-keyword">template</span>&lt;<span class="hljs-keyword">class</span> <span class="hljs-title class_">T</span>, class = std::<span class="hljs-keyword">enable_if_t</span>&lt;!impl::is_streamable_v&lt;<span class="hljs-keyword">const</span> T &amp;&gt; &amp;&amp; std::is_convertible_v&lt;<span class="hljs-keyword">const</span> T &amp;, std::wstring_view&gt;&gt;&gt;
+<span class="hljs-keyword">template</span>&lt;<span class="hljs-keyword">class</span> <span class="hljs-title class_">T</span>, <span class="hljs-type">class</span> = std::<span class="hljs-type">enable_if_t</span>&lt;!impl::is_streamable_v&lt;<span class="hljs-keyword">const</span> T &amp;&gt; &amp;&amp; std::is_convertible_v&lt;<span class="hljs-keyword">const</span> T &amp;, std::<span class="hljs-type">wstring_view</span>&gt;&gt;&gt;
```
joshgoebel, line 10: I don't believe class is a type here?  We should probably remove it from the list of types and (in a follow-up PR) add it instead with a `class [IDENT]` multi-matcher rule as done in many other languages - allowing us to also flag the class name as a `title.class`, etc...

joshgoebel:
Very very close. :)  I think just fix the class issue and this is good to go.

test/markup/cpp/template-complexity.expect.txt
joshgoebel: Also pretty sure `class` is not a type but a keyword... you use class to specify classes, it's not uses (as a type) to declare a variable `class a`... you'd say `NameOfClass a`... so in that usage:

```
class Booger { 
}
```

It would be a keyword.

joshgoebel:
@klmr Would you mind quickly reviewing the list of keywords/types/hints just as a sanity check incase there are other issues I should flag but I'm not because I don't know C++?

test/markup/cpp/function-title.txt
krisvanrens: Yeah I'm sorry, I think I was confused die to the outcome of the auto detection in the developer markup tool.

test/markup/cpp/template-complexity.expect.txt
krisvanrens: Yeah you're right. I'm still trying to get used to the internal conflicts between the language lawyer and the highlight mindset. I'll fix this.

krisvanrens:
> > Uhm, I'm not sure what I did wrong then -- this was not my intention. I
> 
> No worries could just be me reading it wrong... a lot of the more git-savvy contributors do that and often it makes reviews harder.

Ah, yes I see what you mean. I was not aware of the merge strategy that you have. For some reason I assumed you don't like merge commits. Will you squash all the commits from this PR at merge-time?

test/markup/cpp/template-complexity.expect.txt
krisvanrens: I'm going to move `auto`, `class`, `struct`, `union` and `enum`, as they're all somewhere in the twilight zone of being a keyword/storage specifier and a type specifier. I don't have a strong opinion about this regarding the highlighting class, but language-strictly spoken they're probably more type than keyword. But at least they're considered 'special' for the highlighter -- this is the most important fact.

krisvanrens: And indeed in templates `class` serves a double role as a typename specifier. With some regular expression magic this could be made a special case, but it's fine like it is.

joshgoebel: > I'm going to move auto, class, struct, union and enum,

I'd say broadly these are all keywords (from our highlighting POV and historically how we treat them across many grammars).  We don't currently differentiate between "type of storage" vs "type of type"... I think the context of their usage is relevant as well... lets take a made up class-based language:

Next is me just thinking about this seat of my pants:

```js
// define a class
class Person { }

// a new variable of class Person
var bob : class<Person> = Person.new(...)
```

The first `class` is a `keyword`.  You could argue the second `class` is a bit closer to `type` (and that could be worth discussing), but for convenience we'd probably just call it a keyword, rather than try to contextually parse the context.  `Person` of course would be `title.class`.

krisvanrens: > I don't believe class is a type here? We should probably remove it from the list of types and (in a follow-up PR) add it instead with a `class [IDENT]` multi-matcher rule as done in many other languages - allowing us to also flag the class name as a `title.class`, etc...

Could you point me at a language definition file that uses these multi-matcher rules, or at least a language definition that is a very good example of how a language definition should be done? I'm going to try to study this stuff for the next PR. Thanks!

joshgoebel:
> Ah, yes I see what you mean. I was not aware of the merge strategy that you have.

Well, typically the scope of the PRs is small enough that squash works well.  It's not set in stone.  If someone has a very clean commit history and I see value in every individual commit then I'll rebase/merge it.  95% of the time we squash.

> For some reason I assumed you don't like merge commits. 

I don't like messy commit histories.

> Will you squash all the commits from this PR at merge-time?

Most likely. 

---

About rebasing though:

What is hard about rebasing is it's hard to review the changes on GitHub _because they keep changing_... when I'm working with someone on a PR and they push a tiny change/fix I want to see JUST that change (often it might be 1-2 lines)... that way I can confirm it matches what I was expecting.  Often when someone instead rebases you lose track of "what just changed" because they smartly merged it back into their prior commits.

Rebasing can be great (as the very last step)... and along the way at `fixup: ` or `squash: ` commits... sometimes it can just be hard to follow if it's done over and over in the middle of working on a PR. :)

joshgoebel:
I would probably have given you "auto" as a type (I flagged it before in my head but didn't deal it worth mentioning), but I don't have a strong argument either way. :-)

joshgoebel:
@krisvanrens You should consider joining our Discord if you want to dig into C++ further in the future.  I'd be happy to explain what I mean by a "flatter" syntax.  I don't know if C++ is "wrong/outdated" now so much as "overly-complex".  Certainly possible it's also both. :)

krisvanrens:
> @krisvanrens You should consider joining our Discord if you want to dig into C++ further in the future. I'd be happy to explain what I mean by a "flatter" syntax. I don't know if C++ is "wrong/outdated" now so much as "overly-complex". Certainly possible it's also both. :)
> 
> The `c` grammar could also use much of the same thinking if you were willing.

I will definitely join the Discord channel for info. I've got a very busy month ahead, but after that there's time for this. Thanks for your patience anyway :-)

My guess is that if I take on the C++ grammar, the C-one should be doable as well.

joshgoebel:
Still needs changelog entries. :)

CHANGES.md
```diff
@@ -93,6 +93,9 @@ Grammars:
 - enh(haskell) add support for HexFloatLiterals (#3150) [Martijn Bastiaan][]
 - fix(c,cpp) allow declaring multiple functions and (for C++) parenthetical initializers (#3155) [Erik Demaine][]
 - enh(rust) highlight raw byte string literals correctly (#3173) [Nico Abram][]
+- enh(cpp) add various keywords and commonly used types (#3178) [Kris van Rens][]
+- enh(cpp) improve and cleanup parsing of reserved keywords and types (#3178) [Kris van Rens][]
+- fix(cpp) fix detection of common functions that are function templates (#3178) [Kris van Rens][]
```
joshgoebel, line 98: Also need to add link for Kris van Rens on line ~129.

krisvanrens: Ugh yes :-)

klmr:
> things like `optional` and `variant` which are common vocabulary types since 2017 should be in there in my opinion.

As Josh alluded to I strongly disagree with this. I’ve explained the reasons for this here: https://github.com/highlightjs/highlight.js/issues/3093#issuecomment-811504182 — tl;dr: standard library types and functions are not reserved words, and highlighting them in builtins leads to numerous false positives. They also don’t have any special semantics compared to user-defined types and functions, so giving them special highlighting *even where correct* serves little purpose in making code more readable; on the contrary, it creates a distraction by emphasising an irrelevant distinction.

Compare this with IDEs and source code editors: the good ones don’t highlight these identifiers either (or rather, they highlight them the same as user-defined types and functions).

In fact, previous revisions of the C++ grammar underwent a rigorous cleanup to get rid of arbitrary identifiers that get highlighted incorrectly and I strongly advocate keeping it that way.

krisvanrens:
Hi Konrad, thanks for your remark.

> > things like `optional` and `variant` which are common vocabulary types since 2017 should be in there in my opinion.
> 
> As Josh alluded to I strongly disagree with this. I’ve explained the reasons for this here: [#3093 (comment)](https://github.com/highlightjs/highlight.js/issues/3093#issuecomment-811504182) — tl;dr: standard library types and functions are not reserved words, and highlighting them in builtins leads to numerous false positives. They also don’t have any special semantics compared to user-defined types and functions, so giving them special highlighting _even where correct_ serves little purpose in making code more readable; on the contrary, it creates a distraction by emphasising an irrelevant distinction.
> 
> Compare this with IDEs and source code editors: the good ones don’t highlight these identifiers either (or rather, they highlight them the same as user-defined types and functions).
> 
> In fact, previous revisions of the C++ grammar underwent a rigorous cleanup to get rid of arbitrary identifiers that get highlighted incorrectly and I strongly advocate keeping it that way.

OK, fair enough. I'm still getting familiar with the 'highlighting mindset', and this makes sense. I checked a couple of editors/IDEs for this and it seems they're of mixed standpoint on this. I feel that it also has to do with the fact if the IDE/editor is using 'simple' regex-based highlighting or some form of static analysis using a compiler/'intellisense-like engine'.

Then I will clean out the list of commonly-used types. However, as I see it from your remark, one could say the list that is now called `TYPES` can go entirely? Or would you suggest to keep the container-types? What's the standpoint on this?

Also: what about the free functions that are listed here now? `std::move`, `std::memcpy` etc. this essentially is a list that is never really finished because there are so many 'common' functions.

I would also suggest based on your remark that I add a comment for anyone in the future wanting to do the same thing as I did. Because before I started, the `KEYWORDS` list was heavily polluted already.

krisvanrens:
@klmr BTW strongly agree with you to keep the grammar/highlight listings simple in favor of the fact that any special list is never finished. We would end up with hundreds of types and functions that would have to be rigorously updated every C++ release. But this would practically mean we have to lose the `COMMON_FUNCTIONS` and `TYPES` arrays altogether.

joshgoebel:
> But this would practically mean we have to lose the COMMON_FUNCTIONS and TYPES arrays altogether.

_No._  `COMMON_FUNCTIONS` exists for auto-detect NOT for highlighting.  It's ok if it's incomplete.  It's not used for coloring, only for guessing that the language is C++.

Why should TYPES go away again? @klmr   `int`, `char`, etc. are are built-in types.  This is a helpful distinction we make for many languages regardless of technicalities of the C++ spec itself.  So I'm not seeing the need to remove TYPES. (though perhaps some items may need to go?)

joshgoebel:
>  I feel that it also has to do with the fact if the IDE/editor is using 'simple' regex-based highlighting or some form of static analysis using a compiler/'intellisense-like engine'.

If we were a full parser highlighting some of these contextual keywords would make a lot of sense, but that's not what we strive to be... so unless it's very easy to pick off a few common edge cases contextually (we do this commonly for `class`, etc) then we'd rather avoid the false positives that come from flagging things incorrectly.

krisvanrens:
> > But this would practically mean we have to lose the COMMON_FUNCTIONS and TYPES arrays altogether.
> 
> _No._ `COMMON_FUNCTIONS` exists for auto-detect NOT for highlighting. **It's ok if it's incomplete.** It's not used for coloring, only for guessing that the language is C++.

Yes, right 😏  One day I'll get it right..


> Why should TYPES go away again? @klmr `int`, `char`, etc. are are built-in types. This is a helpful distinction we make for many languages regardless of technicalities of the C++ spec itself. So I'm not seeing the need to remove TYPES. (though perhaps some items may need to go?)

As far as I understand from Konrads' response: if used to actually highlight a 'common' type (which is what `TYPES` really is now, language-reserved types are in `RESERVED_TYPES`) this list suffers from the fact that it's never really finished as there are hundreds of common types (leading to false positives in custom code etc.). They are only special in the sense that they're provided by the standard library, but not from a more strict language-perspective. I guess for hinting purposes they can still be used, but I'm not sure if this is necessary.

It's this more strict view on what should be highlighted and when that hasn't been clear to me, especially not from the original implementation I started with a few weeks ago.

joshgoebel:
>  if used to actually highlight a 'common' type (which is what TYPES really is now, language-reserved types are in RESERVED_TYPES

Ah, I'm back on the page now.  I juggle 180 languages in my head. :-)  So TYPES shouldn't necessarily go away either, it should instead be renamed `TYPES_HINTS` (or some such) and added to keywords as `_typeHints` or some such so that it can serve the same purpose as COMMON_FUNCTIONS.  It should also be pruned significantly so that it only includes the "weirder" "strong signal" C++ types... things like "array" or "async" (keywords common to MANY other languages) should be removed.

joshgoebel:
When this is ready I'll go back in and add comments to the different lists so it's clear what purpose they serve to future contributors.

klmr:
Agreed, neither `COMMON_FUNCTIONS` nor `TYPES` should go away — but they shouldn’t be used for highlighting, just for language detection. As for whether the built-in types (`int` etc.) should be treated as *keyword* or as a *type* I defer to Josh’s preference. As far as C++ is concerned they’re keywords, and that’s what they were previously; but having them highlighted separately can be sensible.

However, they should be complete: currently `RESERVED_TYPES` is missing many keyword types that were previously part of the grammar (e.g. all the atomics).

krisvanrens:
> Agreed, neither `COMMON_FUNCTIONS` nor `TYPES` should go away — but they shouldn’t be used for highlighting, just for language detection. As for whether the built-in types (`int` etc.) should be treated as _keyword_ or as a _type_ I defer to Josh’s preference. As far as C++ is concerned they’re keywords, and that’s what they were previously; but having them highlighted separately can be sensible.
> 
> However, they should be complete: currently `RESERVED_TYPES` is missing many keyword types that were previously part of the grammar (e.g. all the atomics).

Right, I'll re-add the atomic types because they were there before and aren't now. But as far as the language definition is concerned they are not actually reserved keywords, they're simply library-introduced type aliases of `std::atomic<...>`. But then again, this array `RESERVED_TYPES` is used for special highlighting, not just for hinting. Where's the line of reason then not to add say, the iterator types? Or the threading synchronization primitives etc.?

I'm trying to get a proper understanding of your vision on this. Thanks!

klmr:
> But as far as the language definition is concerned they are not actually reserved keywords

You’re entirely right, my bad. I don’t remember the rationale for adding them here, I think this was an oversight on my part, resulting from confusion with C (where these types *are* privileged in the language grammar, even though they’re still only typedefs to `_Atomic` types).

joshgoebel:
> You’re entirely right, my bad. I don’t remember the rationale for adding them here, 

Lets leave them out for now then so we can start winding down this PR. :-)

krisvanrens:
> > But as far as the language definition is concerned they are not actually reserved keywords
> 
> You’re entirely right, my bad. I don’t remember the rationale for adding them here, I think this was an oversight on my part, resulting from confusion with C (where these types _are_ privileged in the language standard, even though they’re still only typedefs to `_Atomic` types).

Ok, no problem ;-) That's very confusing indeed.

I updated the implementation, and also removed the array `BUILT_INS` because those were C-builtins too. In C++ there's only `_Pragma`, but that's already covered in the `PREPROCESSOR` section.

AFAIK, we also have to remove the regular expression `CPP_PRIMITIVE_TYPES` now. It matches all `_t` types, which is considered incorrect now (e.g. there's false positives on trait helper alias templates). @klmr Do you concur? This would be the last thing then.

krisvanrens:
> > You’re entirely right, my bad. I don’t remember the rationale for adding them here,
> 
> Lets leave them out for now then so we can start winding down this PR. :-)

Ok, I updated the PR now (perhaps there's one more change -- see my other remark to Konrad).

I pruned the list of type hints (`string`, `array`, `async`, etc.) but perhaps there's a few more that can go, please check it quickly with your 180-language brain ;-)

krisvanrens:
> > > But as far as the language definition is concerned they are not actually reserved keywords
> > 
> > 
> > You’re entirely right, my bad. I don’t remember the rationale for adding them here, I think this was an oversight on my part, resulting from confusion with C (where these types _are_ privileged in the language standard, even though they’re still only typedefs to `_Atomic` types).
> 
> Ok, no problem ;-) That's very confusing indeed.
> 
> I updated the implementation, and also removed the array `BUILT_INS` because those were C-builtins too. In C++ there's only `_Pragma`, but that's already covered in the `PREPROCESSOR` section.
> 
> AFAIK, we also have to remove the regular expression `CPP_PRIMITIVE_TYPES` now. It matches all `_t` types, which is considered incorrect now (e.g. there's false positives on trait helper alias templates). @klmr Do you concur? This would be the last thing then.

Ugh, the highlighting makes me feel weird not seeing `int8_t` et al. being highlighted as types. Perhaps we must change the regex into an additional `TYPES` array containing every type from [here](https://en.cppreference.com/w/cpp/header/cstdint) eligible for highlighting.

src/languages/cpp.js
```diff
@@ -402,18 +396,22 @@ export default function(hljs) {
...
-    type: [...RESERVED_TYPES, ...TYPES],
+    type: RESERVED_TYPES,
     keyword: RESERVED_KEYWORDS,
     built_in: BUILT_INS,
-    literal: LITERALS
+    literal: LITERALS,
+    keywords: {
+      // Only for relevance, not highlighting.
+      _hint: TYPE_HINTS
+    }
```
joshgoebel, line 396:
```suggestion
    literal: LITERALS,
    _type_hints: TYPE_HINTS
```

You're already inside a keywords object here, are you not?

krisvanrens: Right, fixed it. Got an error with the Arduino highlighting markup test now though -- it seems all highlighting is missing from the result.

klmr:
> @klmr Do you concur? This would be the last thing then.

I do, for the reason you’ve mentioned. However, I acknowledge that this is a clash of standards: in POSIX all names with trailing `_t` are reserved. Of course that’s “irrelevant” as far as standard C++ is concerned but in practice a lot of C++ code uses POSIX libraries. And it’s confusing that in C++ `int8_t` isn’t a keyword type (and even more so because `char8_t` *is*).

krisvanrens:
> > @klmr Do you concur? This would be the last thing then.
> 
> I do, for the reason you’ve mentioned. However, I acknowledge that this is a clash of standards: in POSIX all names with trailing `_t` are reserved. Of course that’s “irrelevant” as far as standard C++ is concerned but in practice a lot of C++ code uses POSIX libraries. And it’s confusing that in C++ `int8_t` isn’t a keyword type (and even more so because `char8_t` _is_). @joshgoebel Are you OK with this? I'll fix it.

OK; then I suggest we add the aforementioned types from the [cstdint header](https://en.cppreference.com/w/cpp/header/cstdint), but drop the regular expression to prevent false positives in type trait helpers. I guess at this point in time it's a concession on our side to not have all POSIX reserved types highlighted.

Any future discussion on [POSIX reserved names](https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html) should also include type trait helper conventions (i.e. `_t` and `_v`) and maybe some more C++-related conventions (like the `atomic_` types).

Then at least for now there's a consistent set of base keywords and types.

src/languages/cpp.js
krisvanrens: Fixed this error as well, just a sec..

krisvanrens: (test is fixed now as well)

joshgoebel:
I'm wrapping this up and merging now that we seem at a nice spot... This PR is already too large.  Any remaining work can fall into a second PR.

krisvanrens:
> I'm wrapping this up and merging now that we seem at a nice spot... This PR is already too large. Any remaining work can fall into a second PR.

OK. I had a change ready for the remark I made to Konrad about the list of additional types. I can put it into a future PR.

joshgoebel (approved):

joshgoebel:
> AFAIK, we also have to remove the regular expression CPP_PRIMITIVE_TYPES now. It matches all _t types, which is considered incorrect now (e.g. there's false positives on trait helper alias templates).

What is a "trait helper template", can someone show me an example?

klmr:
[Traits](https://en.cppreference.com/w/cpp/header/type_traits) in C++ are types which describe properties of other types. They’re used for compile-time metaprogramming; e.g. `std::is_integral_v<int>` inherits from `true_type`, whereas `std::is_integral_v<float>` inherits from `false_type`.

There are other traits which describe types rather than values (e.g. `std::make_unsigned_t<int>`, which resolves to `int`), and those end in `_t` rather than `_v`.

joshgoebel:
Is the variety such that we couldn't handle this with a smarter regex?  IE, exclude `is*`,`*make`, etc but allow the other `_t` types to shine thru?

krisvanrens:
Perhaps we could detect the template arguments and the `std::` namespace indication, but then still if someone writes `using namespace std;` (kind of saying: I don't have to write `std::` anymore, it is implied) and then defines a custom type alias that ends in `_t` it would lead to a false positive. I don't know what your opinion about corner cases like this is? There's similar cases in other languages I would suspect?

joshgoebel:
>  I don't know what your opinion about corner cases like this is?

That we shouldn't overly concern ourselves with them, perhaps they never come up at all.

_I worry we're being too pre-emptive here._  That can often result in bloated grammars with questionable benefits.  I'm not yet convinced we need ANY further work here on `_t` types.  We have had the `_t` rule for a long time now with no complaints about false positives that I recall.  Perhaps they aren't happening, perhaps no one cares, or perhaps the existing behavior is somehow preferred?

I'd prefer to wait until someone raises an actual concern vs adding another list of keywords... there is a cost to all these explicit lists, they increase the file size of our bundle, etc...

klmr:
> Is the variety such that we couldn't handle this with a smarter regex? IE, exclude `is*`,`*make`, etc but allow the other `_t` types to shine thru?

I’m not sure I understand what that should accomplish: users are free to define their own names ending in `_v` or `_t`, and these obviously shouldn’t be highlighted. And the names in the standard library also shouldn’t be highlighted, analogous to my argument against highlighting other standard library names.

The clash between POSIX and C++ naming conventions is unfortunate but, ultimately, not something that can be solved (nor IMHO a real issue).

joshgoebel:
> And the names in the standard library also shouldn’t be highlighted, analogous to my argument against highlighting other standard library names.

These situations are different though... in that the problem with standard library is we can't (or don't want to) include a large enough list... so we will always fail to highlight SOME standard library things **making the highlighting appear inconsistent**... false negatives.  Resolving that doesn't add any complexity to the grammar.

This `_t` problem seems to be a case of false positives (of which no one has even filed an issue)... and it's remaining as a mode prevents us from adding yet another list of types from the `cstdint` header, (_which do look like things I would expect to see highlighted_).  It's behavior is also 100% predictable and consistent even if not 100% correct, making it different than the highlighting of standard library functions.

test/markup/cpp/template-complexity.expect.txt
joshgoebel: Look at Wren.

krisvanrens:
> > And the names in the standard library also shouldn’t be highlighted, analogous to my argument against highlighting other standard library names.
> 
> These situations are different though... in that the problem with standard library is we can't (or don't want to) include a large enough list... so we will always fail to highlight SOME standard library things **making the highlighting appear inconsistent**... false negatives. Resolving that doesn't add any complexity to the grammar.
> 
> This `_t` problem seems to be a case of false positives (of which no one has even filed an issue)... and it's remaining as a mode prevents us from adding yet another list of types from the `cstdint` header, (_which do look like things I would expect to see highlighted_). It's behavior is also 100% predictable and consistent even if not 100% correct, making it different than the highlighting of standard library functions.

Mmm. Yes. I'm not sure what the signal "no one has filed an issue" in the light of type traits in C++ really means. It's an advanced feature not everyone dabbles with (from what I can tell looking around me), multiplied with the fact that there's only a few people filing issues (again, my feeling) 😏 

I would very much agree to keep the grammar as simple as possible. However I'm still divided as to why or why not highlight other conventions if the `*_t` types are highlighted (e.g. the `atomic_*` types or `*_v` values). Choosing only one of them seems inconsistent, to me they feel equally-important. I would very much vouch for a regex-only approach here. Consistent highlighting for all standard library facilities is utopia if the grammar must be short.

Regarding the `cstdint` list; this must be done with a list..but it's not a long list and the types in it are quite relevant.

Is there a difference in how important the highlighting of 'types' vs. 'functions' vs. 'constants' etc. is considered for Highlight.js?

klmr:
> the problem with standard library is we can't (or don't want to) include a large enough list

That’s *one* problem. The other being that highlighting standard library names is not useful and creates many false positives when these names aren’t reserved and are routinely overridden in user code.

Regarding the `cstdint` names: for the same reason, I don’t particularly think they need to be highlighted, nor would I *expect* them to be highlighted. But I can see how others might want that (some editors *do* highlight e.g. `size_t` and `int16_t`) so I think it’s defensible. But user-defined type aliases ending in `_t` are idiomatic and *extremely common* in C++ code, so highlighting them all would lead to *many* false positives.

If you don’t want to include all names from `cstdint` manually, why not use something like this:

```js
/u?int(_fast|_least)?(8|16|32|64)_t|u?int(max|ptr)_t|size_t|ptrdiff_t|nullptr_t/
```

That captures all types defined in `cstdint` and three related types (`size_t`, `ptrdiff_t`, `nullptr_t`) defined elsewhere (but often treated identically).

joshgoebel:
> But user-defined type aliases ending in _t are idiomatic and extremely common in C++ code

I think we're conflating things a bit here.  _Our goal is not to only highlight built-ins or reserved types._  If we know `bob_data_t` is a type alias (by common convention) then it can be highlighted with `type` - this is a positive benefit.  `type` is NOT reserved to only built-in or reserved types. The same way we highlight classes in Ruby regardless of whether they are core library classes or user-defined classes, a class is a class.

The false positive argument of trait helpers does have some merit... could we exclude those with just a negative look-ahead for `<`?

klmr:
> If we know `bob_data_t` is a type alias (by common convention) then it can be highlighted with `type` - this is a positive benefit. 

Oh I agree, that would be awesome. The issue is that this introduces (substantial, not just occasional) inconsistencies since most user-defined types (/type aliases) are would *not* be highlighted for C++, since there is simply *no way*, based on syntax alone, to construct a heuristic with (IMHO) acceptable error rate to recognise type names in C++. There is just no sufficiently widespread convention. Even code that uses `_t` suffixes for some type aliases will commonly use many types and type aliases without that suffix.

klmr:
> The false positive argument of trait helpers does have some merit... could we exclude those with just a negative look-ahead for `<`?

Yes, that should be possible. There might be cases where this fails but these would be *very* rare and wouldn’t be catastrophic.

joshgoebel:
>  There is just no sufficiently widespread convention. Even code that uses _t suffixes for some type aliases will commonly use many types and type aliases without that suffix.

Well, now you've moved me right back to the middle again. LOL.  If `_t` truly isn't a common or widespread convention then getting rid of the `*_t` rule makes more sense.  Does `_t` signify anything useful at all?  In a mixed codebase why would someone use `_t` for some type aliases but not all of them?

krisvanrens:
> > There is just no sufficiently widespread convention. Even code that uses _t suffixes for some type aliases will commonly use many types and type aliases without that suffix.
> 
> Well, now you've moved me right back to the middle again. LOL. If `_t` truly isn't a common or widespread convention then getting rid of the `*_t` rule makes more sense. Does `_t` signify anything useful at all? In a mixed codebase why would someone use `_t` for some type aliases but not all of them?

The typename postfix `_t` is a convention in the _standard_ for many types (but not all, e.g. `vector`/`string`/etc.). What Konrad means here (correct me if I'm wrong) is that there's no "general consensus" on this convention in _user land_. Some code bases do it, most don't.

joshgoebel:
> /u?int(_fast|_least)?(8|16|32|64)_t|u?int(max|ptr)_t|size_t|ptrdiff_t|nullptr_t/

I'd be ok with this in this use case, but generally I try and stay away from this for lack of maintainability... but this is clear enough. We'd probably want `\b` on both ends though with the middler wrapped in `()`?

klmr:
> We'd probably want `\b` on both ends

Definitely, the regex wasn’t meant to be production ready since I wasn’t sure in which context it would be applied. In general we would want word boundaries, correct.

krisvanrens:
> > /u?int(_fast|_least)?(8|16|32|64)_t|u?int(max|ptr)_t|size_t|ptrdiff_t|nullptr_t/
> 
> I'd be ok with this in this use case, but generally I try and stay away from this for lack of maintainability... but this is clear enough. We'd probably want `\b` on both ends though with the middler wrapped in `()`? We could also use `variant` or `regex.either` to split it up just a little if we so desired.

In my opinion this lack of maintainability, or rather: static definition of additional types is also a signal to future maintainers: "there's no room for an extensive list of other types here, we're being conservative about it."

I would have worked for me 😉

joshgoebel:
I read it more as:

> bugger off unless your a regex expert, we're elite RegEx ninjas here

Though as I said this one ain't so bad.  If we want to signal future maintainers though I'd prefer we write a readable note in the comments. :-)

PR #3180, opened 2021-05-07
Add support for inline code highlighting

### LINKED ISSUE #945: Highlight.JS Inline Code Syntax Highlightning
Hello!

Highlight.js is a great highlightning tool. But now i want to know if it is also possible to make inline highlighting, so not like a block (like at http://tekkkz.com/posts/articles/arch_installation.html) but inside a text paragraph?

If it is possible, how?

JamesTheAwesomeDude:
Resolves #945

### Changes

This pull request adds an optional parameter to `highlightAll()`, which can make it use a query selector other than `pre code`. It does not change any existing behavior of the software.

### Checklist
- [x] Not a grammar change, markup tests not needed.
- [x] I could not identify which section of the changelog to file this under, either:

This wasn't a `deprecation` or `removal`; I'm unclear if it's big enough to bother listing as an `API change` high-level bullet; it doesn't belong under `security`, `themes`, `language grammars`, `parser`, `grammars`, `new languages`, `theme improvements`, `new themes`, or `dev improvements` (that last one looks like it's to do with changes to your in-house tooling system.)

It's a tiny, optional, backwards-compatible, mostly-user-facing `enhancement`. But it looks like that's not currently a header/category in the changelog. Should I create it?


JamesTheAwesomeDude:
I could not identify which section of the changelog to file this under, either:

This wasn't a `deprecation` or `removal`; I'm unclear if it's big enough to bother listing as an `API change` high-level bullet; it doesn't belong under `security`, `themes`, `language grammars`, `parser`, `grammars`, `new languages`, `theme improvements`, `new themes`, or `dev improvements` (that last one looks like it's to do with changes to your in-house tooling system.)

It's a tiny, optional, backwards-compatible, mostly-user-facing `enhancement`. But it looks like that's not currently a header/category in the changelog. Should I create it?

joshgoebel:
This type of item would fall under `parser`, that's the catch all for overall improvements to the parser, it's API, config, etc...

---

While selector does seem a natural choice [as an argument] here one could also suggest that other items pertaining directly to how we perform "highlight all" (such as `noHighlightRe`, `classPrefix`) might be useful to pass as well, but that is not desirable. We'd prefer to keep this API clean and simple.  Configuration should happen via our `configure` API, ie:

```js
hljs.configure({cssSelector: "code"});
hljs.highlightAll();
```

---

Are you willing to re-work your PR to instead add a `cssSelector` configuration option?

joshgoebel:
It's then also worth discussing if our default CSS may need to be changed to offer both the `pre code` and `code` variants of the top-level `.hljs` selector.

JamesTheAwesomeDude:
> It's then also worth discussing if our default CSS may need to be changed to offer both the a `pre code` and `code` variant of the top-level `.hljs` selector.

Yes, I think that may also require a lot of refactoring.

I'm [currently jury-rigging it](http://www.ishygddt.xyz/~blog/2021/05/syntax-highlighting-for-inline-code-in-wordpress) with a `:not` selector, but this just screams "FIXME":

```css
code.hljs:not(pre>code) {
  padding: initial;
  display: initial;
}
```

(This may be idiopathic to WordPress, which has a sort of "convention" — and by "sort of", I mean it's nearly immutably how the HTML is operated while using the software — that Block `<code>` elements are always designated by being wrapped in `<pre>`, and that their Inline counterparts are designated by nothing more than not being so wrapped.)

-----

> Configuration should happen via our existing configure API, ie:
> 
> ```js
> hljs.configure({cssSelector: "code"});
> hljs.highlightAll();
> ```
> 
> Are you willing to re-work your PR to instead add a `cssSelector` configuration option?

Certainly! Since I'm currently unfamiliar with your API, you've done 90% of the work for me just there by defining the implementation! :grin: (I've just pushed the changes; how do they look?)

src/highlight.js
```diff
@@ -70,7 +70,9 @@ const HLJS = function(hljs) {
     languages: null,
     // beta configuration options, subject to change, welcome to discuss
     // https://github.com/highlightjs/highlight.js/issues/1086
-    __emitter: TokenTreeEmitter
+    __emitter: TokenTreeEmitter,
+    // https://github.com/highlightjs/highlight.js/pull/3180
+    allCodeSelector: 'pre code',
```
joshgoebel, line 75: Let's go with `cssSelector` the scope of all these settings is implied to be all, they are global.

docs/api.rst
```diff
@@ -111,7 +112,7 @@ Accepts an object representing options with the values to updated. Other options
 highlightAll
 ------------
 
-Applies highlighting to all ``<pre><code>...</code></pre>`` blocks on a page.
+Applies highlighting to all elements on a page matching ``allCodeSelector``.
```
joshgoebel, line 115:
```suggestion
Applies highlighting to all elements on a page matching the configured ``cssSelector``.
The default ``cssSelector`` value is ``pre code``, which highlights all code blocks. 
```

joshgoebel:
>  that Block <code> elements are always designated by being wrapped in <pre>, and that their Inline counterparts are designated by nothing more than not being so wrapped.)

That sounds 100% semantically correct to me.  So why can't that just be accomplished with two rules?

```
pre code.hljs {
  display: block;
  overflow-x: auto;
  padding: 0.5em;
}

code.hljs { 
  padding: ...;
}
```

The block level items then only apply to pre/code as they should... and we only add a small amount of padding to inline elements.

docs/api.rst
JamesTheAwesomeDude: > highlights all code blocks

This phrasing seems to suggest that `<code>` tags _not_ wrapped in `<pre>` are **not code blocks**; would that be a good justification to remove `display: block;` from the `.hljs` rule (possibly migrating it to a rule for `pre .hljs`) in a future PR?

joshgoebel: > to suggest that <code> tags not wrapped in <pre> are not code blocks;

They aren't - they are just inline code.  A few words of code does not a block make.  We could add:

```
If you wish to highlight inline `<code>` elements also, try `cssSelector: "code"`.
```

Or some such... to make it clearer.

joshgoebel: I think that makes it clear that block is used in the CSS sense of the the meaning.

src/highlight.js
```diff
@@ -72,7 +72,7 @@ const HLJS = function(hljs) {
     // https://github.com/highlightjs/highlight.js/issues/1086
     __emitter: TokenTreeEmitter,
     // https://github.com/highlightjs/highlight.js/pull/3180
```
joshgoebel, line 74:
```suggestion
```

Token emitter is special, not every item here needs a link to a PR.

docs/api.rst
JamesTheAwesomeDude: > They aren't - they are just inline code.

I take it, then, that you would be amenable to (in a future PR) removing or moving [the line that transforms them into blocks](https://github.com/highlightjs/highlight.js/blob/11.0.0-beta0/src/styles/default.css#L11-L12)?

joshgoebel:
>  to remove display: block; from the bare .hljs rule

I don't think we need a separate PR for that, the scope of this one is small enough to finish this work I think.

This will merit a mention in breaking changes now since it would break anyone using `div.hljs` or something weird (they'll need to update their CSS)... but I can handle that if you don't know where to put it.

JamesTheAwesomeDude:
Yeah, I could use a hand with the documentation / I'm not qualified to tear down chesterton's fence in this case; thanks.

docs/api.rst
```diff
@@ -120,7 +122,7 @@ initHighlighting
 
 *Deprecated as of 10.6:* Please use ``highlightAll()`` instead.
 
-Applies highlighting to all ``<pre><code>...</code></pre>`` blocks on a page.
+Applies highlighting to all elements on a page matching ``allCodeSelector``.
```
joshgoebel, line 125:
```suggestion
Applies highlighting to all elements on a page matching ``cssSelector ``.
```

CHANGES.md
```diff
@@ -54,6 +54,7 @@ Parser:
 - adds `title.class` sub-scope support (#3078) [Josh Goebel][]
 - adds `title.function` sub-scope support (#3078) [Josh Goebel][]
 - adds `beforeMatch` compiler extension (#3078) [Josh Goebel][]
+- adds `allCodeSelector` option (#3180) [James Edington][]
```
joshgoebel, line 57:
```suggestion
- adds `cssSelector ` configuration option (#3180) [James Edington][]
```

CHANGES.md
```diff
@@ -129,6 +130,7 @@ Dev Improvements:
 [NullVoxPopuli]: https://github.com/NullVoxPopuli
 [Mike Watling]: https://github.com/Pickysaurus
 [Nico Abram]:https://github.com/nico-abram
```
JamesTheAwesomeDude, line 132:
```suggestion
[Nico Abram]: https://github.com/nico-abram
```

src/styles/default.css
```diff
@@ -9,7 +9,6 @@
 */
 
 .hljs {
-  display: block;
   overflow-x: auto;
   padding: 1em;
```
JamesTheAwesomeDude, line 13: The padding on these is quite aggressive when applied to inline elements: 
Screenshot from 2021-05-07 05-43-20

Should that be popped out onto its own rule for "code blocks"?

```diff
 .hljs {
   overflow-x: auto;
-  padding: 1em;
   background: #F0F0F0;
   color: #444;
 }
 
+pre .hljs {
+  padding: 1em;
+}
```

Doing so yields the following result:

Screenshot from 2021-05-07 05-55-36

joshgoebel: I thought you were done - I just picked it up and ran with it.  Does it look like I covered things?

JamesTheAwesomeDude:
(Of note: I'd left the trailing comma on the new line in order for future diffs to be cleaner; I don't know whether the current setting of the linter is correct.)

src/highlight.js
```diff
@@ -71,7 +71,7 @@ const HLJS = function(hljs) {
     // beta configuration options, subject to change, welcome to discuss
     // https://github.com/highlightjs/highlight.js/issues/1086
     __emitter: TokenTreeEmitter,
-    cssSelector: 'pre code',
```
JamesTheAwesomeDude: (I'd left the trailing comma on the new line in order for future diffs to be cleaner; I don't know whether the current setting of the linter is correct.)

src/styles/default.css
JamesTheAwesomeDude: Yes, that looks alright.

src/styles/default.css
```diff
@@ -8,15 +8,28 @@
...
+Typically this "required" baseline CSS is added by `makestuff.js` during build.
+*/
+pre code.hljs {
   display: block;
   overflow-x: auto;
   padding: 1em;
+}
+
+code.hljs {
+  padding: 3px 5px;
```
JamesTheAwesomeDude, line 24: People use all sorts of different high-res (or low-res) displays these days; should this be `pt`-based instead of `px`-based?

```suggestion
  padding: 4pt 6pt;
```

src/highlight.js
joshgoebel: We like clean diffs but we don't like extra commas. :)  They used to **break** things, but that's no longer true.  Perhaps something to revisit in the future, but for now we obey the linter.

src/styles/default.css
joshgoebel: Curious, what cases/devices are you thinking of here?  Everything I'm aware of accounts for Retina (2x, 3x, etc) screens automagically so the 3px will already be 6px or 9px or whatever is needed to derive the correct 3px-ish appearance regardless of screen pixel density.

joshgoebel (approved):

joshgoebel:
@JamesTheAwesomeDude Thanks so much for working on this.

PR #3181, opened 2021-05-07
enh(sqf) improve function name highlighting (slightly broader)

JonBons:
I've tweaked the function matching regex for the sqf language so that it can match functions with the `myTag_component_fnc_initVehicle` naming convention which is common with popular mods for arma, CBA, ACE, etc.

### Changes
Changed regex for sqf functions to `\w+_fnc_\w+` which matches the whole word now instead of just one section.

### Checklist
- [x] Updated markup tests,

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -126,6 +126,7 @@ Dev Improvements:
 [Josh Goebel]: https://github.com/joshgoebel
 [Ryan Mulligan]: https://github.com/ryantm
 [R3voA3]: https://github.com/R3voA3
+[JonBons]: https://github.com/JonBons
 [Wei Su]: https://github.com/swsoyee
 [Jared Luboff]: https://github.com/jaredlll08
 [NullVoxPopuli]: https://github.com/NullVoxPopuli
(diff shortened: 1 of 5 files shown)


JonBons:
I have not updated the changelog yet as I did not see a Version next or whatever to place it under and I don't want to cause a merge conflict. Where should I place it?

joshgoebel:
Thanks!

The topmost section is the v11 stuff it would go there, at the bottom of that section with all the other grammar improvements.

src/languages/sqf.js
```diff
@@ -19,7 +19,7 @@ export default function(hljs) {
   // https://community.bistudio.com/wiki/Functions_Library_(Arma_3)#Adding_a_Function
   const FUNCTION = {
     className: 'title',
-    begin: /[a-zA-Z][a-zA-Z0-9]+_fnc_\w*/
+    begin: /\w+_fnc_\w+/
```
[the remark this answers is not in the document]
JonBons: Yeah, tested and I believe this should be good.

CHANGES.md
```diff
@@ -82,6 +82,7 @@ Grammars:
 - enh(nginx) improving highlighting of some sections [Josh Goebel][]
 - fix(vim) variable names may not be zero length [Josh Goebel][]
 - enh(sqf) Updated keywords to Arma 3 v2.02 (#3084) [R3voA3][]
+- enh(sqf) Refactored function regex to match CBA component func naming scheme (#3181) [JonBons][]
```
joshgoebel, line 85: You need to also add the link on line ~125 or so.

JonBons: :+1:

joshgoebel (approved):

PR #3193, opened 2021-05-17
Update sqf.js to Arma 3 v2.05

Leopard20:
Updates sqf.js to match the latest version of Arma 3 (v2.05)

### Changes

- Added new commands
- Removed invalid commands
- Removed `#` as an illegal character (legal as of Arma 3 v1.82)


joshgoebel:
@R3voA3 Any chance you could review this to see if it feels about right?

joshgoebel:
> Removed # as an illegal character (legal as of Arma 3 v1.82)

How so?  How is # used now in the language?

Leopard20:
> 
> 
> > Removed # as an illegal character (legal as of Arma 3 v1.82)
> 
> How so? How is # used now in the language?

https://community.bistudio.com/wiki/a_hash_b

R3voA3:
@dedmen

src/languages/sqf.js
```diff
@@ -2473,6 +2473,6 @@ export default function(hljs) {
       STRINGS,
       PREPROCESSOR
     ],
-    illegal: /^\$ /
+    illegal: /\w\$|\?|\@/
```
joshgoebel, line 2476: Please break these out into an array and add a comment explaining why each of these is illegal to help future readers/maintenance.

src/languages/sqf.js
```diff
@@ -1858,7 +1840,7 @@ export default function(hljs) {
     'setCurrentTask',
     'setCurrentWaypoint',
     'setCustomAimCoef',
-    'setCustomMissionData',
+    'SetCustomMissionData',
```
joshgoebel, line 1843: These changes (at a glance) seem almost random.  Are you carefully making them by hand or is there some script you are using to fetch the keyword list directly from Arma?

joshgoebel: And the whole grammar is flagged `case_insensitive`, so is there truly a correct casing here at all - that keeps changing from version to version?

R3voA3: We use this highlighter for the Arma 3 Discord where casing doesn't really matter.
The casing is set in the keyword description in the engine and are fetched via a script. The casing seems random and inconsistent because they are set like this in engine.

joshgoebel: Can we get a link to the script you are using added to the source in comments for the aid of future maintainers? (or if it's super short we could include the whole script in a comment)

Leopard20: I wrote it myself

joshgoebel: ~4 lines is pretty short.   You can attach it here so we can have a look at it.

Leopard20: @joshgoebel It was a bit longer than what I thought
```sqf
/*
	# Author: Leopard20
	
	# Description:
	This script can be used to dump all commands to clipboard..
	Make sure you're using the Diag EXE to dump all of the commands.
	
	# How to use:
	Simply copy the array of KEYWORDS and LITERALS from this sqf.js file and paste it in the script.
	Execute the script from the debug console.
	All commands will be copied to clipboard
*/

_KEYWORDS = ['if'];                                                //Array of all KEYWORDS
_LITERALS = ['west'];                                              //Array of all LITERALS
_allCommands = createHashMap;
{
	_type = _x select [0,1];
	if (_type != "t") then {
		_command_lowerCase = ((_x select [2]) splitString " ")#(((["n", "u", "b"] find _type) - 1) max 0);
		_command_upperCase = supportInfo ("i:" + _command_lowerCase) # 0 # 2;
		_allCommands set [_command_lowerCase, _command_upperCase];
	};
} forEach supportInfo "";
_allCommands = _allCommands toArray false;
_allCommands sort true;                                            //sort by lower case
_allCommands = ((_allCommands apply {_x#1}) -_KEYWORDS)-_LITERALS; //remove KEYWORDS and LITERALS
copyToClipboard str (_allCommands select {_x regexMatch "\w+"});
```

src/languages/sqf.js
Leopard20: @joshgoebel is this ok?
```js
illegal: [
      '\w\$',  //$ is only valid when used with Hex numbers (e.g. $FF) or at the beginning of localized strings (e.g. "$STR_HELLO")
      '\?',    //There's no ? in SQF
      '\@/'    //There's no @ in SQF
    ]
```

src/languages/sqf.js
```diff
@@ -2473,6 +2503,10 @@ export default function(hljs) {
       STRINGS,
       PREPROCESSOR
     ],
-    illegal: /\w\$|\?|\@/
+    illegal: [
+      '\w\$',  //$ is only valid when used with Hex numbers (e.g. $FF) or at the beginning of localized strings (e.g. "$STR_HELLO")
+      '\?',    //There's no ? in SQF
+      '\@/'    //There's no @ in SQF
```
joshgoebel: These should be regex:

```suggestion
      /\w\$/,  //$ is only valid when used with Hex numbers (e.g. $FF) or at the beginning of localized strings (e.g. "$STR_HELLO")
      /\?/,    //There's no ? in SQF
      /@\//    //There's no @ in SQF
```

Why does @ need to be escaped?

Leopard20: I don't know. It doesn't?

Leopard20: @joshgoebel 
```js
illegal: [
      /\w\$/,  //$ is only valid when used with Hex numbers (e.g. $FF) or at the beginning of localized strings (e.g. "$STR_HELLO")
      '?'   ,  //There's no ? in SQF
      '@'      //There's no @ in SQF
    ]
```
is this better?

joshgoebel: No, they should be regex (not strings).  The `?` does need to be escaped (as I showed)I since `?` is a quantifier.  Strings are just "raw regex" (escaping is not performed) and we prefer actual regex whenever possible to avoid this type of confusion.

joshgoebel: Does `@` mean something special in regex that I'm just not thinking of?

Leopard20: I don't know, but both escaped and non-escaped versions match correctly. So I'm assuming it's not needed as you said.

joshgoebel:
Can you try to sort out the failing test case or if you have no idea I can pick this up and most likely resolve it for you.

Leopard20:
Sorry, I tried looking at it but I have no idea what's wrong.

src/languages/sqf.js
dedmen: correct script command casing is supposed to be camelCase.
Sometimes in engine we do CamelCase for incomplete things, and convert it to correct one later.
Sometimes thats also forgotten

dedmen:
I think whats wrong is this
https://github.com/highlightjs/highlight.js/blob/main/test/detect/smalltalk/default.txt
file being detected as SQF
I guess easy workaround would be to add `|` to not allowed things?
`||` is valid, `|` with whitespace around it is not.

https://github.com/highlightjs/highlight.js/blob/main/test/detect/smalltalk/default.txt#L6
`.` is also not allowed in SQF
https://github.com/highlightjs/highlight.js/blob/main/test/detect/smalltalk/default.txt#L17
`:=` also not
https://github.com/highlightjs/highlight.js/blob/main/test/detect/smalltalk/default.txt#L27
`[` also not

This is kinda brute force fixing, but.. why not?

src/languages/sqf.js
```diff
@@ -47,21 +77,23 @@ export default function(hljs) {
   };
 
   const KEYWORDS = [
-    'case', 'catch',
+    'break', 'breakWith', 'breakOut', 'breakTo',
```
dedmen, line 80: are breakWith, breakOut, breakTo, continue, continueWith really keywords?
They are really just script  commands.
But then again waitUntil and such are in here too... Guess its fine

joshgoebel:
Per @dedmen lets beef up the `illegal` to rule out some common things you find in Smalltalk but not SQF:

https://github.com/highlightjs/highlight.js/pull/3193#issuecomment-843088729

That should do the trick.

> This is kinda brute force fixing, but.. why not?

`illegal` is the best way to fix auto-detect issue because usually it's only a line or two and it tends to stay fixed over the long-haul as long as the languages don't add new features that use those illegals, etc.

dedmen:
With the illegal its just important that it doesn't detect these things in strings, but I assume thats already handled.
No idea how that works ^^

joshgoebel:
> With the illegal its just important that it doesn't detect these things in strings, but I assume thats already handled.

Every mode has it's own `illegal` regex... so a string mode would not inherit the global illegal, so no issue.

joshgoebel:
Ping.  I think where we were at on this is:

> Per @dedmen lets beef up the illegal to rule out some common things you find in Smalltalk but not SQF:

Need to beef up illegal further to fix the failing auto-detect tests...

joshgoebel:
Closing until someone has time to return to this and finish it up.  Happy to reopen this whenever you're ready.

PR #3195, opened 2021-05-18
enh(r) Add operators and punctuation

### LINKED ISSUE #3194: (R) Numbers a not highlighted in multiple circumstances
**Describe the issue**

1. Numbers aren’t highlighted at the very start of the document
2. Numbers aren’t highlighted after <code>&nbsp;<-&nbsp;</code> (and possibly after other constructs)
3. `->` is declared as illegal (to distinguish it from Haskell) *despite being legal in R*; this *might* be an OK *heuristic*, but the documentation makes it sound as if `illegal` is more than just a gentle nudge in the weighting, and this wouldn’t be appropriate.

R

`highlight`

**Sample Code to Reproduce**

```r
1
a <- 1
```

**Expected behavior**

```html
<span class="hljs-number">1</span>
a &lt;- <span class="hljs-number">1</span>
```

I’ve got a PR to fix these issues and also implements operators and punctuation highlighting at the same time. But, alas, it now fails to highlight numbers directly after operators/punctuation, because the rule for numbers eats the preceding character. I thought we had a solution for this that didn’t require look-behind, but that solution was removed in 0e9fb35f1 (ostensibly this shouldn’t have caused a behavioural change, so it’s possible that the previous version already had the errors described above).

<details>
<summary>Here’s the patch I used to add operators (which fixes problem (2)) and allow numbers at the beginning of the document, but this now causes numbers directly after operators to fail highlighting.</summary>

```patch
diff --git a/src/languages/r.js b/src/languages/r.js
index a4348678..5cb90f3e 100644
--- a/src/languages/r.js
+++ b/src/languages/r.js
@@ -144,32 +140,42 @@ export default function(hljs) {
           {
             match: [
-              /[^a-zA-Z0-9._]/, // not part of an identifier
+              /[^a-zA-Z0-9._]|^/, // not part of an identifier
               /0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*[pP][+-]?\d+i?/
             ]
           },
           {
             match: [
-              /[^a-zA-Z0-9._]/, // not part of an identifier
+              /[^a-zA-Z0-9._]|^/, // not part of an identifier
               /0[xX][0-9a-fA-F]+(?:[pP][+-]?\d+)?[Li]?/
             ]
           },
           {
             match: [
-              /[^a-zA-Z0-9._]/, // not part of an identifier
+              /[^a-zA-Z0-9._]|^/, // not part of an identifier
               /(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?[Li]?/
             ]
           }
         ]
       },
       {
-        // infix operator
-        begin: '%',
-        end: '%'
+        className: 'operator',
+        variants: [
+          {
+            // custom infix operator
+            begin: '%',
+            end: '%'
+          },
+          {
+            // Base operators minus replacement functions and punctuation,
+            // plus aliases.
+            match: /[=!<>:]=|\|\||&&|:::?|<-|<<-|[-+*\/?!$&|:<=>@^~]|->|->>|\*\*/
+          }
+        ]
       },
-      // relevance boost for assignment
       {
-        begin: regex.concat(SIMPLE_IDENT, "\\s+<-\\s+")
+        className: 'punctuation',
+        match: /\[\[|[(){}[\]]/
       },
       {
         // escaped identifier
```

</details>

klmr:
This change adds highlighting for operators and punctuation, and fixes #3194.

### Changes

* Correct syntax highlighting for numbers adjacent to operators or at the beginning of the document
* Remove incorrect declaration of `->` as illegal
* Add support for highlighting operators and punctuation
* Update language definition to use v11 names (`className` => `scope`)

*Unfortunately* this introduces a slight regression, hence a draft PR: the auto-detection sample now fails for R, since `<-` is no longer used as a hint, instead it now has relevance 0 along with all other operators (without this, many languages incorrectly auto-detect as R, presumably since they don’t define operators).

Is there a way of adding an auto-detection hint for `<-` assignment *without* adding yet another variant to the two separate definitions of operators (L161 and L202)?

Also, is the formatting OK as it is? I’m no JS guy, and the linter I’m using is going slightly crazy, though it doesn’t seem entirely consistent in its complaints.

### Checklist
- [x] Added/updated markup tests, or they don't apply here because...
- [x] Fix failing auto-detection


src/languages/r.js
```diff
@@ -118,7 +126,7 @@ export default function(hljs) {
       hljs.HASH_COMMENT_MODE,
 
       {
-        className: 'string',
+        scope: 'string',
```
joshgoebel, line 129: This noise makes the PR much harder to review when it's now separated in it's own commit.

klmr: Gah, true. I can pull it out if you want.

joshgoebel:
> Is there a way of adding an auto-detection hint for <- assignment without adding yet another variant to the two separate definitions of operators (L161 and L202)?

Does it really need to be in BOTH? Often spaces will be used and we'll match (and color it) either way with the number rule.. so perhaps the relevancy boost can only handle the "on it's own" case... in which case it's just 2-3 lines of code much like exactly what we had before.

klmr:
> Perhaps that would fix the auto-detect?

I’ll try!

src/languages/r.js
joshgoebel: Didn't turn out to be so terrible, but thought for the future...

src/languages/r.js
```diff
@@ -131,48 +139,80 @@ export default function(hljs) {
...
+              NUMBER_TYPES
             ]
           },
           {
+            scope: { 2: 'number' },
+            relevance: 0,
             match: [
-              /[^a-zA-Z0-9._]/, // not part of an identifier
-              /(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?[Li]?/
+              /[^a-zA-Z0-9._]|^/, // not part of an identifier
```
joshgoebel, line 189: Please document the use of `^`.

src/languages/r.js
```diff
@@ -131,48 +139,80 @@ export default function(hljs) {
...
+          {
+            scope: {
+              1: 'punctuation',
+              2: 'number'
+            },
             match: [
-              /[^a-zA-Z0-9._]/, // not part of an identifier
-              /0[xX][0-9a-fA-F]+(?:[pP][+-]?\d+)?[Li]?/
+              PUNCTUATION,
+              NUMBER_TYPES
```
joshgoebel, line 182:
```suggestion
              PUNCTUATION_RE,
              NUMBER_TYPES_RE
```

src/languages/r.js
```diff
@@ -18,13 +18,20 @@ export default function(hljs) {
...
+  const NUMBER_TYPES = regex.either(
+    // Special case: only hexadecimal binary powers can contain fractions
+    /0[xX][0-9a-fA-F]+\.[0-9a-fA-F]*[pP][+-]?\d+i?/,
+    // Hexadecimal numbers without fraction and optional binary power
+    /0[xX][0-9a-fA-F]+(?:[pP][+-]?\d+)?[Li]?/,
+    // Decimal numbers
+    /(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?[Li]?/
+  );
+  const OPERATORS = /[=!<>:]=|\|\||&&|:::?|<-|<<-|->>|->|\|>|[-+*\/?!$&|:<=>@^~]|\*\*/;
+  const PUNCTUATION = /\[\[|[(){}[\]\\,]/;
```
joshgoebel, line 30: This just makes my head hurt, can we use regex either and have an array of punctuation?  For some reason the operators is a bit easier to read.

src/languages/r.js
```diff
@@ -178,8 +178,8 @@ export default function(hljs) {
               2: 'number'
             },
             match: [
-              PUNCTUATION,
-              NUMBER_TYPES
+              PUNCTUATION_RE,
+              NUMBER_TYPES_RE
```
joshgoebel, line 182: You'l have to actually rename them everywhere though :)

klmr: On it, on it!

src/languages/r.js
klmr: Like this? 252483ca4a0e1cd407b5787d9deb470d1ce5dece

src/languages/r.js
```diff
@@ -131,48 +146,84 @@ export default function(hljs) {
...
         ]
       },
+
+      // Operators/punctuation when they're not directly followed by numbers
       {
-        // infix operator
-        begin: '%',
-        end: '%'
+        scope: 'operator',
+        match: /<-/
```
joshgoebel, line 206: Please add the full assignment regex back or is there some reason this had to change?  More explicit is better than less typically.

klmr: The previous assignment regex was only used as a relevance boost, and only matches a small subset of all possible assignments, namely a subset of all legal identifiers, plus whitespace, plus assign, plus whitespace.

We can’t use this for highlighting since it misses many cases:

* Legal identifiers not included in the subset used here
* Lack of whitespace before or after the assignment
* Assignment to a complex expression rather than an identifier (e.g., at its simplest, `x[1] <- 2`)

I’m actually not sure why this regex was so overly specific even for relevance boosting: was it to avoid clashing with other languages? At any rate, assignment expressions in R can be (and routinely are, in actual code) quite complex.

klmr: Sorry, commit bf1dc80fecfba5d8cca6c5f63494bf139b197d44 included an unnecessary, redundant declaration, I’ve removed that in cb7a392ce026805205a3adfed295885a839e8766. The result should be alright: `<-` is now handled in all situations, but the signal boost is only applied with the previous situation (except using the full `IDENT_RE`).

src/languages/r.js
```diff
@@ -131,48 +146,91 @@ export default function(hljs) {
...
           },
           {
+            scope: {
+              1: 'operator',
+              2: 'number'
+            },
             match: [
-              /[^a-zA-Z0-9._]/, // not part of an identifier
-              /0[xX][0-9a-fA-F]+(?:[pP][+-]?\d+)?[Li]?/
+              /%[^%]*%/,
```
joshgoebel, line 178: What is this again in R?  And %% is also valid?

joshgoebel:
Looking good!

src/languages/r.js
klmr: Any `%…%` sequence is a valid binary operator, `%%` included.

src/languages/r.js
```diff
@@ -131,48 +146,91 @@ export default function(hljs) {
...
-      // relevance boost for assignment
+
+      {
+        scope: 'operator',
+        variants: [
+          {
+            relevance: 0,
+            match: OPERATORS_RE
+          },
+          { match: /%[^%]*%/ }
```
joshgoebel, line 222: > Any %…% sequence is a valid binary operator, %% included (and this also includes whitespace! So a % foo % b is a valid expression with two operands, a and b).

So I'm worried about %%... I've had similar issues with `||` in the past where a Ruby rule for `|param|` was eating ors (`||`) in other languages because it was coded too widely.  I wonder/worry about comments/patterns in some other languages.  I've seen issues with strange C headers interacting with certain rules in the past.

Made up example:

```
= %%%%%%%%%%%%%%%%%%%%%%%%%%%%%
= Header
= %%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

I've seen weird stuff like this... I think we may need:

- `%%` relevance: 0
- `%...%` 

---

You know... just make the relevance 0 period... I'm not sure that `%` operators are worthy of special relevance... it's used as modulus in many languages so % not % % as a pattens is rife for false positives with simple math formulas.  I'm going to say that this having relevance in the past has likely been a mistake.

klmr: Ah, shame, I was hoping that we could precisely use `%…%` operators for a nice relevance boost since it’s so specific to R, but of course this overlooked its use as a comment fence.

I think rather than do any more whacky matching I’ll go with your last suggestion and just dial its relevance down to 0 everywhere.

klmr: … hmm, nope, that leads to a failing detect test, the sample now gets detected as Vala. Which is weird, since that’s actually a *very* different language (looks a lot like Java), but hl.js has constant issues telling the two apart.

joshgoebel: >  but of course this overlooked its use as a comment fence.

And it's use as modulus...  if there were a super common R pattern perhaps there is still room for that:

```
% IDENT %
```

But I don't think just % anything % later works (and was far too overly broad before)...

klmr: Why does Vala have such a high boost for preprocessor comments?

https://github.com/highlightjs/highlight.js/blob/6e182e0f306114d68dab261a63404ab50e40cb87/src/languages/vala.js#L51-L56

That’s completely unjustified, it’s an extremely common syntax for comments. Removing that relevance boost fixes the detection issue.

joshgoebel: My issues with `%...%` still stand (they aren't resolved by a change to Vala), but yes I agree that's Vala being misbehaved and would welcome a PR to fix that.  The default relevance of 1 should apply there.

joshgoebel: Hahaha, I guess we can include it here, please mention it in changelog though. :-)

klmr: > My issues with `%...%` still stand

Are you sure ? It now has `relevance: 0` everywhere.

test/markup/r/names.expect.txt
```diff
@@ -1,26 +1,26 @@
 <span class="hljs-comment"># Valid names</span>
 
-a1_foo, A1_FOO, .foo_, ._foo, Bar.42, foo..1, ., ._, .., ..., ..1, <span class="hljs-built_in">c</span>, <span class="hljs-built_in">T</span>, <span class="hljs-built_in">F</span>, ._1
+a1_foo<span class="hljs-punctuation">,</span> A1_FOO<span class="hljs-punctuation">,</span> .foo_<span class="hljs-punctuation">,</span> ._foo<span class="hljs-punctuation">,</span> Bar.42<span class="hljs-punctuation">,</span> foo..1<span class="hljs-punctuation">,</span> .<span class="hljs-punctuation">,</span> ._<span class="hljs-punctuation">,</span> ..<span class="hljs-punctuation">,</span> ...<span class="hljs-punctuation">,</span> ..1<span class="hljs-punctuation">,</span> <span class="hljs-built_in">c</span><span class="hljs-punctuation">,</span> <span class="hljs-built_in">T</span><span class="hljs-punctuation">,</span> <span class="hljs-built_in">F</span><span class="hljs-punctuation">,</span> ._1
```
joshgoebel, line 3: Long-term I do worry this makes the tests much harder to read, but I don't have a quick solution.

src/languages/r.js
joshgoebel: No no  I meant that just changing Vala wouldn't solve the problem if that's what you were saying but I don't think it was. :-) So we're good.

test/markup/r/names.expect.txt
klmr: Yeah, it hasn’t been great during debugging. :-(

Maybe I should just replace the commas by spaces?

joshgoebel (approved):
Looks good.  You happy?

test/markup/r/names.expect.txt
joshgoebel: Maybe.  That wouldn't help in actual code examples though... :-)

klmr:
Thanks for the patient review! There were a lot of details that required attention.

joshgoebel:
How does this kind of solution feel to the original look-behind problem?

klmr:
It’s not too bad but of course there’s still quite a bit of duplication here. I also keep racking my brain whether I’ve missed a situation in which numbers are preceded by something other than operators and whitespace (I don’t *think* so).

joshgoebel:
> It’s not too bad but of course there’s still quite a bit of duplication here. 

I'm more interested in overall complexity than boilerplate.  Feels like a pretty clean solution overall I think.  Not perfect, but not ugly.

> I also keep racking my brain whether...

Problem for another day. :-)

PR #3197, opened 2021-05-19
Update GML to GML 2.3+

gnysek:
Updating GML syntax Higlighting to be up to date with latest 2.3.2 GameMaker Studio 2 release (as of May 2021).

Based on file found in `ProgramData/GameMakerStudio2/Cache/runtimes/runtime-2.3.2.XXX/fnames`

### Changes
sync of function list based on "fnames" file, with same order as they are defined by GML 2.3+

### DIFF
--- a/src/languages/gml.js
+++ b/src/languages/gml.js
@@ -11,393 +11,348 @@ export default function(hljs) {
     keyword: 'begin end if then else while do for break continue with until ' +
       'repeat exit and or xor not return mod div switch case default var ' +
       'globalvar enum function constructor delete #macro #region #endregion',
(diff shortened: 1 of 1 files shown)


joshgoebel:
This old way of having tons of keywords mixed makes PRs impossible to review so this will need to be rebased after #3198 gets merged.  Lets move the discussion over to that thread though as I think there are probably some other things we should do first as well to bring this up-to-date with our modern scopes, etc.

joshgoebel:
Ping.

joshgoebel:
Can we rebase this on #3198 now that it's merged? Thanks!

PR #3198, opened 2021-05-19
gml cleanup

joshgoebel:
### Changes

- flattens keywords
- adds `self` to `variable.language`

### Checklist


joshgoebel:
@gnysek Typically we'd also order these lists by alpha now... is there any reason we shouldn't do that for GML as well?

joshgoebel:
Also can you help me understand what these lists are in the context of GML?

- built ins
- literals
- symbols

Does GML refer to these as "symbols" or was that just the CSS scope chosen to represent them in our highlighter?

Languages typically do NOT have so many literals... I wonder if these are mis-categorized... perhaps they are language variables or language constants?

gnysek:
- built ins are functions (so they all have round brackets after them)
- things marked as literals are generally constatns, but some of them might be read-only global variables, or their values might change between compilations
- things marked as symbols are global variables, or read-only global variables

joshgoebel:
So it's clear that these aren't literals... so what *actual* literals does GML support?  Things that have a discrete literal value like `yes`, `no`, `true`, `null`, `infinity` etc?

>  generally constatns, but some of them might be read-only global variables

What is the difference between constants and read-only global variables in GML?

joshgoebel:
https://github.com/highlightjs/highlight.js/pull/3198/commits/dc17919d1659bbe051553241137bafba3257ebdd

My attempt to clean up the classification:

- LITERALS is now much tighter
- Many literals (which appear to be largely but not exclusively enumerables) are now `symbol`
- `symbols` from before are not `variable.language`

At a glance this seems better, though I'm still not sure it's perfect since I don't know GML. Thoughts?

joshgoebel:
Does GML have user defined functions?

gnysek:
Yes GML allows to create user defined functions, their syntax is:

`function name(arguments) {....`
`name = function(arguments) {...`
`function name(arguments) constructor {....`
`function childname(arguments) : name(arguments) constructor {...`

Read-only variables are those which value changes during program execution, but you cannot write any value to them.

Constants in GM are mostly replaced on compile time to literals (like in compiled code, there's no "pi" anymore, but 3.14, true becomes 1 and false becomes 0 - there's no separate boolean type here).

joshgoebel:
> Yes GML allows to create user defined functions, their syntax is:

Are you interested in improving the GML grammar to support more language features, such as function titles, etc?

joshgoebel:
Does dc17919 look workable to you?

gnysek (approved):

gnysek (approved):

PR #3200, opened 2021-05-19
(chore) fix dual package hazard

aduh95:
To avoid dual package hazard in Node.js, it's important that the instance of hljs is shared across CJS and ESM modules.

Refs: https://github.com/highlightjs/highlight.js/pull/3188

### Changes

This PR is disabling use of ESM for the `core` module in Node.js, as ESM is able to import CJS modules but not the oother way around.

### DIFF
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -15,7 +15,7 @@ const safeImportName = (s) => {
 };
 
 async function buildESMIndex(name, languages) {
-  const header = `import hljs from './core.js';`;
+  const header = `import hljs from 'highlight.js/lib/core';`;
   const footer = "export default hljs;";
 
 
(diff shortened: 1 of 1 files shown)


tools/build_node.js
```diff
@@ -119,7 +119,10 @@ async function buildPackageJSON(options) {
     ".": dual("./lib/index.js"),
     "./package.json": "./package.json",
     "./lib/common": dual("./lib/common.js"),
-    "./lib/core": dual("./lib/core.js"),
+    "./lib/core": {
+      get node() { return this.require; },
```
joshgoebel, line 123: Whoa.  Does `JSON.stringify` automatically loop over getters or how does this work?

aduh95: Yep, it loops over all enumerable keys, and read its value by calling the getter in this case.

joshgoebel:
Closing, as mentioned elsewhere our `cdn` package is where "web-ready" build assets should live, not our primary NPM.  So this is not a relevant issue.

PR #3204, opened 2021-05-23
fix(R) over-eager matching of doc comment examples

klmr:
This PR fixes a highlighting bug where highlighting of examples in doc comments continued even after the comment stopped. E.g.:

```r
#' @examples
#' a
b
```

#### Highlighting before:

```xml
<span class="hljs-comment">#&#x27; <span class="hljs-doctag">@examples</span>
#&#x27; a</span>
b
```

#### Highlighting with this fix:

```xml
<span class="hljs-comment">#&#x27; <span class="hljs-doctag">@examples</span>
#&#x27; a
b</span>
```

### Checklist
- [x] Added markup tests


src/languages/r.js
```diff
@@ -85,18 +85,17 @@ export default function(hljs) {
...
-                  { begin: /\n/ },
                   {
-                    begin: /#'\s*(?=@[a-zA-Z]+)/,
-                    endsParent: true,
+                    match: /\n#'\s*(?=@[a-zA-Z]+)/,
+                    endsParent: true
                   },
                   {
-                    begin: /#'/,
+                    begin: /\n#'/,
```
joshgoebel, line 96: This looks weird, why not `^`?

klmr: The intent is to match the line break itself, not the start of the line. This is supposed to be a match across multiple lines, causing the parent mode to continue until the terminating condition is met (either explicitly via `endsParent: true`, or implicitly after a line break not followed by `#'`). Using `^` doesn’t do that.

However, I don’t seem to actually require `excludeBegin: true` — the tests stay green after removing it.

… There might be another, clever way of telling hl.js to continue the parent mode onto the next line but I haven’t been able to figure out how.

joshgoebel: Add a separate rule to match `\n` with a lookahead for `#`... less magical and you can commit what it's doing to make it clear.?

klmr: Ah, I think I got you. How’s 63760293270a6320e898ac2dedededa8fbbcada0?

joshgoebel:
This was still bugging me (want to review further) but I've been busy with other things, will make time to circle back to this after v11 is released.

joshgoebel:
@klmr Thoughts?

klmr:
Yep, looks good, much cleaner. I wasn’t aware that lookaheads could be nested.

joshgoebel (approved):

Issue #3199, opened 2021-05-19
Discuss: CDN builds should include ESM ready modules


joshgoebel:
> We already have a ton of infrastructure, history, documentation around making our web builds available via CDN (and secondarily) via NPM via `con-assets`... if we do have an ESM library that is intended for use on the web then it should be built with `cdn` and packaged along with the CJS version in our CDN distribution.  It would be very strange for it to not be included there.  That may require some additional work/thought... so if you want to have that discussion...

_Originally posted by @joshgoebel in https://github.com/highlightjs/highlight.js/pull/3188#issuecomment-844207270_

joshgoebel:
@aduh95 If you're interesting I think this is quite likely as simple and taking the ESM build code I removed from `node` build before and migrating it over to the `cdn` build...

jimmywarting:
I would like a very minimal cdn bundle...
I would also want an async version where non of the langauges is imported but where all the languages is supported/listed/mapped and only imported dynamically on demand.

```js
import hljs from 'highlight.js/lib/core';

if (hljs.supports(language)) {
  hljs.highlight(code, { language }).then(html => {
    xyz.innerHTML = html
  })
}

if (hljs.supports('js') || hljs.supports( 'javascript' )) {
  hljs.highlight(code, { language: 'js' }).then(html => {
    xyz.innerHTML = html
  })
}
```
Right now i think that it only supports like 32 with the default bundle. but you do support more langauges than that... i want to support every language that you support (lazily).

joshgoebel:
> I would like a very minimal cdn bundle...

This bundle already exists (CDN build): `es/core.js` 

> I would also want an async version 

This currently does not exist. This would require several complexities be resolved before it could exist in core:

- whole new async API for `highlight*` and perhaps `registerLanguage` / `getLanguage`
- nested grammar dependencies also need to "just work" (`http -> html -> (css/js)`), including circular dependencies `(html => js => html)`
- need a way to extract & register aliases at compile time rather than runtime (aliases are currently unknown until AFTER grammars are loaded)

I'm probably forgetting things, you'd have to skim several of these threads - I've definitely written on this subject before.  

Also, see #2405.  Right now I'm unconvinced that this isn't better handled by a wrapper outside of core such as say a new `highlights-async` library that simply wrapped the core library with async niceness. (you'd have to handle aliases and the dependency graph)

PR #3209, opened 2021-05-30
(feat) include ESM in CDN build

aduh95:
Currently, this PR doesn't include an ES module pre-compiled with common languages. ESM user need to load both `core` and each individual language module. Should I change this?

Fixes: https://github.com/highlightjs/highlight.js/issues/3199

### Checklist


tools/build_cdn.js
```diff
@@ -1,37 +1,93 @@
...
+const { buildPackageJSON, writePackageJSON } = require("./build_node");
+const { rollupCode } = require("./lib/bundling.js");
 const path = require("path");
 const bundling = require('./lib/bundling.js');
 
 async function installPackageJSON(options) {
-  await buildPackageJSON(options);
-  const json = require(`${process.env.BUILD_DIR}/package`);
+  const json = buildPackageJSON(options, {
+    ".": {
```
joshgoebel, line 19: Why would this matter for web usage?

joshgoebel:
> Currently, this PR doesn't include an ES module pre-compiled with common languages.

I believe we'd still want a common build - and I think it would be a **monolith** (ie, languages actually compiled in)... someone just wanting the "default it just works" setup  but also wanting ESM should still be able to have a single source to import to get started, etc...

tools/build_config.js
```diff
@@ -44,14 +44,10 @@ module.exports = {
...
-        plugins: [
-          cjsPlugin(),
-          jsonPlugin()
-        ]
+        plugins: []
       },
       output: {
-        format: "iife",
-        outro: "return module.exports.definer || module.exports;",
+        format: "es",
```
joshgoebel, line 50: These changes confuse me... we still want the IIFE build... I would expect to see a `browser_esm` or perhaps just `pure_esm` section added, not the browser section altered...

joshgoebel: Oh I should have read it all. hmmmm..

tools/build_cdn.js
joshgoebel: I see now that the npm `exports` has been drug along into the v11 CDN package unintentionally, but I'm not sure this is necessary at all.  There are two common uses cases for CDN:

- Purely a delivery system to get our files onto CDNs, where they are directly loaded/imported (most common)
- Via `npm` when someone really wants JUST the build assets to directly copy into their project

I'm not sure `exports` is necessary for either of these cases.

Thoughts?

aduh95: I don't think it's necessary indeed, but I thought it wouldn't hurt to keep it I think. It could be helpful for an hypothetical future tool that can generate an import map from a `package.json` `"exports"`, and humans can also read it to understand the structure of the files to understand which url they should import.

tools/build_config.js
aduh95: Yep it's a hack to avoid compiling each language module twice. I'm not too proud of it, but it works 😅 

I wasn't able to figure out what that `outro` option was doing, maybe it wasn't used after all?

tools/build_browser.js
```diff
@@ -14,7 +14,8 @@ const log = (...args) => console.log(...args);
 const { rollupCode } = require("./lib/bundling.js");
 const bundling = require('./lib/bundling.js');
 const Table = require('cli-table');
-const { result } = require('lodash');
+
+const textEncoder = new TextEncoder();
```
joshgoebel, line 18: Why is this necessary? Rollup is supposed to be generating ASCII.

tools/build_browser.js
```diff
@@ -254,14 +248,71 @@ async function buildBrowserHighlightJS(languages, { minify }) {
   return {
     core: coreSize,
     core_min: core_min,
-    minified: Buffer.byteLength(minifiedSrc, 'utf8'),
+    minified: minifiedSrc.length,
```
joshgoebel, line 252: Is there a reason this was changed?

joshgoebel:
Also currently failing tests.

tools/build_browser.js
aduh95: `minifiedSrc` is now a `UInt8Array` in this implementation, not a string. So they should be exactly the same, expect that it should take a bit less time to compute. I could use `minifiedSrc.byteLength` to make it clearer – or I could simply revert this bit if you prefer the previous way of doing this.

tools/build_browser.js
joshgoebel: I'm sure. See `ascii_only: true` in build config.

It's also possible that `length` would work now where-as it didn't work before... the whole reason was so that the byte counts match the size on disk (which at the time they did not)... but I don't think we need this added TextEncoder stuff...

Side note: You don't need to rebase constantly either, it'd be easier for me to see what changes as we work on something if you just lay new commits on top of old ones.

tools/build_browser.js
joshgoebel: Yeah we don't need the encode... but we may not need bytelength anymore either, see my previous comment...

tools/build_browser.js
aduh95: I'm +1 to remove the `TextEncoder` stuff, it's way easier to deal with strings. I'd like to note that `ascii_only: true` is a terser-only option, so the non-minified files may contain arbitrary UTF-8 chars. Not really an issue since the size that really matters is the gziped one and this one is reported appropriately.

joshgoebel:
Doesn't look like there is any way to test this with jsdom?

https://github.com/jsdom/jsdom/issues/2475

joshgoebel:
@oscarotero  Could you pull this, do a `cdn` build and see if Deno like the resulting build files in `build/es`?

aduh95:
@joshgoebel I've added tests for Deno to test the ESM build. I'm not suggesting we should merge it, but I thought it could be interesting as a POC. wdyt?

joshgoebel:
My question about testing was specifically related to a browser-like environment, not Deno per se.  Yes, it's obvious we could perhaps add a few Deno tests... but let's have that discussion elsewhere (perhaps in the new issue I just created).

Could you please open a second PR with those last few changes and link it to the new issue?

aduh95:
> Could you please open a second PR with those last few changes and link it to the new issue?

I actually need the changes from this PR to run the Deno test (because it needs plain ESM to work), should I remove it from this PR and open a new one when/if this one lands, or should I leave it in (after, it tests that the ESM output is somewhat valid so I suppose that's valuable).

joshgoebel:
> I actually need the changes from this PR to run the Deno test (because it needs plain ESM to work)

Yes, you'd just push your changes as-they-were to another branch.  I wasn't suggesting start form scratch on `main` or anything... then when this is merged that branch could get rebased - and at that point only your test changes would remain.

I don't think we need to do this any longer though... _as I don't think these are good baseline tests for Deno_ as they don't prove anything truly useful.  They might serve as regression tests for the API, but they don't provide any real confidence that Deno has the same behavior as say Node.js since they don't exhaustively test the engine (or really test it much at all).  If Deno is compatible in the least then we'd expect basic function dispatch to work.

>  it tests that the ESM output is somewhat valid so I suppose that's valuable).

I don't believe these Deno tests are a desirable way to "test ESM output in general" either... that would be better done by trying a sample bundle/build with ESM during our CI as we do already to test some bundler edge cases.  The **only** reason to add Deno specific tests at this juncture is if they provide value in pushing us towards perhaps officially supporting Deno in the future.

As I mentioned on the other thread if the desire is to move towards "Deno is officially supported" then we should start with porting `test/markup/index.js` to Deno... that is the most bang for the buck we could get and would provide a *high* confidence that  our engine has the same overall behavior running in Node, in the browser, or in Deno.

joshgoebel:
To be clear there's nothing wrong with the tests per se, I just don't feel they have any value *on their own* right now... if you wanted to go a step further and add the markup tests as well _then_ the API tests would definitely be great to have *in addition*... as a regression check for the API... but just not as our *only* Deno tests.

aduh95:
Thanks for the clarification, I'll open a draft PR with the Deno tests after this one lands.

joshgoebel:
This still seems broken when a language uses `module.exports`...

```
hljs.registerLanguage("cshtml-razor",(()=>{"use strict";module.exports=n=>{
var e="built_in",s={},a={begin:"}",className:e,endsParent:!0},i={begin:"{",
```

It also breaks the ESM build since it now includes `module.exports` which only works on node, not in the browser...

Previously this was dealt with since we compiled to IIFE first (which had the effect of normalizing this)... now that you are directly compiling to ES it seems to just leave module.exports rather than rewriting it to `export default`... is there perhaps some Rollup flag to require the exports to the simpler syntax?

aduh95:
@joshgoebel where do you see this `module.exports`? I'm unable to reproduce:

```console
$ yarn build-cdn
✨  Done in 8.57s.
$ grep -r 'module.exports' build/   
build//highlight.js:if (typeof exports === 'object' && typeof module !== 'undefined') { module.exports = hljs; }
build//highlight.min.js:;"object"==typeof exports&&"undefined"!=typeof module&&(module.exports=hljs);
```

joshgoebel:
Just add a 3rd party module that uses `module.exports` in `extra`, such as cshtml-razor in my example.

aduh95:
I didn't know about extra languages, I'm surprised this didn't come up in earlier discussions about switching to ESM. In a separate discussion, we might want to discuss deprecating the use of CJS for those maybe?

In any case, I've added support for detection of the `"type": "module"`, if they are written as CJS I'm adding back the Rollup CJS plugin which fixes the issue.

I've also added support for `"exports"` field, that might be a bit off-topic, tell if you prefer I open another PR for this – or if you are simply not interested in supporting it for now.

joshgoebel:
> I didn't know about extra languages, I'm surprised this didn't come up in earlier discussions about switching to ESM. 

Didn't think of it.  Of course 3rd party languages go straight into the build pipeline just like 1st party ones so it should have "just worked" had you not also changed the input processing in what I presume was an effort to improve it.

> In a separate discussion, we might want to discuss deprecating the use of CJS for those maybe?

I'm in no hurry to deprecate CJS - and still think it's the simplest way to get started for many of our users, but I've added an `.es.min.js` build to `dist` now as well.  If we can continue to support CJS by simply using one additional plugin with Rollup then I feel need to pressure our 3rd party grammar authors to do extra work.  If they want to switch to ESM, they can - we've always supported authoring using ESM for 3rd party modules.

> I've added support for detection of the "type": "module" ...  I've also added support for "exports" field

This is complexity we do not need at the present time and also I don't think guaranteed to be backwards compatible. `module` is not a great heuristic since some grammars have a package.json, but aren't even real NPM packages (or following the overall NPM spec).  Expecting grammars authors to get this 100% right is something we could decide to enforce in the future, but right now it could be a breaking change.

We also have to support support grammars with no `package.json` at all just by finding their files in the typical `src/languages/*.js` location. Those files could be CJS or ESM.  So the input pipeline needs to be agnostic.

I've reverted back to the much simpler IIFE build process we used before.  ESM modules are just IIFE + added export at the end.  Since variables should be file local I think this works, am I mistaken?

```
 const esm = `${data};\nexport default hljsGrammar;`;
```

joshgoebel:
> I've also added support for "exports" field, 

I'm also not sure becoming further tangled with NPM in this way is a step in the right direction.  It would be much simpler for us to drop `package.json` scanning altogether in favor of "one true way" :

- grammar files go in `src/languages`, CJS or ESM.

I don't find this an onerous restriction, and of course someone is always free to use their own build process.  BUT since what we do now (supporting `src/languages` or `package.json`) hasn't been a burden (yet), I left it in place with v11.  I'm not in any rush to add a lot more variants.

aduh95:
Good work on simplifying the IIFE build process, I think I had some kind of misconception about how the CJS plugin was working, I ended up doing much more work than necessary, oh well.

> I'm also not sure becoming further tangled with NPM in this way is a step in the right direction. It would be much simpler for us to drop `package.json` scanning altogether in favor of "one true way"

Fair enough, I'd argue that `package.json` is no longer an NPM-only thing – Node.js itself now uses it, as well as most modern bundlers – but I agree that there is no need to jump on board.

Let me know if there's something more for me to do in this PR.

tools/build_cdn.js
```diff
@@ -125,7 +125,7 @@ async function installLanguages(languages, options) {
 
   await Promise.all(
     languages.filter((l) => l.third_party)
-      .map((lang) => buildDistributable(lang, options))
+      .map(async(lang) => await buildDistributable(lang, options))
```
aduh95, line 128: Out of curiosity, what's the reason for using async/await here?

aduh95:
Sorry for the back and forth, but I think it's more efficient to ask rollup to output an ES module here:

Before:

```js
var hljsGrammar=(()=>{"use strict";return e=>{
/*...*/}})();export default hljsGrammar;
```

After:

```js
function ada(e){
/*...*/}export default ada;
```

It saves a few bytes (which is always nice for a CDN build), and produces the exact same output for the IIFE files.

tools/build_cdn.js
joshgoebel: Uncertainty about how async works? :-) Are you going to tell me it's effectively the same either way?

joshgoebel:
> but I think it's more efficient to ask rollup to output an ES module here:

Maybe but I've kind of hand verified things the old way (before your last commit) and getting kind of burnt out on this PR... so we'll go with it as it is and then we can circle back in the future.  

What you're suggesting though is the REAL problem before was your use/lack of use of plugins, not the actual switch from IIFE to ESM? That was breaking the default.exports?

joshgoebel (approved):

aduh95:
> What you're suggesting though is the REAL problem before was the use/lack of use of plugins when you switched to ESM, not the actual switch from IIFE to ESM that was breaking the default.exports?

Yes that's exactly it, I had removed the plugins because I didn't know about extra languages written in CJS.

> Maybe but I've kind of hand verified things the old way (before your last commit) and getting kind of burnt out on this PR... so we'll go with it as it is and then we can circle back in the future.

No problem, take care!

joshgoebel:
Thanks for all your work on the PR and providing momentum to help get this done. 👍 I plan to push a release this week I think.

PR #3213, opened 2021-05-30
Edn support

### LINKED ISSUE #3205: Add support for .edn / EDN syntax (identical to Clojure)
Hello! I'm a Clojure programmer, and it seems like highlight.js could easily support EDN syntax highlighting by adding "edn" to the Clojure aliases. EDN is to Clojure what JSON is to JS. The syntax highlighting is equivalent. Is this reasonable? I use Gitea and the EDN files do not get any syntax highlighting.

https://github.com/highlightjs/highlight.js/blob/8358762bbf3f92bd45919a7893071944fc8574fd/src/languages/clojure.js#L146

stelcodes:
I added an alias for clojure, "edn", so edn tagged markup will get the syntax highlighting as clojure. I also added a markup test using the [deps.edn](https://github.com/babashka/babashka/blob/master/deps.edn) file from [babashka](https://github.com/babashka/babashka). I definitely need some review and advice because this is my first contribution. Very new to the code base. Ran `npm run build_and_test` and all the tests passed. I couldn't get `npm run test` or `npm run test-markup` to run, however. Getting `Error: Cannot find module '../build'` errors.

Resolves #3205 
### Changes

Added "edn" to clojure aliases, added edn markup test
### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+## Version 11.1.0 (pending)
+
+Grammars:
+
+- enh(clojure) added `edn` alias (#3213) [Stel Abrego][]
+
+[Stel Abrego]: https://github.com/stelcodes
+
+
 ## Version 11.0.0
 
 **This is a major release.**  As such it contains breaking changes which may require action from users.  Please read [VERSION_11_UPGRADE.md](https://github.com/highlightjs/highlight.js/blob/main/VERSION_11_UPGRADE.md) for a detailed summary of all breaking changes.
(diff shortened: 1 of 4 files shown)


joshgoebel:
>  npm run test or npm run test-markup

Because you must build first before you can run tests.

>  I also added a markup test using the deps.edn file from babashka. 

Awesome, but is there a smaller file that still would show that everything works?  Large markup tests only make reviews slower then large changes to a grammar happen later.  If a 10-20 line EDN file would show that this works just as effectively then it would be much preferred.

Otherwise this looks good.  Still need the changelog entry of course.

stelcodes:
Hey @joshgoebel should I update the version 11 changelog? Something like this under `Grammars:`?
`- enh(processing) added `edn` alias (#3213) [Stel Abrego][]`

joshgoebel (approved):

PR #3217, opened 2021-06-01
(HIGHLIGHT) - Add support to highlight to "BACKFILL" keyword in SQL

rajmaddy89:
Add support to highlight to "BACKFILL" keyword in SQL
So basically if you see this issue, https://github.com/yugabyte/yugabyte-db/issues/7330, BACKFILL keyword is not highlighted in case of SQL even though other commands like CREATE, UPDATE are. Hence adding BACKFILL to the list of keywords

### Changes
Add "BACKFILL" to list of reserved keywords in SQL.js

### Checklist
- [x] Added markup tests, or they don't apply here because

### DIFF
--- a/src/languages/sql.js
+++ b/src/languages/sql.js
@@ -120,6 +120,7 @@ export default function(hljs) {
     "atomic",
     "authorization",
     "avg",
+    "backfill",
     "begin",
     "begin_frame",
     "begin_partition",


joshgoebel:
Where can some information be found on BACKFILL, Google returns no search results in combination with `sql`.

rajmaddy89:
BACKFILL is an internal command that runs when indexes are created.

When an index is created on an existing table, YugabyteDB will automatically backfill existing data into the index in an online manner. BACKFILL command is internal, not run by users.  It is still visible in queries view of YUGABYTE DB because it's helpful to know that your create index is busy backfilling. BACKFILL INDEX queries will only be visible during long running CREATE INDEX that uses index backfill in YUGABYTE.

joshgoebel:
> BACKFILL is an internal command that runs when indexes are created.

Our `SQL` grammar is for official SQL specification keywords only, not vendor specific extensions. The only other reason we might add this is is EVERY major database supported it as a keyword, but based on my Google results I'm guessing that is not the case.  If I'm mistaken, please let me know.  

I would suggest creating a 3rd party grammar for YUGABYTE DB that extends (or replaces) the SQL grammar - or perhaps writing a small plugin to just find `BACKFILL` and highlight it after all the Highlight.js highlighting pass is complete.  You could use `after:highlightElement` for such a thing.

rajmaddy89:
@joshgoebel : Can you help me understand or point me to an example of how to write an plugin within HIGHLIGHT.JS to find any keyword like BACKFILL and highlight it after all the Highlight.js highlighting pass is complete

joshgoebel:
```
text.replace("BACKFILL ", html_to_use)
```

Of course you may want to use a regex for a more precise match, etc...  Though it might break with strings, etc... so perhaps my first recommendation to make a custom grammar is better.  But the above is the simplest way to go about it... I'd suggest looking up `after:highlightElement` in our documentation.

rajmaddy89:
@joshgoebel : Thanks for the suggestion, should the plugin be written in this file ? https://github.com/highlightjs/highlight.js/blob/main/src/highlight.js, please confirm

joshgoebel:
No, you can put the plugin code anywhere (as long as the JS gets loaded). You just have to register it with `addPlugin`.

joshgoebel:
Closing PR as this is not mergeable as previously noted.

PR #3221, opened 2021-06-02
fix(python) identifiers starting with underscore not highlighted

anlambert:
Hello and thanks for maintaining that great library.

I noticed a regression in `highlightjs.js 11.0`, Python identifiers starting with an underscore
character (`__init__` for instance) are no longer highlighted (introduced in 952fa0a).

That PR fixes that issue and add new markup tests to cover it.

On a related note, Python 3.0 introduces additional characters from [outside the ASCII range](https://docs.python.org/3/reference/lexical_analysis.html#identifiers) for identifiers. Adding support for those in highlight.js is out of that PR scope and does not seem easy to implement but I could try to have a look when I have some time.

### DIFF
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -5,7 +5,7 @@ Website: https://www.python.org
 Category: common
 */
 
-import { IDENT_RE } from '../lib/modes.js';
+import { UNDERSCORE_IDENT_RE } from '../lib/modes.js';
 import * as regex from '../lib/regex.js';
 
 export default function(hljs) {
@@ -379,7 +379,7 @@ export default function(hljs) {
       {
         match: [
           /def/, /\s+/,
-          IDENT_RE
+          UNDERSCORE_IDENT_RE
         ],
         scope: {
           1: "keyword",
@@ -392,14 +392,14 @@ export default function(hljs) {
           {
             match: [
               /class/, /\s+/,
-              IDENT_RE, /\s*/,
-              /\(\s*/, IDENT_RE,/\s*\)/
+              UNDERSCORE_IDENT_RE, /\s*/,
+              /\(\s*/, UNDERSCORE_IDENT_RE,/\s*\)/
             ],
           },
           {
             match: [
               /class/, /\s+/,
-              IDENT_RE
+              UNDERSCORE_IDENT_RE
             ],
           }
         ],
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,11 +2,13 @@
 
 Grammars:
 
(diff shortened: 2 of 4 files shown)


src/languages/python.js
```diff
@@ -392,14 +392,14 @@ export default function(hljs) {
...
-              /\(\s*/, IDENT_RE,/\s*\)/
+              UNDERSCORE_IDENT_RE, /\s*/,
+              /\(\s*/, UNDERSCORE_IDENT_RE,/\s*\)/
             ],
           },
           {
             match: [
               /class/, /\s+/,
-              IDENT_RE
+              UNDERSCORE_IDENT_RE
```
joshgoebel, line 402: Is this just for completeness?  Are classes often named in such a fashion?  I thought convention was `CamelCase`?

joshgoebel:
> On a related note, Python 3.0 introduces additional characters from outside the ASCII range for identifiers. 

If it's along the lines of #2756 then it's not a high priority and needs some considerable effort and further research.

src/languages/python.js
anlambert: Class names in Python follow the global identifiers specification, so yes class names starting with an underscore are totally allowed by the language, see example [here](https://forge.softwareheritage.org/source/swh-web/browse/master/swh/web/tests/conftest.py$170).

anlambert:
> > On a related note, Python 3.0 introduces additional characters from outside the ASCII range for identifiers.
> 
> If it's along the lines of #2756 then it's not a high priority and needs some considerable effort and further research.

That's why I thought too, ECMA 2018 could help here with its [unicode category support in RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes) but it is only supported in recent  browsers. So yes, this needs considerable effort and testing on the subject.

joshgoebel:
>  only supported in recent browsers

We're mostly fine with supporting only more recent browsers.  Safari is a hold out on look-behind regex, but we're mostly ES2015/2018 already otherwise in several ways...

> unicode category support in RegExp 

If you skim the other issue you'll see the issue is `/u` seemed significantly slower in testing - and for a feature that no one is asking for (you make only the second person to mention it) it doesn't seem worth a slow down. So it's pretty much just on hold until someone who cares a lit pours some time into it.

joshgoebel (approved):

joshgoebel:
Please add changelog entry.

anlambert:
> Please add changelog entry.

Done. I also updated new markup test with the class starting with an underscore case.

joshgoebel (approved):

joshgoebel:
Thanks so much!

PR #3225, opened 2021-06-03
fix(parser) Resolve issue with missing TS property

jswanner:
Using `highlightElement` with version 10.7.2 causes the following TypeScript build error:

```
error TS2339: Property 'highlightElement' does not exist on type 'HLJSApi'.
```

### Changes
fix(parser) Resolves issue with missing TypeScript property

### Checklist
- [x] Added markup tests, or they don't apply here because only modified types definitions

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## Version 10.7.3 (pending)
+
+- fix(parser) Resolves issue with missing TypeScript property [Jacob Swanner][]
+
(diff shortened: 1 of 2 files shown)


joshgoebel (approved):

joshgoebel:
Fixed with release of 10.7.3.

jswanner:
Awesome, thanks!

PR #3233, opened 2021-06-07
fix(markdown) Images with empty alt or links with empty text

### LINKED ISSUE #3232: (markdown) Images with empty alt or links with empty text do not have highlight
**Describe the issue**
Images with empty alt or links with empty text do not have highlight

markdown

`highlight`

**Sample Code to Reproduce**
```markdown
![](https://example.com)

![alt](https://example.com)

[](https://example.com)

[text](https://example.com)
```

gives out HTML result

```html
![](https://example.com)

![<span class="hljs-string">alt</span>](<span class="hljs-link">https://example.com</span>)

[](https://example.com)

[<span class="hljs-string">text</span>](<span class="hljs-link">https://example.com</span>)
```

**Expected behavior**
```html
![](<span class="hljs-link">https://example.com</span>)

![<span class="hljs-string">alt</span>](<span class="hljs-link">https://example.com</span>)

[](<span class="hljs-link">https://example.com</span>)

[<span class="hljs-string">text</span>](<span class="hljs-link">https://example.com</span>)
```

**Additional context**
https://github.com/highlightjs/highlight.js/blob/9350caf515c0983fcdb44866c94a71b9ab9df425/src/languages/markdown.js#L88-L110

Looks like it will only match square brackets with something in between.

joshgoebel:
Resolves #3232.

### Changes

- [x] Properly highlights links with empty alt or text descriptions

### Checklist


allejo (commented):
Fixes the issue at hand :+1: Do we want to take into consideration linkable images, which is currently not accurate?

image

joshgoebel:
>  Do we want to take into consideration linkable images, which is currently not accurate?

We could open a new issue for that if you'd like.  I have no idea what the nesting rules even are or how common that is.

Issue #3237, opened 2021-06-11
enh(css-like) custom attributes should highlight as `attr`, including the `--`


thanoskrg:
Hello and thanks for this amazing library! I am using it for my blog to highlight code snippets, and recently I used CSS highlighting on a code example which included CSS variables.

More specifically I have this case:

```css
/* default context vars */
body {
  --color: black;
  --background-color: lightgrey;
}

/* dark theme context vars */
body.dark {
  --color: lightgrey;
  --background-color: black;
}

/* selector with styles using vars */
main {
  color: var(--color);
  background-color: var(--background-color);
  transition: background-color 0.2s ease-in-out;
}
```

The output looks like this:

![Screenshot 2021-06-11 at 10 05 44 AM](https://user-images.githubusercontent.com/22836902/121645284-b082c080-ca9c-11eb-999a-cb353af58f47.png)

Normally, I would expect that the prefixing `--` of the variables would have been handled as part of the attribute.

I am thinking of forking this repo to introduce a PR addressing this issue. What do you think?

joshgoebel:
Any patch here would need to keep in mind https://github.com/highlightjs/highlight.js/issues/2269 and provide patches for ALL 4 grammars potentially or explain clearly why this does not apply to some (I'm not sure if it does or not).  A patch that only fixes just CSS would not be accepted.

I'm not sure what to do here is so clear cut though.  This reminds me a little of our refusal to highlight vendor prefixes, which I have argued is actually a feature, not a bug.  There are a few different things to consider here:

- Should we highlight `--` as part of the name?
- Should we use `variable` or `attr` or `attribute`?  I think technically these would be `variable.definition` or some such.
- Inside of `var(--blah-blah)` should we then use `variable`? (if we decide not to at the top-level)

While Github's choices are interesting I don't want to fall into the trap that it's somehow "canonical" or that we always must do the same thing just because they (or anyone else) does.

@allejo Thoughts?

thanoskrg:
Hey @joshgoebel, each one of these grammars, have their own syntax for defining variables. 

My example is explicit for the `--blah-blah` syntax under the context of CSS highlighting. Regarding your bullet points:

> Should we highlight -- as part of the name?

I would say yes.

> Inside of var(--blah-blah) should we then use variable? 

I don't think so.

Do you think it would make sense to open a PR as a proposal? I would really like to contribute to `highlight.js` library!

joshgoebel:
[...]

PR #3239, opened 2021-06-12
feat(css): handle css variables syntax

thanoskrg:
This PR addresses the https://github.com/highlightjs/highlight.js/issues/3237 so that CSS variables syntax is properly highlighted.

### Changes
* Update `src/languages/css.js` definition
* Update test scenarios for CSS markup

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -72,6 +72,7 @@ export default function(hljs) {
       //   end: /\)/,
       //   contains: [ hljs.CSS_NUMBER_MODE ]
       // },
+      modes.CSS_VARIABLE,
       {
         className: 'attribute',
         begin: '\\b(' + css.ATTRIBUTES.join('|') + ')\\b'
--- a/src/languages/lib/css-shared.js
+++ b/src/languages/lib/css-shared.js
@@ -30,6 +30,10 @@ export const MODES = (hljs) => {
         '|dpi|dpcm|dppx' +
         ')?',
       relevance: 0
+    },
+    CSS_VARIABLE: {
+      className: "attr",
+      begin: /--[A-Za-z][A-Za-z0-9_-]*/
     }
   };
 };
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Grammars:
 
 - fix(csharp) add missing `catch` keyword (#3251) [Konrad Rudolph][]
 - add additional keywords to csp.js (#3244) [Elijah Conners][]
+- feat(css) handle css variables syntax (#3239) [Thanos Karagiannis][]
 - fix(markdown) Images with empty alt or links with empty text (#3233) [Josh Goebel][]
 - enh(powershell) added `pwsh` alias (#3236) [tebeco][]
 - fix(r) fix bug highlighting examples in doc comments [Konrad Rudolph][]
@@ -24,6 +25,7 @@ Grammars:
 [tebeco]: https://github.com/tebeco
 [Pankaj Patil]: https://github.com/patil2099
 [Benedikt Wilde]: https://github.com/schtandard
+[Thanos Karagiannis]: https://github.com/thanoskrg
 
 
 ## Version 11.0.0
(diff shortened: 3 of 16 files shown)


joshgoebel:
A visual snap is alwyas nice with PRs like this.

src/languages/css.js
```diff
@@ -25,6 +25,10 @@ export default function(hljs) {
     hljs.APOS_STRING_MODE,
     hljs.QUOTE_STRING_MODE
   ];
+  const CSS_VARIABLE = {
+    className: "variable",
+    begin: /--[A-Za-z_-](?:(?!\s)[A-Za-z0-9_-])*/,
```
joshgoebel, line 30: This looks a bit overly complex a regex... what are we trying to accomplish with the `(?!\s)`? Must variables be two letters at least?

thanoskrg:
> A visual snap is alwyas nice with PRs like this.

@joshgoebel here it is how it looks like on *base16/nord* theme.

demo-css-variables

src/languages/css.js
thanoskrg: > Must variables be two letters at least?

No, there is no such constraint.

I think you are right about the complexity of the regex. Maybe we could start with a simpler regex for now. 

What do you think of this?
```suggestion
    begin: /--[A-Za-z][A-Za-z0-9_-]*/,
```

joshgoebel: That looks much more like what I would have expected to see.

thanoskrg:
That's how it looks now after the latest changes:

demo-css-variables-2

src/languages/lib/css-shared.js
```diff
@@ -30,6 +30,10 @@ export const MODES = (hljs) => {
         '|dpi|dpcm|dppx' +
         ')?',
       relevance: 0
+    },
+    CSS_VARIABLE: {
+      className: "attr",
+      begin: /--[A-Za-z][A-Za-z0-9_-]*/
```
thanoskrg, line 36: [note] @joshgoebel I thought it would make sense to place this under `lib/css-shared.js` for DRY.

joshgoebel:
Theme maintainers: Here is a case where we're increasing the semantic density of our highlighting, but at the same time perhaps making existing theme behavior worse.  I have a suspicion many/some themes only think about `attribute` when it comes to CSS... and haven't given any thought to `attr` at all (in the context of CSS).  Given how some themes almost contain "micro-themes" for different types of markup (Markdown vs diffs vs Code).  I'm not sure of the perfect solution.

Just sticking with `attribute` here for theme behavior has the best short-term behavior but I'm not sure is a great long-term plan.  Perhaps the real problem is our historic distinction between `attr` and `attribute`... already having a clear semantic scope to reach for precludes something more complex like `attribute.custom`... thought perhaps some would argue that's better than breaking themes in the short-term?  Though it also (in the short or mid-term) makes the distinction meaningless until some future themes comes along to express the styles differently.

I could also see many themes wanting to just merge these two styles (in the case of CSS), yes I hesitate to do that at the grammar level since I always try to be as accurate as possible when it comes to scopes - and then let themes decide HOW to highlight those scope.


So perhaps the semantic difference here between attributes and custom attributes (specifically w.r.t. CSS) just doesn't matter? Thoughts?

CC @highlightjs/theme-maintainers  @highlightjs/core

thanoskrg:
@joshgoebel any updates on this? Is this currently blocked until we get an answer on your last [question](https://github.com/highlightjs/highlight.js/pull/3239#issuecomment-860904286)?

joshgoebel:
I was hoping for some thoughts, yes.  Eventually if there is nothing this'll probably merge as-is...  I don't usually get in a  rush until I have a new release in sight.  I need to go figure out when 11.1 will be.

thanoskrg:
> I was hoping for some thoughts, yes. Eventually if there is nothing this'll probably merge as-is... I don't usually get in a rush until I have a new release in sight. I need to go figure out when 11.1 will be.

Thanks for your swift reply.

joshgoebel (approved):

PR #3246, opened 2021-06-16
Update SUPPORTED_LANGUAGES.md

wolfiex:
### Checklist

### DIFF
--- a/SUPPORTED_LANGUAGES.md
+++ b/SUPPORTED_LANGUAGES.md
@@ -4,6 +4,16 @@ The table below shows the full list of supported languages (and corresponding cl
 
 Languages that listed a **Package** below are 3rd party languages and are not bundled with the core library.  You can find their repositories by following the links.
 
+### Using 3rd Party Languages: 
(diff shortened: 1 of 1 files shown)


SUPPORTED_LANGUAGES.md
```diff
@@ -4,6 +4,16 @@ The table below shows the full list of supported languages (and corresponding cl
 
 Languages that listed a **Package** below are 3rd party languages and are not bundled with the core library.  You can find their repositories by following the links.
 
+### Using 3rd Party Languages:
```
joshgoebel, line 7: If anything this should link to the appropriate section of our README (where we already show how to do this). Also, just FYI, most 3rd party plugins include these instructions already in their own READMEs.

joshgoebel (changes requested):
Should refer to more complete instructions elsewhere, not add them inline.

joshgoebel:
I clarified this a bit in: 50a680877b2777ce2fe30f083037652f9decb5c9

PR #3255, opened 2021-06-22
JS - change CLASS_REFERENCE regex to also accept numbers and underscore

### LINKED ISSUE #3252: (Javascript) Float32Array broken after 11.0.1
**Describe the issue**

`Float32Array` is being broken in 3 token `Float` `32` and `Array`. This used to work fine on 10.7.3. Here is a [side-by-side result jsfiddle](http://jsfiddle.net/xbzwndrp/5/).

javascript

`highlightElement`

**Sample Code to Reproduce**

```
new Float32Array();
```

also repros with more complicated lines, like `const x = new Float32Array(3);`.

**Expected behavior**

What used to happen on 10.7.3. `Float32Array` highlighted as a builtin.

**Additional context**

rbrishabh:
fixes #3252 

Hey @joshgoebel! How does this look?
Let me know if this needs any changes! Thanks


joshgoebel:
Adding a markup test for this case to`test/markup/javascript/class` would be a good idea.

rbrishabh:
What do you think? @joshgoebel
Thank you for your quick reviews!

fserb:
I appreciate the quick fix, but is this correct? `/\b[A-Z][a-z0-9_]+([A-Z][a-z0-9_]+)*/`

There are a few problems that I can see here:
- there's this assumption that all classes must start with an uppercase, so `new myClass()` won't work. Is this on purpose?
- it's also assuming that it must be followed by a lowercase. So `new CSSThing()` won't work.
- it's also missing `$`, which is a valid character for JS class name (think JQuery's `$`), but really: `new my$c`  is valid JS.
- it's also missing `$` and `_` as valid initial characters. `new _X()` or `new $X()`. For the `$` you probably need to remove the `\b`.
- Unless I'm missing something about grouping, the first `+` is wrong and the `(....+)*` could be replaced by `*`.

I think those points above could be addressed with something like `/[a-zA-Z\$_][a-zA-Z\$_0-9]*/`

It's also missing unicodes, but this may be a bit beyond the scope of this change. This [stack overflow](https://stackoverflow.com/a/9337047/3702) answer has a complete solution, but not what highlight.js needs, because builtins seem to be cleared up before).

joshgoebel:
> there's this assumption that all classes must start with an uppercase, so new myClass() won't work. Is this on purpose?

This is a heuristic rule, not an absolute one.  Uppercase classes is a hugely common convention, but a lowercase symbol could be anything. That said if someone wanted to add a mode specifically for `new blah`, that would be ok.  But this rule cannot be expanded because we'd have too many false positively.

> it's also assuming that it must be followed by a lowercase. So new CSSThing() won't work.

This should indeed be taken into account.  So we need a few more markup tests.

> it's also missing $, which is a valid character for JS class name (think JQuery's $), but really: new my$c is valid JS.

This seems an edge case and same problem as first point, false positives.

> it's also missing $ and _ as valid initial characters. new _X() or new $X(). For the $ you probably need to remove the \b.

Ditto.

Unless I'm missing something about grouping, the first + is wrong and the (....+)* could be replaced by *.

Good point, but the whole thing may need to be slightly rethought now to handle cases like `CSSHandler`... so

joshgoebel:
@rbrishabh You need to actually run the test suite, and update the test expectations also, etc... see our docs for building and testing.  Now we also need to add test for class names like `ASTParser`... Perhaps something like `[A-Z][A-Za-z0-9_]+` ?

rbrishabh:
Sure! I can look into this later today!

rbrishabh:
hey @joshgoebel! How does this look now?

src/languages/javascript.js
```diff
@@ -252,7 +252,7 @@ export default function(hljs) {
 
   const CLASS_REFERENCE = {
     relevance: 0,
-    match: /\b[A-Z][a-z]+([A-Z][a-z]+)*/,
+    match: /\b[A-Z][A-Za-z0-9_]+)*/,
```
fserb, line 255: does this even work? I think you have a mismatched parenthesis.
you can remove the `+)`.

rbrishabh: woops! My bad! thanks for the quick review! how does it look now?

src/languages/javascript.js
```diff
@@ -252,7 +252,7 @@ export default function(hljs) {
 
   const CLASS_REFERENCE = {
     relevance: 0,
-    match: /\b[A-Z][a-z]+([A-Z][a-z]+)*/,
+    match: /\b[A-Z][A-Za-z0-9_]+/,
```
fserb, line 255: This is okey, although technically it should be  `*` and not `+` (otherwise `new X()` won't work). But I'm not sure if this breaks some other logic on HighlightJS.

joshgoebel: I think one letter class names is an edge case.  We really want to see the camel case pattern to trigger this heuristic.

joshgoebel:
I'm still looking for an updated `test/markup/javascript/class.expect.txt ` but don't see it... if you want to let HLJS update the tests for you (and confirm them by inspection) there is a line you can uncomment in `test/markup/index` to do this. (just uncomment it and run the markup tests once to write out the new expectations)

src/languages/javascript.js
joshgoebel: Let's focus more on the tests... at the very end I might take one last look and tweak this a variety of diff ways.

joshgoebel: I think we need BOTH... I played with this briefly.

`CRV` with zero context could be a constant (all CAPS is a constant) or a class name... Changing this rule as it is now incorrectly identifies constants as classes (or vice versa)... I think perhaps we need to handle both:

- `CRV` (constant)
- `CRV(` (class)

So we'd use the camel case rule (with zero other context) PLUS the newer new rule proposed here, but only if a look-ahead sees a `(`.  I think that would cover a lot of cases.  Thoughts?

joshgoebel:
Ping.

rbrishabh:
hi! I have a busy schedule and cant keep up with this now! Please feel free to continue it from here. I can close it for now

joshgoebel:
Thanks anyways!  It'll be here if anyone needs to reference it when there is time to pick this back up.

PR #3264, opened 2021-07-05
Fix registered language name for grammar names with dash, fixes #3263.

### LINKED ISSUE #3263: Breaking language name change for languages with dash (python-repl is now python_repl)
Languages with a dash such as `julia-repl`, `python-repl` are now registered as `julia_repl`, `python_repl`.

Without digging too much into this I believe https://github.com/highlightjs/highlight.js/commit/82e02ee4d39f6aad26d7c7902e214398869e1496 caused it (cc @joshgoebel )

fredrikekre:
Grammars with `-` in the name (`julia-repl`) now correctly register as that instead of `julia_repl`. The replacement here should reverse the modification in https://github.com/highlightjs/highlight.js/blob/55f68f72be2cb69d14561353ad851d6b2242dfcc/tools/build_browser.js#L190

fixes #3263

### DIFF
--- a/src/stub.js
+++ b/src/stub.js
@@ -5,6 +5,10 @@ import * as builtIns from "builtInLanguages";
 const hljs = HighlightJS;
 
 for (const key of Object.keys(builtIns)) {
+  // our builtInLanguages Rollup plugin has to use `_` to allow identifiers to be 
+  // compatible with `export` naming conventions, so we need to convert the 
+  // identifiers back into the more typical dash style that we use for language 
+  // naming via the API
   const languageName = key.replace("grmr_", "").replace("_", "-");
   hljs.registerLanguage(languageName, builtIns[key]);
 }
--- a/src/stub.js
+++ b/src/stub.js
@@ -5,9 +5,9 @@ import * as builtIns from "builtInLanguages";
 const hljs = HighlightJS;
 
 for (const key of Object.keys(builtIns)) {
-  // our builtInLanguages Rollup plugin has to use `_` to allow identifiers to be 
-  // compatible with `export` naming conventions, so we need to convert the 
-  // identifiers back into the more typical dash style that we use for language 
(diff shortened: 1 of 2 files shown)


test/builds/browser_build_as_commonjs.js
```diff
@@ -21,4 +21,13 @@ API.forEach(n => {
   assert(_ => keys.includes(n), `API should include ${n}`);
 });
 
+// See e.g. highlightjs/highlight.js#3263
+const langs = ["python", "python-repl"]
```
fredrikekre, line 25: Added tests here since it is specific to the brower build.

joshgoebel:
> The replacement here should reverse the modification in

Do you know of any way to just quote the export names so we could avoid this encode/decode dance?  I couldn't find anything looking at the export documentation.

test/builds/browser_build_as_commonjs.js
joshgoebel: Yes, this is a helpful addition for sure and regression test. :)

fredrikekre:
I have no idea, sorry.

src/stub.js
```diff
@@ -5,7 +5,7 @@ import * as builtIns from "builtInLanguages";
 const hljs = HighlightJS;
 
 for (const key of Object.keys(builtIns)) {
-  const languageName = key.replace("grmr_", "");
+  const languageName = key.replace("grmr_", "").replace("_", "-");
```
joshgoebel, line 8:
```suggestion
  // our builtInLanguages Rollup plugin has to use `_` to allow the identifiers to be compatible
  // with `export` naming conventions, so we need to convert the identifiers back into the
  // more typical dash style that we use for language naming via the API
  const languageName = key.replace("grmr_", "").replace("_", "-");
```

joshgoebel (approved):

joshgoebel:
@fredrikekre Thanks for much for spotting this and the PR so now 11.1 won't be broken!!!

Issue #3273, opened 2021-07-14
Declared types don't allow features of existing language definitions


jsharkey13:
**Describe the issue/behavior that seems buggy**

I am attempting to declare and register a new language definition for an internal pseudolanguage (aimed at students, and very simplistic), but doing so in TypeScript and not plain JS. I am using `11.1.0`.
I declare the new rules in a function with the type `LanguageFn`, and am attempting to add highlighting to function definition [using an example copied from `languages/javascript.js`](https://github.com/highlightjs/highlight.js/blob/11.1.0/src/languages/javascript.js#L273-L298). There are examples which use similar config in [Java](https://github.com/highlightjs/highlight.js/blob/11.1.0/src/languages/java.js#L165-L175) and in [C++](https://github.com/highlightjs/highlight.js/blob/11.1.0/src/languages/cpp.js#L573-L584) so I assumed this was a well-supported feature.

However, this won't compile because the `ModeDetails` interface expects the `match` property to be `RegExp | string` and does not allow an array of these. [The documentation for `match`](https://highlightjs.readthedocs.io/en/latest/mode-reference.html?highlight=ident_re#match) does say "type: regexp or array of regexp".

If I `// @ts-ignore` the whole function, then it _does_ work and correctly highlight each section with the right class applied, so this is just an issue with the declared types.

**Sample Code or Instructions to Reproduce**

A minimal TypeScript example would be (slimmed down to the absolute minumum from the real JavaScript language definition referenced above):

```typescript
import {LanguageFn} from "highlight.js";

const myLanguageRules: LanguageFn = function(_hljs) {
  return {
    name: 'myLanguage',
    contains: [
      {
        match: [
          /function/,
          /\s+/,
          /[a-z][A-Z]/,    // avoid using hljs.IDENT_RE for this example
          /(?=\s*\()/
        ],
        scope: {           // swap deprecated "className" to "scope" to avoid another type error
          1: "keyword",
          3: "title.function"
        },
      }
    ]
  }
};
```
This has the type error described above for `match` (and would have a type error on `className` if I had not swapped it to `scope` for version 11 compatibility).

**Expected behavior**

Since the example comes basically directly from one of the provided language definitions, I expected it to be allowed by the type definitions.

I think changing the ModeDetails interface to allow this case (if it is indeed valid!) is not too difficult and strictly an expansion of what is already allowed:
``` typescript
interface ModeDetails {
    // ...
    match?: RegExp | string | (RegExp | string)[]
[...]

PR #3274, opened 2021-07-14
fix(types) Fix some type definition issues

### LINKED ISSUE #3272: Type errors declaring a new language definition in TypeScript

I am attempting to declare and register a new language definition for an internal pseudolanguage (aimed at students, and very simplistic), but doing so in TypeScript and not plain JS. I am using `11.1.0`.
I declare the new rules in a function I want to pass into `hljs.registerLanguage(...)` [using the `LanguageFn` type you export from `index.d.ts`](https://github.com/highlightjs/highlight.js/blob/11.1.0/types/index.d.ts#L77). However `LanguageFn` annotates the `hljs` argument as optional, which means that I cannot use something like `hljs.QUOTE_STRING_MODE` in my function without a guard against `hljs` being undefined as the type annotation says it might be. 

A minimal TypeScript example would be:

```typescript
import {LanguageFn} from "highlight.js";

const myLanguageRules: LanguageFn = function(hljs) {
     return {
        name: 'myLanguage',
        contains: [
            hljs.QUOTE_STRING_MODE,
            hljs.NUMBER_MODE,
        ]
    };
}
```
where the references to `hljs.QUOTE_STRING_MODE` and number mode will cause type errors since `hljs` has the type `HLJSApi | undefined` (the error on trying to use `hljs` is `TS2532: Object is possibly 'undefined'.`). This is such a simple case that you can see it will cause a problem without even needing to use the compiler.

The only way to do this guard in a way that allows me to use `hljs` in many places in my definition seems to be the incredibly hacky `if (!hljs) return {} as Language;` at the start, otherwise the type system complains about the return type of the function being incorrect. However I don't understand why `hljs` is marked as optional in the definition `export type LanguageFn = (hljs?: HLJSApi) => Language` in the first place? Surely it is always passed in by the library?

**Expected behavior**

I would expect the `LanguageFn` type to actually be `export type LanguageFn = (hljs: HLJSApi) => Language` without the argument being optional, and then if I choose not to use the `hljs` argument I can just name it `_hljs` in my own function and TypeScript knows this means I do not intend to use the argument.

I could foresee changing this signature causing issues for anyone using TypeScript to define languages of their own if (and probably _only_ if) they do not have a  `hljs` argument at all in their function. Otherwise it seems safe to change. Was there a reason it is declared the way it is? Is it autogenerated, or written by hand?

joshgoebel:
Should resolve #3272 and #3273.

### Changes

- fix(types) begin, match, end all now take arrays
- fix(types) _emit now a scope attribute, not mode attribute
- fix(types) do not forget about endScope
- remove dead code
- fix(types) make hljs not optional

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+## Version 11.2.0 (pending)
+
+Parser:
+
+- fix(types) Fix some type definition issues (#3274) [Josh Goebel][]
+
+[Josh Goebel]: https://github.com/joshgoebel
+
 ## Version 11.1.0
 
 Grammars:
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -428,7 +428,7 @@ const HLJS = function(hljs) {
         }
       }
       do {
-        if (top.scope && !top.isMultiClass) {
+        if (top.scope) {
           emitter.closeNode();
         }
         if (!top.skip && !top.subLanguage) {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -212,9 +212,11 @@ declare module 'highlight.js' {
         begin?: RegExp | string | (RegExp | string)[]
         match?: RegExp | string | (RegExp | string)[]
         end?: RegExp | string | (RegExp | string)[]
+        // deprecated in favor of `scope`
         className?: string
         scope?: string | Record<number, string>
         beginScope?: string | Record<number, string>
+        endScope?: string | Record<number, string>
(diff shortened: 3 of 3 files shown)


joshgoebel:
@jsharkey13

jsharkey13 (approved):
I can't easily build and test this, but it definitely contains the changes I expected to fix the two issues. Thanks!

joshgoebel:
@jsharkey13 Thanks!

PR #3281, opened 2021-07-20
Trusted Types compatibility

vrana:
https://web.dev/trusted-types/ disallows string assignments to `innerHTML`. This change makes the code compatible by converting the string to `TrustedHTML` if the browser supports Trusted Types.

### Checklist
- [X] Added markup tests, or they don't apply here because the change is not markup related.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ## Version 11.1.1 (pending)
 
-- Trusted Types compatibility [Jakub Vrána][]
+- Trusted Types compatibility (#3281) [Jakub Vrána][]
 
 [Jakub Vrána]: https://github.com/vrana
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## Version 11.1.1 (pending)
+
+- Trusted Types compatibility [Jakub Vrána][]
+
+[Jakub Vrána]: https://github.com/vrana
+
 ## Version 11.1.0
 
 Grammars:
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -709,7 +709,7 @@ const HLJS = function(hljs) {
   }
 
   let ttPolicy = {createHTML: s => s};
-  if (typeof trustedTypes != 'undefined') {
+  if (typeof trustedTypes != 'undefined' && trustedTypes.createPolicy) {
     ttPolicy = trustedTypes.createPolicy('highlight.js', ttPolicy);
   }
 
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -708,6 +708,11 @@ const HLJS = function(hljs) {
     element.classList.add(`language-${language}`);
   }
 
+  let ttPolicy = {createHTML: s => s};
+  if (typeof trustedTypes != 'undefined') {
+    ttPolicy = trustedTypes.createPolicy('highlight.js', ttPolicy);
+  }
+
   /**
    * Applies highlighting to a DOM node containing code.
    *
@@ -734,7 +739,7 @@ const HLJS = function(hljs) {
     const text = node.textContent;
     const result = language ? highlight(text, { language, ignoreIllegals: true }) : highlightAuto(text);
 
-    element.innerHTML = result.value;
+    element.innerHTML = ttPolicy.createHTML(result.value);
     updateClassName(element, language, result.language);
     element.result = {
       language: result.language,


joshgoebel:
> (typeof trustedTypes != 'undefined')

Is this truly the best/only way to check for compatibility?

---

Ref: https://w3c.github.io/webappsec-trusted-types/dist/spec/

This seems very, very new.  Is this currently only a Chrome thing? I'm not sure we should add features that only a single browser supports.   I don't think a Chrome only world is a good world and I'm hesitant to do things that might encourage [or hasten] that... are there indications that other browsers engines also plan to adopt this?

CC @highlightjs/core

joshgoebel:
From Draft spec:

> Note: Trusted Type objects wrap values that are explicitly trusted by the author. As such, creating a Trusted Type object instance becomes a de facto injection sink, and hence code that creates a Trusted Type instances is security-critical.

Also if we're considering  ratcheting up the security here it's worth discussing how this relates to plugins.  We're very safe with our own HTML generation, but a plugin can VERY easily replace `result.value` with any value, such that we could no longer make any guarantees about... so what is the correct behavior in that case?

I assume the plugin would itself need to certify it's HTML with it's own Trust Type policy... in which case I think this is the wrong layer to add this - it would truly belong in perhaps `html_renderer.js`, etc... and this PR would then also require documentation updates explaining this, etc...

allejo:
image

No Firefox or Safari support is a hard pass for me. Once there's greater browser adoption, then I'd be open to this. I don't see the benefit of hardening XSS protection on a single browser and leaving the rest to need bulky polyfills.

> Is this truly the best/only way to check for compatibility?

Looks like [that's whey offer as a tiny polyfill](https://github.com/w3c/webappsec-trusted-types#tinyfill) but they do have [a complete (but heavier) polyfill](https://github.com/w3c/webappsec-trusted-types#browsers).

vrana:
While Chromium based browsers (Chrome, Edge, Opera) are the only ones currently supporting Trusted Types, it's still a valid defense against DOM XSS in these browsers. My code is written in a way that it doesn't require a specific browser. If the browser doesn't support Trusted Types then the code will work the same as before - a string will be simply assigned to `innerHTML`. There's no need for using a polyfill here.

You are right that this code doesn't improve the security of Highlight.js. The main purpose is to unblock projects using Highlight.js wishing to adopt Trusted Types. Until all dependencies are compatible with TT, a project can't enforce them. It means that a future code in the project might introduce a vulnerability because TT couldn't be enforced.

You are also right that adding the `createHTML` call to `HTMLRenderer` would be better because the code there is very simple and almost provably safe (except not escaping `className` in `span()`). I've considered that option but decided against that because calling it here is less intrusive. If someone is expecting that `HTMLRenderer.value()` would return a string then returning TrustedHTML from it would break that code. TrustedHTML stringifies well but it doesn't support calling string methods like `trim()`. However, if you are willing to take this more complicated path then I'd be happy to move the code there.

For projects depending on Highlight.js wishing to adopt TT (which currently happens in multiple projects at least in Google) there are these options:

1. Do not adopt TT. (doesn't improve security posture of the project while accepting that there still might be a risk in Highlight.js or its plugins)
2. Patch Highlight.js. (complicating updates)
3. Remove the dependency on Highlight.js.

Accepting this PR seems like the best option - it adds only a tiny amount of code, it's backwards compatible and it works in all browsers.

> Is this truly the best/only way to check for compatibility?

Do you have something specific in mind? I can e.g. add `&& trustedTypes.createPolicy`.

joshgoebel:
> If the browser doesn't support Trusted Types then the code will work the same as before 

Unless the spec changes, which is quite possible (as the draft itself says).

> There's no need for using a polyfill here.

_There is technically (IMHO), only it's a very tiny one._  But my issue here isn't with the size of the change.

> except not escaping className in span()

Thanks for pointing this out.  I may make a separate issue to track this but as all grammars are themselves fully executable JS code we already sort of consider them a "fully trusted" context - if someone does not fully trust a 3rd party grammar they should not use it - period.  Perhaps Trusted Types might allow better granularity here in the future though... (I hadn't read the whole spec).

> I've considered that option but decided against that ... If someone is expecting that HTMLRenderer.value() would return a string then returning TrustedHTML from it would break that code. 

This is a private/internal API so I don't think this is an issue.   But I'd also be fine adding a `trustedHTML` getter or something long term that simply returned the TrustedHTML wrapped value.

---

Thanks for the detailed response, but for now I say we wait for the spec to mature and to be implemented by additional browser engines.  We do not generally rush to add bleeding edge features supported by a single browser - _even if_ the change *appears* minimal. As mentioned above I also worry a draft spec is a "moving target"... having Highlight.js potentially break if things should change out from under us does not seem like much fun.

I'd suggest `2. Patch Highlight.js` if this is an absolute necessity for Google.  Such a tiny patch should be easy to manage and maintain until the spec matures and perhaps this lands in the core library.

joshgoebel:
You also might consider a plugin that hooked `after:highlightElement` though I'm not sure if that would be helpful in your use case or not... we'd still call `element.innerHTML`...  so if that is going to raise of blow up then that might be a problem... if it would instead just silently log (or do nothing) then you could set the innerHTML using TrustedTypes from within the after callback - preventing any need to maintain a patched library.

vrana:
> There is technically (IMHO), only it's a very tiny one.

This code works as is, there's really no need for a polyfill anywhere.

koto:
Thanks Josh! That's fair, we'll look into internal patches or the `after:highlightElement` hook. 

To respond to your concern though regarding the API volatility, it's of course probable that the API will evolve and change (even HTML is now a living standard), though it's highly unlikely that it would in a backwards-incompatible way. To protect against that, every integration feature tests the API before using it (in this particular patch, we probably could also test for `trustedTypes.createPolicy` existence or wrap the API usage in `try..catch`, falling back to regular `innerHTML` assignment to be extra sure).

Backwards compatibility is of a huge concern to us (and I'm speaking here as TT editor), for principled, but also practical reasons. Quite large [existing integrations](https://github.com/w3c/webappsec-trusted-types/wiki/Integrations) and applications use TT API already.

lweichselbaum:
Hi Josh, thank you for taking the time for reviewing this request.
As koto said, we can definitely patch internally and are not blocked on this change. But we'd like to address this issue for all highlight.js users so they're not running into the same issues as we did.

I'd also love to see Trusted Types be supported natively by more browser engines as I strongly believe it's the best tool against DOM-based XSS we currently have.
Even though Trusted Types is right now only natively supported by Chromium based browsers it is already broadly used (see https://mitigation.supply/#trusted-types, a prominent example is VSCode) and supported in popular frameworks like React and Angular (blogpost about using Trusted Types in SPAs: https://auth0.com/blog/securing-spa-with-trusted-types/). Therefore only non-breaking API changes should be expected.
Developers interested in Trusted Types can also unlock many of the benefits by just using the polyfill, runtime enforcement by the browser is just one of the Trusted Types benefits. 

The downside of this change is rather small and the risk of breakage in the unlikely case of an API change can be fully mitigated by testing for `trustedTypes.createPolicy` existence or wrap the API usage in a `try..catch` as @koto wrote above.
But the upside is really substantial as it would unblock also users of highlight.js from adopting Trusted Types and would help them to get rid of DOM-based XSS.

PR #3289, opened 2021-07-26
fix(csharp) Fix assignments flagging as functions

joshgoebel:
Attemps to resolve https://github.com/exercism/v3-beta/issues/127#issuecomment-885352710

### Changes

- Disallowed `=` inside type expressions`<...>` to prevent them from being too greedy

If this isn't sufficient then we'll need a complex recursive regex like Java I think?

### Checklist


joshgoebel:
Also requested review from Exercism team.

ErikSchierboom (approved):
It appears to be working! It's somewhat hard to see, but I think it's good!

allejo (approved):
LGTM :+1:

image

image

ErikSchierboom:
Thanks @joshgoebel!

PR #3293, opened 2021-07-29
Add support for X# programming language

PatrickKru:
Adds the 3rd party X# language definition.

### Changes

Adds X# to `SUPPORTED_LANGUAGES.md`
Updates changelog at `CHANGES.md`

### Checklist
- [x] Added markup tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,12 @@ Parser:
 
 - fix(types) Fix some type definition issues (#3274) [Josh Goebel][]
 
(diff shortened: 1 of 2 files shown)


joshgoebel:
A few notes:

- You really should use `scope` everywhere, not `className`.
- I'm not sure why you're using `hljs.inherit(STRINGS`, you shouldn't need to inherit... the data you're merging is the same as the data that's already there?

Quite nice!

PR #3318, opened 2021-08-31
fix(abnormal): curly braces show exceptions, {} show &#123; And the &…

Lauweal:
… # 125;

### Checklist


joshgoebel:
These are valid HTML characters, why would they need to be escaped at all?

PR #3319, opened 2021-08-31
fix(abnormal): curly braces show exceptions, {} show &#123; And the &…

Lauweal:
image


joshgoebel:
`{}` are completely valid characters in HTML, there is no need to escape them.  We only escape characters that need to be for security reasons (to prevent code injection).

PR #3320, opened 2021-09-01
fix(js/ts) regex inside `SUBST` is no longer highlighted

### LINKED ISSUE #3288: (Javascript) Slashes in template literal break parsing
**Describe the issue**

A slash inside a template literal, after a slash inside an interpolation expression inside the template literal, breaks the recognition of the template literal end delimiter. This doesn't happen if there is no slash in the expression, or if the string slash comes before the interpolated expression.

Both JavaScript and TypeScript. Also when autodetected.

I dunno, I've come across this [somewhere on StackOverflow](https://stackoverflow.com/questions/68521841/catch-doesnt-work-when-the-api-cant-be-reached) - but specifying the language of the code block didn't help either. In the sample code below, I tried to make a minimal reproducible example, which might not autodetect as JS - but if you put that snippet in a larger code block that *is* successfully autodetected as JS, the same problem occurs.

**Sample Code to Reproduce**

```js
`${dividend/divisor} / `;
const oops = 1;
```
Screenshot from output of above code, rendered on StackOverflow

**Expected behavior**

Github gets it right - the template literal ends with the `` ` ``, and the next line is highlighted normally not seen as a continuation of the template string. The `const` declaration should be recognised, such as in the following example:

same code with ellipsis instead of slash

joshgoebel:
Resolves #3288.

### Checklist
- [x] Added markup tests, or they don't apply here because we don't typically test negatives

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Parser:
 
 Grammars:
 
+- fix(js/ts) regex inside `SUBST` is no longer highlighted [Josh Goebel][]
 - fix(python) added support for unicode identifiers (#3280) [Austin Schick][]
 - enh(css/less/stylus/scss) improve consistency of function dispatch (#3301) [Josh Goebel][]
 - enh(css/less/stylus/scss) detect block comments more fully (#3301) [Josh Goebel][]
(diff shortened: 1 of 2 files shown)


allejo (approved):
Works as intended :+1: I hope it's no big loss for having regex support removed from `{}`

PR #3323, opened 2021-09-03
(scala) Add `do` and `then` keywords

nicolasstucki:
### Changes

Add `then` and `do` as keywords. These keywords are also keywords in Scala 2. `then` was reserved but not used as a keyword. `do` was part of `do .. while` keyword combination.

See https://github.com/scala/vscode-scala-syntax/blob/57e0829cd46980699570101a68c320a20330d36c/src/typescript/Scala.tmLanguage.ts#L595

### Checklist


joshgoebel (approved):

joshgoebel:
@nicolasstucki Thanks so very much!

PR #3324, opened 2021-09-04
enh(scala) remove symbol syntax and fix quoted code syntax

nicolasstucki:
### Changes

Scala 2 had only symbols using this syntax. Scala 3 introduced quoted code using the same syntax, which should not be highlighted. Removing this special case makes all use cases consistent.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Grammars:
 - enh(css/less/stylus/scss) add support for CSS Grid properties [monochromer][]
 - enh(java) add support for Java Text Block (#3322) [Teletha][]
 - enh(scala) add missing `do` and `then` keyword (#3323) [Nicolas Stucki][]
+- enh(scala) remove symbol syntax and fix quoted code syntax (#3324) [Nicolas Stucki][]
 
 [Austin Schick]: https://github.com/austin-schick
 [Josh Goebel]: https://github.com/joshgoebel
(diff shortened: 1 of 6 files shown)


joshgoebel:
What exactly is the fix to quoted code syntax exactly?  I don't see anything in the PR about that (other than the new test).

joshgoebel:
Nevermind, I get it now. :)  Just had to read it twice. :)  Quoted code is using the `'` now, not just symbols...

So what exactly is quoted code... is that not a string then?

nicolasstucki:
> So what exactly is quoted code... is that not a string then?

These are part of the metaprogramming syntax of Scala 3. These represent quoted sections of code, in other words, the contents are also normal source code.

They follow syntactical rules similar to string interpolation. String interpolators like `s"..."` and `f"..."` can contain `${x}` or `$x` to escape the string. In quoted code a `'{...}` can be escaped with `${x}` or `$x`. As well as a splice `${..}` can be escaped with `'{x}` or `'x`. The `'x` is the one we want to fix in this PR.

For more info see https://docs.scala-lang.org/scala3/guides/macros/quotes.html. This document also shows some instances of `'x` that should not have been highlighted.

joshgoebel (approved):

joshgoebel:
Scala 2 had only symbols using this syntax. Scala 3 introduced quoted code using the same syntax, which should not be highlighted. Removing this special case makes all use cases consistent.

joshgoebel:
@nicolasstucki Thanks fo much!

PR #3326, opened 2021-09-06
enh(scala) add Scala 3 extension soft keyword

nicolasstucki:
### Changes

Add support for Scala 3 extension soft keyword (see https://docs.scala-lang.org/scala3/reference/soft-modifier.html).

Use the same heuristic as in https://github.com/scala/vscode-scala-syntax/blob/57e0829cd46980699570101a68c320a20330d36c/src/typescript/Scala.tmLanguage.ts#L1134.

### Checklist

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -112,6 +112,17 @@ export default function(hljs) {
     contains: [ NAME ]
   };
 
+  const EXTENSION = {
+    begin: [
+      /^\s*/, // Is first token on the line
+      'extension',
+      /\s+(?=[[(])/, // followed by at least one space and `[` or `(`
+    ],
+    beginScope: {
+      2: "keyword",
+    }
+  };
+
   return {
     name: 'Scala',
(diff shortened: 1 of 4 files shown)


src/languages/scala.js
```diff
@@ -117,6 +117,11 @@ export default function(hljs) {
     contains: [ NAME ]
   };
 
+  const EXTENSION = {
+    className: 'keyword',
+    begin: /(?<=^\s*)extension(?=\s+[\[\(])/
```
joshgoebel, line 122: Sorry, we can't use look-behind, Safari still doesn't support this.

joshgoebel: Consider using a multi-matcher instead and only highlighting the part you care about.

<https://highlightjs.readthedocs.io/en/latest/mode-reference.html#beginscope>

nicolasstucki: Will do this for other PRs as well

PR #3327, opened 2021-09-06
enh(scala) add Scala 3 `end` and `extension` soft keywords

nicolasstucki:
### Changes

Add Scala 3 `end` of definition or expression marker (see https://docs.scala-lang.org/scala3/reference/soft-modifier.html)

Simplified version of https://github.com/scala/vscode-scala-syntax/blob/main/src/typescript/Scala.tmLanguage.ts#L599-L634. Here we only need one kind of keyword.

Based on #3326

### Checklist

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -112,6 +112,17 @@ export default function(hljs) {
     contains: [ NAME ]
   };
 
+  const EXTENSION = {
+    begin: [
+      /^\s*/, // Is first token on the line
+      'extension',
+      /\s+(?=[[(])/, // followed by at least one space and `[` or `(`
+    ],
+    beginScope: {
+      2: "keyword",
+    }
+  };
+
   return {
     name: 'Scala',
     keywords: {
@@ -126,6 +137,7 @@ export default function(hljs) {
       METHOD,
       CLASS,
       hljs.C_NUMBER_MODE,
+      EXTENSION,
       ANNOTATION
     ]
   };
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -123,6 +123,19 @@ export default function(hljs) {
     }
   };
 
+  const END = [{
+    begin: [
+      /^\s*/, // Is first token on the line
+      /end/,
+      /\s+/,
+      /(extension\b)?/, // `extension` is the only marker that follows an `end` that cannot be captured by another rule.
+    ],
+    beginScope: {
+      2: "keyword",
+      4: "keyword",
+    }
+  }];
+
   return {
     name: 'Scala',
     keywords: {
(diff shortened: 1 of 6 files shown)


src/languages/scala.js
```diff
@@ -117,6 +117,16 @@ export default function(hljs) {
...
   };
 
+  const EXTENSION = {
+    className: 'keyword',
+    begin: /(?<=^\s*)extension(?=\s+[\[\(])/
+  };
+
+  const END = {
+    className: 'keyword',
+    begin: /(?<=^\s*)end(\s+extension|\b)/
```
joshgoebel, line 127: Is there some reason both these are special and can't simply be added to the keywords list?  Is there a reason they absolutely need mode rules instead?

Are these not reserved words everywhere?

nicolasstucki: > Are these not reserved words everywhere?

Unfortunately, they are not reserved keywords. It is not part of the Scala 2 syntax and Scala 3 only allows them in this particular context. Otherwise, they are considered normal identifiers. We can have names `end`/`extension` that can be defined with `def extension: FileExtension = ...` or members called `end` such as in `line.end`.

> Is there some reason both these are special and can't simply be added to the keywords list? Is there a reason they absolutely need mode rules instead?

This is to avoid breaking code in Scala 2 and code in Scala 3 (in particular contexts).

joshgoebel: Ok, I figured once I saw you using the normal keyword feature in your other PR.  We'll still need to remove the look-behind here though and find another way.

joshgoebel (approved):

joshgoebel (approved):

joshgoebel:
@nicolasstucki Thanks for the work on this!

nicolasstucki:
It is a draft only because it depends on #3326 and any changes that might happen to that PR. Though we could skip/close #3326 and merge this one instead.

nicolasstucki:
Managed to join the `END` rules to avoid having to modes.

src/languages/scala.js
```diff
@@ -112,6 +112,30 @@ export default function(hljs) {
     contains: [ NAME ]
   };
 
+  const EXTENSION = {
+    begin: [
+      /^\s*/, // Is first token on the line
+      'extension',
+      /\s+(?=[[(])/, // followed by at least one space and `[` or `(`
```
nicolasstucki, line 119: Should I use escapes for `(` and `[` as in the other PR?
The rule would be `/\s+(?=\[|\()/`.

joshgoebel: This is fine, it's a character group.  But we don't generally use a single character group to avoid escapes, hence my flagging the other.

PR #3329, opened 2021-09-06
enh(scala) add `inline` soft keyword

nicolasstucki:
### Changes

Make `inline` a keyword when it is not part of a member selection of application.

Basic idea form https://github.com/scala/vscode-scala-syntax/blob/main/src/typescript/Scala.tmLanguage.ts#L665, but without the need to distinguish the two kinds of keyword.
Also see https://docs.scala-lang.org/scala3/reference/soft-modifier.html

### Checklist

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -112,6 +112,16 @@ export default function(hljs) {
     contains: [ NAME ]
   };
 
+  // TODO: use negative look-behind in future
+  //       /(?<!\.)\binline(?=\s)/
+  const INLINE_MODES = [{
+    match: /\.inline\b/
+  },
+  {
+    begin: /\binline(?=\s)/,
+    keywords: 'inline'
+  }];
+
   return {
     name: 'Scala',
     keywords: {
@@ -126,6 +136,7 @@ export default function(hljs) {
       METHOD,
       CLASS,
       hljs.C_NUMBER_MODE,
+      ...INLINE_MODES,
       ANNOTATION
     ]
   };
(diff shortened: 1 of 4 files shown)


src/languages/scala.js
```diff
@@ -117,6 +117,11 @@ export default function(hljs) {
     contains: [ NAME ]
   };
 
+  const INLINE = {
+    className: 'keyword',
+    begin: /(?<!\.)\binline(?=\s)/
```
joshgoebel, line 122: Again, can't use look-behind.  To solve "false positive" cases like this you can often introduce a second dummy rule like:

```js
// consume inline as a property access
{ match: /\.inline/ } 
```

The dummy rule matches the proper access case, preventing your highlighting rule from seeing that content.

---

Also, `beginKeywords` has this magic built-in but you'd need to be inside of a separate scope to take advantage of that, so I think my first suggestion may be what you want.

nicolasstucki: `beginKeywords` did not work with the `` `inline` + ...`` case.

src/languages/scala.js
```diff
@@ -112,6 +112,14 @@ export default function(hljs) {
...
   };
 
+  const INLINE_NAME_SELECT = {
+    match: /\.inline\b/
+  };
+  const INLINE = {
+    begin: /\binline(?=\s)/,
+    keywords: 'inline'
+  };
+
```
joshgoebel, line 122:
```suggestion
  // TODO: use negative look-behind in future
  const INLINE_MODES = [{
    match: /\.inline\b/
  },
  {
    begin: /\binline(?=\s)/,
    keywords: 'inline'
  }];
```

src/languages/scala.js
```diff
@@ -126,6 +134,8 @@ export default function(hljs) {
       METHOD,
       CLASS,
       hljs.C_NUMBER_MODE,
+      INLINE_NAME_SELECT,
+      INLINE,
```
joshgoebel, line 138:
```suggestion
      ...INLINE_MODES,
```

src/languages/scala.js
joshgoebel: Lets keep them grouped I think.

src/languages/scala.js

joshgoebel (approved):

PR #3330, opened 2021-09-06
enh(scala) add `using` soft keyword

nicolasstucki:
### Changes

`using` is a keyword if it appears at the start of a parameter or argument list.
See https://docs.scala-lang.org/scala3/reference/soft-modifier.html

Based on https://github.com/scala/vscode-scala-syntax/blob/57e0829cd46980699570101a68c320a20330d36c/src/typescript/Scala.tmLanguage.ts#L504

### Checklist

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -6,6 +6,8 @@ Contributors: Erik Osheim <d_m@plastic-idolatry.com>
 Website: https://www.scala-lang.org
 */
 
+import * as regex from '../lib/regex.js';
+
 export default function(hljs) {
   const ANNOTATION = {
     className: 'meta',
@@ -107,8 +109,7 @@ export default function(hljs) {
   const METHOD = {
     className: 'function',
     beginKeywords: 'def',
-    end: /[:={\[(\n;]/,
-    excludeEnd: true,
+    end: regex.lookahead(/[:={\[(\n;]/),
     contains: [ NAME ]
   };
 
@@ -146,6 +147,17 @@ export default function(hljs) {
     keywords: 'inline'
   }];
 
+  const USING_PARAM_CLAUSE = {
+    begin: [
+      /\(\s*/, // Opening `(` of a parameter or argument list
+      /using/,
+      /\s+(?!\))/, // Spaces not followed by `)`
+    ],
+    beginScope: {
+      2: "keyword",
+    }
+  };
+
   return {
     name: 'Scala',
     keywords: {
@@ -163,6 +175,7 @@ export default function(hljs) {
       EXTENSION,
       END,
       ...INLINE_MODES,
+      USING_PARAM_CLAUSE,
       ANNOTATION
     ]
   };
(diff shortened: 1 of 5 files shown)


src/languages/scala.js
```diff
@@ -107,11 +107,22 @@ export default function(hljs) {
   const METHOD = {
     className: 'function',
     beginKeywords: 'def',
-    end: /[:={\[(\n;]/,
+    end: /(?=[:={\[(\n;])/,
```
joshgoebel, line 110: Please import `regex.js` tools and use `regex.lookahead` here for clarity.

src/languages/scala.js
```diff
@@ -107,11 +107,22 @@ export default function(hljs) {
...
     excludeEnd: true,
     contains: [ NAME ]
   };
 
+  const USING_PARAM_CLAUSE = {
+    begin: [
+      /[(]\s*/, // Opening `(` of a parameter or argument list
```
joshgoebel, line 117:
```suggestion
      /\(\s*/, // Opening `(` of a parameter or argument list
```

Prefer escape vs character group.

joshgoebel (changes requested):
Looks pretty good now, just tiny changes.

src/languages/scala.js
nicolasstucki: Now I'm only using the negative-lookahead. Should I define `regex.negativelookahead`?

src/languages/scala.js

src/languages/scala.js
joshgoebel: Huh, this looks positive?   I didn't flag any places you were using negative (or if I did please ignore)... I'm not sure we want to have tons of JS instead of regex but I have found the super common functions we already have more readable, such as positive `lookahead`, etc..

nicolasstucki: I got mixed up with the negative-lookahead below (`/\s+(?!\))/`). Now I use `regex.lookahead` here. Sorry for the confusion.

src/languages/scala.js
```diff
@@ -107,7 +109,7 @@ export default function(hljs) {
   const METHOD = {
     className: 'function',
     beginKeywords: 'def',
-    end: /[:={\[(\n;]/,
+    end: regex.lookahead(/[:={\[(\n;]/),
     excludeEnd: true,
```
joshgoebel, line 113: If you're using lookahead then you shouldn't need `excludeEnd` anymore, as there will be no match to exclude.

nicolasstucki: Removed

src/languages/scala.js
joshgoebel: No problem.

joshgoebel (approved):

joshgoebel:
@nicolasstucki Anything else or is scala perfect now? :-)

Psoffie009:
> ### Changes
> `using` is a keyword if it appears at the start of a parameter or argument list.
> See https://docs.scala-lang.org/scala3/reference/soft-modifier.html
> 
> Based on https://github.com/scala/vscode-scala-syntax/blob/57e0829cd46980699570101a68c320a20330d36c/src/typescript/Scala.tmLanguage.ts#L504
> 
> ### Checklist
> * [x]  Added markup tests, or they don't apply here because...
> * [x]  Updated the changelog at `CHANGES.md`

Psoffie009:
> @nicolasstucki Anything else or is scala perfect now? :-)

PR #3344, opened 2021-09-29
 enh(bash) added gnu core utilities 

### LINKED ISSUE #3342: [Request] Imporove bash's colours.
I'm always frustrated when I write a bash code block but it does not support the gnu coreutils so some of the commands are not highlighted like `grep and cat` for example.
it also doesn't recognize executables that i installed like `nmap` or any other program.

A clear and concise description of how you imagine we might solve this (if you have any ideas).
add it to the built_in of bash.js
for the other executables that are not in the gnu coreutils or sh buildins i don't know a easy fix, you could make everything coloured that starts at the beginning of the line also include `./<executable name>` or if it comes after the pipe `echo "cGxlYXNlIGZpeAo=" | base64 -d` so echo and base64 are yellow because currently base64 is not.
[gnu core utils](https://en.wikipedia.org/wiki/List_of_GNU_Core_Utilities_commands)

I don't have an alternative solution.

**Additional context...**
```bash
./test "3"
echo "test"
cat test.txt | grep "lol"
```
only test and echo are coloured, im missing the cat and grep.
also it would be nice to see the ./ also getting a colour.

katzeprior:
I added the gnu core utils commands to bash for future changes.

Almost Resolves #3342 

### Changes
1. Added gnu core utilities commands to the bash list.
2. fixed bash tests to reflect changes.
3. fixed 1/3 test errors in shell and docker tests.

### Checklist
fixed most not added


joshgoebel:
For a quick fix (I'm not sure how much work you want to put in here) you could just remove `install` from the list and then fixup the remaining tests for the last amount of effort.

joshgoebel:
I think that's about as close as we're going to get for now unless you want to look into adding individual rules for all the edge cases:

- `&&`, `|`, etc...

That would be much easier to do in the future though once we support look-behind.

joshgoebel:
> Don't know how the look-behind feature will look

I'm speaking of look-behind in regex.  Useful to say "something that was immediately proceeded by"... but it's not supported in all browsers yet, so we wait... without it you have to do a more complex multi-matcher to handle those cases.

> ssh doesn't fit in the list but still should be highlighted i think.
I don't necessarily disagree but  I'm not sure we should highlight "every single useful linux command" under the sun... so I feel like we may need further justification before just adding the top 50 random linux tools.  :-)  Feels like we need some kind of criteria.

PR #3348, opened 2021-10-03
enh(fsharp) Global overhaul

### LINKED ISSUE #3347: (F#,fsharp) Wrong tokenization
```fsharp
open System.Text

let base64Charset =
      [ 'A' .. 'Z' ]
    @ [ 'a' .. 'z' ]
    @ [ '0' .. '9' ]
    @ [ '+'; '/'   ]
    |> Array.ofList
```

In the above F# sample (using `language-fsharp`), highlight.js is unable to tokenize all the character definitions (in single quotes: `'x'`) properly.

Using `language-ocaml` — which has a very similar syntax — seems to work (but then other F# specific things start to break).

See repro here: https://codepen.io/yaurthek/pen/jOwRRoz (directly referencing highlightjs.org highlight.js version so it's up to date and has support for F#)

*Side note: the jsfiddle suggested in the issue template (https://jsfiddle.net/ajoshguy/gfzujpyt/) is a dead link*

mlaily:
Fixes #3347 as well as a lot of other things F# related.

### Changes
Here is the high level list of things that changed (Also listed in the commit description):
* Fix character literals not being parsed
* Add missing keywords
* Fix parsing of generic type symbols
* Fix parsing of ML comments without space within
* Improve attributes parsing
* Add parsing of computation expressions
* Add parsing of preprocessor and FSI directives
* Add parsing for many operators
* Add parsing of binary number literals
* Add unit tests

### Checklist
- [x] Added markup tests

---

I have two questions left:
- What should I put in the changelog?
- It seems I have broken some language auto-detection tests. What's the best way to fix it? (I suspect it's part of the "`relevance` black magic", but I'm not sure where to start)

### DIFF
--- a/src/languages/fsharp.js
+++ b/src/languages/fsharp.js
@@ -1,20 +1,24 @@
 /*
 Language: F#
 Author: Jonas Follesø <jonas@follesoe.no>
-Contributors: Troy Kershaw <hello@troykershaw.com>, Henrik Feldt <henrik@haf.se>
+Contributors: Troy Kershaw <hello@troykershaw.com>, Henrik Feldt <henrik@haf.se>, Melvyn Laïly <melvyn.laily@gmail.com>
 Website: https://docs.microsoft.com/en-us/dotnet/fsharp/
 Category: functional
 */
 
 /** @type LanguageFn */
 export default function(hljs) {
+
+  const GENERICTYPESYMBOL = {
+    begin: /('|\^)[a-zA-Z0-9_]+/,
+    scope: 'symbol'
+  };
+  
   const TYPEPARAM = {
     begin: '<',
     end: '>',
     contains: [
-      hljs.inherit(hljs.TITLE_MODE, {
-        begin: /'[a-zA-Z0-9_]+/
-      })
+      GENERICTYPESYMBOL
     ]
   };
 
@@ -39,6 +43,7 @@ export default function(hljs) {
     "extern",
     "false",
     "finally",
+    "fixed",
     "for",
     "fun",
     "function",
@@ -57,6 +62,7 @@ export default function(hljs) {
     "mutable",
     "namespace",
     "new",
+    "not",
     "null",
     "of",
     "open",
@@ -66,7 +72,6 @@ export default function(hljs) {
     "public",
     "rec",
     "return",
-    "sig",
     "static",
     "struct",
     "then",
@@ -94,30 +99,21 @@ export default function(hljs) {
     illegal: /\/\*/,
     contains: [
       {
-        // monad builder keywords (matches before non-bang kws)
-        className: 'keyword',
-        begin: /\b(yield|return|let|do)!/
+        // monad builder keywords (matches before non-bang keywords)
+        scope: 'keyword',
+        match: /\b(yield|return|let|do|match|use)!/
       },
       {
-        className: 'string',
-        begin: '@"',
-        end: '"',
-        contains: [
-          {
-            begin: '""'
-          }
-        ]
+        scope: 'string',
+        // matches triple quote strings, verbatim strings (@""), character literals...
+        match: /(?:"""[\s\S]*?"""|@"(?:""|[^"])*"|"(?:\\[\s\S]|[^\\"])*")|'(?:[^\\']|\\(?:.|\d{3}|x[a-fA-F\d]{2}|u[a-fA-F\d]{4}|U[a-fA-F\d]{8}))'/
       },
-      {
-        className: 'string',
-        begin: '"""',
-        end: '"""'
-      },
-      hljs.COMMENT('\\(\\*(\\s)', '\\*\\)', {
+      hljs.COMMENT(/\(\*(?!\))/, /\*\)/, {
         contains: ["self"]
       }),
       {
-        className: 'class',
+        // type definitions:
+        scope: 'title.class',
         beginKeywords: 'type',
         end: '\\(|=|$',
         excludeEnd: true,
@@ -127,20 +123,47 @@ export default function(hljs) {
         ]
       },
       {
-        className: 'meta',
-        begin: '\\[<',
-        end: '>\\]',
-        relevance: 10
+        // computation expressions:
+        beginScope: "emphasis",
+        match: /\b[_a-z]\w*(?=\s*\{)/i
       },
       {
-        className: 'symbol',
-        begin: '\\B(\'[A-Za-z])\\b',
-        contains: [hljs.BACKSLASH_ESCAPE]
+        // preprocessor directives and fsi commands:
+        scope: 'meta',
+        begin: '#',
+        end: '$',
+        keywords: {
+          keyword: 'if else endif line nowarn light r i I load time help quit'
+        }
       },
+      {
+        // [<Attributes("")>]
+        scope: 'meta',
+        begin: /^\s*\[<(?=[<\w])/,
+        excludeBegin: true,
+        end: />\]/,
+        excludeEnd: true,
+        relevance: 10,
+        contains: [
+          {
+            scope: 'string',
+            begin: /"/,
+            end: /"/
+          },
+          hljs.C_NUMBER_MODE
+        ]
+      },
+      {
+        scope: 'operator',
+        // only non arithmetic operators that we can confidently match:
+        match: /\|{1,3}>|<\|{1,3}|->|<-|(?<!<)<<(?!<)|(?<!>)>>(?!>)|:>|:\?>?|\.\.|::|(?<!\|)\|(?!\|)/
+      },
+      GENERICTYPESYMBOL,
       hljs.C_LINE_COMMENT_MODE,
       hljs.inherit(hljs.QUOTE_STRING_MODE, {
         illegal: null
       }),
+      hljs.BINARY_NUMBER_MODE,
       hljs.C_NUMBER_MODE
     ]
   };
--- a/test/markup/fsharp/attributes.expect.txt
+++ b/test/markup/fsharp/attributes.expect.txt
@@ -1,0 +1,4 @@
+<span class="hljs-comment">// Strings and numbers are highlighted inside the attribute</span>
+[&lt;<span class="hljs-meta">Foo</span>&gt;]
+[&lt;<span class="hljs-meta">Bar(<span class="hljs-string">&quot;bar&quot;</span>); Foo(<span class="hljs-number">1</span>, <span class="hljs-number">2</span>)</span>&gt;]
+<span class="hljs-keyword">let</span> x () = ()
--- a/test/markup/fsharp/attributes.txt
+++ b/test/markup/fsharp/attributes.txt
@@ -1,0 +1,4 @@
+// Strings and numbers are highlighted inside the attribute
+[<Foo>]
+[<Bar("bar"); Foo(1, 2)>]
+let x () = ()
--- a/test/markup/fsharp/comments.expect.txt
+++ b/test/markup/fsharp/comments.expect.txt
@@ -3,12 +3,42 @@
 <span class="hljs-comment">(*
     here is a multi-line comment on
     multiple lines
+
+trying to break it:
+<span class="hljs-comment">(**)</span>
+/*
+asdf
+*/
+
+<span class="hljs-comment">(* *)</span>
+
 *)</span>
 
 <span class="hljs-keyword">let</span> index =
     len
-    |&gt; float
-    |&gt; Operators.(*) <span class="hljs-number">0.1</span>      <span class="hljs-comment">// (*) here is not comment</span>
-    |&gt; Operators.(+) <span class="hljs-number">1</span>        <span class="hljs-comment">// (+) here is not comment</span>
-    |&gt; Operators.(-) len      <span class="hljs-comment">// (-) here is not comment</span>
-;;
+    <span class="hljs-operator">|&gt;</span> float
+    <span class="hljs-operator">|&gt;</span> Operators.(*) <span class="hljs-number">0.1</span>      <span class="hljs-comment">// (*) here is not comment</span>
+    <span class="hljs-operator">|&gt;</span> Operators.(+) <span class="hljs-number">1</span>        <span class="hljs-comment">// (+) here is not comment</span>
+    <span class="hljs-operator">|&gt;</span> Operators.(-) len      <span class="hljs-comment">// (-) here is not comment</span>
+
+
+<span class="hljs-comment">// foobar</span>
+<span class="hljs-comment">//bar</span>
+<span class="hljs-comment">(**)</span>
+<span class="hljs-comment">(*nospace*)</span>
+<span class="hljs-comment">(* space *)</span>
+<span class="hljs-comment">/// &lt;summary&gt;</span>
+<span class="hljs-comment">/// Class level summary documentation goes here.</span>
+<span class="hljs-comment">/// &lt;/summary&gt;</span>
+<span class="hljs-comment">/// &lt;remarks&gt;</span>
+<span class="hljs-comment">/// Longer comments can be associated with a type or member through</span>
+<span class="hljs-comment">/// the remarks tag.</span>
+<span class="hljs-comment">/// &lt;/remarks&gt;</span>
+<span class="hljs-keyword">let</span> x = ()
+
+<span class="hljs-comment">// the next one is not a comment</span>
+(*) (*)
+
+/*
+this one is <span class="hljs-keyword">not</span> a valid comment either
+*/
--- a/test/markup/fsharp/comments.txt
+++ b/test/markup/fsharp/comments.txt
@@ -3,6 +3,15 @@
 (*
     here is a multi-line comment on
     multiple lines
+
+trying to break it:
+(**)
+/*
+asdf
+*/
+
+(* *)
+
 *)
 
 let index =
@@ -11,4 +20,25 @@ let index =
     |> Operators.(*) 0.1      // (*) here is not comment
     |> Operators.(+) 1        // (+) here is not comment
     |> Operators.(-) len      // (-) here is not comment
-;;
+
+
+// foobar
+//bar
+(**)
+(*nospace*)
+(* space *)
+/// <summary>
+/// Class level summary documentation goes here.
+/// </summary>
+/// <remarks>
+/// Longer comments can be associated with a type or member through
+/// the remarks tag.
+/// </remarks>
+let x = ()
+
+// the next one is not a comment
+(*) (*)
+
+/*
+this one is not a valid comment either
+*/
--- a/test/markup/fsharp/computation-expressions.expect.txt
+++ b/test/markup/fsharp/computation-expressions.expect.txt
@@ -1,0 +1,23 @@
+<span class="hljs-meta">#<span class="hljs-keyword">r</span> &quot;nuget: Ply&quot;</span>
+<span class="hljs-keyword">open</span> FSharp.Control.Tasks
+<span class="hljs-keyword">open</span> System.Threading.Tasks
+
+<span class="hljs-comment">// Single line, and contains a capital letter</span>
+<span class="hljs-keyword">let</span> unitTask = <span class="hljs-emphasis">unitTask</span> { <span class="hljs-keyword">return</span> () }
+
+<span class="hljs-keyword">let</span> work =
+    <span class="hljs-emphasis">async</span> {
+        <span class="hljs-keyword">let</span> delayTask () =
+            <span class="hljs-comment">// Nested computation</span>
+            <span class="hljs-emphasis">task</span> {
+                printfn <span class="hljs-string">&quot;Delay...&quot;</span>
+                <span class="hljs-keyword">do!</span> Task.Delay <span class="hljs-number">1000</span>
+                <span class="hljs-keyword">return</span> <span class="hljs-number">42</span>
+            }
+        <span class="hljs-keyword">let!</span> result = delayTask () <span class="hljs-operator">|&gt;</span> Async.AwaitTask
+        printfn <span class="hljs-string">&quot;Async F# sleep...&quot;</span>
+        <span class="hljs-keyword">do!</span> Async.Sleep <span class="hljs-number">1000</span>
+        <span class="hljs-keyword">return</span> result
+    }
+
+<span class="hljs-keyword">let</span> result = work <span class="hljs-operator">|&gt;</span> Async.RunSynchronously
--- a/test/markup/fsharp/computation-expressions.txt
+++ b/test/markup/fsharp/computation-expressions.txt
@@ -1,0 +1,23 @@
+#r "nuget: Ply"
+open FSharp.Control.Tasks
+open System.Threading.Tasks
+
+// Single line, and contains a capital letter
+let unitTask = unitTask { return () }
+
+let work =
+    async {
+        let delayTask () =
+            // Nested computation
+            task {
+                printfn "Delay..."
+                do! Task.Delay 1000
+                return 42
+            }
+        let! result = delayTask () |> Async.AwaitTask
+        printfn "Async F# sleep..."
+        do! Async.Sleep 1000
+        return result
+    }
+
+let result = work |> Async.RunSynchronously
--- a/test/markup/fsharp/fsi-and-preprocessor.expect.txt
+++ b/test/markup/fsharp/fsi-and-preprocessor.expect.txt
@@ -1,0 +1,10 @@
+
+<span class="hljs-meta">#<span class="hljs-keyword">r</span> &quot;file.dll&quot;;;                               // Reference (dynamically <span class="hljs-keyword">load</span>) the given DLL</span>
+<span class="hljs-meta">#<span class="hljs-keyword">i</span> &quot;package source uri&quot;;;                     // Include package source uri when searching for packages</span>
+<span class="hljs-meta">#<span class="hljs-keyword">I</span> &quot;path&quot;;;                                   // Add the given search path for referenced DLLs</span>
+<span class="hljs-meta">#<span class="hljs-keyword">load</span> &quot;file.fs&quot; ...;;                         // Load the given file(s) as <span class="hljs-keyword">if</span> compiled and referenced</span>
+<span class="hljs-meta">#<span class="hljs-keyword">time</span> [&quot;on&quot;|&quot;off&quot;];;                          // Toggle timing on/off</span>
+<span class="hljs-meta">#<span class="hljs-keyword">help</span>;;                                       // Display <span class="hljs-keyword">help</span></span>
+<span class="hljs-meta">#<span class="hljs-keyword">r</span> &quot;nuget:FSharp.Data, 3.1.2&quot;;;               // Load Nuget Package &#x27;FSharp.Data&#x27; version &#x27;3.1.2&#x27;</span>
+<span class="hljs-meta">#<span class="hljs-keyword">r</span> &quot;nuget:FSharp.Data&quot;;;                      // Load Nuget Package &#x27;FSharp.Data&#x27; with the highest version</span>
+<span class="hljs-meta">#<span class="hljs-keyword">quit</span>;;</span>
--- a/test/markup/fsharp/fsi-and-preprocessor.txt
+++ b/test/markup/fsharp/fsi-and-preprocessor.txt
@@ -1,0 +1,10 @@
+
+#r "file.dll";;                               // Reference (dynamically load) the given DLL
+#i "package source uri";;                     // Include package source uri when searching for packages
+#I "path";;                                   // Add the given search path for referenced DLLs
+#load "file.fs" ...;;                         // Load the given file(s) as if compiled and referenced
+#time ["on"|"off"];;                          // Toggle timing on/off
+#help;;                                       // Display help
+#r "nuget:FSharp.Data, 3.1.2";;               // Load Nuget Package 'FSharp.Data' version '3.1.2'
+#r "nuget:FSharp.Data";;                      // Load Nuget Package 'FSharp.Data' with the highest version
+#quit;;
--- a/test/markup/fsharp/generic-types.expect.txt
+++ b/test/markup/fsharp/generic-types.expect.txt
@@ -1,0 +1,36 @@
+<span class="hljs-comment">// Primarily testing for generic parameters highlighting...</span>
+
+<span class="hljs-title class_"><span class="hljs-keyword">type</span> <span class="hljs-title">Ref</span>&lt;<span class="hljs-symbol">&#x27;a</span>&gt; </span>=
+{ <span class="hljs-keyword">mutable</span> contents: <span class="hljs-symbol">&#x27;a</span> }
+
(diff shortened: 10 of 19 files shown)


src/languages/fsharp.js
```diff
@@ -1,20 +1,24 @@
...
+  const GENERICTYPESYMBOL = {
+    begin: /('|\^)[a-zA-Z0-9_]+/,
+    scope: 'symbol'
+  };
+  
   const TYPEPARAM = {
     begin: '<',
     end: '>',
     contains: [
-      hljs.inherit(hljs.TITLE_MODE, {
```
joshgoebel, line 15: Anytime you replace a built-in mode (`hljs.TITLE_MODE`) with a custom one first check it's relevance for example titles are just user-defined identifiers have `0` relevance to which language something is... so your new `GENERICTYPESYMBOL` rule should also have no relevance.

Plus there is already relevance here from the outer `<>` rule.  Is any amount of whitespace allowed here or could we make this rule tighter?

src/languages/fsharp.js
```diff
@@ -94,30 +99,21 @@ export default function(hljs) {
...
+        scope: 'keyword',
+        match: /\b(yield|return|let|do|match|use)!/
       },
       {
-        className: 'string',
-        begin: '"""',
-        end: '"""'
+        scope: 'string',
+        // matches triple quote strings, verbatim strings (@""), character literals...
+        match: /(?:"""[\s\S]*?"""|@"(?:""|[^"])*"|"(?:\\[\s\S]|[^\\"])*")|'(?:[^\\']|\\(?:.|\d{3}|x[a-fA-F\d]{2}|u[a-fA-F\d]{4}|U[a-fA-F\d]{8}))'/
```
joshgoebel, line 109: We really need to use some of our regex tools (or `variants`) to make this easier to read... we try not to encourage large unreadable regex in the codebase - we'd rather have the rules written out separately for 10 types of strings so it's easier to maintain, etc.

joshgoebel: Things like escaping inside strings would likely be matched in a `contains` mode, not inside the primary regex.

src/languages/fsharp.js
```diff
@@ -94,30 +99,21 @@ export default function(hljs) {
...
       },
-      hljs.COMMENT('\\(\\*(\\s)', '\\*\\)', {
+      hljs.COMMENT(/\(\*(?!\))/, /\*\)/, {
         contains: ["self"]
       }),
       {
-        className: 'class',
+        // type definitions:
+        scope: 'title.class',
```
joshgoebel, line 116: `title.class` is for names of classes such as `WordParser`, etc... (and `class` alone is deprecated)  you may want a multi-matcher here... what is the syntax you're trying to cover?

src/languages/fsharp.js
```diff
@@ -127,20 +123,47 @@ export default function(hljs) {
         ]
       },
       {
-        className: 'meta',
-        begin: '\\[<',
-        end: '>\\]',
-        relevance: 10
+        // computation expressions:
+        beginScope: "emphasis",
```
joshgoebel, line 127: No, this is a `markup` scope... it's not allowed in code definition.

src/languages/fsharp.js
```diff
@@ -127,20 +123,47 @@ export default function(hljs) {
...
+        match: /\b[_a-z]\w*(?=\s*\{)/i
+      },
+      {
+        // preprocessor directives and fsi commands:
+        scope: 'meta',
+        begin: '#',
```
joshgoebel, line 133: Can this be anywhere on a line or only the beginning?

src/languages/fsharp.js
```diff
@@ -127,20 +123,47 @@ export default function(hljs) {
...
+        }
+      },
+      {
+        // [<Attributes("")>]
+        scope: 'meta',
+        begin: /^\s*\[<(?=[<\w])/,
+        excludeBegin: true,
+        end: />\]/,
+        excludeEnd: true,
+        relevance: 10,
```
joshgoebel, line 146: Perhaps 2, we don't really hand out 10s anymore unless we're sure no other language could ever contain such a thing.

src/languages/fsharp.js
```diff
@@ -127,20 +123,47 @@ export default function(hljs) {
...
+          hljs.C_NUMBER_MODE
+        ]
       },
       {
-        className: 'symbol',
-        begin: '\\B(\'[A-Za-z])\\b',
-        contains: [hljs.BACKSLASH_ESCAPE]
+        scope: 'operator',
+        // only non arithmetic operators that we can confidently match:
+        match: /\|{1,3}>|<\|{1,3}|->|<-|(?<!<)<<(?!<)|(?<!>)>>(?!>)|:>|:\?>?|\.\.|::|(?<!\|)\|(?!\|)/
```
joshgoebel, line 159: Please use `regex.either` from utils and turn this into a nice easy to read array of operators, etc.

src/languages/fsharp.js
joshgoebel: This likely needs to be more precise as right now it'll match say comments in many other languages.  You may need to match the keywords with a lookahead to ensure it's truly a preprocessor and not a false positive.

src/languages/fsharp.js
mlaily: > for example titles are just user-defined identifiers have 0 relevance to which language something is

Noted!

> Is any amount of whitespace allowed here or could we make this rule tighter?


Any amount of whitespace is allowed, as long as it's properly indented. The following is a valid construct:
```fsharp
type Blah<  
         'a 
         > = { x: 'a }
```

src/languages/fsharp.js
mlaily: I changed this because I noticed `class` was deprecated. I thought `title.class` was the replacement.

The syntax I'm trying to cover is any type definition:
```fsharp
type A = int * int // type abbreviation
type B = { FirstName: string; LastName: string } // record
type C = Circle of int | Rectangle of int * int // discriminated union
type D = Day | Month | Year // another discriminated union
type E<'a> = Choice1 of 'a | Choice2 of 'a * 'a // a generic type definition

type MyClass(initX: int) = // class
   let x = initX
   member _.Method() = printf "x=%i" x
```
It's currently parsed as follows:
image

I'm not sure I understand the meaning behind all the different standard scopes. What should I do?

src/languages/fsharp.js
mlaily: Ah that's too bad, I liked the subtle difference!
(it's hard to pick a scope, it does not seem to completely fit in any existing one... I think I'll go for `keyword`, even though it's possible to define custom computation expressions)

src/languages/fsharp.js
mlaily: > Can this be anywhere on a line or only the beginning?

It has to be the first non-whitespace thing on a line.

I used the same matching used for `csharp` as I believe it's similar.

Would the following be better?
```js
  {
    // preprocessor directives and fsi commands:
    scope: 'meta',
    begin: /^\s*#(if|else|endif|line|nowarn|light|r|i|I|load|time|help|quit)/i,
    end: /\s|$/
  },
```

(I'm not sure if there is a better way to do a look-ahead without repeating all the keywords)

src/languages/fsharp.js
joshgoebel: I think you need a look-ahead regex on the `<` just to make sure it's followed by `whitespace '[char]`... otherwise this is set going to find tons of false positives. In fact this whole thing might be better as a multi-match (where you pass an array of regex and then pass scopes for each - see the docs) - if the type symbols always follow this pattern so closely.

src/languages/fsharp.js
joshgoebel: > The syntax I'm trying to cover is any type definition (the part before the first = or ():

We do not  care/desire this type of "semantic" flagging (that what we deprecated when we deprecated `class`).  IE we don't care about "class definitions".  We only care about "class name/title" (which may or may not be in a definition).   So above there should be NO nesting. 

- `type` is a keyword 
- `A` is a `title.class`
- `type A = int * int` (overall is nothing)

src/languages/fsharp.js
joshgoebel: What is it called exactly in F#? What is the closed equivalent in other languages?

src/languages/fsharp.js
joshgoebel: Yes, but please put them all in a nice tidy array and then use our regex tools to build the regex expression.  And you might want a multi-matcher here because technically the `\s*` should not be part of the highlighted content. (we do try to be strict about our whitespace boundaries when it's easy to do so).

`end` should probably be a look-ahead here for the same reason, to not consume the space inside the scope.

src/languages/fsharp.js
mlaily: It's called a [computation expression](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/computation-expressions). I don't know of an equivalent in other languages.

They are F#'s way to implement syntactic sugar for monads, and they also can be used to create simple but powerful DSL.
For example, they are used to implement async workflows:
```fsharp
let work =
    async { // async computation expression
        let delayTask () =
            task { // task computation expression (compatible with C# Tasks)
                printfn "Delay..."
                // bang keywords can be used inside computation expressions.
                // The semantic of this depends on its implementation.
                do! Task.Delay 1000
                return 42
            }
        let! result = delayTask () |> Async.AwaitTask
        printfn "Async F# sleep..."
        do! Async.Sleep 1000
        return result
    }

let result = work |> Async.RunSynchronously
```

As another and completely different example, they are used to create a DSL to initialize a web application with the Saturn F# web framework: https://saturnframework.org/reference/Saturn/saturn-application-applicationbuilder.html
```fsharp
let app = application {
    pipe_through endpointPipe
    use_router topRouter
    url "http://0.0.0.0:8085/"
    memory_cache
    use_static "static"
    use_gzip
}
```

The previous `application` computation expression defines several new keywords that can be used in the computation (`pipe_through`, `use_router`, `url`...)

joshgoebel: Wouldn't `{}` be punctuation (or operators) and the name perhaps `title.function`?  If someone can define DSLs with them like this then it sounds a lot like a type of function/proc/lambda, etc...

mlaily: Semantically it has nothing to do with a function (it does not define something callable in any way).
Do you mean syntactically maybe?

I feel like it would fit better as a keyword (a keyword before a block that can contain statements from a DSL). 
It seems other code highlighting libraries supporting F# interpret it that way.

mlaily:
I think I addressed most (all?) of your remarks.

Sorry about the big diff.

Most notable changes:
- I replaced my string regex with different modes (I used csharp as a starting point)
- In the process, I added some support for interpolated strings (`$"...{1+1}..." // prints ...2...`).
  - Some edge cases do not parse correctly, but it's better than before so I left it.
  - `test/markup/fsharp/strings.txt` contains some breaking examples.
- I split all the big remaining regexes
- I used multi-matching where it was possible and beneficial
- I added detection for many builtin types and other literals (initially to help with auto-detection, but it is also nicer to look at)
- I managed to fix auto-detection by fiddling with relevance a bit (notably by splitting operators detection and assign a zero relevance to some)

Since I wasn't sure whether you squash your PR or not, I added another commit with a more or less proper message.

Thanks for your feedback, and tell me if I have missed something!

src/languages/fsharp.js
joshgoebel: > Do you mean syntactically maybe?

Yes.

I'm uncertain.  Lets use [classNameAliases](https://highlightjs.readthedocs.io/en/latest/mode-reference.html#classnamealiases) to give it the proper F# name/scope `computation-expression.`, and then alias it back down to keyword for now.

Curious, do you know what scope TextMate/VSCode applies to these?

src/languages/fsharp.js
```diff
@@ -89,52 +78,317 @@ export default function(hljs) {
...
+        match: /\{\{/
+      },
+      {
+        match: /\}\}/
+      },
+      hljs.BACKSLASH_ESCAPE,
+      SUBST
+    ]
+  };
+  // $@"...{1+1}..."
```
joshgoebel, line 259: Double braces or single braces? Your regex includes two...

src/languages/fsharp.js
```diff
@@ -89,52 +78,317 @@ export default function(hljs) {
...
+        begin: [
+          /type/,
+          /\s+/,
+          hljs.UNDERSCORE_IDENT_RE
+        ],
+        beginScope: {
+          1: 'keyword',
+          3: 'title.class'
+        },
+        end: /\(|=|$/,
```
joshgoebel, line 367: Technically probably should be a look-ahead to not consume these, yes?

src/languages/fsharp.js
mlaily: Placeholders are defined between single braces, but doubling a brace escapes it.
```fsharp
$"this is {{an example}} and here is an actual placeholder: {1+1}"
$@"this is {{an example}} and here is an actual placeholder: {1+1}"
$"""this is {{an example}} and here is an actual placeholder: {1+1}"""
// All the above are interpreted as "this is {an example} and here is an actual placeholder: 2"
```

src/languages/fsharp.js
```diff
@@ -89,52 +78,317 @@ export default function(hljs) {
...
       },
       {
         // computation expressions:
-        beginScope: "emphasis",
-        match: /\b[_a-z]\w*(?=\s*\{)/i
+        begin: [
+            /\b[_a-z]\w*/,
+            /(?=\s*\{)/
+          ],
+        beginScope: { 1: 'keyword' }
```
joshgoebel, line 379: Traditionally we'd just use `regex.concat` or inline it (as it was before)... I'll give it some thought but not sure we want to start using multi-matchers just in this way (just to split out look-aheads).

src/languages/fsharp.js
mlaily: So, `/(?=\(|=|$)/`? Isn't `excludeEnd: true` enough?
(I don't notice any difference in the parsing, that's why I'm asking)

src/languages/fsharp.js
```diff
@@ -89,52 +78,317 @@ export default function(hljs) {
...
-        keywords: {
-          keyword: 'if else endif line nowarn light r i I load time help quit'
-        }
+        begin: [
+          /^\s*/,
+          regex.concat(/#/, regex.either(...PREPROCESSOR_KEYWORDS)),
+          /\b/
+        ],
+        beginScope: { 2: 'meta' },
+        end: /\s|$/,
```
joshgoebel, line 389: Shouldn't this be a look-ahead?  Remember `\s` also matches newlines. The newline itself shouldn't be part of this match...

src/languages/fsharp.js
```diff
@@ -154,15 +408,37 @@ export default function(hljs) {
...
+        // split in two because we don't want the same relevance for all operators.
+        // warning: the order of the two modes is important, because |> collides with |
+        variants: [
+          {
+            // fsharp specific operators:
+            match: regex.either(
+              /\|{1,3}>/, // |> ||> |||>
+              /<\|{1,3}/, // <| <|| <|||
+              /(?<!<)<<(?!<)/, // << (but not < or <<<)
+              /(?<!>)>>(?!>)/, // >> (but not > or >>>)
```
joshgoebel, line 422: We don't yet support negative lookbehind because not all browsers support it well yet.

src/languages/fsharp.js
```diff
@@ -154,15 +408,37 @@ export default function(hljs) {
...
+            relevance: 2
+          },
+          {
+            // operators common to many languages:
+            match: regex.either(
+              /->/,
+              /<-/,
+              /\.\./,
+              /::/,
+              /(?<!\|)\|(?!\|)/ // | (but not || or |||)
```
joshgoebel, line 435: We don't yet support negative lookbehind because not all browsers support it well yet.

src/languages/fsharp.js
joshgoebel: Ah, I didn't see that, still I'd prefer we change this... long-term `excludeEnd` is probably on it's way out since regex already has a built-in way to do this: look-ahead - and we could simplify our engine to avoid building built-in regex behavior.

src/languages/fsharp.js
joshgoebel: I'm wondering if this should be:

```
      {
        begin: /\{\{/
        end: /\}\}/
      },
```

?

src/languages/fsharp.js
joshgoebel: Ah yet another case where I missed excludeEnd, but the same thougth would apply here: prefer look-ahead. (probably slightly faster also)

joshgoebel (changes requested):
This is looking pretty sharp.  I didn't focus on the relevance stuff as I'll take a final pass at that myself and likely change a few things.  Traditionally we don't give operators relevance at all because it would be too hard to balance all the grammars. And there are other unspoken rules I don't expect you to be aware of. :)

src/languages/fsharp.js
mlaily: Mmh, I don't think so: an escaped brace does not imply another matching escaped brace (something like `$"a brace: {{. another brace: {{"` is entirely possible, and evaluated as `"a brace: {. another brace: {"`)

joshgoebel: Ah, I'll blame it on the fact that the examples made them appear paired.  I understand now. :-)

joshgoebel:
> Sorry about the big diff.

No worries, language "reboots" are always a lot of effort...

> Since I wasn't sure whether you squash your PR or not, 

Almost always squash (myself), but during dev I prefer to see the individual commits to make reviewing easier for long-running RPs.

> I added another commit with a more or less proper message.

That is appreciated!

joshgoebel:
> What should I put in the changelog?

A one-liner saying the grammar has been rewritten and greatly improved sounds about right.

src/languages/fsharp.js
```diff
@@ -84,63 +78,368 @@ export default function(hljs) {
...
-        className: 'meta',
-        begin: '\\[<',
-        end: '>\\]',
-        relevance: 10
+        // [<Attributes("")>]
+        scope: 'meta',
+        begin: /^\s*\[<(?=[<\w])/,
+        excludeBegin: true,
+        end: />\]/,
+        excludeEnd: true,
```
mlaily, line 399: I can also replace this `excludeEnd` with a look-ahead in the `end` regex, what do you think?

Should I also replace the `excludeBegin` with a `begin` regex in a look behind like so (with a look ahead for the expected content): `/(?<=^\s*\[<)(?=\s*\w)/`?

(and I just noticed the current regex has an extra `<` in the content look ahead that I inadvertently added when copy pasting from the csharp version...)

mlaily:
> This is looking pretty sharp. I didn't focus on the relevance stuff as I'll take a final pass at that myself and likely change a few things. Traditionally we don't give operators relevance at all because it would be too hard to balance all the grammars. And there are other unspoken rules I don't expect you to be aware of. :)

Thanks. I feel like some operators are an important way to distinguish F# from other languages (for example the `|>` operator used everywhere), but I'll leave it to you.

If you don't want to add relevance to any operator, my split between "relevant" and "irrelevant" operators should be removed, as it seems rather fragile and might cause some problems if we are not careful (because of overlapping matches, e.g. between `|` and `|>`)

> No worries, language "reboots" are always a lot of effort...

Honestly, I didn't expect I would spend so much time on this, but now I'm in too deep to go back :)

src/languages/fsharp.js
mlaily: Hmm, honestly I don't know what to do about that. What do you suggest?

- Match the variants I didn't want to match and accept having no real logic to what is highlighted or not? (`>>` is used to combine functions, `>>>` is a binary right shift)
- Do some magic I don't know about to replace the negative look behind?
- Do some hack by matching space before/after?
- ...?

(Do you know how it behaves on browsers that don't support it? Can't we keep it if it fails more or less gracefully)

src/languages/fsharp.js
```diff
@@ -84,63 +78,368 @@ export default function(hljs) {
...
+    "None",
+    "Ok",
+    "Error",
+    "infinity",
+    "infinityf",
+    "nan",
+    "nanf",
+    "__LINE__",
+    "__SOURCE_DIRECTORY__",
+    "__SOURCE_FILE__"
```
joshgoebel, line 111: Did you consider `variable.language` or `variable.constant`?  These look like contstants to me, not literals.

src/languages/fsharp.js
```diff
@@ -84,63 +78,368 @@ export default function(hljs) {
...
+    "sprintf",
+    "eprintf",
+    "eprintfn",
+    "fprintf",
+    "fprintfn",
+    "failwith",
+    "failwithf"
+  ];
+
+  // (* potentially multi-line ML style comment *)
```
joshgoebel, line 195: What is "multi-line ML" is that not redundant?

src/languages/fsharp.js
mlaily: Oops, I didn't notice these scopes!

I'm not sure I understand all the nuances between `keyword`, `built_in`, `literal`, `variable.language`, and `variable.constant`... ([the doc](https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html) is sometimes confusing, e.g. for `built_in` it mentions it could be used for constants?)

Which one should be used for user-defined constants, and which one for language level constants/macro?
As I understand it,  `__LINE__` and co are some kind of magical constants whose values are set by the compiler. Does that fit with `variable.constant`?

src/languages/fsharp.js
mlaily: It's ML as in "Metal Language", as in "[ML family language](https://en.wikipedia.org/wiki/ML_(programming_language))". It's unfortunate it can also be interpreted as multi-line.

(these languages all have in common `(* a specific syntax for comments *)`).

Would you prefer I replaced the acronym with the full name?

src/languages/fsharp.js
joshgoebel: > I'm not sure I understand all the nuances between keyword, built_in, literal, variable.language, and variable.constant... 

Me either, LOL... it's a moving target sometimes (with the new ones).  The more nuanced ones are newer scopes and we're trying to use them more.  In the past the older scopes did more "work"...

With regard to current understanding of the older scopes:

- keyword - key language reserved words (if, else, break, continue, etc)
- built_in (built in library functions or variables) - though now this can intermingle with `variable.language` as well as `title.function` and can sometimes be nebulous
- literals - the name is literally the value, `true`, `false`, `null`


> Which one should be used for user-defined constants

`variable.constant`

> which one for language level constants/macro?

I think `variable.constant` as `variable.language` often refers to things such as `this`, `self`, etc...

joshgoebel: Looks like F# allows you to define your own self identifier so sadly that's not something we can highlight.

src/languages/fsharp.js
joshgoebel: I went ahead and resolved these in my commit.

src/languages/fsharp.js
joshgoebel: > Would you prefer I replaced the acronym with the full name?

Yep!

src/languages/fsharp.js
mlaily: > Curious, do you know what scope TextMate/VSCode applies to these?

No sorry, I only checked prismJS and github...

src/languages/fsharp.js
mlaily: > If you don't want to highlight >>> just match it without a scope. 

Ah that makes sense! I guess it didn't occur to me it was an option :)

> Though why are we highlighting some operators and not others?

Mostly because I am afraid to highlight things in places where they should not be highlighted.
[F# allows to define new operators, using sequences of certain characters](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/operator-overloading), and I didn't want to risk only partially highlighting a custom operator for example. So I thought sticking with standard non-arithmetic operators would be a good compromise for now.

joshgoebel:
> Thanks. I feel like some operators are an important way to distinguish F# from other languages (for example the |> operator used everywhere),

Except it's not exclusive.  The pipe operator `|>` (as I know it) is used everywhere in Elixir and there are proposals to add it to Wren also - and that's just two other languages I know of off the top of my head.

The problem is we have almost 200 first-party languages and unless FULL operator support is added to all of them simultaneously (which is a HUGE undertaking, and quite possibly also impossible for some) then auto-detect is going to be just entirely broken if we start giving a select few relevance points.

For example allowing F# to score relevance would lead to many false positives of Elixir code (which doesn't yet include operator relevance)... etc, etc.

So the first task would be: _Add operator highlighting to ALL grammars._ 

After that was done one could circle back and look at the pros and cons of using it for relevance.

---

There are still other issues... even if you think you have a cool exclusive operator I've seen the weirdest ASCII art in comments that can also be problematic, such as a `*=` operator  in a hash comment language for example and then you have C code like:

```
/* =*=*=*=*=*=*=*=*=*= main code =*=*=*=*=*=*=*=*=*= */
```

Yes I've seen such things. :-)  This may be slightly improved since now we have `MAX_KEYWORD_HITS` to protect us against keyword stuffing, but until someone shows me otherwise I'm going to be conservative on operator relevance for now.

src/languages/fsharp.js
mlaily: Yes, I saw, but what about the `excludeBegin`? Should we also strive to replace them with look behinds?

src/languages/fsharp.js
mlaily: Ok, thanks for the explanations!

> Looks like F# allows you to define your own self identifier so sadly that's not something we can highlight.

Are you refering to `this` or `self`? If so, I think it's more that F# does not define such a keyword, and allows the developer to use whatever variable name they want to refer to the current object: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/classes#self-identifiers

mlaily:
About operator relevance: ok I see, thanks for the explanation, it was probably a wrong assumption/rookie mistake on my part :)

mlaily:
I added several commits with fixes.

I replaced the `excludeBegin` with a look-behind in a separate commit, feel free to discard it if you prefer `excludeBegin`.


I still need to address the negative look-behinds for operators, but it's proving trickier than just matching some without assigning them a scope...

src/languages/fsharp.js
joshgoebel: Hmmm, now that I'm looking closer not highlighting the basic math operators seems very strange to me... couldn't we just detect "operator like expressions" with a regex and just highlight ALL potential custom defined operators?  I'm not sure that would have been my first choice, but I really dislike not highlighting the basic operators.

Would doing this conflict with other language features we're not already parsing?

src/languages/fsharp.js
joshgoebel: > Should we also strive to replace them with look behinds?

No, because _we can't use look-behinds at all yet_ - see prior discussion.   When we support look-behind that's definitely something we'd have a look at.

src/languages/fsharp.js
mlaily: > Would doing this conflict with other language features we're not already parsing?

That's part of what worries me... For example I think I removed explicit parsing of the generic type definition brackets to simplify things. They would need to be re-added, along with some more matching for this pattern when it's not part of a type definition (e.g. `let f (x: inref<SomeStruct>) = x.SomeField`)

I fear there are a lot of patterns like that I don't necessarily have in mind...

src/languages/fsharp.js
mlaily: > we can't use look-behinds at all yet

Ah I did not understand that, I thought it was only *negative* look behind we couldn't use, my bad.

Why does [the doc](https://highlightjs.readthedocs.io/en/latest/language-guide.html#regular-expression-features) seems to indicate look-behinds in `begin` are supported?

src/languages/fsharp.js
mlaily: I initially wanted to highlight only the most idiomatic operators (maybe `|>`, `<|`, `>>`, `<<`), but it seemed too arbitrary, that's why I settled on the set I chose...

src/languages/fsharp.js
```diff
@@ -369,11 +376,8 @@ export default function(hljs) {
       },
       {
         // computation expressions:
-        begin: [
-            /\b[_a-z]\w*/,
-            /(?=\s*\{)/
-          ],
-        beginScope: { 1: 'keyword' }
+        beginScope: 'computation-expression',
```
joshgoebel, line 379: Just `scope` should also work and reads a bit nicer.

mlaily: Oops, sorry!

src/languages/fsharp.js
joshgoebel: The key is in the wording:

> look-behind matching (**when JS supports it**) for begin (#2135)

The key being "when JS [browsers] supports it".  _Nothing in our engine prevents them from working for begin - if the JS engine includes support._ (as you've seen when using them initially)   If you wanted to fork the library or use a 3rd party grammar that uses them - go for it - you just lose support for Mac OS Safari.  At least one 3rd party library does feature detection - resulting in inconsistent behavior across browsers - but thats' not a tradeoff we're interested in for Core.

However,  the engine itself would break if look-behind was used in `end` because of the way our engine is currently designed - the end matchers do not always have access to the entire string.

src/languages/fsharp.js
mlaily: Well... now that you mention it, `<>` is the boolean not operator in F# ^^'

It's a bit sad, but maybe not matching any operator is safer then...

It's really frustrating, especially after I thought I managed to do what I wanted with look-behinds...

joshgoebel:
Does F# have a default formatter or well enforced code/lint standards?  Sometimes this can help as I don't mind taking that into account if it's common enough as to be a non-brainer.

src/languages/fsharp.js
mlaily: Ok. I'm still a bit confused by the subtlety, but ok ^^

joshgoebel: https://github.com/highlightjs/highlight.js/blob/dea3cde209d0fe5a164607a61da8c42d796fa459/docs/language-guide.rst#regular-expression-features

mlaily:
> 
> 
> Does F# have a default formatter or well enforced code/lint standards? Sometimes this can help as I don't mind taking that into account if it's common enough as to be a non-brainer.

I think [fantomas](https://github.com/fsprojects/fantomas/) is becoming the de-facto code formatter (It's mentioned in the [F# code formatting guidelines](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting))

But I'm not sure everyone agrees on the details, and the guidelines mention some formatting that is accepted but that tools will override anyway...

(I'm not sure what comment you were replying to and what you have in mind?)

mlaily:
> Sometimes if it helps we can "ignore" poorly formatted code and often that makes our job easier since then there are often guaranteed space boundaries around some things that make detection slightly simple. Until the code golf people show up. :-)

Ah yes, I don't feel comfortable relying on this but I see the appeal :)

joshgoebel:
> Ah yes, I don't feel comfortable relying on this but I see the appeal :)

It can sometimes work well when it's scoped to certain parts of the grammar... "Oh your operators aren't always highlighted, then perhaps write more idiomatic code with nice whitespace, etc..."... and if it means the difference between not highlighting operators at all for anyone or highlighting them for 95% of people who write clean code, then I'd err on the latter.

PR #3351, opened 2021-10-06
enh(nsis) adds NSIS v3.08 commands, some small improvements

idleberg:
### Changes

Adds NSIS v3.08 commands
Fixes case sensitivity option

### Checklist


joshgoebel:
Also, what is `nsExec::ExecToLog "calc.exe"` here exactly?  `class` as a scope has been deprecated, so we need a new scope.

Screen Shot 2021-10-06 at 7 22 57 PM

joshgoebel:
> In NSIS terms, it's a command (ExecToLog) provided by a plugin (nsExec).

I'm thinking:

```
<title.function>nsExec::ExecToLog</>
```

or

```
<title.class>nsExec</>::<title.function>ExecToLog</>
```

Thoughts?

joshgoebel:
@idleberg That was an either/or question. :)

allejo:
Hello 👋

Coming from other programming languages, having a differentiation between a class name and a function is helpful in terms of highlighting but that's because they each carry a separate meaning on their own.

Based solely on this conversation, it sounds to me that since you can't have a class by itself in NSIS then highlighting them separately wouldn't really carry any useful meaning. To me, this behavior sounds a lot more like a namespace rather than a class, like you'd see in C++ with `std::string` and `std::string::npos`; I'd expect both of these values to be highlighted as a single value instead of `std` being a "class."

My vote would be to highlight it as,

```
<title.function>nsExec::ExecToLog</>
```

PR #3353, opened 2021-10-07
fix(js/ts) Float32Array higlighted incorrectly

### LINKED ISSUE #3252: (Javascript) Float32Array broken after 11.0.1

joshgoebel:
### Changes

Resolves #3252.

- Improves the class name detection regex to handle more cases like `Float32Array` or `CSSParser`

### Checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,8 +29,7 @@ module.exports = {
     quotes: "off",
     // this is the style we are already using
     "operator-linebreak": ["error", "before", { overrides: {
-      "=": "after",
-      "+": "after"
+      "=": "after"
     }
     }],
     // sometimes we declare variables with extra spacing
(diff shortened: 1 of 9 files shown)


joshgoebel:
> The only notable exception to this would be the JSON object 

I may go ahead and hard-code that as an exception to the constant rule. :-)  I think that makes sense for a core library class.

PR #3354, opened 2021-10-09
Removed backslashes as escape characters for MATLAB

### LINKED ISSUE #3349: (MATLAB) Should not recognize backslash in strings as escape
## Describe the issue

<https://github.com/highlightjs/highlight.js/blob/934f2e78be0d3b9bf579664f4ccb27987441d1a0/src/languages/matlab.js#L79-L99>

The only special things in MATLAB strings are the quotes themselves. A string terminates immediately on a matched quote, unless the quote is doubled. Everything else should be treated literally.

Therefore, those `BACKSLASH_ESCAPE` should be removed from `contains`.

### References

* <https://www.mathworks.com/help/matlab/matlab_prog/represent-text-with-character-and-string-arrays.html>

* <https://stackoverflow.com/a/2045223>

## Which language seems to have the issue?

`matlab`

## Are you using `highlight` or `highlightAuto`?

`highlight`

## Sample code to reproduce

```matlab
Oir2OmeTiff(OirPath="C:\Example.oir",OutputDirectory="D:\");
% Force
```

From <https://github.com/microsoft/vscode/issues/125618>

## Expected behavior

On the first line, the second string is `"D:\"`. GitHub handles it correctly.

deepto98:
Resolves #3349 

### Checklist
- [ ]Removed  BACKSLASH_ESCAPE  from **contains** for relevant string definitions in **matlab.js**


joshgoebel:
Seems you'll also need to update the tests as well.  Look for the commented line in `markup/index.js` if you'd like us to rewrite the tests for you (which might be exactly fine).

joshgoebel:
Ping.

joshgoebel (approved):

PR #3357, opened 2021-10-11
FSharp fixes

mlaily:
Hello !

This is a follow up PR for #3348.

I realized some things were not optimal or needed fixing...

I did not create a new issue. Let me know if this is ok or if you prefer I create a new one...

### Changes

- fix: Stop highlighting type names everywhere. Only match them in type annotations.
  - Since the names are not reserved and can be re-used for variable/binding names shadowing, it can be confusing and misleading.
- enh: Highlight operators. Operators are a big part of F#, so highlighting them brings a lot of value for legibility, etc...
- enh: Support quoted identifiers (` ``...`` `. They can contain pretty much anything inside the double backquotes)
- fix: Fix type names and symbols matching
   * Type names can contain apostrophes.
   * Type symbols can contain quoted identifiers, but no apostrophes.
- enh: Match attributes in more places (not just at the start of a line), and match more of the things they can contain (not just basic strings and numbers).

### Checklist
- [x] Added markup tests


mlaily:
Well, I tried to continue matching type annotations without a hard-coded list of types while keeping it simple, but I don't think I will be able to get to something that does not break without parsing much more of the F# grammar.

So I decided to follow your advice and stop trying to be perfect. I went back to a hard-coded list of known types that I only highlight when a type annotation context is detected.

The parsing logic is much more simple, and it does a reasonable job for known types.
It leaves all the unknown types as is, which I think is a better alternative than trying to parse everything and failing partially for no good reason xx% of the time (from the point of view of a user).

---

I'm awaiting your feedback. BTW there is still the `// TODO: choose one of the alternative definitions:` that should be addressed line 256, if you could tell me which way you prefer.

src/languages/fsharp.js
```diff
@@ -221,16 +225,134 @@ export default function(hljs) {
...
+        COMMENT,
+        GENERIC_TYPE_SYMBOL,
+        hljs.inherit(QUOTED_IDENTIFIER, { scope: null }), // match to avoid strange patterns inside that may break the parsing
+        OPERATOR
+      ]
+    };
+  }
+
+  const TYPE_ANNOTATION = makeTypeAnnotationMode(/:/, 'operator');
+  // DU stands for Discriminated Union
```
joshgoebel, line 322: Lets spell it out vs hiding it in a comment. :-)  The comment doesn't exist at the reference point so someone is still left asking "what's a DU".

src/languages/fsharp.js
```diff
@@ -221,16 +225,134 @@ export default function(hljs) {
...
+      hljs.inherit(QUOTED_IDENTIFIER, {
+        scope: null // Use parent scope
+      }),
+      // Cannot contain another apostrophe
+      hljs.inherit(hljs.UNDERSCORE_TITLE_MODE, {
+        scope: null // Use parent scope
+      })
+    ],
+    relevance: 0
+  };
```
joshgoebel, line 251: I think this has gotten too complex... we're trying to match single identifiers here, yes?  Are there tests for these?

If so, then we probably need a single rule or a `variant` for each perhaps... to match the entire type symbol (quoted or unquoted)... we don't need to rely on `contains` for this as contains can match MULTIPLE times which can lead to false positives in other languages or incorrect parsing:

For example what I think you meant to allow:

- `^[quoted identifier]`
- `^[underscore title]`

But what you actually allow:

```
^[quoted identifier][quoted identifier][underscore_title][quoted identifier]
```

This can be solved with `endsParent` also but I think what we likely need here is just to break out the two cases individually without resorting to children modes.

joshgoebel:
Sorry to only comment on small stuff for now, I think this is looking better overall but I'll need more time to more carefully review it later.

src/languages/fsharp.js
mlaily: Ok will do. I thought it was getting a bit too long, hence the short version.

src/languages/fsharp.js
mlaily: Indeed, you are right. This mode should only match one identifier syntax, prefixed with `'` or '^'.

There are tests in test/markup/fsharp/types.txt and also in other markup tests since it's often used, but I didn't try to break it with invalid edge cases.

If I use `variants` I can't use `begin` in the parent mode though.

Is there a way to reuse the existing `UNDERSCORE_TITLE_MODE` and `QUOTED_IDENTIFIER` and indicate they should be prefixed with another regex? Or did you mean I should actually not reuse these and rewrite a complete regex for both those variants?

joshgoebel: > Is there a way to reuse the existing UNDERSCORE_TITLE_MODE and QUOTED_IDENTIFIER and indicate they should be prefixed with another regex?

Just reference their attributes directly:

```
QUOTED_IDENTIFIER.match
```

And be consistent with `match/begin`... if there is only a single expression (no end), prefer `match`.

joshgoebel: > Something like this?:

Exactly.

> but I didn't try to break it with invalid edge cases.

I wasn't suggesting that, only looking for positive examples in a single file, but I didn't see one.

src/languages/fsharp.js
```diff
@@ -221,16 +225,130 @@ export default function(hljs) {
...
+  // List or symbolic operator characters from the FSharp Spec 4.1, minus the dot, and with `?` added, used for nullable operators.
+  const OPERATOR_CHARS = Array.from("!%&*+-/<=>@^|~?");
+  const OPERATOR_CHAR_RE = regex.concat('[', ...OPERATOR_CHARS.map(regex.escape), ']');
+  // TODO: choose one of the alternative definitions:
+  // const OPERATOR_CHAR_RE = regex.either(...OPERATOR_CHARS.map(regex.escape));
+  // const OPERATOR_CHAR_RE = /[!%&*+\-\/<=>@^|~?]/;
+  // The lone dot operator is special. It cannot be redefined, and we don't want to highlight it. It can be used as part of a multi-chars operator though.
+  const OPERATOR_CHAR_OR_DOT_RE = regex.either(OPERATOR_CHAR_RE, /\./);
+  // When a dot is present, it must be followed by another operator char:
+  const OPERATOR_FIRST_CHAR_OF_MULTIPLE_RE = regex.concat(OPERATOR_CHAR_OR_DOT_RE, regex.lookahead(OPERATOR_CHAR_OR_DOT_RE));
```
joshgoebel, line 259: So is `..` intended to be valid here?

mlaily: The second part of `OPERATOR_FIRST_CHAR_OF_MULTIPLE_RE ` is a look-ahead, but the goal is to match `.` when followed by any operator char, so `..` must be matched yes. (It's the built-in operator to define ranges, e.g. `[1..10]`)

src/languages/fsharp.js
```diff
@@ -221,16 +225,130 @@ export default function(hljs) {
...
+        match: /<|>/
+      },
+      TYPE_ANNOTATION // generic types can have constraints, which are type annotations. e.g. type MyType<'T when 'T : delegate<obj * string>> =
+    ]
+  };
+
   const COMPUTATION_EXPRESSION = {
     // computation expressions:
     scope: 'computation-expression',
+    // BUG: might conflict with record deconstruction. e.g. let f { Name = name } = name // will highlight f
```
joshgoebel, line 351: Can we just capture this edge case with a separate rule to handle `let [ident] {` ?

mlaily: We certainly could, but it's only one edge case among others.
Examples:
```fsharp
type Record = { Prop: int }
// f2 is a function with 3 parameters (a, b, c). a and c are both extracted from records
let private f2 { Prop = a } b { Prop = c } = () // `private` can be used to the function private...
let value = 42
let x = f2 { Prop = 42 } value { Prop = 0 } // Call f2 with some values
```
With the current grammar it will highlight as follows: 
image
(Ideally, only the words `type`, `int`, `let` and `private` should be in red)

I've seen other projects use a list of known computation expressions to avoid this problem, but IMO it's not very good either: anyone can create a new one and there is no consensus on a list of "official" computer expressions...

Even knowing all these shortcomings, I still feel like it's preferable to continue using the current regex to highlight `identifier {`.
I don't think it's warranted to make an exception just for `let ident {`.

What do you think?

joshgoebel: > Even knowing all these shortcomings, I still feel like it's preferable ...

But why?  

I see you refer to these as "edge cases", is that strictly true?

mlaily: > I see you refer to these as "edge cases", is that strictly true?

It's an edge case in the sense that computation expressions are probably more frequent than something followed by a record (which wrongly triggers the computation expression highlighting).

I think you started referring to this bug as an edge case in your first comment :p

> But why?

Because I think it's important to highlight computation expressions, even if it means there are false positives, and because I think it's better to have failures that are consistent (at least the user can try to figure out why it breaks and use a different syntax).
But all things considered — and since it appears we can't really disambiguate between a computation expression and something followed by a record — it might be better to drop support for computation expression highlighting altogether... Would that be preferable?

(Note the problem was already present before the current PR)

joshgoebel: > (Note the problem was already present before the current PR)

Since you point this out we'll leave it for now and see how it goes... feels like it's right on the border though.

test/markup/fsharp/operators.expect.txt
```diff
@@ -1,28 +1,77 @@
...
+
+<span class="hljs-comment">// Code quotation</span>
 
 <span class="hljs-keyword">open</span> Microsoft.FSharp.Quotations
 <span class="hljs-comment">// A typed code quotation.</span>
-<span class="hljs-keyword">let</span> expr : Expr&lt;<span class="hljs-type">int</span>&gt; = &lt;@ <span class="hljs-number">1</span> + <span class="hljs-number">1</span> @&gt;
+<span class="hljs-keyword">let</span> expr <span class="hljs-operator">:</span> Expr<span class="hljs-operator">&lt;</span><span class="hljs-type">int</span><span class="hljs-operator">&gt;</span> <span class="hljs-operator">=</span> <span class="hljs-operator">&lt;@</span> 1 <span class="hljs-operator">+</span> 1 <span class="hljs-operator">@&gt;</span>
```
joshgoebel, line 65: What happened to `number` here getting lost?

mlaily: Oops! Well spotted, this is a little embarrassing...

This should be fixed now.

mlaily:
Hello,

Sorry for the delay.
I missed some github notifications and didn't realize I had things to change before this PR could be merged!

It should be ok now, hopefully...

src/languages/fsharp.js
```diff
@@ -247,35 +247,43 @@ export default function(hljs) {
...
+        /:=/, // Reference cell assignment
+        /::?/, // : or ::
+        /\$/), // A single $ can be used as an operator
+      relevance: 0
+    };
+  }
+
+  const OPERATOR = makeOperatorMode(true);
+  // This variant is used when matching '=' should end a parent mode:
+  const OPERATOR_WITHOUT_EQUAL = makeOperatorMode(false);
```
joshgoebel, line 286: `true` and `false` have no meaning (in context) and as such are difficult to read... use named arguments/object literals instead:

```
const OPERATOR = makeOperatorMode({ includeEqual: false}); 
```

mlaily: Ok!

It's the first time I use destructuring like that, so let me know if I did something wrong.

joshgoebel:
@mlaily  Thanks for all the time and effort!

joshgoebel:
Can you confirm your change list (in the PR notes) is correct [or update it] so I can use that as the content for the squashed commit comments...

joshgoebel (approved):

mlaily:
> Can you confirm your change list (in the PR notes) is correct [or update it] so I can use that as the content for the squashed commit comments...

If you are referring to the description of the PR/the first comment, I reviewed and updated it.

> @mlaily Thanks for all the time and effort!

Thank you very much for reviewing this PR. I'm sorry it ended up this big and took so many back and forths.

PR #3371, opened 2021-10-25
(docs) make demo page more keyboard accessible

Hirse:
### Changes
Adds basic keyboard accessibility to the [demo page](https://highlightjs.org/static/demo) by adding `a`-tags to react to keydown events.

hljs-demo-keyboard

### Checklist
- [ ] ~Added markup tests~, or they don't apply here because no highlight changes
- [ ] ~Updated the changelog at `CHANGES.md`~ No published changes

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -1,20 +1,13 @@
 hljs.debugMode();
 hljs.highlightAll();
 
-const addEventListeners = (element, listener) => {
-  element.addEventListener("click", listener);
-  element.addEventListener("keydown", (event) => {
-    if (event.key === "Enter") {
-      listener(event);
-    }
-  });
-};
+document.querySelectorAll(".categories a").forEach((category) => {
+  category.addEventListener("click", (event) => {
+    event.preventDefault();
 
-document.querySelectorAll(".categories > li").forEach((category) => {
-  addEventListeners(category, (event) => {
     const current = document.querySelector(".categories .current");
-    const currentCategory = current.dataset.category;
-    const nextCategory = event.target.dataset.category;
+    const currentCategory = current.name;
+    const nextCategory = event.target.name;
 
     if (currentCategory !== nextCategory) {
       current.classList.remove("current");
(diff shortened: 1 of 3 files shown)


joshgoebel:
> The onclick event handler (and click event) is triggered when the mouse is pressed and released when over a page element or when the Enter key is pressed while a keyboard-navigable element has focus. 

Shouldn't the existing `onclick` handler just work?  A target of "#" could be added if necessary...

Hirse:
Added `<a>`-tags and removed custom keydown handler.

demo/demo.js
```diff
@@ -1,20 +1,13 @@
...
+  category.addEventListener("click", (event) => {
+    event.preventDefault();
 
-document.querySelectorAll(".categories > li").forEach((category) => {
-  addEventListeners(category, (event) => {
     const current = document.querySelector(".categories .current");
-    const currentCategory = current.dataset.category;
-    const nextCategory = event.target.dataset.category;
+    const currentCategory = current.name;
+    const nextCategory = event.target.name;
```
joshgoebel, line 10: Using dataset seems entirely appropriate here to name the type of data vs using the generic HTML `name`... was there a reason for this change?

joshgoebel:
I really like that all the tabindex=0 disappears, feels like a step in the correct direction.

demo/demo.js
```diff
@@ -23,15 +25,21 @@ document.querySelectorAll(".categories > li").forEach((category) => {
   });
 });
 
-document.querySelectorAll(".styles > li").forEach((style) => {
+document.querySelectorAll(".styles a").forEach((style) => {
```
joshgoebel, line 28:
```suggestion
document.querySelectorAll(".styles li a").forEach((style) => {
```

More descriptive as these are still a list.

joshgoebel: Or perhaps just `ul.styles` ?

demo/demo.js
Hirse: No, just copying from my testing. :D

demo/demo.js
Hirse: Changed to your first suggestion.

joshgoebel (approved):

PR #3374, opened 2021-10-26
fixed type of `either` in `src/lib/regex.js`

jaacko-torus:
and added types to `types.d.ts`

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -23,6 +23,15 @@ declare module 'highlight.js' {
         install: (vue: any) => void
     }
 
+    // perhaps make this an interface?
+    type RegexOptions = {
+        capture?: boolean
+    }
+
+    type RegexEitherArgs =
+        (RegExp | string)[] | [RegexOptions] |
+        [...(RegExp | string)[], RegexOptions]
+
     interface PublicApi {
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -23,6 +23,15 @@ declare module 'highlight.js' {
...
     }
 
+    // perhaps make this an interface?
+    type RegexOptions = {
+        capture?: boolean
+    }
+
+    type RegexEitherArgs =
+        (RegExp | string)[] | [RegexOptions] |
+        [...(RegExp | string)[], RegexOptions]
```
joshgoebel, line 33: You can't have just options... so minimally this should be:

```
  (RegExp | string)[] | 
  [...(RegExp | string)[], RegexEitherOptions]
```

types/index.d.ts
```diff
@@ -23,6 +23,15 @@ declare module 'highlight.js' {
         install: (vue: any) => void
     }
 
+    // perhaps make this an interface?
+    type RegexOptions = {
```
joshgoebel, line 27:
```suggestion
    type RegexEitherOptions = {
```

types/index.d.ts
joshgoebel: And we can probably just inline this type, I don't think we need a custom type for a single function.

jaacko-torus:
That should fix it :D

joshgoebel (approved):

PR #3375, opened 2021-10-27
Cleanup spacing and other styles of demo

Hirse:
Before 
old

After
new

### Changes
* Rework and align spacing
* Align padding of code blocks with theme padding (and remove schoolbook padding)
* Give lists a bit more width to have fewer line breaks.

### Checklist
- [ ] ~Added markup tests~, or they don't apply here because no markup changes
- [ ] ~Updated the changelog at `CHANGES.md`~ only website changes

### DIFF
--- a/demo/style.css
+++ b/demo/style.css
@@ -97,7 +97,7 @@ aside ul::-webkit-scrollbar-thumb:active {
 aside ul a {
   display: block;
   line-height: 1.375em;
-  padding: 1px 0.5em 1px 1em;
+  padding: 0 0.5em 0 1em;
   position: relative;
   text-decoration: none;
 }
--- a/demo/style.css
+++ b/demo/style.css
@@ -67,7 +67,6 @@ aside ul {
 aside ul {
   margin: 0 0 1.5em;
   overflow-y: auto;
-  position: relative;
   scrollbar-width: thin;
 }
 
@@ -96,8 +95,10 @@ aside ul::-webkit-scrollbar-thumb:active {
 }
 
 aside li {
-  padding: 1px 0.5em 1px 1em;
   cursor: pointer;
+  line-height: 1.375em;
+  padding: 0 0.5em 0 1em;
+  position: relative;
 }
 
 aside li:hover {
--- a/demo/style.css
+++ b/demo/style.css
@@ -1,27 +1,26 @@
 @import url(https://fonts.googleapis.com/css?family=PT+Sans:400,700,400italic);
 
-/* Base styles */
 body {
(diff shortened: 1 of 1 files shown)


joshgoebel (approved):

joshgoebel:
Can we line up the arrow better? Looks like it's 1-2px off?

Hirse:
> 
> 
> Can we line up the arrow better? Looks like it's 1-2px off?

Does this look better to you now?
image

joshgoebel:
Better. Are you checking all this in Safari also?

Hirse:
> 
> Better. Are you checking all this in Safari also?

No, I don't have a Mac.

demo/style.css
```diff
@@ -122,13 +116,13 @@ aside ul a.current:before {
...
-  margin-left: 18em;
+  margin-left: 19em;
   min-width: 36em;
-  padding: 1px 4em;
+  padding: 1.5em 3em 0;
 }
 
 main > div {
-  margin: 2.5em 0 3em;
+  margin-bottom: 3em;
```
joshgoebel, line 125: Please revert, the top deserves more margin.

demo/style.css
```diff
@@ -122,13 +116,13 @@ aside ul a.current:before {
 }
 
 main {
-  margin-left: 18em;
+  margin-left: 19em;
   min-width: 36em;
-  padding: 1px 4em;
+  padding: 1.5em 3em 0;
```
joshgoebel, line 121: Or this is possibly the problem... hard to say.

joshgoebel: I don't think the two top titles are equatable in any way so they do not need to "line up" if that is what you were going for.

joshgoebel:
This simply includes "too much" and too intermingled... there seem to be CSS changes other than the ones explicitly documented in the issue text and it's difficult to properly review "looks better to me" PRs like this (hence the delay, but my apologies for that)... often I'll miss something and then just have to revert it later, or someone will break something in Safari because they were only proofing with Chrome etc... I'm also not convinced all of these changes were needed - though it's hard to judge with them all bundled together.

I did try to look at the commit history but it just seems to be a progression rather than "one change per commit", which would make this easier to follow.  If you feel a few of these changes were critical I'd be happy to see a new PR with individual commits (for easy review - to make sure we aren't changing more than intended)... Or also happy to take a second look at this if you wanted to go back and refactor it into a bunch of tiny individual commits detailing each change and the thinking behind it.

One quick example making the example font larger isn't a terrible idea... but that could be decided separately from other layout changes, etc...

PR #3377, opened 2021-10-29
hljs playground

Hirse:
Adding a playground as a public version of `tools/developer.html` to be added to the website.
Currently does not include the localstorage, so not a complete replacement.
Generated using the same process as the demo page.
Working version (using CDN sources): https://codesandbox.io/s/hljs-playground-jomgi

playground

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -89,7 +89,7 @@ module.exports = {
       }
     },
     {
-      files: ["demo/**/*.js"],
+      files: ["demo/**/*.js", "playground/**/*.js"],
       globals: {
         hljs: "readonly"
       }
--- a/playground/index.html
+++ b/playground/index.html
@@ -1,0 +1,58 @@
+<!DOCTYPE html>
+<html lang="en">
+<!-- This file is intended to be built using the instructions here: http://highlightjs.readthedocs.io/en/latest/building-testing.html#building -->
+<head>
+  <meta charset="utf-8" />
+  <title>highlight.js Playground</title>
+  <link rel="shortcut icon" type="image/png" href="https://highlightjs.org/static/favicon.png" />
+  <link rel="stylesheet" href="style.css" />
+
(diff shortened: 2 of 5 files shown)


joshgoebel:
This type of idea has come up in the past - I also think we already have our own (much simpler) take on this for the new website (that's not yet published).  I also have concerns whether we want this [the FULL utility] to be user facing for ALL users of the library - whether that's it's own support burden or not.  I feel like at the very least if we went this direction (fully public tool) it would need it's own small dedicated page of docs, etc.

I'll discuss with @highlightjs/core.

joshgoebel:
Another idea (if we go this more ambitious route) is constantly saving the code into the URL (vs local storage) as a `#hashtag` so that problematic code can be tested, then the URL copied right into a GitHub issue, etc... which would link to the developer tool with that snippet already loaded (and the correct language, theme, etc)... might reduce the need for the JSFiddle we currently use for such things.

Hirse:
> Another idea (if we go this more ambitious route) is constantly saving the code into the URL (vs local storage) as a `#hashtag` so that problematic code can be tested, then the URL copied right into a GitHub issue, etc... which would link to the developer tool with that snippet already loaded (and the correct language, theme, etc)... might reduce the need for the JSFiddle we currently use for such things.

Yes, I had that idea too, but wanted to showcase my current version before spending more time on it.

I'd be happy to work more on it if you decide to go this direction.

joshgoebel:
I think the new website makes this redundant, yes?

https://highlightjs-web.herokuapp.com/demo

Hirse:
> I think the new website makes this redundant, yes?
> 
> https://highlightjs-web.herokuapp.com/demo

Yeah, looks like it. :)

joshgoebel:
Sorry this took so long to resolve - please don't feel bad. (I'm glad you're smiling)  If you see room for improvement on the new site's similar feature I'm sure we'd be open to ideas/PRs.

PR #3380, opened 2021-10-31
(chore) apply lint to our tooling scripts also

Hirse:
Add linting for `tools/**/*.js` with the exception of the `vendor` folder.

Does not change any rules, but fixes an ESLint style violation in the `.eslintrc.js` file itself.

Most changes were autofixed, I only manually added `.js` file endings, changed `==` to `===`, and removed two unused imports.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,9 +28,10 @@ module.exports = {
     // for now ignore diff between types of quoting
     quotes: "off",
     // this is the style we are already using
-    "operator-linebreak": ["error", "before", { overrides: {
-      "=": "after"
-    }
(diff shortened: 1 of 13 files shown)


joshgoebel:
I'm not sure I want our CI to fail because linting of `tools` failed, but we'll give it a shot and see if it proves annoying in practice.

joshgoebel (approved):

Hirse:
@joshgoebel Could you also add `hacktoberfest-accepted` here? 
Since highlight.js doesn't have the `hacktoberfest` tag itself (any more), merged PRs don't count by themselves. :(

PR #3382, opened 2021-11-01
enh(swift) add SE-0290 unavailability condition

bradleymackey:
New language feature implemented in Swift 5.6, this conforms to the specification described in [SE-0290](https://github.com/apple/swift-evolution/blob/main/proposals/0290-negative-availability.md). Note this only includes `#unavailable`, not `@unavailable`.

### Changes

- Adds `#unavailable` keyword, using the same keyword rules as existing availability attribute.

### Checklist

### DIFF
--- a/test/markup/swift/availability.expect.txt
+++ b/test/markup/swift/availability.expect.txt
@@ -1,0 +1,11 @@
+<span class="hljs-keyword">#available</span>()
+<span class="hljs-keyword">#available</span>(<span class="hljs-keyword">iOS</span> <span class="hljs-number">15.0</span>, <span class="hljs-operator">*</span>)
+<span class="hljs-keyword">@available</span>()
+<span class="hljs-keyword">@available</span>(<span class="hljs-keyword">iOS</span> <span class="hljs-number">15.0</span>, <span class="hljs-operator">*</span>)
+<span class="hljs-keyword">@available</span>(<span class="hljs-keyword">iOS</span> <span class="hljs-number">14</span>, deprecated: <span class="hljs-string">&quot;reason&quot;</span>, <span class="hljs-operator">*</span>)
+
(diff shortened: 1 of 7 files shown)


test/markup/swift/availability.expect.txt
```diff
@@ -0,0 +1,11 @@
...
+<span class="hljs-keyword">#available</span>(<span class="hljs-keyword">iOS</span> <span class="hljs-number">15.0</span>, <span class="hljs-operator">*</span>)
+<span class="hljs-keyword">@available</span>()
+<span class="hljs-keyword">@available</span>(<span class="hljs-keyword">iOS</span> <span class="hljs-number">15.0</span>, <span class="hljs-operator">*</span>)
+<span class="hljs-keyword">@available</span>(<span class="hljs-keyword">iOS</span> <span class="hljs-number">14</span>, deprecated: <span class="hljs-string">&quot;reason&quot;</span>, <span class="hljs-operator">*</span>)
+
+<span class="hljs-keyword">#unavailable</span>()
+<span class="hljs-keyword">#unavailable</span>(<span class="hljs-keyword">iOS</span> <span class="hljs-number">15.0</span>, <span class="hljs-operator">*</span>)
+<span class="hljs-comment">// not a keyword</span>
+<span class="hljs-meta">@unavailable</span>()
+<span class="hljs-meta">@unavailable</span>(iOS <span class="hljs-number">15.0</span>, <span class="hljs-operator">*</span>)
```
joshgoebel, line 11: Is this still the expected behavior then, this is acting as a user defined attribute?

bradleymackey: I haven't tested it (it may give a compiler error) but I think it's a sensible to fallback to this behaviour as it's simply not defined in the proposal.

joshgoebel (approved):

PR #3386, opened 2021-11-02
enh(java) add `sealed` and `non-sealed` keywords

### LINKED ISSUE #3346: (Java) Java 17 sealed keywords not implemented
With the release of Java 17 being an **LTS**, it seems proper to support the new `sealed` keyword.
However, the new `non-sealed` keyword brings some issues, that [while I have attempted to fix](https://github.com/thiccaxe/highlight.js/commit/e285198f222a48a363338b54e3c077b6a18db18d), seem best suited for someone more familiar with the highlightjs library.

[sample js fiddle](https://jsfiddle.net/Lv2hmbpt/2/)

**Expected behavior**
the `sealed` keyword should be highlighted, and as more hyphenated keywords may come to java in the future, some structure
 may be useful for `non-sealed` and the like.

bradleymackey:
Adds support for these missing keywords, using a different matching pattern for hyphenated keywords as a workaround. [Source for suggested workaround](https://github.com/thiccaxe/highlight.js/commit/e285198f222a48a363338b54e3c077b6a18db18d#r57401045).

### Changes

- Adds `sealed` and `non-sealed` keywords
- Fixes #3346

### Checklist

### DIFF
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -73,7 +73,8 @@ export default function(hljs) {
     'module',
     'requires',
     'exports',
-    'do'
+    'do',
+    'sealed'
   ];
 
   const BUILT_INS = [
@@ -177,6 +178,15 @@ export default function(hljs) {
         className: {
           1: "keyword",
           3: "title.class"
+        }
+      },
+      {
+        // Hypenated keywords
+        match: [
+          /non-sealed/
+        ],
+        className: {
+          1: "keyword"
         }
       },
       {
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -181,7 +181,7 @@ export default function(hljs) {
         }
       },
       {
-        // Hypenated keywords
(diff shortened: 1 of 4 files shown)


src/languages/java.js
```diff
@@ -179,6 +180,15 @@ export default function(hljs) {
...
       },
+      {
+        // Exceptions for hyphenated keywords
+        match: [
+          /non-sealed/
+        ],
+        className: {
+          1: "keyword"
+        }
+      },
```
joshgoebel, line 191: Lets use the simple form?
```suggestion
        match:  /non-sealed/,
        scope: "keyword"
      },
```

joshgoebel (approved):
Looks good but lets make the one change suggested.

src/languages/java.js
bradleymackey: Got it.

joshgoebel (approved):

PR #3396, opened 2021-11-09
(chore) add `sideEffects: false` to enable tree-shaking in esbuild & others

shakyShane:
I have an application where I want to run code highlighting in Node, like this: 

```ts
import hljs from "highlight.js";

export const loader = async () => {
  return hljs.highlight("console.log('hello world!')", { language:"ts" }).value;
}
```

Part of the build process then strips `loader` so that it's not present in the browser build. Without `sideEffects: false`, bundlers cannot be sure if this library can be stripped during the build.

### Checklist


joshgoebel:
I think I follow this but would love feedback from someone else before we merge.

joshgoebel:
> Part of the build process then strips loader so that it's not present 

Why - `hljs` is obviously referenced - is `loader` not?  Wouldn't the default value of `sideEffects` be `true` (for compatibility)... and isn't sideEffects false going to lead to more stripping than the opposite?  I'm not understanding this completely.

joshgoebel:
Marking this as draft until we have someone else to review it and provide more info.

joshgoebel (approved):

joshgoebel:
This will land in 11.5.

PR #3397, opened 2021-11-10
fix(clj) Fix broken/inconsistent highlighting for `$`, Numbers, namespaced maps, HINT mode. Add charachter, regex and punctuation modes.

MrEbbinghaus:
Hey, it's me again. 

I added some more fixes to Clojure highlighting. Sorry if the scope of this PR is too big, but I couldn't be bothered to open 7 different PRs. :-)

Here is a write-down in your format. I don't know if I should add this to the `CHANGES.md` in this format? Or everything together with my last line into one `enh(clj) Fixup a lot of things.`?
```
fix(clj) `$` in symbol breaks highlighting
fix(clj) Add complete regex for number detection
enh(clj) Add character mode for character literals
fix(clj) Inconsistent namespaced map highlighting
enh(clj) Add `regex` mode to regex literal
fix(clj) Remove inconsistent/broken highlighting for metadata
enh(clj) Add `punctuation` mode for commas.
```
### Checklist
- [x] Added markup tests.
----

Following is the list of stuff this PR fixes, together with reference images and my reasoning.

## `$` in symbol breaks highlighting
Clojure symbols can contain `$`. I fixed it.
Current:
image

GitHub:
```clojure
(.getEnclosingClass java.util.Map$Entry)
(java.util.Map$Entry. .getEnclosingClass)
```

## Number highlighting just matched integer.
This is how valid Clojure numbers are highlighted currently:
image

I added a regex to include them all.

GitHub does it wrong as well:
```clojure
; integer ; BigInt
00          42N
42          0N

; octal   ; hex
052       0x2a
00N       0x0N

; radix.  ; radix BigInt
2r101010  2r101010N
8r52      8r52N
16r2a     16r2aN
36r16     36r16N


;; ratios
1/2
-1/2

;; floats
42.0
-42.0
42.

; BigDecimal ; with Exponent
42.0M        42.0E2
-42M         -42.0E+9
42.M         42E-0
42M          42.0E2M
             42E+9M
```

## Missing character literals
Clojure has different character literals, they were highlighted not at all or wrong. I added the `character` mode to them.
Current:
image

## Namespaced map keys highlighting inconsistent with itself and other highlighters
Current:
`#:ship` and `#:person` have different colours. 
image

I changed it to be always not coloured. 

GitHub:
```clojure
#:person{:first "Han"
         :last "Solo"
         :ship #:ship{:name "Millennium Falcon"}}
#::{:a 1, :b 2}
```
Cursive for IntelliJ: 
image
Calva for VS Code highlights them in the colour of the rainbow parentheses:
image

## Add Regex Highlighting
Clojure has a regex literal `#"My\sRegex!"`, I added the `regex` scope for this.

GitHub:
```clojure
#"\bMy\sRegex!\b"   ; Special regex stuff is highlighted
# "\bMy\sRegex!\b"  ; # and string have different highlighting.
```

## Remove HINT mode.
Metadata (which can be a keyword, symbol, collection, ...) was highlighted like a comment, and it wasn't consistent.
image

I removed this mode. You can have huge nested maps of metadata if you like. Removing highlighting from them doesn't make sense.

GitHub:
```clojure
; different highlighting for meta data
^String
^{:tag String}
^{:tag java.lang.String}
^{:any ["collection" {:I #{:like}}]}

^:private
^{:private true}

; deprecated but still working:
#^String
#^{:tag String}
#^:private
```
Cursive: 
image
Calva:
image

## Add `punctuation` scope for `,`
In Clojure, the comma `,`  is treated like a whitespace. It is sometimes used to make compact maps more readable:
`{:a 1, :b 2}` is equal to: `{:a 1 :b 2}` 
I have seen editor themes that made them slightly opaque.


test/markup/clojure/globals_definition.txt
```diff
@@ -6,7 +6,7 @@
 (defn clojure-function [args]
   (let [string   "multiline\nstring"
         regexp   #"regexp"
-        number   100,000
+        number   100000
```
joshgoebel, line 9: Is this just invalid and it's wrong that we've always had it here or what?

src/languages/clojure.js
```diff
@@ -8,8 +8,8 @@ Category: lisp
 
 /** @type LanguageFn */
 export default function(hljs) {
-  const SYMBOLSTART = 'a-zA-Z_\\-!.?+*=<>&#\'';
-  const SYMBOL_RE = '[' + SYMBOLSTART + '][' + SYMBOLSTART + '0-9/;:]*';
+  const SYMBOLSTART = 'a-zA-Z_\\-!.?+*=<>&\'';
```
joshgoebel, line 11: `#` meant to be removed?

src/languages/clojure.js
```diff
@@ -45,15 +45,15 @@ export default function(hljs) {
       'lazy-seq spread list* str find-keyword keyword symbol gensym force rationalize'
   };
 
-  const SIMPLE_NUMBER_RE = '[-+]?\\d+(\\.\\d+)?';
+  const NUMBER_RE = /[-+]?(((0[xX][0-9a-fA-F]+|0[0-7]+|[1-9][0-9]?[rR][0-9a-zA-Z]+)N?)|[0-9]+(\/[0-9]+|N|((\.[0-9]*)?([eE][+-]?[0-9]+)?M?)))/
```
joshgoebel, line 48: Please break this up into separate regex using `variants` as seen in other grammars, like `javascript`, etc...

joshgoebel:
> Remove HINT mode.

What is the purpose of metadata hints in Clojure?

joshgoebel:
>  I don't know if I should add this to the CHANGES.md

Please, the more detail the better though if you did a "lots of fixes" with a second level bullet list under that that would be even better.

test/markup/clojure/globals_definition.txt
MrEbbinghaus: , is treated like a whitespace.
So this is actually 100 000, two separate numbers. (let) needs an even number of forms, with binding targets on the left.

There is no thousands separator in  Clojure.

src/languages/clojure.js
MrEbbinghaus: I readded it in line 12 because it interfered with my fix for namespaced maps.

```js
const SYMBOL_RE = '[#]?[' + SYMBOLSTART + '][' + SYMBOLSTART + '0-9/;:$#]*';
```

If a symbol begins with `#` it can't be followed by a number, and it can't stand alone.

`#::{}`, `#::foo{}`, `#:foo{}` are valid maps. `#::` shouldn't be matched as a symbol, but the start of a namespaced map. 


The "SYMBOL" mode matches way more than the Clojure term "symbol", maybe we should split it into two? 
symbols and specials? 

> **Symbols**
Symbols begin with a non-numeric character and can contain alphanumeric characters and *, +, !, -, _, ', ?, <, > and = (other characters may be allowed eventually).
>
> '/' has special meaning, it can be used once in the middle of a symbol to separate the namespace from the name, e.g. my-namespace/foo. '/' by itself names the division function.
>
>'.' has special meaning - it can be used one or more times in the middle of a symbol to designate a fully-qualified class name, e.g. java.util.BitSet, or in namespace names. Symbols beginning or ending with '.' are reserved by Clojure. Symbols containing / or . are said to be 'qualified'.
>
> Symbols beginning or ending with ':' are reserved by Clojure. A symbol can contain one or more non-repeating ':'s.

https://clojure.org/reference/reader#_symbols

src/languages/clojure.js
MrEbbinghaus: I initially did, but I couldn't get an order where everything matched like it should. 
Maybe negative lookaheads resolve that problem, though. 
I will look at this again.

MrEbbinghaus:
> What is the purpose of metadata hints in Clojure?

General meta information. Anything really.

The `test` function in clojure.core executes the `:test` meta key.

```clojure
(defn my-function
  "this function adds two numbers"
  {:test #(do
            (assert (= (my-function 2 3) 5))
            (assert (= (my-function 4 4) 8)))}
  ([x y] (+ x y)))
```

clojure.core uses it for documentation and compiler hints. 
image

image


CIDER the Clojure Emacs IDE uses it for breakpoints.
```clojure
(dotimes [i 10]
  #dbg ^{:break/when (= i 7)}
  (prn i))
```

[koacha](https://github.com/lambdaisland/kaocha), a test runner, uses them to focus on specific sets of tests 
```clojure
(deftest ^:yyy other-test
  (is (= 3 3)))
```

[meta-merge](https://github.com/weavejester/meta-merge) uses keys in metadata to specify per-key merge behaviour.

```clojure
(meta-merge {:a [:b :c]} {:a ^:prepend [:d]})
; => {:a [:d :b :c]}
(meta-merge {:a [:b :c]} {:a ^:replace [:d]})
; => {:a [:d]}
(meta-merge {:a [:b :c]} {:a ^:displace [:d]})
; => {:a [:b :c]}
```




https://clojure.org/reference/reader#_metadata
> ## Metadata (^)
> Metadata is a map associated with some kinds of objects: Symbols, Lists, Vector, Sets, Maps, tagged literals returning an IMeta, and record, type, and constructor calls. The metadata reader macro first reads the metadata and attaches it to the next form read (see with-meta to attach meta to an object):
> ^{:a 1 :b 2} [1 2 3] yields the vector [1 2 3] with a metadata map of {:a 1 :b 2}.
> 
> A shorthand version allows the metadata to be a simple symbol or string, in which case it is treated as a single entry map with a key of :tag and a value of the (resolved) symbol or string, e.g.:
> ^String x is the same as ^{:tag java.lang.String} x
> 
> Such tags can be used to convey type information to the compiler.
> 
> Another shorthand version allows the metadata to be a keyword, in which case it is treated as a single entry map with a key of the keyword and a value of true, e.g.:
> ^:dynamic x is the same as ^{:dynamic true} x
> 
> Metadata can be chained in which case they are merged from right to left.

CHANGES.md
```diff
@@ -5,7 +5,15 @@ Grammars:
...
-- fix(clojure) `comment` macro catches more than it should [Björn Ebbinghaus][]
+- fix(clojure) Several issues with Clojure highlighting (#3397) [Björn Ebbinghaus][]
+  - fix(clojure) `comment` macro catches more than it should (#3395)
+  - fix(clojure) `$` in symbol breaks highlighting
+  - fix(clojure) Add complete regex for number detection
+  - enh(clojure) Add character mode for character literals
+  - fix(clojure) Inconsistent namespaced map highlighting
+  - enh(clojure) Add `regex` mode to regex literal
+  - fix(clojure) Remove inconsistent/broken highlighting for metadata
+  - enh(clojure) Add `punctuation` mode for commas.
```
MrEbbinghaus, line 16: @joshgoebel Like so?

src/languages/clojure.js
joshgoebel: The order should be the same as they appear inside the larger regex as the variants will eventually get recompiled that way anyways... but for maintenance/readability it's much simpler to split these out.

src/languages/clojure.js
joshgoebel: This PR is big enough but if you want to circle back later with more improvements. :)

src/languages/clojure.js
```diff
@@ -45,20 +45,44 @@ export default function(hljs) {
...
   };
+  const CHARACTER = {
+    scope: 'character',
+    variants: [
+      {match: /\\o[0-3]?[0-7]{1,2}/},                             // Unicode Octal 0 - 377
+      {match: /\\u[0-9a-fA-F]{4}/},                               // Unicode Hex 0000 - FFFF
+      {match: /\\(newline|space|tab|formfeed|backspace|return)/}, // special characters
+      {match: /\\\S/}                                             // any non-whitespace char
+    ]
+  }
```
MrEbbinghaus, line 72: @joshgoebel I refactored the Number and Character regex to be more maintainable.

joshgoebel: Now how would you write 2 - -3? Is there no confusion because it's lisp like?

MrEbbinghaus:
highlight.js does not have regex languages, does it?

Maybe an idea for the future, as GitHub supports regex highlighting.
```clojure
#"\bMy\sRegex!\b"   ; Special regex stuff is highlighted
"\bMy\sRegex!\b"  
```

But not in all languages 🤔 
```javascript
const RE = /\bMy\sRegex!\b/;
```

src/languages/clojure.js
MrEbbinghaus: Clojure uses prefix notation, and `-` is actually a function like any other. There are no special operators symbols like in other languages. 

```clojure
(- 2 -3)
; or even more
(- 2 -3 +5) ; <- GitHub has broken highlighting. :-)
```

joshgoebel: What does that bottom one do?

joshgoebel:
> Special regex stuff is highlighted

See the history behind #2751.  I'm open to this, but _we do not need (or want) a whole regex grammar_. (there is sadly no good way to reuse it properly at the moment).  If we picked a few key things to focus on and decided to do that within our existing regex support, I think that could be nice.  And it would only involve changing: `REGEXP_MODE` and most every grammar would benefit.

One easy start would be start by moving the existing escapes we recognize into `char.escape` scope.  If you wanted to tackle that separately...

src/languages/clojure.js
```diff
@@ -45,20 +45,44 @@ export default function(hljs) {
...
+      {match: /[-+]?([1-9][0-9]*|0)N?/},                             // int (don't match leading 0) // 42 42N
+    ]
   };
+  const CHARACTER = {
+    scope: 'character',
+    variants: [
+      {match: /\\o[0-3]?[0-7]{1,2}/},                             // Unicode Octal 0 - 377
+      {match: /\\u[0-9a-fA-F]{4}/},                               // Unicode Hex 0000 - FFFF
+      {match: /\\(newline|space|tab|formfeed|backspace|return)/}, // special characters
+      {match: /\\\S/}                                             // any non-whitespace char
```
joshgoebel, line 70: This is so tiny a match... can we do any looking forward to qualify it?  I'm imagining false positives with any language that allows `\SOMETHING` style tags...  if we can't qualify it I think we should make it `relevance: 0`.

src/languages/clojure.js
MrEbbinghaus: `(+ 1 2 3)` is like `1 + 2 + 3` in other languages. 

GitHub doesn't recognise numbers beginning with `+`. 

```clojure
42
-42
+42
```

I will send a PR to them as well. :-) While working on highlight.js's parsing, I started to notice errors and edgecases in other highlighters aswell.

joshgoebel:
Looking good, almost there!

src/languages/clojure.js
MrEbbinghaus: It is a literal and can appear anywhere, so you are right with `relevance: 0`. 
I didn't put any thought in this or the other modes, what the `relevance` may be. What is "typically Clojure"?

It would be a nice exercise to generate possible values for modes and try to match them against other languages to automatically calculate the relevance. 🤔

MrEbbinghaus: I added `relevance: 0`.

Although, I am not sure how many programming languages support the full first Unicode plane. `\⠓` is a valid Clojure literal.

joshgoebel (approved):

joshgoebel:
@MrEbbinghaus Thanks so much for help on this!

MrEbbinghaus:
@joshgoebel Thank you for the clear and fast review. :-)

PR #3400, opened 2021-11-13
(chore) add gzip size compression report

### LINKED ISSUE #3097: Task: Add reporting the size of build changes on PRs

Prism does this and it's super cool.  We should start reporting the size of the compressed web build changes with PRs.  Just to start to get a feel for when we increase and shrink the final build size.

https://github.com/preactjs/compressed-size-action

Just the first one I found.  GitHub actions sounds about right.

None.

**Additional context...**

None.

bradleymackey:
This will report the gzip'd size of the CDN build output on every pull request, being run via GitHub actions.
Currently is applied to all Javascript output files.

### Changes

- Add new GitHub actions workflow to report compressed size changes.
- Resolves #3097

### Checklist
- [x] Markup tests not needed

### DIFF
--- a/.github/workflows/compressed_size.yml
+++ b/.github/workflows/compressed_size.yml
@@ -1,0 +1,21 @@
+# This workflow computes the size of a CDN build's output on all Javascript files.
+# Reported file sizes are after the result of gzip compression.
+# Compression action used: https://github.com/preactjs/compressed-size-action
+
+name: CDN Size (gzip)
+
+on: [pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


joshgoebel:
Can we have just a list vs wildcard?  I think what we'd want indivdually:

- /*.min.js
- /es/*.min.js

And perhaps also (and only if the wildcard is aggregated):

- /languages/*.min.js
- /es/languages/*.min.js

Thoughts?

bradleymackey:
I think that can be done with the following wildcard: `./build/{*.min.js,es/*.min.js,languages/*.min.js,es/languages/*.min.js}`

bradleymackey:
Yep it seems to work. See [here for example](https://github.com/bradleymackey/highlight.js/pull/2).

joshgoebel (approved):

PR #3402, opened 2021-11-16
(enh) improve default theme accessibility

joshgoebel:
- first pass at accessibility report in checkTheme
- themes have badges
- light and dark
- resolve CSS color names when necessary
- (enh) default Dark theme is WCAG AAA wrt contrast

### Checklist


tools/checkTheme.js
```diff
@@ -184,6 +191,79 @@ function check_group(group, rules) {
...
+    if (!this.rule.declarations) return false;
+    return this.fg || this.bg;
+  }
+  toString() {
+    return ` ${this.foreground} on ${this.background}`
+  }
+
+  contrastRatio() {
+    if (!this.foreground) return "unknown (no fg)"
+    if (!this.background) return "unknown(no bg)"
```
allejo, line 234:
```suggestion
    if (!this.background) return "unknown (no bg)"
```

Merely for consistency

tools/checkTheme.js
```diff
@@ -184,6 +191,79 @@ function check_group(group, rules) {
...
+  rules.forEach(rule => {
+    var color = new CSSRule(rule, body);
+    if (!color.hasColor) return;
+    table.push([
+      color.contrastRatio(),
+      rule.selectors,
+      color.foreground,
+      color.background
+    ])
+    // console.log(r.selectors[0], color.contrastRatio(), color.toString());
```
allejo, line 262:
```suggestion
```

Left behind on accident?

tools/checkTheme.js
```diff
@@ -184,6 +191,79 @@ function check_group(group, rules) {
...
+    if (!this.foreground) return "unknown (no fg)"
+    if (!this.background) return "unknown(no bg)"
+    return round2(wcagContrast.hex(this.foreground, this.background));
+  }
+}
+
+function contrast_report(rules) {
+  console.log("Accessibility Report".yellow);
+
+  var hljs = rules.find (x => x.selectors && x.selectors.includes(".hljs"));
```
allejo, line 242:
```suggestion
  var hljs = rules.find(x => x.selectors?.includes(".hljs"));
```

nitpick

tools/checkTheme.js
```diff
@@ -184,6 +191,79 @@ function check_group(group, rules) {
...
 }
 
+const round2 = (x) => Math.round(x*100)/100;
+
+class CSSRule {
+  constructor(rule, body) {
+    this.rule = rule;
+    if (rule.declarations) {
+      this.bg = rule.declarations.find(x => x.property =="background")?.value;
+      this.fg = rule.declarations.find(x => x.property =="color")?.value;
```
allejo, line 201: Two notes on this one:

1. Some styles use `background-color` to differentiate the background for specific elements. I think we should add support for checking `background-color` for WCAG contrast.
	https://github.com/highlightjs/highlight.js/blob/0bf3adeecd5ba76cef8ee843e38722089b232da1/src/styles/foundation.css#L74-L77

2. Should we have a minimum Node defined somewhere for dev environments? Optional chaining was added in Node 14 and our `package.json` requires Node 12+ (correctly so, for our users).

tools/checkTheme.js
```diff
@@ -184,6 +191,79 @@ function check_group(group, rules) {
...
+    return this.fg || this.bg;
+  }
+  toString() {
+    return ` ${this.foreground} on ${this.background}`
+  }
+
+  contrastRatio() {
+    if (!this.foreground) return "unknown (no fg)"
+    if (!this.background) return "unknown(no bg)"
+    return round2(wcagContrast.hex(this.foreground, this.background));
```
allejo, line 235: We have some themes that use `rgba()`, will `.hex()` work on those values?

https://github.com/highlightjs/highlight.js/blob/0bf3adeecd5ba76cef8ee843e38722089b232da1/src/styles/arduino-light.css#L52-L54

allejo (commented):
Because I dealt with this in highlight.php, just some nitpicks if we want this tool to be robust enough to scan all of our themes.

tools/checkTheme.js
joshgoebel: I don't strictly adhere to "comment policing"... if something was useful for debugging might be so in the future it doesn't bother me if a comment lives on until it's removed in a future pass.

tools/checkTheme.js
joshgoebel: > Should we have a minimum Node defined somewhere for dev environments? 

Ah, we probably should, but where... damn old versions of Node. :-)

> Some styles use background-color  

I'll break this out into a separate issue... the tool doesn't have to be perfect for the main thrust of this to be mergeable.

tools/checkTheme.js
joshgoebel: As you point out this isn't supported everywhere so I'm going to slow down with it until we figure our how/where to specify that dev work requires newer versions...

PR #3004, opened 2021-02-14
[PHP] Add support for Enum

Ayesh:
Related to #2997.

PHP 8.1 will add [support for Enums with a syntax](https://php.watch/versions/8.1/enums#enum-syntax) similar to this:
```php
enum Foo: string {}
```

At the moment, the [vote](https://wiki.php.net/rfc/enumerations) has two more days to end, but it already has the 2/3 majority it requires to pass. This PR includes changes that did not make it to #2997 because the vote has yet to end. Once the vote is concluded, I will mark mark this PR ready for review, although early reviews are received with so much gratitude!

### Changes
 - Add Enum syntax support as a class-name variant.
 - Update CHANGES.md file.

### Checklist
- [x] Added markup tests, or they don't apply here because...
- [x] Updated the changelog at `CHANGES.md`
- [x] Added myself to `AUTHORS.txt`, under Contributors

PR #3404, opened 2021-11-21
fix(php) Remove duplicated code from default.txt

wkania:
[Code](https://github.com/highlightjs/highlight.js/blob/main/test/detect/php/default.txt#L47-L53) was duplicated in the [PR](https://github.com/highlightjs/highlight.js/pull/3004/files#diff-67d2c64508a7b38a3a651a39b292605749674d9f932d4a11c6ab21e369ae2a3eR59-R65)

### Changes
none

### DIFF
--- a/test/detect/php/default.txt
+++ b/test/detect/php/default.txt
@@ -56,14 +56,6 @@ enum Foo: string {
     case Test = 'test';
 }
 
-match ($key) {
-    1 => 'Integer 1',
-    '1' => 'String 1',
-    true => 'Bool true',
-    [] => 'Empty array',
-    [1] => 'Array [1]',
-};
-
 echo URI::ME . URI::$st1;
 
 __halt_compiler () ; datahere


joshgoebel:
I do not see where or how this is duplicated.

wkania:
@joshgoebel 
[Code](https://github.com/highlightjs/highlight.js/blob/main/test/detect/php/default.txt#L47-L53) is the same [as](https://github.com/highlightjs/highlight.js/blob/main/test/detect/php/default.txt#L59-L65) .
Looks like so merge/rebase conflict resolving typo. [PR](https://github.com/highlightjs/highlight.js/pull/3004/files#diff-67d2c64508a7b38a3a651a39b292605749674d9f932d4a11c6ab21e369ae2a3eR59-R65) was about _enum_, not _match_.
I hope this description is more clear.
Found it while working on something else.

joshgoebel (approved):

PR #3407, opened 2021-11-23
enh(php) Add support for Attributes

wkania:
### Changes
PHP 8 added support for [Attributes](https://www.php.net/releases/8.0/en.php#attributes).  Nested attributes were added to PHP 8.1.
They are similar to the ones from [Rust](https://github.com/highlightjs/highlight.js/blob/main/src/languages/rust.js#L240).

image

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -12,13 +12,15 @@ Category: common
  * */
 export default function(hljs) {
   const regex = hljs.regex;
-  const IDENT_RE_CORE = '[a-zA-Z0-9_\x7f-\xff]*' +
+  const IDENT_RE = '[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
     // negative look-ahead tries to avoid matching patterns that are not
     // Perl at all like $ident$, @ident@, etc.
-    '(?![A-Za-z0-9])(?![$]))';
-  const IDENT_RE = regex.concat("([a-zA-Z_\\x7f-\\xff]", IDENT_RE_CORE);
+    '(?![A-Za-z0-9])(?![$])';
   // Will not detect camelCase classes
-  const PASCAL_CASE_CLASS_NAME_RE = regex.concat("([A-Z]", IDENT_RE_CORE);
+  const PASCAL_CASE_CLASS_NAME_RE = '([\\\\?A-Z][a-zA-Z0-9_\x7f-\xff]*' +
+  // negative look-ahead tries to avoid matching patterns that are not
+  // Perl at all like $ident$, @ident@, etc.
+  '(?![A-Za-z0-9])(?![$])){1,}';
   const VARIABLE = {
     scope: 'variable',
     match: '\\$+' + IDENT_RE,
@@ -317,8 +319,7 @@ export default function(hljs) {
           regex.concat(WHITESPACE, "+"),
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-          regex.concat(/\\?/, IDENT_RE),
-          regex.concat(WHITESPACE, "*", /\(/),
+          PASCAL_CASE_CLASS_NAME_RE,
         ],
         scope: {
           1: "keyword",
@@ -431,10 +432,57 @@ export default function(hljs) {
   };
   PARAMS_MODE.contains.push(FUNCTION_INVOKE);
 
+  const ATTRIBUTES = {
+    begin: regex.concat(/#\[\s*/, PASCAL_CASE_CLASS_NAME_RE),
+    beginScope: "meta",
+    end: /]/,
+    endScope: "meta",
+    keywords: {
+      literal: LITERALS,
+      keyword: [
+        'new',
+        'array',
+      ]
+    },
+    contains: [
+      {
+        begin: /\[/,
+        end: /]/,
+        keywords: {
+          literal: LITERALS,
+          keyword: [
+            'new',
+            'array',
+          ]
+        },
+        contains: [
+          'self',
+          NAMED_ARGUMENT,
+          LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+          hljs.C_BLOCK_COMMENT_MODE,
+          STRING,
+          NUMBER,
+          CONSTRUCTOR_CALL,
+        ]
+      },
+      NAMED_ARGUMENT,
+      LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+      hljs.C_BLOCK_COMMENT_MODE,
+      STRING,
+      NUMBER,
+      CONSTRUCTOR_CALL,
+      {
+        scope: 'meta',
+        match: PASCAL_CASE_CLASS_NAME_RE
+      }
+    ]
+  };
+
   return {
     case_insensitive: false,
     keywords: KEYWORDS,
     contains: [
+      ATTRIBUTES,
       hljs.HASH_COMMENT_MODE,
       hljs.COMMENT('//', '$'),
       hljs.COMMENT(
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -12,15 +12,16 @@ Category: common
  * */
 export default function(hljs) {
   const regex = hljs.regex;
-  const IDENT_RE = '[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
-    // negative look-ahead tries to avoid matching patterns that are not
-    // Perl at all like $ident$, @ident@, etc.
-    '(?![A-Za-z0-9])(?![$])';
-  // Will not detect camelCase classes
-  const PASCAL_CASE_CLASS_NAME_RE = '([\\\\?A-Z][a-zA-Z0-9_\x7f-\xff]*' +
   // negative look-ahead tries to avoid matching patterns that are not
   // Perl at all like $ident$, @ident@, etc.
-  '(?![A-Za-z0-9])(?![$])){1,}';
+  const NOT_PERL_ETC = /(?![A-Za-z0-9])(?![$])/;
+  const IDENT_RE = regex.concat(
+    /[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/,
+    NOT_PERL_ETC);
+  // Will not detect camelCase classes
+  const PASCAL_CASE_CLASS_NAME_RE = regex.concat(
+    /(\\?[A-Z][a-z0-9_\x7f-\xff]+|\\?[A-Z]+(?=[^A-Z])){1,}/,
+    NOT_PERL_ETC);
   const VARIABLE = {
     scope: 'variable',
     match: '\\$+' + IDENT_RE,
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -20,7 +20,7 @@ export default function(hljs) {
     NOT_PERL_ETC);
   // Will not detect camelCase classes
   const PASCAL_CASE_CLASS_NAME_RE = regex.concat(
-    /(\\?[A-Z][a-z0-9_\x7f-\xff]+|\\?[A-Z]+(?=[^A-Z])){1,}/,
+    /(\\?[A-Z][a-z0-9_\x7f-\xff]+|\\?[A-Z]+(?=[A-Z][a-z0-9_\x7f-\xff])){1,}/,
     NOT_PERL_ETC);
   const VARIABLE = {
     scope: 'variable',
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -433,6 +433,15 @@ export default function(hljs) {
   };
   PARAMS_MODE.contains.push(FUNCTION_INVOKE);
 
+  const ATTRIBUTE_CONTAINS = [
+    NAMED_ARGUMENT,
+    LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+    hljs.C_BLOCK_COMMENT_MODE,
+    STRING,
+    NUMBER,
+    CONSTRUCTOR_CALL,
+  ];
+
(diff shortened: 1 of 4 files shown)


src/languages/php.js
```diff
@@ -102,11 +103,50 @@ export default function(hljs) {
...
+            className: 'keyword',
+            begin: '\\b(new)\\b'
+          },
+        ]
+      },
+      hljs.C_BLOCK_COMMENT_MODE,
+      STRING,
+      NUMBER,
+      {
+        className: 'literal',
```
joshgoebel, line 134: Why can `keywords` not be used here?

src/languages/php.js
```diff
@@ -102,11 +103,50 @@ export default function(hljs) {
     // <https://www.php.net/manual/en/reserved.classes.php>
     'Directory __PHP_Incomplete_Class parent php_user_filter self static stdClass'
   };
+  const ATTRIBUTES = {
+    className: 'meta',
```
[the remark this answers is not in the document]
wkania: @joshgoebel 

Answers:
1. There will be. They are better than annotations, replace them and annotations are just code in comments. Attributes are part of the language.
2. I made complex cases, to test the code, but yes.   [Api Platform](https://api-platform.com/docs/core/graphql/#filters) , [Doctrine](https://www.doctrine-project.org/2021/05/24/orm2.9.html)
3. PhpStorm is one of the most popular and highlights them like [this](https://blog.jetbrains.com/phpstorm/2020/10/phpstorm-2020-3-eap-4/) and [this](https://github.com/JetBrains/phpstorm-attributes#deprecated). Opening `#[`, closing `]` and the name of the attribute is meta
4. When I was working on it, I checked how Highlight js presents attributes in other languages, for example: [C#](https://docs.microsoft.com/pl-pl/dotnet/csharp/programming-guide/concepts/attributes/creating-custom-attributes), [Java](https://en.wikipedia.org/wiki/Java_annotation) and [Rust](https://doc.rust-lang.org/reference/attributes.html). I guess they don't have such complex nested cases, so probably I used too much meta tag.
In [this blog](https://stitcher.io/blog/attributes-in-php-8#rundown) the name of the attribute is also a 'title'.
In my opinion, the class names should be 'title' like in PhpStorm and the name of the attribute should be meta. 
I have no idea about the words like `allowMissingFields`.  Is any other language use some other classes than: keywords, title, meta ...?

src/languages/php.js
wkania: Of all the defined keywords, only `new` is allowed in the attributes.
if there is some cleaner and easier way to make nesting, then it's beyond my knowledge.

src/languages/php.js
```diff
@@ -102,11 +103,50 @@ export default function(hljs) {
...
+    begin: '^\\s*#\\[',
+    end: '\\]',
+    contains: [
+      {
+        begin: /\[/,
+        end: /\]/,
+        contains: [
```
joshgoebel, line 114:
```suggestion
        end: /\]/,
        keywords: { 
          literals: LITERALS,
          keyword: "new",
        },
        contains: [
```

Keywords should work just fine here.

src/languages/php.js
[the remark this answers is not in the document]
joshgoebel: > PhpStorm is one of the most popular and highlights them like this and this. Opening #[, closing ] and the name of the attribute is meta

I think this would be a good thing to model, and should be doable.

> In this blog the name of the attribute is also a 'title'.

Looks like `type` to be me which surely is not correct.

> I have no idea about the words like allowMissingFields

I would assume code inside attributes would fall back to using the same highlighting as outside code... so perhaps this would be highlighted as some type of hash attribute, or not at all.

src/languages/php.js
joshgoebel: See above.

src/languages/php.js
wkania: Will see.

src/languages/php.js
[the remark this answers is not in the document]
wkania: @joshgoebel Ok so I will go with the PhpStorm highlighting.

About `allowMissingFields` this could be the same as [named arguments](https://stitcher.io/blog/new-in-php-8#named-arguments-rfc) (which is also not implemented in highlight js)

I will pick [something](https://github.com/highlightjs/highlight.js/blob/main/src/styles/a11y-dark.css).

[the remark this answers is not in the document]
joshgoebel: We do have `attribute` and `attr` for such things, see docs.

joshgoebel:
Ping. Any luck?

wkania:
> Ping. Any luck?

Complex regex and no support for Lookbehind :(. 
Medium MR by size and complexity, now is hard in both metrics.

That's why I plan to make smaller merges with missing functionality like:
- constants
- function invoke
- named arguments
Then I will go back to this one.

wkania:
> Was work on this going to continue or was it superseded by the other PRs?

I didn't work on it for the last 2 weeks.  Updated code but still I don't know how to highlight with meta tag only the beginning `<span class="hljs-meta">#[MyAttribute<span>`  and end `<span class="hljs-meta">]<span>`. Any suggestion welcome.

joshgoebel:
>  Any suggestion welcome.

Can't you do this with `beginScope` and `endScope`, added in v11?

wkania:
> > Any suggestion welcome.
> 
> Can't you do this with `beginScope` and `endScope`, added in v11?


So easy with this one :)

src/languages/php.js
```diff
@@ -12,13 +12,13 @@ Category: common
...
-  const IDENT_RE_CORE = '[a-zA-Z0-9_\x7f-\xff]*' +
+  const IDENT_RE_CORE = 'a-zA-Z0-9_\x7f-\xff]*' +
     // negative look-ahead tries to avoid matching patterns that are not
     // Perl at all like $ident$, @ident@, etc.
     '(?![A-Za-z0-9])(?![$]))';
-  const IDENT_RE = regex.concat("([a-zA-Z_\\x7f-\\xff]", IDENT_RE_CORE);
+  const IDENT_RE = regex.concat("([a-zA-Z_\\x7f-\\xff][", IDENT_RE_CORE);
   // Will not detect camelCase classes
-  const PASCAL_CASE_CLASS_NAME_RE = regex.concat("([A-Z]", IDENT_RE_CORE);
+  const PASCAL_CASE_CLASS_NAME_RE = regex.concat("([A-Z][\\\\", IDENT_RE_CORE);
```
joshgoebel, line 21: Please create a second constant (to allow `\`) if you must, but don't split the expressions up this way, this is hard to follow and is likely to increase maintenance in the future.

src/languages/php.js
```diff
@@ -431,11 +430,59 @@ export default function(hljs) {
   };
   PARAMS_MODE.contains.push(FUNCTION_INVOKE);
 
+  const ATTRIBUTES = {
+    begin: regex.concat(/^\s*#\[\s*\\?/, PASCAL_CASE_CLASS_NAME_RE),
```
joshgoebel, line 434: This should not match whitespace... our HTML scope output should be clean:


```html
<!-- bad -->
<span class="highlight">             keyword</span>
<!-- good -->
<span class="highlight">keyword</span>
```

Could this be confused for something else, what if we just matched starting at the `#[`?

src/languages/php.js
```diff
@@ -431,11 +430,59 @@ export default function(hljs) {
...
+  };
+
   return {
     case_insensitive: false,
     keywords: KEYWORDS,
     contains: [
-      hljs.HASH_COMMENT_MODE,
+      hljs.inherit(hljs.HASH_COMMENT_MODE, {
+        begin: /^\s*#(?!\[)/
+      }),
```
joshgoebel, line 485: Same issue with whitespace, why was this change necessary?

joshgoebel: Oh you're adding a guard... but would it work just fine without the whitespace?  This is usually solved with match order not negative look-aheads... we should be matching attributes before comments so the comment rule can't mismatch them... is there a reason we can't do that?

src/languages/php.js
wkania: Hmm, it's correct to have a class name preceded by a namespace like: `new Assert\Count()`  
Just fixing what I missed earlier.  
I wanted to eliminate duplicate code. Two almost the same regex might be harder to maintain in the future.

src/languages/php.js
wkania: So I should highlight this with two meta tags?
```    
#[
         Assert\Collection(
```

src/languages/php.js
wkania: > Oh you're adding a guard... but would it work just fine without the whitespace? This is usually solved with match order not negative look-aheads... we should be matching `ATTRIBUTE` before `COMMENT` so the comment rule can't mismatch them... is there a reason we can't do that?


Moving `ATTRIBUTES` above the old comment code was enough. Fixed.

src/languages/php.js
joshgoebel:
```html
<span>#[
         Assert\Collection(</span>
```

Is acceptable.  This is what I"m objecting to:

```html
<span>
<!-- tons of whitespace -->

#[
         Assert\Collection(</span>
```

src/languages/php.js
joshgoebel: If you have another idea, I'm listening, but splitting the character group up and hiding the `[` all over the code, nope. :)

joshgoebel: A more precise regex would be preferable... because were not really trying to match:

`A\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\`

We're trying to match `Ident\Indent` or some such... it'd be better to do that that the imprecise thing.

src/languages/php.js
wkania: > No.
> 
> ```
> <span>#[
>          Assert\Collection(</span>
> ```
> 
> Is acceptable. This is what I'm objecting to:
> 
> ```
> <span>
> 
> 
> #[
>          Assert\Collection(</span>
> ```

Fixed

src/languages/php.js
wkania: > A more precise regex would also be better... because were not really trying to match:
> 
> `A\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\`
> 
> We're trying to match `Ident\Ident` or some such... it'd be better to simple do that vs extending the character group like this.

Fixed

joshgoebel:
In the future if you could commit your changes in a fixup commit (from the last review point) rather than rewriting the whole history it'd make it MUCH easier for me to review.  Thanks.

src/languages/php.js
```diff
@@ -12,13 +12,15 @@ Category: common
  * */
 export default function(hljs) {
   const regex = hljs.regex;
-  const IDENT_RE_CORE = '[a-zA-Z0-9_\x7f-\xff]*' +
+  const IDENT_RE = '[a-zA-Z_\\x7f-\\xff][a-zA-Z0-9_\x7f-\xff]*' +
```
joshgoebel, line 15: Is it a single `\` or `\\` for `\x7f`?  Looks like we're doing it different in different spots.  Is that intentional?

- L15/20 etc. Lines 15 has both forms.

joshgoebel (changes requested):
Much better, think this is almost to the finish line!

src/languages/php.js
wkania: Good catch. Smart paste tried to be helpful. Fixed

wkania: I made this typo in earlier PR :(

joshgoebel:
How does that look?

wkania:
> How does that look?

Didn't think about it because I treat abbreviations and acronyms like any other word (camelCase / PascalCase). Of course there is a lot of code out there where all characters of abbreviation or acronym are capitalized.

All changes look good.

joshgoebel:
>  a lot of code out there where all characters of abbreviation or acronym are capitalized.

Yeah usually in other places where we deal with the pattern we support all caps abbreviations, eventually I need to abstract these rules into MODES.

wkania:
> > a lot of code out there where all characters of abbreviation or acronym are capitalized.
> 
> Yeah usually in other places where we deal with the pattern we support all caps abbreviations, eventually I need to abstract these rules into MODES.

Ok, So everything is good with PR, because I see the change requested?

src/languages/php.js
```diff
@@ -431,10 +433,57 @@ export default function(hljs) {
...
+      },
+      NAMED_ARGUMENT,
+      LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+      hljs.C_BLOCK_COMMENT_MODE,
+      STRING,
+      NUMBER,
+      CONSTRUCTOR_CALL,
+      {
+        scope: 'meta',
+        match: PASCAL_CASE_CLASS_NAME_RE
```
joshgoebel, line 477: Is it on purpose that this last rule isn't deeply nested like the other contains rules? (line 461-466 vs 469-474)?

wkania: > Is it on purpose that this last rule isn't deeply nested like the other contains rules? (line 461-466 vs 469-474)?

It is fix for the [second MyAttribute](https://github.com/highlightjs/highlight.js/pull/3407/files#diff-b6b04ba70bc0b19c207aeef76a2e31596393452dba4c39e2c78833d89b63c070R13). Without it, only the first attribute name was highlighted with the meta.

joshgoebel: Makes sense - and that can never be deeply nested? It has to be top-level?

wkania: > Makes sense - and that can never be deeply nested? It has to be top-level?

Yes, attributes are only top-level. They can only contain nested classes. Example how IDE highlight them:

image

joshgoebel:
@wkania Thanks so much!

brzuchal:
@joshgoebel when this can be released?

joshgoebel:
Whenever the next release is, there isn't a fixed scheduled, often every month or two.

PR #3412, opened 2021-11-29
(chore) es6 via CDN

### LINKED ISSUE #3383: [Request] Module exports for CDN-hosted libraries
The [CDN-hosted libraries](https://highlightjs.org/usage/#importing-the-library) does not provide ECMAScript 6 module exports even thought they are included in the source here on GH. Possibly for compatibility- or some other logical reason, I can't say for sure without delving deeper into the code.

```js
import hljs from "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.3.1/highlight.min.js";
```
*Will throw a `SyntaxError` due to a missing `export` on the requested resource.*

CDN-hosted ECMAScript 6 modules. 
The actual implementation/CDN-provider doesn't really matter as long as it's `import`able.

GH Pages provides its own CDN and could be set up to serve JS files directly from this repo.
[Some restrictions apply](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#usage-limits) but requires relatively low maintenance and initial effort. For small-scale, not-at-all business critical projects it's more than acceptable that the file isn't pre-processed (minified etc.).

```js
import hljs from "https://highlightjs.github.io/highlight.js/src/highlight.js";
```
*A dyslectic nightmare to look at, but it would get the job done*

**Additional context...**
N/A

joshgoebel:
Resolves #3383.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,15 @@ detection.
   - [ES6 Modules / `import`](#es6-modules--import)
 - [Getting the Library](#getting-the-library)
   - [Fetch via CDN](#fetch-via-cdn)
+    - [**cdnjs** (link)](#cdnjs-link)
+      - [Common JS](#common-js)
+      - [ES6 Modules](#es6-modules)
+    - [**jsdelivr** (link)](#jsdelivr-link)
+      - [Common JS](#common-js-1)
+      - [ES6 Modules](#es6-modules-1)
+    - [**unpkg** (link)](#unpkg-link)
+      - [Common JS](#common-js-2)
+      - [ES6 Modules](#es6-modules-2)
     - [Download prebuilt CDN assets](#download-prebuilt-cdn-assets)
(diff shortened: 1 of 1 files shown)


allejo (approved):

README.md
```diff
@@ -268,13 +271,16 @@ const highlightedCode = hljs.highlight('<span>Hello World!</span>', {language: '
 
 ### ES6 Modules / `import`
 
+*Note: You can also import directly from fully static URLs, such as our every own pre-built
+ES6 Module CDN resources. See [Fetch via CDN](#fetch-via-cdn) for specific examples.*
```
allejo, line 275:
```suggestion
*Note: You can also import directly from fully static URLs, such as our very own pre-built
ES6 Module CDN resources. See [Fetch via CDN](#fetch-via-cdn) for specific examples.*
```

PR #3413, opened 2021-11-30
Stan updates

### LINKED ISSUE #3410: Stan syntax updates
**Information**
- Language: Stan

**Description**
Lots of updates to Stan over the last few months

Type updates:
* `array[]` is now a valid keyword
* `complex` is a valid keyword literal

The higher order functions that are valid are:

          ## Algebraic equation solver
          "algebra_solver", "algebra_solver_newton",

          ## Ordinary differential equation
          "ode_rk45", "ode_rk45_tol", "ode_ckrk", "ode_ckrk_tol", "ode_adams",
          "ode_adams_tol", "ode_bdf", "ode_bdf_tol", "ode_adjoint_tol_ctl",

          ## 1D integrator
          "integrate_1d",

          ## Reduce-sum function
          "reduce_sum", "reduce_sum_static",

It's probably easier to just have all the functions. You can view the updated [recent pr request in the rouge library ](https://github.com/rouge-ruby/rouge/blob/ce05db529558fcbb413d8bcff9e30323f9b52dbf/lib/rouge/lexers/stan.rb)
<details>
<summary>Long list of functions</summary>
          # Integer-Valued Basic Functions

          ## Absolute functions
          "abs", "int_step",

          ## Bound functions
          "min", "max",

          ## Size functions
          "size",

          # Real-Valued Basic Functions

          ## Log probability function
          "target", "get_lp",

          ## Logical functions
          "step", "is_inf", "is_nan",

          ## Step-like functions
          "fabs", "fdim", "fmin", "fmax", "fmod", "floor", "ceil", "round",
          "trunc",

          ## Power and logarithm functions
          "sqrt", "cbrt", "square", "exp", "exp2", "log", "log2", "log10",
          "pow", "inv", "inv_sqrt", "inv_square",

          ## Trigonometric functions
          "hypot", "cos", "sin", "tan", "acos", "asin", "atan", "atan2",

          ## Hyperbolic trigonometric functions
          "cosh", "sinh", "tanh", "acosh", "asinh", "atanh",

          ## Link functions
          "logit", "inv_logit", "inv_cloglog",

          ## Probability-related functions
          "erf", "erfc", "Phi", "inv_Phi", "Phi_approx", "binary_log_loss",
          "owens_t",

          ## Combinatorial functions
          "beta", "inc_beta", "lbeta", "tgamma", "lgamma", "digamma",
          "trigamma", "lmgamma", "gamma_p", "gamma_q",
          "binomial_coefficient_log", "choose", "bessel_first_kind",
          "bessel_second_kind", "modified_bessel_first_kind",
          "log_modified_bessel_first_kind", "modified_bessel_second_kind",
          "falling_factorial", "lchoose", "log_falling_factorial",
          "rising_factorial", "log_rising_factorial",

          ## Composed functions
          "expm1", "fma", "multiply_log", "ldexp", "lmultiply", "log1p",
          "log1m", "log1p_exp", "log1m_exp", "log_diff_exp", "log_mix",
          "log_sum_exp", "log_inv_logit", "log_inv_logit_diff",
          "log1m_inv_logit",

          ## Special functions
          "lambert_w0", "lambert_wm1",

          ## Complex Conversion Functions
          "get_real", "get_imag",

          # Complex-Valued Basic Functions

          ## Complex Construction Functions
          "to_complex",

          # Array Operations

          ## Reductions
          "sum", "prod", "log_sum_exp", "mean", "variance", "sd", "distance",
          "squared_distance", "quantile",

          ## Array size and dimension function
          "dims", "num_elements",

          ## Array broadcasting
          "rep_array",

          ## Array concatenation
          "append_array",

          ## Sorting functions
          "sort_asc", "sort_desc", "sort_indices_asc", "sort_indices_desc",
          "rank",

          ## Reversing functions
          "reverse",

          # Matrix Operations

          ## Integer-valued matrix size functions
          "num_elements", "rows", "cols",

          ## Dot products and specialized products
          "dot_product", "columns_dot_product", "rows_dot_product", "dot_self",
          "columns_dot_self", "rows_dot_self", "tcrossprod", "crossprod",
          "quad_form", "quad_form_diag", "quad_form_sym", "trace_quad_form",
          "trace_gen_quad_form", "multply_lower_tri_self_transpose",
          "diag_pre_multiply", "diag_post_multiply",

          ## Broadcast functions
          "rep_vector", "rep_row_vector", "rep_matrix",
          "symmetrize_from_lower_tri",

          ## Diagonal matrix functions
          "add_diag", "diagonal", "diag_matrix", "identity_matrix",

          ## Container construction functions
          "linspaced_array", "linspaced_int_array", "linspaced_vector",
          "linspaced_row_vector", "one_hot_int_array", "one_hot_array",
          "one_hot_vector", "one_hot_row_vector", "ones_int_array",
          "ones_array", "ones_vector", "ones_row_vector", "zeros_int_array",
          "zeros_array", "zeros_vector", "zeros_row_vector", "uniform_simplex",

          ## Slicing and blocking functions
          "col", "row", "block", "sub_col", "sub_row", "head", "tail",
          "segment",

          ## Matrix concatenation
          "append_col", "append_row",

          ## Special matrix functions
          "softmax", "log_softmax", "cumulative_sum",

          ## Covariance functions
          "cov_exp_quad",

          ## Linear algebra functions and solvers
          "mdivide_left_tri_low", "mdivide_right_tri_low", "mdivide_left_spd",
          "mdivide_right_spd", "matrix_exp", "matrix_exp_multiply",
          "scale_matrix_exp_multiply", "matrix_power", "trace", "determinant",
          "log_determinant", "inverse", "inverse_spd", "chol2inv",
          "generalized_inverse", "eigenvalues_sym", "eigenvectors_sym",
          "qr_thin_Q", "qr_thin_R", "qr_Q", "qr_R", "cholseky_decompose",
          "singular_values", "svd_U", "svd_V",

          # Sparse Matrix Operations

          ## Conversion functions
          "csr_extract_w", "csr_extract_v", "csr_extract_u",
          "csr_to_dense_matrix",

          ## Sparse matrix arithmetic
          "csr_matrix_times_vector",

          # Mixed Operations
          "to_matrix", "to_vector", "to_row_vector", "to_array_2d",
          "to_array_1d",

          # Higher-Order Functions

          ## Algebraic equation solver
          "algebra_solver", "algebra_solver_newton",

          ## Ordinary differential equation
          "ode_rk45", "ode_rk45_tol", "ode_ckrk", "ode_ckrk_tol", "ode_adams",
          "ode_adams_tol", "ode_bdf", "ode_bdf_tol", "ode_adjoint_tol_ctl",

          ## 1D integrator
          "integrate_1d",

          ## Reduce-sum function
          "reduce_sum", "reduce_sum_static",

          ## Map-rect function
          "map_rect",

          # Deprecated Functions
          "integrate_ode_rk45", "integrate_ode", "integrate_ode_adams",
          "integrate_ode_bdf",

          # Hidden Markov Models
          "hmm_marginal", "hmm_latent_rng", "hmm_hidden_state_prob"
        ]
</details>

The BNF grammars page was updated. I see that this is referenced in the prism stan code. The updated file is [stan bnf grammars 2.28](https://mc-stan.org/docs/2_28/reference-manual/bnf-grammars.html).

spinkney:
Resolves #3410 

### Changes
Updated Stan highlighter for Stan version 2.28
* complex and array types
* complex numbers
* added all new functions and distributions

Improvements over previous
* highlighting of constants
* highlighting for user defined distributions
* highlighting for truncation

### Checklist


src/languages/stan.js
```diff
@@ -484,14 +325,14 @@ export default function(hljs) {
 
   const INCLUDE = {
     className: 'meta',
-    begin: /^#include\b/,
+    begin: /#\s*[a-z]+\b/,
```
joshgoebel, line 328: This seems wrong. The contains string rule only make sense for SOME directives, likely not all... so this would have to be much more nuanced if you wished to make it generic, no?

src/languages/stan.js
```diff
@@ -484,14 +325,14 @@ export default function(hljs) {
...
     end: /$/,
     relevance: 0, // relevance comes from keywords
-    keywords: "include",
+    keywords: '#include',
```
joshgoebel, line 331: Pretty sure this won't work without using `$pattern` as # isn't a normal keyword character.

src/languages/stan.js
```diff
@@ -504,16 +345,38 @@ export default function(hljs) {
       $pattern: hljs.IDENT_RE,
       title: BLOCKS,
       keyword: STATEMENTS.concat(VAR_TYPES).concat(SPECIAL_FUNCTIONS),
-      built_in: FUNCTIONS
+      built_in: FUNCTIONS.concat(DISTRIBUTIONS).concat(expanded),
```
joshgoebel, line 348: Now lets find a better named for `expanded`... I only was showing you the idea in the example, not the naming.

src/languages/stan.js
```diff
@@ -504,16 +345,38 @@ export default function(hljs) {
...
       hljs.HASH_COMMENT_MODE,
       BLOCK_COMMENT,
+     {
+       className: 'built_in',
+       variants: [
+         {
+          begin: /(?<=~)\s+(\w+)/
+         },
+         {
+          begin: /(?<=\+=)\s+(\w+(_lpdf|_lupdf|_lpmf|_cdf|_lcdf|_lccdf|_qf))/
```
joshgoebel, line 363: Don't we already have a manual list of these?

src/languages/stan.js
```diff
@@ -523,18 +386,26 @@ export default function(hljs) {
...
+      {
+        className: 'keyword',
+        begin: /\btarget\s*/,
       },
       {
         className: 'number',
         variants: [
           {
-            begin: /\b\d+(?:\.\d*)?(?:[eE][+-]?\d+)?/
+            begin: /(?:\b\d+(?:_\d+)*(?:\.(?:\d+(?:_\d+)*)?)?|\B\.\d+(?:_\d+)*)(?:[eE][+-]?\d+(?:_\d+)*)?i?(?!\w)/i
```
joshgoebel, line 408: I see a top-level `|` here do I not?  If so please break this out into multiple variant and add comments on what we're trying to match to aid future maintenance.

src/languages/stan.js
```diff
@@ -504,16 +345,38 @@ export default function(hljs) {
...
+          begin: /(?<=~)\s+(\w+)/
+         },
+         {
+          begin: /(?<=\+=)\s+(\w+(_lpdf|_lupdf|_lpmf|_cdf|_lcdf|_lccdf|_qf))/
+         }
+        ] 
+     },
+     {
+      className: 'symbol',
+      begin: /\s+(pi|e|sqrt2|log2|log10)(?=\()/,
```
joshgoebel, line 369: `\s+` likely should not be part of the symbol match... we do not include outside space in scopes - we try to be precise like that.  If you need to match the space before you likely need a multi-matcher (passing an array of regex to `match`, see the docs).

src/languages/stan.js
```diff
@@ -504,16 +345,38 @@ export default function(hljs) {
...
+      relevance: 0
+     },
+     {
+       begin: /(?<=\))\s*T(?=\s*\[)/,
```
joshgoebel, line 373: `(?<`

We cannot use regex look-behind anywhere because Safari does not yet support look-behind, sorry.

src/languages/stan.js
```diff
@@ -523,18 +386,26 @@ export default function(hljs) {
...
-        begin: '~\\s*(' + hljs.IDENT_RE + ')\\s*\\(',
-        keywords: DISTRIBUTIONS
+        // hack: in range constraints, upper must follow either , or <
+        // <offset = ..., multiplier = ...> or <multiplier = ...>
+        begin: /[<,]\s*multiplier\s*=/,
+        keywords: 'multiplier'
+      },
+      {
+        className: 'keyword',
+        begin: /\btarget\s*/,
```
joshgoebel, line 402: If the `+=` specificity isn't needed can this not just be moved into the normal keyword list?

src/languages/stan.js
```diff
@@ -523,18 +386,26 @@ export default function(hljs) {
...
-        className: 'keyword',
-        begin: /\btarget\s*\+=/
+        // hack: in range constraints, upper must follow either , or <
+        // <multiplier = ..., offest = ...> or <offset = ...>
+        begin: /[<,]\s*offset\s*=/,
+        keywords: 'offset'
       },
       {
-        begin: '~\\s*(' + hljs.IDENT_RE + ')\\s*\\(',
-        keywords: DISTRIBUTIONS
```
joshgoebel, line 531: Why was this removed? Can distributions be used with out the trailing `(`?

src/languages/stan.js
spinkney: `+=` has to be there

src/languages/stan.js
spinkney: Ah, I'll comment this out and can revisit once Safari gets look-behind. Not sure how to get this to work otherwise
```
vector[N] T[N); // do not highlight T
X ~ normal(mu, sigma) T[1, ]; // highlight T 
```

src/languages/stan.js
spinkney: I'm not sure, this seems to work for all my tests

src/languages/stan.js
spinkney: It's parsing as what I want:
Screen Shot 2021-11-30 at 1 45 47 PM

Feel free to make changes. I don't know javascript or regex.

src/languages/stan.js
spinkney: Will update to DISTRIBUTIONS_EXPANDED

src/languages/stan.js
spinkney: Yes, but this is for user defined functions. We will not know what comes before for those. I will add a comment

src/languages/stan.js
spinkney: Shoot, this comes from prism.js and how the author handled complex numbers see https://github.com/PrismJS/prism/blob/c53ad2e65b7193ab4f03a1797506a54bbb33d5a2/components/prism-stan.js#L56. Let me add a reference to that. I definitely don't understand this regex.

src/languages/stan.js
spinkney: I wasn't sure what this was exactly doing. Distributions must end with the parenthesis so `example_dist();`

src/languages/stan.js
```diff
@@ -504,16 +345,38 @@ export default function(hljs) {
       $pattern: hljs.IDENT_RE,
       title: BLOCKS,
       keyword: STATEMENTS.concat(VAR_TYPES).concat(SPECIAL_FUNCTIONS),
-      built_in: FUNCTIONS
+      built_in: FUNCTIONS.concat(DISTRIBUTIONS).concat(expanded),
+      punctuation: DELIMIT
```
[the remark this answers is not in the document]
spinkney: I didn't understand how to do this, I'll comment out the delimit for now as I don't think it's needed.

src/languages/stan.js
spinkney: thanks, removed

src/languages/stan.js
joshgoebel: What is the full list of # directives Stan has?

src/languages/stan.js
joshgoebel: You only seeing the `meta` highlighting... the `keywords` isn't working there at all because the "#" is not matched by IDENT_RE, which is what Stan uses for keyword pattern matching.

src/languages/stan.js
joshgoebel: You can use multi-match sometimes, but it will **consume** the prior content (so you can't highlight it, or it gets very complicated)... ie there is never a way to look-behind, you're always looking forward... not sure that would help us here though so perhaps better to just wait as you suggest.

src/languages/stan.js
joshgoebel: Please ask vs just removing things you don't understand. :-)

This makes sure that distributions are ONLY highlighted when followed by a `(` which is good because it can prevent false positives - or conflicts with other languages that might sharing some of the same naming, but with a different syntax.  Generally the more precise we can match a thing, the better.

So I'd add this back and remove DISTRIBUTIONS from the generic `keywords` list...

src/languages/stan.js
joshgoebel: We prefer our regex split out and properly documented, not just a huge mess of unreadable regex (as found in some other codebases)... if you don't understand this I would revert it and we will wait for someone else to come along and fix it properly.

src/languages/stan.js
```diff
@@ -27,21 +28,41 @@ export default function(hljs) {
...
     'return'
   ];
+
   const SPECIAL_FUNCTIONS = [
     'print',
     'reject',
-    'increment_log_prob|10',
-    'integrate_ode|10',
-    'integrate_ode_rk45|10',
-    'integrate_ode_bdf|10',
```
joshgoebel, line 36: Why did 10 get lost?

src/languages/stan.js
```diff
@@ -27,21 +28,41 @@ export default function(hljs) {
...
   ];
   const VAR_TYPES = [
+    `array`,
+    'complex',
     'int',
     'real',
     'vector',
     'ordered',
-    'positive_ordered',
+    'positive_ordered|10',
```
joshgoebel, line 65: Justification for 10 here?

src/languages/stan.js
joshgoebel: @gibson042 sometimes finds these things fun... not sure if they'd be willing to help here.

spinkney: this does allow for proper handling of complex numbers

joshgoebel: The question isn't whether it works but whether it's readable and maintainable... our policy generally is we require the different types of numbers to be properly broken out (see many of the other newer grammars, like Swift, etc)... we don't just copy a huge regex from somewhere else because it seems to work (at the current moment).

spinkney: I was adding context for who ever wants to tackle this.

spinkney: I do not think this should be merged unless complex numbers are highlighted appropriately.

src/languages/stan.js
spinkney: so #include can include anything really as long as the cpp compiler can pick it up. This includes .cpp, .stan. stanfunctions endings.

joshgoebel: I'm not taking about `#include`, but since you changed it to `#whatever` (but yet didn't touch the string contains rules) so I'm asking what directives Stan supports - or is this just using the C pre-compiler?

```
#ifdef SUPER_MODE
```

For example, SUPER_MODE here is not necessarily a string... 

Was this your intention? What were you trying to accomplish with this change?

src/languages/stan.js
joshgoebel: Entirely disagree and it would be sad to see you stop now since there are a multitude of other improvements we could benefit from.

src/languages/stan.js
spinkney: yes, let me get those back in

src/languages/stan.js
spinkney: are there other languages which use `positive_ordered` types? I thought this was one of the few

src/languages/stan.js
spinkney: done

src/languages/stan.js
spinkney: I removed keywords and just use the meta highlighting as I think it looks good. Let me know if this goes against the package design philosophy.

src/languages/stan.js
spinkney: got it, updated to put include back in

src/languages/stan.js
```diff
@@ -53,421 +74,241 @@ export default function(hljs) {
...
-    'normal_cdf',
-    'normal_lccdf',
-    'normal_lcdf',
-    'normal_lpdf',
-    'normal_rng',
-    'not_a_number',
-    'num_elements',
-    'ordered_logistic_lpmf',
-    'ordered_logistic_rng',
+    // Integer-Valued Basic Functions
```
joshgoebel, line 77: Did you get this categorized list ready made from somewhere (or a script)?  If so we should link to it... it's nice to read for sure... but if not, who is going to maintain this over time - and would not a simple alphabetical list be preferable?  "Nice" lists like this often become a mess as no one knows where to add "one-offs" as often happens during real maintenance of a grammar over the long-term.

src/languages/stan.js
joshgoebel: This is also another case of look-behind... I only mentioned it once, but you'll have to make sure NONE of your regex are using it across the entire grammar...

joshgoebel: Should the suffix version not also include the `(?=\())` lookahead filter?

src/languages/stan.js
joshgoebel: It's probably ok... I'll make a final pass and clean it up later if necessary.  The relevancy may have to be tuned now, etc... but it likely looked weird before for some themes so I'm ok changing it.

src/languages/stan.js
spinkney: yes, you're right. Added back

src/languages/stan.js
spinkney: You can get all the signatures this way:

clone the [stan-docs repo](https://github.com/stan-dev/docs) then cd into it run the python script with the major and minor version numbers:
```bash
python3 extract_function_sigs.py 2 28
```
then run a bash command to extract only the unique names
```bash
cut -d ';' -f1 stan-functions-2_28.txt | uniq -u > stan_functions_list-2_28.txt
```
I'll update with the alphabetical list and then I can remove the `_lpdf` stuff as well since this can be updated easily

joshgoebel: > You can get all the signatures this way:

Lets comment this in the source to aid future maintenance.

src/languages/stan.js
spinkney: I parsed the regex out into two pieces and commented on them, it's actually a lot simpler than it looks

spinkney:
I'm going to have to break out distributions. Hold on to review until I've pushed the next update.

src/languages/stan.js
joshgoebel: You can't do this, you can't run raw JS code on the input.

As I mentioned before you can use multi-matchers to achieve things *similar* to this (match groups)... see the docs for `beginScope` and `endScope` which can be keyed just like TM grammars, etc... as to whether it'd be allowed it would depend on how complex the actual implementation ended up being.

spinkney:
the only thing this doesn't have anymore is highlighting for user defined distributions but I think I can try to tackle that in a different PR

spinkney:
This is good to review. That multi-match stuff is great. Got truncation working and UDFs

src/languages/stan.js
```diff
@@ -512,41 +410,74 @@ export default function(hljs) {
       hljs.HASH_COMMENT_MODE,
       BLOCK_COMMENT,
       {
-        // hack: in range constraints, lower must follow "<"
-        begin: /<\s*lower\s*=/,
-        keywords: 'lower'
+        scope: 'symbol',
+        match: /\s(pi|e|sqrt2|log2|log10)(?=\()/,
```
joshgoebel, line 414: What makes these special?  Looks like a function call to me, which would traditionally be a built-in or title.function...

src/languages/stan.js
```diff
@@ -512,41 +410,74 @@ export default function(hljs) {
...
+          2: "built_in",
+          4: "keyword"
+        }
       },
       {
-        begin: '~\\s*(' + hljs.IDENT_RE + ')\\s*\\(',
-        keywords: DISTRIBUTIONS
+        // highlights distributrions that end with special endings
+        scope: 'built_in',
+        match: /\w+(_lpdf|_lupdf|_lpmf|_cdf|_lcdf|_lccdf|_qf)(?=\s*[\(.*\)])/
```
joshgoebel, line 441: Are we sure `(?=\s*[\(.*\)])` this is correct?

What are we trying to match here again?  The `.*` inside of [] looks very strange.

src/languages/stan.js
```diff
@@ -512,41 +410,74 @@ export default function(hljs) {
...
-          {
-            begin: /\.\d+(?:[eE][+-]?\d+)?\b/
-          }
+        // cannot ever have these endings with the ~
+        match: [
+          /~\s*/,
+          /\w+(_lpdf|_lupdf|_lpmf|_cdf|_lcdf|_lccdf|_qf)(?=\s*[\(.*\)])/
         ],
-        relevance: 0
+        scope: { 2: "illegal" }
```
joshgoebel, line 449: @highlightjs/core  Any interest in getting into illegal?  I know I've mentioned it in some thread somewhere, but we never have in the past and I'm thinking that now likely isn't the time either...

src/languages/stan.js
spinkney: I'm treating them a bit differently because someone could name these from a type. So `e` by itself should not be highlighted but `e()` should. I'm open to these being put somewhere else.

src/languages/stan.js
spinkney: they must only be highlighted if there is a parenthesis and I want it to be closed with any character or space in between

joshgoebel: Also if we always highlight these (regardless of prefix), can all of these suffixes just be removed from the keyword lists entirely since we have this rule?

spinkney: I've kind of half-way allowed user defined functions getting highlighted. This only allows user defined distributions which must have one of those endings. I wanted to make sure that if someone inputs something that looks like a function with `()` but hasn't defined it, it doesn't get highlighted and make them think it's a valid function. It seems much more likely that if the function has one of those endings that it is already defined in the function block. 

Do we allow this? Or should we only allow highlighting of things defined explicitly in the language? I'm ok with including the `keywords: DISTRIBUTIONS` here and not highlighting user defined distributions.

src/languages/stan.js
spinkney: I'm also ok if this just doesn't get highlighted. Do you have a suggestion for that and I can remove the illegal stuff?

src/languages/stan.js
joshgoebel: > So e by itself should not be highlighted but e() should

I don't have any problem with that... the `(` conditional look-ahead is GREAT... I was asking why is this symbol vs built-in or title.function... generally symbol is for a label or string like object such as symbol in ruby:

```rb
a = "string"
b = :symbol 
```

src/languages/stan.js
joshgoebel: > If someone inputs something that looks like a function with () but hasn't defined it

We highlight *snippets* (often), not full code blocks.  Plus we have no way of knowing if someone "defined" something earlier or not since we don't keep track of any context like that. Unless we detect it as it's being defined, but then later we can't detect a "reference to something defined in the past"...

I feel like this should perhaps be all or nothing.  My thinking:

- distributions and other built in functions are `built_in`
- all function dispatch `doSomething(`, ie `[validFuncName][lookahead("(")]`) are `title.function`.

Is a `identifier(` guaranteed to be a function dispatch in Stan?

src/languages/stan.js
joshgoebel: If it's not guarding another rule, just remove it, if it is then document what it's protecting against with a comment and remove the `scope` - which will cause it to be matched and consumed, but not highlighted.

allejo: > @highlightjs/core Any interest in getting into illegal?

I don't think now's the time to get into this, this would be better suited as a separate issue or PR

joshgoebel: Ok, lets rip it out then. :)

spinkney (commented):
hey @joshgoebel, this commit broke the stan syntax highlighting for me. Nothing gets highlighted, reverting this and stan highlighting works.

src/languages/stan.js
spinkney: made it `built_in`

src/languages/stan.js
spinkney: I've removed the `illegal` and changed to `no-highlight`. Let me know if that is the appropriate use. It is protecting against highlighting when there is a ~ and a distribution, that cannot ever happen and it is documented on line 439.

src/languages/stan.js
spinkney: I added all udfs as title.function and I broke out distributions so they are only used with `built_ins`. I'm digging what this looks like.

joshgoebel:
Oh line 10 is broken that's weird.

src/languages/stan.js
```diff
@@ -410,7 +410,7 @@ export default function(hljs) {
       hljs.HASH_COMMENT_MODE,
       BLOCK_COMMENT,
       {
-        scope: 'symbol',
+        scope: 'built_in',
```
joshgoebel, line 413: Better I think.

src/languages/stan.js
```diff
@@ -436,30 +436,36 @@ export default function(hljs) {
...
+        scope: 'title.function',
+        begin: /\w*(_lpdf|_lupdf|_lpmf|_cdf|_lcdf|_lccdf|_qf)(?=\s*[\(.*\)])/
+      },
+      {
+        // cannot ever have these endings with the ~
+        begin: [
+          /~\s*/,
+          /\w*(_lpdf|_lupdf|_lpmf|_cdf|_lcdf|_lccdf|_qf)(?=\s*[\(.*\)])/
+        ],
+        scope: { 2: "no-highlight" }
```
joshgoebel, line 473: This is not  valid scope... if you don't want to highlight them, then just have no scope at all.  But what is this rule accomplishing, acting as a guard for which other rule?

src/languages/stan.js
```diff
@@ -442,13 +442,22 @@ export default function(hljs) {
...
+      },
+      {
+        // highlights user defined distributions after ~
+        begin: [
+          /~/,
+          /\s*\w+(?=\s*[\(.*\)])/,
+          '(?!.*/\b(' + regex.either(...DISTRIBUTIONS) + ')\b)'
         ],
-        scope: { 3: "title.function" },
+        scope: { 2: "title.function" }
```
joshgoebel, line 460: This split between built_in and title.function looks much nicer.

src/languages/stan.js
```diff
@@ -512,41 +410,89 @@ export default function(hljs) {
...
-        begin: /[<,]\s*upper\s*=/,
-        keywords: 'upper'
+        match: regex.concat(/[<,]\s*/, regex.either(...RANGE_CONSTRAINTS), /\s*=/),
+        keywords: RANGE_CONSTRAINTS
       },
       {
-        className: 'keyword',
-        begin: /\btarget\s*\+=/
+        scope: 'keyword',
+        match: /\btarget\s*/,
```
joshgoebel, line 423: Should this `\s*` be lookahead?

src/languages/stan.js
spinkney: it's acting as a guard to not highlight distribution that has those endings *after* a `~`. So should I just delete line 464?

src/languages/stan.js
spinkney: yep, let me fix that and push

src/languages/stan.js
joshgoebel: > it's acting as a guard to not highlight distribution that has those endings after a ~

A comment would be helpful to explain this.

But what is the deal with distributions that come after `~`? (curious)

> So should I just delete line 464?

Yes, though then you may need to collapse the multi-matcher to a  single regex, can't recall if it'll force ya to do that or not.

spinkney: I removed the scope. I had to keep the `scope: {}` just without anything inside or else it breaks.

spinkney: > But what is the deal with distributions that come after `~`? (curious)

it's basically syntactic sugar for an unnormalized log density. Stan didn't have `_lupdf` before to indicate this so if you wanted an unnormalized density (which is faster but you cannot do certain things with the posterior samples, like the bridgesampling R package needs).

src/languages/stan.js
```diff
@@ -420,7 +420,7 @@ export default function(hljs) {
       },
       {
         scope: 'keyword',
-        match: /\btarget\s*/,
+        match: /\btarget\s*(?=\+=)/,
```
joshgoebel, line 423:
```suggestion
        match: /\btarget(?=\s*\+=)/,
```

No?

src/languages/stan.js
joshgoebel: > it's basically syntactic sugar for an unnormalized log density.

Why do we not want to highlight "unnormalized log density" distributions?

spinkney: You just can't have an ending `_lpdf, lcdf, lupdf, etc.` with the ~. You can only do it with `target +=` hence it should not be highlighted.

spinkney: You can either do
```stan
target += some_density_in_stan_lupdf(y | x);
```
or
```
y ~ some_density_in_stan(x);
```
you cannot write
```
y ~ some_density_in_stan_lupdf( ? | x);
```

joshgoebel: > you cannot write

_So why would it exist in the code at all as something we need to match for?_ - why would it be something we have to worry about at all then?

spinkney: I just didn't want to highlight something that was invalid. Otherwise this will get highlighted as valid.

spinkney: The issue is that the distributions are keywords and these will match regardless of where they are. I need them to match in many places except you cannot have a distribution with those endings after a ~.

joshgoebel: I think I'm going to yank this (it can be revisited in the future with more data or concrete examples).  Our job isn't to decide if code is "valid" or not...  we highlight *correct* code - it is expected that code we are asked to highlight will be "correct".  Results for *incorrect* code are unpredictable.

We do sometimes try to avoid false positives (with guards like this), but not to the extent of going out of our way to avoid matching code we KNOW is invalid.

If someone says "hey this shouldn't be highlighted" we say "sorry, but your code is buggy, so fix it first".

spinkney:
🎉

src/languages/stan.js
spinkney: cool, sounds good

joshgoebel (approved):

rok-cesnovar:
Thanks all for the great work!

PR #3414, opened 2021-11-30
enh(js/ts) fix => async function title highlights

### LINKED ISSUE #3405: (js) Async arrow function should be highlighted as function
**Describe the issue**
Async arrow function should be highlighted as function, instead it is not highlighted at all. This is the most used shape for functions in modern Javascript syntax (instead of `function name() {}`)

js / ts

`highlight` specifically asking for `javascript` language

**Sample Code to Reproduce**

In the examples below, `init` is a function name. The function is declared as arrow function in Javascript.

**✅  Example without `async` (works well) :**

```ts
const init = () => {
  // something
}
```

Screen Shot 2021-11-21 at 22 31 56
Screen Shot 2021-11-21 at 22 31 56

**❌ Example with `async` (does not recognize function name) :** 

```ts
const init = async () => {
  // something
}
```

Screen Shot 2021-11-21 at 22 32 35
Screen Shot 2021-11-21 at 22 32 30

**Expected behavior**
The presence or absence of `async` should not change function identification. In the example above, `init` should be colored blue in both cases.

joshgoebel:
Resolves #3405.

### Checklist


allejo (approved):
image

Works for the syntaxes I tried :+1:

PR #3416, opened 2021-12-03
enh(nsis) Update variables pattern

idleberg:
Updates pattern for NSIS variable to allow `.`-characters. The bundled MUI2 header has been [making use of this](https://github.com/kichik/nsis/blob/master/Contrib/Modern%20UI%202/Interface.nsh#L11-L25) for years.

### Changes

Allow `.` in variables regex pattern

### Checklist

### DIFF
--- a/src/languages/nsis.js
+++ b/src/languages/nsis.js
@@ -8,6 +8,7 @@ Website: https://nsis.sourceforge.io/Main_Page
 import * as regex from '../lib/regex.js';
 
 export default function(hljs) {
+  const regex = hljs.regex;
   const LANGUAGE_CONSTANTS = [
     "ADMINTOOLS",
     "APPDATA",
@@ -493,7 +494,7 @@ export default function(hljs) {
     "zlib"
   ];
 
-  const FUNCTION_DEF = {
+  const FUNCTION_DEFINITION = {
     match: [
       /Function/,
       /\s+/,
@@ -502,6 +503,21 @@ export default function(hljs) {
     scope: {
       1: "keyword",
       3: "title.function"
+    }
+  };
+
+  const VARIABLE_NAME_RE = regex.concat(hljs.IDENT_RE, "(\.", hljs.IDENT_RE, ")*");
+  const VARIABLE_DEFINITION = {
+    match: [
+      /Var/,
+      /\s+/,
+      /(?:\/GLOBAL\s+)?/,
+      VARIABLE_NAME_RE
+    ],
+    scope: {
+      1: "keyword",
+      3: "params",
+      4: "variable"
     }
   };
 
@@ -522,7 +538,8 @@ export default function(hljs) {
           relevance: 0
         }
       ),
-      FUNCTION_DEF,
+      VARIABLE_DEFINITION,
+      FUNCTION_DEFINITION,
       {
         beginKeywords: 'Function PageEx Section SectionGroup FunctionEnd SectionEnd',
       },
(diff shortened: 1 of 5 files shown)


src/languages/nsis.js
```diff
@@ -159,7 +159,7 @@ export default function(hljs) {
   const VARIABLES = {
     // $variables
     className: 'variable',
-    begin: /\$+\w+/,
+    begin: /\$+[\w\.]+/,
```
joshgoebel, line 162: Can variables BEGIN with a `.`?  If not we probably need `\w[\w.]+` no?

joshgoebel:
Also adding some `markup` tests for this change would be good.

src/languages/nsis.js
idleberg: Did a quick check and you are right, it's not allowed

joshgoebel:
Please add `markup/nsis/variables.txt` and you can just borrow a snippet from that header if you like.

idleberg:
@joshgoebel I've committed two simple test cases. However, writing them revealed some bugs I'm not sure how to fix.

The following tests will fail because they're in conflict with existing rules:

## Literals

```nsis
Var Custom.Variable
StrCpy $Custom.Variable "Hello World"
MessageBox MB_OK "$$Custom.Variable is: $Custom.Variable"
````

This test will fail since `custom` is detected as a literal

```html
<span class="hljs-keyword">Var</span> <span class="hljs-literal">Custom</span>.Variable<span class="hljs-keyword">StrCpy</span> <span class="hljs-variable">$Custom.Variable</span> <span class="hljs-string">&quot;Hello World&quot;</span>
<span class="hljs-keyword">MessageBox</span> <span class="hljs-params">MB_OK</span> <span class="hljs-string">&quot;<span class="hljs-meta">$$</span>Custom.Variable is: <span class="hljs-variable">$Custom.Variable</span>&quot;</span>
```
## Numbers

```nsis
StrCpy $0 "Hello World"
MessageBox MB_OK "$$0 is: $0"
```

Again, this will fail since 0 is detected as a number

```html
<span class="hljs-keyword">StrCpy</span> $<span class="hljs-number">0</span> <span class="hljs-string">&quot;Hello World&quot;</span>
<span class="hljs-keyword">MessageBox</span> <span class="hljs-params">MB_OK</span> <span class="hljs-string">&quot;<span class="hljs-meta">$$</span>0 is: $0&quot;</span>
```

joshgoebel:
Screen Shot 2021-12-03 at 5 12 00 PM

For `custom` I think _perhaps_ we need a special rule for Var?

```
Var [VARIABLE_NAME_RE]
```

Rule... we can do that with a multi-matcher now:

```
match: [ 
  /Var/,
  /\s+/,
  VARIABLE_NAME_RE
],
scope: { 1: "keyword", 3: "variable" }
```

If we wanted to highlight the definition also... thoughts?  Is Var that simple?

idleberg:
> If we wanted to highlight the definition as `variable` also (others we could simply match it but apply no scope)... thoughts?

I think that would be awesome

> Is Var that simple in actual usage?

There's an optional parameter for `Var`, so it's either one of the following:

```nsis
# This can be declared at top-level only
Var MyVariable

# This can be declared at block-level only
Var /GLOBAL MyVariable
```

joshgoebel:
In that can what is the scope of `/GLOBAL`?

idleberg:
> In that case what is the scope of `/GLOBAL`? params?
> 
> Screen Shot 2021-12-03 at 6 18 43 PM

Should be fine for now, we can add more specific one later (if that's desired)

joshgoebel (approved):

PR #3417, opened 2021-12-03
enh(nsis) Update defines pattern

idleberg:
Updates pattern for NSIS defines to allow `!`-characters. This example from the NSIS wiki has an [example](https://nsis.sourceforge.io/Check_if_a_file_exists_at_compile_time) that uses this.

### Changes

Allow `!` in defines regex pattern

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,7 +32,7 @@ Themes:
 [Björn Ebbinghaus]: https://github.com/MrEbbinghaus
 [Josh Goebel]: https://github.com/joshgoebel
 [Samia Ali]: https://github.com/samiaab1990
-
+[idleberg]: https://github.com/idleberg
 
 ## Version 11.3.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ Grammars:
 - enh(swift) add SE-0290 unavailability condition (#3382) [Bradley Mackey][]
 - fix(fsharp) Highlight operators, match type names only in type annotations, support quoted identifiers, and other smaller fixes. [Melvyn Laïly][]
 - enh(java) add `sealed` and `non-sealed` keywords (#3386) [Bradley Mackey][]
+- enh(nsis) Update defines pattern to allow `!` (#3417) [idleberg][]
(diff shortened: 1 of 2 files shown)


joshgoebel:
I see a lot of `!` there (following your link), can you point me to exactly which part we're talking about?

idleberg:
> I see a lot of `!` there (following your link), can you point me to exactly which part we're talking about?

This is the relevant line:

```nsis
!define !defineifexist "!insertmacro !defineifexist"
```

It's basically a shortcut that assigns the `!insertmacro` instruction to a definition. The following two do the same:

```nsis
# Default !insertmacro syntax
!insertmacro !defineexists [params]

# Shortcut for !insertmacro
${!defineexist} [params]
```

PS: Unlike [variables](https://github.com/highlightjs/highlight.js/pull/3416#discussion_r761610781), defines can begin with any allowed characters – including `!` and `.`

idleberg:
Since I think the syntax in the linked example is kind of terrible, here's a version that only uses `!` where necessary:

```nsis
!macro DefineIfExist _VAR_NAME _FILE_NAME
	# macro code omitted for brevity
!macroend
!define !defineifexist "!insertmacro DefineIfExist"
```

In NSIS, all compile-time commands begin with a `!`. I guess the author's intention to name the definition as `!defineifexists` was to reflect that his macro does compile-time operations. It could be named anything.

CHANGES.md
```diff
@@ -16,6 +16,7 @@ Grammars:
   - enh(clojure) Add `regex` mode to regex literal
   - fix(clojure) Remove inconsistent/broken highlighting for metadata
   - enh(clojure) Add `punctuation` mode for commas.
+- enh(nsis) Update defines pattern to allow `!` (#3417) [idleberg][]
```
joshgoebel, line 19: Please add your name around line ~30 also.  Each version needs the name repeated, even if you've contributed in the past.

joshgoebel (changes requested):
Needs name attribution for changelog.

joshgoebel (approved):

PR #3420, opened 2021-12-04
enh(nsis) Update language string pattern

idleberg:
Updates pattern for NSIS language strings to allow `!`-characters. This is basically the same as #3417, but for language strings. Unfortunately, I can't produce any example to back this up.

### Changes

Allow `!` in language strings regex pattern

### Checklist


joshgoebel:
> Unfortunately, I can't produce any example to back this up.

Then how do we know this is needed at all? I'm a little confused. :)

idleberg:
> Then how do we know this is needed at all? I'm a little confused. :)

It's too unusual to find results in the wild (I tried both the NSIS wiki and GitHub's code search), but it's valid code. I created a [demo repository](https://github.com/idleberg/hljs-nsis-demo) where such code is compiled in a GitHub action. If `!` wasn't valid, the compiler would throw an error.

joshgoebel (approved):

PR #3422, opened 2021-12-06
enh(php) Left and right-side of double colon

wkania:
Double colon aka [Scope Resolution Operator](https://www.php.net/manual/en/language.oop5.paamayim-nekudotayim.php) is used in a few contexts:
- access static variable or method of the class: `Example::$uuid;  self::$a;  Example::test();  $a::test();` 
- constant reference: `Foo::Test;`
- access overridden properties or methods of a class: `parent::$a;  parent::__construct();`
- enum value reference: `Foo::Test;`
- class name reference: `self::class;  static::class;  parent::class;  Foo::class; $a::class`

### Changes
- declare and use `IDENT_RE` for valid labels instead of `w+`
- declaration of PHP [constants](https://www.php.net/manual/en/language.oop5.constants.php) were not highlighted. Now they use the same class as variables. 
- enum and constant reference were not highlighted. Now they use the same class as variables. 
- Class name references are highlighted as `variable.language`.
[class](https://wiki.php.net/rfc/class_name_literal_on_object) is a special case of the constant, it's also a reserved keyword.
- left-side class name highlighting

Info about rules for valid labels ([variables](https://www.php.net/manual/en/language.variables.basics.php), [constants](https://www.php.net/manual/en/language.constants.php), and other names).

> The name of a constant follows the same rules as any label in PHP. A valid constant name starts with a letter or underscore, followed by any number of letters, numbers, or underscores. As a regular expression, it would be expressed thusly: ^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -317,8 +317,8 @@ export default function(hljs) {
           regex.concat(WHITESPACE, "+"),
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-          regex.concat("\\\\?", IDENT_RE),
-          regex.concat(WHITESPACE, "*\\("),
+          regex.concat(/\\?/, IDENT_RE),
+          regex.concat(WHITESPACE, "*", /\(/),
         ],
         scope: {
           1: "keyword",
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -447,15 +447,6 @@ export default function(hljs) {
           3: "variable.constant",
         },
       },
-      {
-        // swallow composed identifiers to avoid parsing them as keywords
-        match: regex.concat(
-          /->+/,
-          IDENT_RE,
-          regex.concat("(?!", WHITESPACE, "*\\()")
-        ),
-        // scope:"wrong"
-      },
       CONSTRUCTOR_CALL,
       {
         scope: 'function',
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -12,12 +12,13 @@ Category: common
  * */
 export default function(hljs) {
   const regex = hljs.regex;
+  const IDENT_RE = '([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
+    // negative look-ahead tries to avoid matching patterns that are not
+    // Perl at all like $ident$, @ident@, etc.
+    '(?![A-Za-z0-9])(?![$]))';
   const VARIABLE = {
     scope: 'variable',
-    match: '\\$+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
-      // negative look-ahead tries to avoid matching patterns that are not
-      // Perl at all like $ident$, @ident@, etc.
-      `(?![A-Za-z0-9])(?![$])`
+    match: '\\$+' + IDENT_RE,
   };
   const PREPROCESSOR = {
     scope: 'meta',
@@ -313,7 +314,7 @@ export default function(hljs) {
           regex.concat(WHITESPACE, "+"),
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-          /\\?\w+/,
+          regex.concat("\\\\?", IDENT_RE, "+"),
           regex.concat(WHITESPACE, "*\\("),
         ],
         scope: {
@@ -330,7 +331,7 @@ export default function(hljs) {
       /\b/,
       // to prevent keywords from being confused as the function title
       regex.concat("(?!fn\\b|function\\b|", normalizeKeywords(KWS).join("\\b|"), "|", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-      /\w+/,
+      regex.concat(IDENT_RE, "+"),
       regex.concat(WHITESPACE, "*"),
       regex.lookahead(/(?=\()/)
     ],
@@ -351,8 +352,8 @@ export default function(hljs) {
         {
           contains: [
             {
-              className: 'doctag',
-              begin: '@[A-Za-z]+'
+              scope: 'doctag',
+              match: '@[A-Za-z]+'
             }
           ]
         }
@@ -382,9 +383,9 @@ export default function(hljs) {
       {
         // swallow composed identifiers to avoid parsing them as keywords
         match: regex.concat(
-          /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/,
-          regex.concat("(?!", WHITESPACE, "*\\()"),
-          /(?![a-zA-Z0-9_\x7f-\xff])/
+          /(::|->)+/,
+          IDENT_RE,
+          regex.concat("(?!", WHITESPACE, "*\\()")
         ),
         // scope:"wrong"
       },
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -12,10 +12,13 @@ Category: common
  * */
 export default function(hljs) {
   const regex = hljs.regex;
-  const IDENT_RE = '([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
+  const IDENT_RE_CORE = '[a-zA-Z0-9_\x7f-\xff]*' +
     // negative look-ahead tries to avoid matching patterns that are not
     // Perl at all like $ident$, @ident@, etc.
     '(?![A-Za-z0-9])(?![$]))';
+  const IDENT_RE = regex.concat("([a-zA-Z_\\x7f-\\xff]", IDENT_RE_CORE);
+  // Will not detect camelCase classes
+  const PASCAL_CASE_CLASS_NAME_RE = regex.concat("([A-Z]", IDENT_RE_CORE);
   const VARIABLE = {
     scope: 'variable',
     match: '\\$+' + IDENT_RE,
@@ -47,7 +50,7 @@ export default function(hljs) {
     end: /[ \t]*(\w+)\b/,
     contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
   });
-  // list of valid whitespaces because non-breaking space might be part of a name
+  // list of valid whitespaces because non-breaking space might be part of a IDENT_RE
   const WHITESPACE = '[ \t\n]';
   const STRING = {
     scope: 'string',
@@ -314,7 +317,7 @@ export default function(hljs) {
           regex.concat(WHITESPACE, "+"),
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-          regex.concat("\\\\?", IDENT_RE, "+"),
+          regex.concat("\\\\?", IDENT_RE),
           regex.concat(WHITESPACE, "*\\("),
         ],
         scope: {
@@ -331,13 +334,64 @@ export default function(hljs) {
       /\b/,
       // to prevent keywords from being confused as the function title
       regex.concat("(?!fn\\b|function\\b|", normalizeKeywords(KWS).join("\\b|"), "|", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-      regex.concat(IDENT_RE, "+"),
+      IDENT_RE,
       regex.concat(WHITESPACE, "*"),
       regex.lookahead(/(?=\()/)
     ],
     scope: {
       3: "title.function.invoke",
     }
+  };
+
+  const CONSTANT_REFERENCE = regex.concat(IDENT_RE, "\\b(?!\\()");
+
+  const LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON = {
+    variants: [
+      {
+        match: [
+          regex.concat(
+            /::/,
+            regex.lookahead(/(?!class\b)/)
+          ),
+          CONSTANT_REFERENCE,
+        ],
+        scope: {
+          2: "variable.constant",
+        },
+      },
+      {
+        match: [
+          /::/,
+          /class/,
+        ],
+        scope: {
+          2: "variable.language",
+        },
+      },
+      {
+        match: [
+          PASCAL_CASE_CLASS_NAME_RE,
+          regex.concat(
+            "::",
+            regex.lookahead(/(?!class\b)/)
+          ),
+        ],
+        scope: {
+          1: "title.class",
+        },
+      },
+      {
+        match: [
+          PASCAL_CASE_CLASS_NAME_RE,
+          /::/,
+          /class/,
+        ],
+        scope: {
+          1: "title.class",
+          3: "variable.language",
+        },
+      }
+    ]
   };
 
   return {
@@ -380,16 +434,17 @@ export default function(hljs) {
       },
       VARIABLE,
       FUNCTION_INVOKE,
+      LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
       {
         match: [
           /const/,
-          regex.concat(WHITESPACE, "+"),
+          /\s/,
           IDENT_RE,
-          regex.concat(WHITESPACE, "*="),
+          /\s*=/,
         ],
         scope: {
           1: "keyword",
-          3: "variable",
+          3: "variable.constant",
         },
       },
       {
@@ -425,6 +480,7 @@ export default function(hljs) {
             contains: [
               'self',
               VARIABLE,
+              LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
               hljs.C_BLOCK_COMMENT_MODE,
               STRING,
               NUMBER
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -450,7 +450,7 @@ export default function(hljs) {
       {
         // swallow composed identifiers to avoid parsing them as keywords
         match: regex.concat(
-          /(::|->)+/,
+          /->+/,
           IDENT_RE,
           regex.concat("(?!", WHITESPACE, "*\\()")
         ),
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -381,6 +381,18 @@ export default function(hljs) {
       VARIABLE,
       FUNCTION_INVOKE,
       {
+        match: [
+          /const/,
+          regex.concat(WHITESPACE, "+"),
+          IDENT_RE,
+          regex.concat(WHITESPACE, "*="),
+        ],
+        scope: {
+          1: "keyword",
+          3: "variable",
+        },
+      },
+      {
         // swallow composed identifiers to avoid parsing them as keywords
         match: regex.concat(
           /(::|->)+/,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@ These changes should be for the better and should not be super noticeable but if
 
 Grammars:
 
+- enh(php) Left and right-side of double colon [Wojciech Kania][]
+- enh(php) add PHP constants [Wojciech Kania][]
 - enh(php) add PHP 8.1 keywords [Wojciech Kania][]
 - fix(cpp) fix `vector<<` template false positive (#3437) [Josh Goebel][]
 - enh(php) support First-class Callable Syntax (#3427) [Wojciech Kania][]
--- a/test/markup/php/functions.expect.txt
+++ b/test/markup/php/functions.expect.txt
@@ -13,7 +13,7 @@
 <span class="hljs-variable">$date</span> = <span class="hljs-keyword">new</span> <span class="hljs-title class_">DateTimeImmutable</span> ();
 <span class="hljs-variable">$date</span>-&gt;<span class="hljs-title function_ invoke__">format</span>(<span class="hljs-string">&#x27;Y-m-d&#x27;</span>);
 
-DateTimeImmutable::<span class="hljs-title function_ invoke__">createFromMutable</span>(<span class="hljs-keyword">new</span> <span class="hljs-title class_">\DateTime</span>(<span class="hljs-string">&#x27;now&#x27;</span>));
+<span class="hljs-title class_">DateTimeImmutable</span>::<span class="hljs-title function_ invoke__">createFromMutable</span>(<span class="hljs-keyword">new</span> <span class="hljs-title class_">\DateTime</span>(<span class="hljs-string">&#x27;now&#x27;</span>));
 
 <span class="hljs-title function_ invoke__">str_contains</span> (\<span class="hljs-title function_ invoke__">strtoupper</span>(<span class="hljs-title function_ invoke__">substr</span>(<span class="hljs-string">&#x27;abcdef&#x27;</span>, -<span class="hljs-number">2</span>), <span class="hljs-string">&#x27;f&#x27;</span>));
 
--- a/test/markup/php/titles.expect.txt
+++ b/test/markup/php/titles.expect.txt
@@ -1,8 +1,8 @@
-<span class="hljs-keyword">final</span> <span class="hljs-class"><span class="hljs-keyword">class</span> <span class="hljs-title">Example</span> </span>{
-    <span class="hljs-keyword">const</span> <span class="hljs-variable">FOO</span>=<span class="hljs-string">&#x27;foo&#x27;</span>;
+<span class="hljs-keyword">final</span> <span class="hljs-class"><span class="hljs-keyword">class</span> <span class="hljs-title">Example</span> <span class="hljs-keyword">extends</span> <span class="hljs-title">Foo</span> </span>{
+    <span class="hljs-keyword">const</span> <span class="hljs-variable constant_">FOO</span>=<span class="hljs-string">&#x27;foo&#x27;</span>;
 
     <span class="hljs-keyword">public</span> <span class="hljs-function"><span class="hljs-keyword">function</span> <span class="hljs-title">__construct</span>(<span class="hljs-params">
-        <span class="hljs-keyword">public</span> <span class="hljs-keyword">readonly</span> <span class="hljs-keyword">string</span> <span class="hljs-variable">$name</span> = <span class="hljs-built_in">self</span>::<span class="hljs-variable">FOO</span>
+        <span class="hljs-keyword">public</span> <span class="hljs-keyword">readonly</span> <span class="hljs-keyword">string</span> <span class="hljs-variable">$name</span> = <span class="hljs-built_in">self</span>::<span class="hljs-variable constant_">FOO</span>
     </span>) </span>{}
 
     <span class="hljs-keyword">public</span> <span class="hljs-function"><span class="hljs-keyword">function</span> <span class="hljs-title">getClass</span>(<span class="hljs-params"></span>): <span class="hljs-title">string</span> </span>{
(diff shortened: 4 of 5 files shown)


src/languages/php.js
```diff
@@ -18,6 +18,28 @@ export default function(hljs) {
...
       `(?![A-Za-z0-9])(?![$])`
   };
+  const CONSTANT = {
+    className: "variable",
+    relevance: 0,
+    variants: [
+      {
+        begin: /const\s*/,
+        end: /\w+/,
+        excludeBegin: true
```
joshgoebel, line 28: Nope, far too broad as this would match `const 100 lines of source code here \w+` not just `const \w+`.  Please see the docs and use a multi-matcher for this: https://highlightjs.readthedocs.io/en/latest/mode-reference.html#beginscope

joshgoebel: See `wren` and other newer grammars for examples.

src/languages/php.js
```diff
@@ -18,6 +18,28 @@ export default function(hljs) {
...
+      },
+      {
+        begin: /::(?!class\b)/,
```
joshgoebel, line 31: Do you have a specific example of the false positive that cause you to see this?

src/languages/php.js
```diff
@@ -18,6 +18,28 @@ export default function(hljs) {
...
+        end: /\w+/,
+        excludeBegin: true
+      },
+      {
+        className: "keyword",
+        begin: /::/,
+        end: /class/,
+        excludeBegin: true
+      },
```
joshgoebel, line 40: Not sure what this rule is try too accomplish?

src/languages/php.js
wkania: Yes, I added [test](https://github.com/highlightjs/highlight.js/pull/3422/files#diff-a1909111fa7ad598f9584d4ed37e338d8dbd46f522d567017bb237df5c58ae64R8).
Without this regex rule, it would be marked as a variable, instead of a keyword.
```
-        <span class="hljs-keyword">return</span> DateTimeImmutable::<span class="hljs-variable">class</span>;
+        <span class="hljs-keyword">return</span> DateTimeImmutable::<span class="hljs-keyword">class</span>;
```

src/languages/php.js
wkania: [class](https://wiki.php.net/rfc/class_name_literal_on_object) is a special case of the constant, it's also a reserved keyword.

src/languages/php.js
wkania: \w+ will stop matching on first `;` or `=` which are mandatory in PHP or any non-word character, like space. I also added `relevance: 0` to not [detect](https://github.com/highlightjs/highlight.js/blob/main/test/detect/javascript/short-plain.txt)  as PHP.

I will look at beginscope tomorrow.

joshgoebel: > which are mandatory in PHP

Except this might not be parsing PHP at all... autodetect works by just parsing with everything and seeing what "works"... (and there is also no guarantee we're given VALID PHP).  Hence every rule must be as "tight" as possible.

joshgoebel: > relevance: 0

This is probably too low since `const` itself is a REAL keyword worth one point, so I'd remove it and let the default relevance of 1 apply (to the keyword, not the variable name).

src/languages/php.js
joshgoebel: Ok, I see what you're doing nwo... I'd use a multi-match here also but I'd consider going even wider and getting the class name as well.  Doesn't `::class` imply we have a class?

```
[DateTimeImmutable][::][class]

DateTimeImmutable => title.class
class => keyword
```

src/languages/php.js
wkania: I thought `relevance: 0` only applies to const name and `const` as a keyword still has relevance 1.

joshgoebel: Your matching it so the keyword engine will never see it. Unless excludeBegin does the trick.  I don't actually know off the top of my head. The style you're using is a bit different. 

Once you use multi match though you'll definitely need to handle the keyword yourself so no relevance hacks should be needed.

src/languages/php.js
wkania: > Doesn't ::class imply we have a class?

Nope. It can be also the keyword `self` or `static` that is an alias for the class in which they are called or an object (variable).



    public function getClassFromSelf(): string {
        return self::class;
    }
    public static function getClassFromStaic(): string {
        return static::class;
    }
    public function getClass(): string {
        $date = new DateTimeImmutable();
        return date::class;
    }

joshgoebel: Does the uppercase _by convention_ generally mean we have a class?  ie `Car::class`?  We do this for other grammars if it's a strong convention even though nothing requires classes to be uppercase in say Javascript.

wkania: > Does the uppercase by convention generally mean we have a class? 



It's a good practice, so I implemented it that way.

wkania: About the `self` and `static` I didn't notice earlier that they are marked as `hljs-built_in` and not keywords.
My current implementation changed this behavior inside the function. Now they are marked as a class. Tomorrow I will try to find this regression.

src/languages/php.js
wkania: I removed `relevance` and changed all `className` to `scope`.

src/languages/php.js
```diff
@@ -11,23 +11,101 @@ Category: common
...
+        beginScope: {
+          1: "keyword",
+          3: "title.class",
+        },
+      }
+    ]
+  };
   const PREPROCESSOR = {
-    className: 'meta',
+    scope: 'meta',
```
joshgoebel, line 100: In the future please do this type of separate "maintenance" renaming of things in a separate commit to make the actual changes easier to see/review.

src/languages/php.js
```diff
@@ -11,23 +11,101 @@ Category: common
...
+  const CONSTRUCTOR = {
+    variants: [
+      {
+        begin: [
+          /new/,
+          /\s+/,
+          /\\?[A-Z]\w*/,
+          /\s*\(/,
+        ],
+        beginScope: {
```
joshgoebel, line 92:
```suggestion
        match: [
          /new/,
          /\s+/,
          /\\?[A-Z]\w*/,
          /\s*\(/,
        ],
        scope: {
```

src/languages/php.js
```diff
@@ -11,23 +11,101 @@ Category: common
...
     begin: '\\$+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
       // negative look-ahead tries to avoid matching patterns that are not
       // Perl at all like $ident$, @ident@, etc.
       `(?![A-Za-z0-9])(?![$])`
   };
+  const CONSTANT = {
+    variants: [
+      {
+        begin: [
+          /const\s+/,
```
joshgoebel, line 26: Please split to exclude the `\s+` from the keyword scope.

src/languages/php.js
```diff
@@ -11,23 +11,101 @@ Category: common
...
+          /::/,
+          /class/,
+        ],
+        beginScope: {
+          2: "keyword",
+        },
+      },
+      {
+        begin: [
+          /[A-Z]\w*/,
```
joshgoebel, line 72: Lets wrap this in a constant since it's re-used multiple times.

src/languages/php.js
```diff
@@ -11,23 +11,101 @@ Category: common
...
+          1: "title.class",
+          2: "variable",
+        },
+      },
+      {
+        begin: [
+          /::/,
+          /class/,
+        ],
+        beginScope: {
```
joshgoebel, line 66:
```suggestion
        match: [
          /::/,
          /class/,
        ],
        scope: {
```

If there is no `end` please change _all_ of these to match/scope pairs.

joshgoebel (changes requested):
Looking pretty good.  I dropped a few notes.

src/languages/php.js
wkania: Fixed everywhere.

src/languages/php.js
wkania: done

src/languages/php.js
wkania: done

src/languages/php.js
wkania: 1.
Because now I have an error, I found the root of my regression. The match is case-insensitive because it uses `i` by default. It totally ignores the uppercase rule. Can I disable it or is it some kind of bug?
``` 
Error: php/[9]/begin: Polynomial backtracking. By repeating any character that matches /\xa0/i, an attack string can be created.

    \s+)([a-zA-Z_-ÿ][a-zA-Z0-9_-ÿ]*
    ^~~              ^~~~~~~~~~~~~~~~

Full pattern:
/(const)(\s+)([a-zA-Z_-ÿ][a-zA-Z0-9_-ÿ]*(?![A-Za-z0-9])(?![$]))(\s*=)/im
```

2. I also added a new const for name and in MR description added info about names.

wkania:
> Looking pretty good. I dropped a few notes.

I fixed issues and added some questions.

src/languages/php.js
```diff
@@ -11,23 +11,105 @@ Category: common
...
+        ],
+        scope: {
+          2: "variable",
+        },
+      },
+      {
+        match: [
+          PASCAL_CASE_CLASS_NAME,
+          regex.concat(
+            "::",
```
joshgoebel, line 55: regex please

src/languages/php.js
```diff
@@ -11,23 +11,105 @@ Category: common
  * @returns {LanguageDetail}
  * */
 export default function(hljs) {
+  const regex = hljs.regex;
+  const LABEL = '[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
```
joshgoebel, line 15: Is this is a valid ident regex then we'd commonly name it `IDENT_RE`

src/languages/php.js
```diff
@@ -11,23 +11,105 @@ Category: common
...
   const VARIABLE = {
-    className: 'variable',
-    begin: '\\$+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
-      // negative look-ahead tries to avoid matching patterns that are not
-      // Perl at all like $ident$, @ident@, etc.
-      `(?![A-Za-z0-9])(?![$])`
+    scope: 'variable',
+    match: '\\$+' + LABEL,
+  };
+  const CONSTANT_CALL = regex.concat(LABEL, "\\b(?!\\()");
```
joshgoebel, line 24: What is the meaning of "CALL" here?  I thought it was a call/dispatch but then your regex seems to say "NOT" followed by `(` Rather than [must be followed by `(`], so I"m confused.

src/languages/php.js
joshgoebel: Can we just make a separate rule for `PASCAL_CASE_CLASS_NAME` entirely to highlight class names in ALL contexts?  That'd remove the need for a few of these, would it not?

src/languages/php.js
```diff
@@ -11,23 +11,105 @@ Category: common
...
+  const CONSTANT = {
+    variants: [
+      {
+        match: [
```
joshgoebel, line 28: As this is a declaration rather than a reference I think it likely needs it's own rule rather than being mixed into this huge `variants`.

src/languages/php.js
wkania: PHP has:
- constant calling
`DateTimeImmutable::ISO8601;`
- static function invoke
`DateTimeImmutable::createFromMutable(new \DateTime());`


I do not want static function invoke to be highlighted as constant

src/languages/php.js
wkania: Can you explain it more or add some link to the code/example?

src/languages/php.js
wkania: > regex please

You mean: change `"::"` to `/::/`?

> Can we just make a separate rule for PASCAL_CASE_CLASS_NAME entirely to highlight class names in ALL contexts? That'd remove the need for a few of these, would it not?

This could highlight code that is not a class. Like enum, constant, or keywords: Closure Generator. 
Doesn't sound good.

src/languages/php.js
wkania: @joshgoebel Do you know anything about this case-insensitive matching?

joshgoebel: PHP is `case_insensitive: true,` so EVERY regex is `/i`, there i no need to specify and it's impossible to have ANY regex that is case-dependent.

So forget about using `CamelCase` rules... I've read not ALL of PHP is case-sensitive though... perhaps we could switch it to "case-sensitive" and try and deal with the exceptions by hand?  How familiar are you with PHP's case sensitivity overall?

src/languages/php.js
joshgoebel: > You mean: change "::" to /::/?

Yes.

> This could highlight code that is not a class. Like enum, constant, or keywords: Closure Generator.


I feel like i'd need examples in context to comment, pretend I'm not familiar with PHP. But lets take enum quickly (because I just googled it).

```
enum Status
{
    case DRAFT;
    case PUBLISHED;
    case ARCHIVED;
}
```

We could match `enum [klass]` with a special rule but also `Status` on it's own (by it's casing) would qualify as a `title.class` (all higher-level class-like things do)... the enum values are UPPER case so they  could be flagged as constants... now if someone wanted to use camel case for enum values then we would have a problem since they would be flagged as `title.class` vs `variable.constant`...

So perhaps this isn't a good idea, I don't know enough PHP to say for certain.

Of course all this talk of casing would require us making PHP case-sensitive and then dealing with the fall-out of that...

src/languages/php.js
joshgoebel: In many other grammars we'd:

- FIRST match the function invoke as `title.function`...
- then match for constants (which now will not require the look-ahead since those are already matched)

I'd rename this to `CONSTANT_REFERENCE` as I think that's more in line with the rest of the codebase... if "call" means something magical in PHP you can add a comment explaining the context further.

src/languages/php.js
joshgoebel: I just mean it doesn't belong with all the rest, make a singular rule:

```
const CONSTANT_DEFINITION = {
   match: [
           /const/,
           /\s+/,
           LABEL,
           /\s*=/,
}
```

joshgoebel: All the rest deal with `x::y`, so it make sense to group them.

src/languages/php.js
joshgoebel: I wonder if we couldn't deal with this by just exhausting all the non-class possibilities... ie:

- `self::class`
- `static::class`
- `$[variable name]::class`
- [whatever]::class` (must be a class, no other options)

Is that perhaps doable?

joshgoebel: There are also per-rule JS callbacks that could manually evaluate the names and make a decision, but we don't encourage their usage (speed)... so I'd likely first want to:

- see if there is some way to make this sorta-work case-insensative
- see if the entire grammar could work as case-sensative

And only if both those fail look into using call-backs to do the class or not-class classification.

joshgoebel:
If you could please open a new PR from `wkania` instead of `wkania-forks` so that I can push changes also that would be greatly appreciated.  GitHub is very picky about such things.  Or else allow pushing, but it's likely the fact that your using an alternative account is the issue.

joshgoebel:
Alternatively you could perhaps add me with repo write permission against your fork `wkania-forks/highlight.js`....

src/languages/php.js
joshgoebel: Also now that I'm thinking about it hard  `static::class` the class here does not look like a keyword... ie, the context it's not being used as a keyword... perhaps `variable.language` or `variable.constant` would be more appropriate... see the CSS class docs. 

IE, in Ruby:

```rb
# here class is a keyword
class Dog
end

# here it's just a built-in property of Dog that returns Dog's meta-class
a = Dog.class
```

The first should be `title.class`, the second `property`.

src/languages/php.js
wkania: So if part of the code will be matched as function invoke then another rule will not replace it with something else?
CONSTANT_REFERENCE looks good.

src/languages/php.js
wkania: > PHP is `case_insensitive: true,` so EVERY regex is `/i`, there i no need to specify and it's impossible to have ANY regex that is case-dependent. I forgot this or I'd have mentioned it earlier.

Now that's news.  How could I miss this `case_insensitive: true,` :)

I checked the current implementation of the PHP highlighter and part of camel case keywords are affected by this. I would separate keywords into case-insensitive (false, if)  and case-sensitive (ArithmeticError ArrayIterator).  Case-insensitive function and class names are not affected by this.

joshgoebel: >  I would separate keywords into case-insensitive (false, if) and case-sensitive 

Except we have no current way to do that [case insensitive in a case sensitive grammar] - the setting is at the grammar level (because we concat every regex into a single huge one)... I could imagine a possible enhancement to JUST `keywords` to allow "as given" or "all lowercase" (would allow the two most common variants, which might in many cases be sufficient)... but that doesn't necessarily help with any custom rules, which would need every letter individually split out... of course we could potentially write a helper for that...

What do you think of the "as given" + all lowercase... ie, supporting "false" and "FALSE" but not "fAlSe"?

I'm trying to figure out what could be done here to wrap this PR up and then circle back to the larger discussion of casing as a sep concern.

src/languages/php.js
joshgoebel: > So if part of the code will be matched as function invoke then another rule will not replace it with something else?

Matching is sequential and "consumes" the content... two rules cannot match the exact same content (though keywords are matched INSIDE other content, so they can be different)... there are exceptions like `returnBegin` but we're slowly deprecating them and don't encourage their use... look-ahead is the only way to match without consuming.

src/languages/php.js
```diff
@@ -313,7 +314,7 @@ export default function(hljs) {
           regex.concat(WHITESPACE, "+"),
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-          /\\?\w+/,
+          regex.concat("\\\\?", IDENT_RE, "+"),
```
joshgoebel, line 317: Idea already includes `[at least one][more*]`... why do we need `+`?

src/languages/php.js
```diff
@@ -339,6 +340,62 @@ export default function(hljs) {
     }
   };
 
+  // This will not detect camelCase classes
+  const PASCAL_CASE_CLASS_NAME = regex.concat("[A-Z]", IDENT_RE);
+
+  const CONSTANT_REFERENCE = regex.concat(IDENT_RE, "\\b(?!\\()");
```
joshgoebel, line 346: How is this a constant reference?  In other languages our constant rule is ALL_CAPS_CASE.

joshgoebel: And if it's truly a constant we'd use `variable.constant`...

src/languages/php.js
```diff
@@ -379,12 +436,25 @@ export default function(hljs) {
       },
       VARIABLE,
       FUNCTION_INVOKE,
+      LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+      {
+        match: [
+          /const/,
+          regex.concat(WHITESPACE, "+"),
```
joshgoebel, line 443: This is super verbose please use `\s` when you can, and only fall back to WHITESPACE when there are issues.

wkania: > This is super verbose please use `\s` when you can, and only fall back to WHITESPACE when there are issues.

Yes, it is. I never saw it (nobody would do that), but the language allows to start the const name with [non-breaking](http://sandbox.onlinephpfunctions.com/code/9780e686f54fad41c69cf310500490cb6eec1384) space. I wanted to use the same rule for whitespace everywhere.  
Also, there is a case of `Polynomial backtracking` (with \s+) and some code formatting rules might have multiple spaces:  
```
const FOO                       = 1;
const SOMETHING                 = 2;

File: src/languages/php.js (hunk 14 above)
Comment (wkania) [reply src/languages/php.js:346]: > How is this a constant reference? In other languages our constant rule is ALL_CAPS_CASE. Is anything following a `::` just a constant in PHP?

Language allows also lower case or any other, but probably almost all use ALL_CAPS_CASE.  
It is just IDENT_RE that can't be followed by (. This distinguishes it from the static method call. I left examples in the PR description. In this context, it's just a helper that is part of a few matching rules.


> And if it's truly a constant we'd use `variable.constant`...

Ok, I guess accessing Enum value is also `variable.constant`, because we can't distinguish those.   

Constant declaration and Enum values also should be marked with `variable.constant`?

File: src/languages/php.js (hunk 15 above)
Comment (joshgoebel) [reply src/languages/php.js:443]: >  but the language allows to start the const name with non-breaking space

I'm not sure we care about this or this matter since the highlighting for all intents and purposes is identical.

> Also, there is a case of Polynomial backtracking (with \s+) 

These can usually be fixed (and you can leave it to me if you want).  So again, let's use `\s` (when it works, which is 99% of the time) and only fall back to more complex usages when it fails to work.

File: src/languages/php.js (hunk 3 above)
Comment (wkania) [reply src/languages/php.js]: @joshgoebel Changed to `variable.constant`

File: src/languages/php.js (hunk 10 above)
Comment (wkania) [reply src/languages/php.js]: Changed

File: src/languages/php.js (hunk 13 above)
Comment (wkania) [reply src/languages/php.js:320]: @joshgoebel Fixed

**Comment by wkania** [comment]
@joshgoebel Pls review.
I added all use cases in the description of the PR. Tests use variations of those cases.

File: src/languages/php.js (hunk 14 above)
Comment (joshgoebel) [reply src/languages/php.js:346]: > It is just IDENT_RE that can't be followed by (.

That's that's a "function dispatch" or something, not a "constant"... constant to me means `const` or read only or `UPPERCASE` by convention in may languages...

> Constant declaration and Enum values also should be marked with variable.constant?

Perhaps if you could detect they are enum values... I dunno what the PHP syntax is... for most languages it's impossible to do this 100% correctly so we usually have a UPPERCASE constant rule for languages where it's a string pattern and then only catch others in very explicit cases like `class name`, etc... where it's clean name is constant despite the weird casing.

File: src/languages/php.js [hunk 16]
```diff
@@ -379,12 +434,25 @@ export default function(hljs) {
       },
       VARIABLE,
       FUNCTION_INVOKE,
+      LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+      {
+        match: [
+          /const/,
+          /\s/,
+          IDENT_RE,
+          /\s*=/,
```
joshgoebel, line 443: Should be a look-ahead, I'm not sure why we need to consume it?

src/languages/php.js
```diff
@@ -379,12 +434,25 @@ export default function(hljs) {
...
+      },
       {
         // swallow composed identifiers to avoid parsing them as keywords
         match: regex.concat(
-          /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/,
-          regex.concat("(?!", WHITESPACE, "*\\()"),
-          /(?![a-zA-Z0-9_\x7f-\xff])/
+          /(::|->)+/,
+          IDENT_RE,
+          regex.concat("(?!", WHITESPACE, "*\\()")
```
joshgoebel, line 455: Was there an issue with line 387?

src/languages/php.js
```diff
@@ -313,7 +317,7 @@ export default function(hljs) {
           regex.concat(WHITESPACE, "+"),
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-          /\\?\w+/,
+          regex.concat("\\\\?", IDENT_RE),
```
joshgoebel, line 320: Can you point me to the test case this is covering?  I didn't see anything with a `\?`

src/languages/php.js
```diff
@@ -313,7 +317,7 @@ export default function(hljs) {
...
           regex.concat(WHITESPACE, "*\\("),
```
joshgoebel, line 321:
```suggestion
          regex.concat(/\\?/, IDENT_RE),
          regex.concat(WHITESPACE, "*", /\(/),
```

Not sure about that top one OTTOMH since I'm not sure what you're trying to match...

Prefer regex always over strings.

src/languages/php.js
wkania: [new \DateTime()](https://github.com/highlightjs/highlight.js/pull/3422/files#diff-60c00e73629094e4af71f4387ed9c01356d51223ba5cf0e8ad21a792426efe07R25)

src/languages/php.js
joshgoebel: Ah I just looked above, yes the enum declarations could be marked that way I suppose...

src/languages/php.js
wkania: We define again the same regex pattern instead of using an already defined one.

src/languages/php.js
joshgoebel: Ah I see  the markup test for it now.

src/languages/php.js
joshgoebel: You changed the regex pattern... did you not?  It was previously ~:

-> ident [ NOT `\s(` AND NOT IDENT ]

You removed the "AND NOT IDENT", portion did you not? Was there a reason?

src/languages/php.js
joshgoebel: Whatever, it'll probably all get squashed in the end...

src/languages/php.js
wkania: Hmm, lines 385 and 387 are the same as the regex rule of IDENT_RE. IDENT_RE only has the suffix `(?![$]))`, which was missing here (in my opinion). [Documentation](https://www.php.net/manual/en/language.variables.basics.php)  states that all labels follow the same pattern:

> Variable names follow the same rules as other labels in PHP. A valid variable name starts with a letter or underscore, followed by any number of letters, numbers, or underscores. As a regular expression, it would be expressed thus: ^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$

I also changed the position of the negative look-ahead from the middle to the end. It didn't break any test and according to my knowledge about regex, it should work the same. If you could provide me with a sample that passes the old code and fail in the new one, then I would fix it.

src/languages/php.js
wkania: Hmm, I don't know enough about the highlight js to know what's the difference.
1. Why can't we consume it? 
2. Why is look-ahead, is a better solution?

joshgoebel: Because we don't just consume things arbitrarily... we aren't highlighting it here, we only need to guarantee it's presence, not consume it.  In general we don't consume things we don't need to.

Later someone may want to add operator matching where we highlight all `=` and consuming it here would make that impossible...  it should be left int he input stream so another rule might match it later.

src/languages/php.js
joshgoebel: > It didn't break any test...

Sadly our tests are far from comprehensive, hence the careful eye I have with reviewing PRs.

This was added with: d49e9fb8fa82a43ed96333510c9c5bc1af776352

The test code added was `, 3 => \Location\Web\URI::class));` so I imagine this was to protect the `::class`... but now you're explicitly handling that... so if we remove this guard entirely, does it break anything?

src/languages/php.js
```diff
@@ -379,12 +434,25 @@ export default function(hljs) {
...
+          3: "variable.constant",
+        },
+      },
       {
         // swallow composed identifiers to avoid parsing them as keywords
         match: regex.concat(
-          /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/,
-          regex.concat("(?!", WHITESPACE, "*\\()"),
-          /(?![a-zA-Z0-9_\x7f-\xff])/
+          /(::|->)+/,
```
wkania, line 453:
```suggestion
          /->+/,
```

src/languages/php.js
wkania: I see. Reverting back [to](https://github.com/highlightjs/highlight.js/commit/d49e9fb8fa82a43ed96333510c9c5bc1af776352#diff-b056c5a26741799f5d40c82d9bfff3d9975a8b63bdb5955cf5e30b3059042bddL76), didn't break anything.

joshgoebel:
@wkania I removed the rule, seems fine.  I pretty sure it's redundant now.  Thanks for all work on this!

PR #3426, opened 2021-12-09
enh(php) Switch highlighter to partially case-insensitive

wkania:
### Changes
After [discussion](https://github.com/highlightjs/highlight.js/pull/3422#discussion_r766006060) related to PHP partial case-insensitive .
- [x] Numbers ([case-insensitive](https://www.php.net/manual/en/language.types.integer.php))
- [x] Variables ([case-sensitive](https://www.php.net/manual/en/language.variables.basics.php))
- [x] class names (case-insensitive)
- [x] function names (case-insensitive)
- [x] keywords (both)

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -303,11 +303,11 @@ export default function(hljs) {
       {
         match: [
           /new/,
-          /\s+/,
+          / +/,
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
           /\\?\w+/,
-          /\s*\(/,
+          / *\(/,
         ],
         scope: {
           1: "keyword",
@@ -320,11 +320,11 @@ export default function(hljs) {
   const FUNCTION_INVOKE = {
     relevance: 0,
     match: [
-      /(?:->|::|\s|\(|\\)/,
+      /\b/,
       // to prevent keywords from being confused as the function title
       regex.concat("(?!fn\\b|function\\b|", normalizeKeywords(KWS).join("\\b|"), "|", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
       /\w+/,
-      /\s*/,
+      / */,
       regex.lookahead(/(?=\()/)
     ],
     scope: {
@@ -374,7 +374,8 @@ export default function(hljs) {
       FUNCTION_INVOKE,
       {
         // swallow composed identifiers to avoid parsing them as keywords
-        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/
+        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?! *\()(?![a-zA-Z0-9_\x7f-\xff])/,
+        // scope:"wrong"
       },
       CONSTRUCTOR_CALL,
       {
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -288,11 +288,31 @@ export default function(hljs) {
     built_in: BUILT_INS.concat([ "Error|0" ]),
   };
 
+  const CONSTRUCTOR_CALL = {
+    variants: [
+      {
+        match: [
+          /new/,
+          /\s+/,
+          // to prevent built ins from being confused as the class constructor call
+          regex.concat("(?!", BUILT_INS.join("\\b|"), "\\b)"),
+          /\\?\w+/,
+          /\s*\(/,
+        ],
+        scope: {
+          1: "keyword",
+          4: "title.class",
+        },
+      }
+    ]
+  };
+
   const FUNCTION_INVOKE = {
     relevance: 0,
     match: [
       /(?:->|::|\s|\(|\\)/,
-      regex.concat("(?!fn\\b|function\\b|match\\b|", KWS.join("\\b|"), "|", BUILT_INS.join("\\b|"), "\\b)"),
+      // to prevent keywords from being confused as the function title
+      regex.concat("(?!fn\\b|function\\b|match\\b|Error\\b|", KWS.join("\\b|"), "|", BUILT_INS.join("\\b|"), "\\b)"),
       /\w+/,
       /\s*/,
       regex.lookahead(/(?=\()/)
@@ -301,6 +321,7 @@ export default function(hljs) {
       3: "function.title.invoke",
     }
   };
+
   return {
     case_insensitive: false,
     keywords: KEYWORDS,
@@ -343,12 +364,9 @@ export default function(hljs) {
       FUNCTION_INVOKE,
       {
         // swallow composed identifiers to avoid parsing them as keywords
-        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?!\()/
+        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/
       },
-      {
-        // swallow create object
-        begin: /new\s\\?[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*\s?\(/
-      },
+      CONSTRUCTOR_CALL,
       {
         className: 'function',
         relevance: 0,
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -46,6 +46,8 @@ export default function(hljs) {
     end: /[ \t]*(\w+)\b/,
     contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
   });
+  // list of valid whitespaces because non-breaking space might be part of a name
+  const WHITESPACE = '[ \t\n]';
   const STRING = {
     className: 'string',
     variants: [
@@ -303,11 +305,11 @@ export default function(hljs) {
       {
         match: [
           /new/,
-          / +/,
+          regex.concat(WHITESPACE, "+"),
           // to prevent built ins from being confused as the class constructor call
           regex.concat("(?!", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
           /\\?\w+/,
-          / *\(/,
+          regex.concat(WHITESPACE, "*\\("),
         ],
         scope: {
           1: "keyword",
@@ -324,7 +326,7 @@ export default function(hljs) {
       // to prevent keywords from being confused as the function title
       regex.concat("(?!fn\\b|function\\b|", normalizeKeywords(KWS).join("\\b|"), "|", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
       /\w+/,
-      / */,
+      regex.concat(WHITESPACE, "*"),
       regex.lookahead(/(?=\()/)
     ],
     scope: {
@@ -374,7 +376,11 @@ export default function(hljs) {
       FUNCTION_INVOKE,
       {
         // swallow composed identifiers to avoid parsing them as keywords
-        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?! *\()(?![a-zA-Z0-9_\x7f-\xff])/,
+        match: regex.concat(
+          /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/,
+          regex.concat("(?!", WHITESPACE, "*\\()"),
+          /(?![a-zA-Z0-9_\x7f-\xff])/
+        ),
         // scope:"wrong"
       },
       CONSTRUCTOR_CALL,
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -11,6 +11,7 @@ Category: common
  * @returns {LanguageDetail}
  * */
 export default function(hljs) {
+  const regex = hljs.regex;
   const VARIABLE = {
     className: 'variable',
     begin: '\\$+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*' +
@@ -145,7 +146,6 @@ export default function(hljs) {
     "isset",
     "iterable",
     "list",
-    "match|0",
     "mixed",
     "new",
     "object",
@@ -172,7 +172,6 @@ export default function(hljs) {
   const BUILT_INS = [
     // Standard PHP library:
     // <https://www.php.net/manual/en/book.spl.php>
-    "Error|0",
     "AppendIterator",
     "ArgumentCountError",
     "ArithmeticError",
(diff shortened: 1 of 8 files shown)


joshgoebel:
Actually let me rewrite the keyword lists to arrays first so the PR will be easier to follow... then you would just need a `bothCases` helper function that would take an array and spit out another array with both variants, ie:

- ["then"]  => ["then", "THEN"]
- ["__FILE__"]  => ["__FILE__", "__file__"]

Assuming the magic constants are case-insensitive.

joshgoebel:
https://github.com/highlightjs/highlight.js/pull/3427

wkania:
> Actually let me rewrite the keyword lists to arrays first so the PR will be easier to follow... then you would just need a `bothCases` helper function that would take an array and spit out another array with both variants, ie:
> 
>     * ["then"]  => ["then", "THEN"]
> 
>     * ["**FILE**"]  => ["**FILE**", "**file**"]
> 
> 
> Assuming the magic constants are case-insensitive.

Great but there are more valid cases like:
 ` If(true) { };   
  iF(true) { };`
  
  Is there an option to overwrite global case sensitivity with the `match: /if/i` ?

joshgoebel:
> Great but there are more valid cases like: If(true) { }; iF(true) { };

But do people do that? That seems very UGLY/bad practice to me. (esp the latter)

We don't have any easy way to do that... all the grammar regex must be `/i` or not `/i`, they can NOT be mixed, the sensitivity is controlled by the setting as I mentioned earlier.

`keywords` is done by raw string matching so it's easy to hack for `BLAH` and `blah`.  We could possibly also calculate a "Blah", but I probably would NOT do it by default... I'd wait and see how common that was in the wild.

wkania:
> But do people do that? That seems very UGLY/bad practice to me. (esp the latter)

People rather use lowercase. I saw some cases of uppercase: TRUE, FALSE, NULL. So we can ignore it.

joshgoebel:
Oh I wonder if it's more common for literals, we could just support those if those are common...

joshgoebel:
Can you give me commit access to your fork?

joshgoebel:
Do you think we could match function dispatch similar to how we do in CPP so we just catch all invocations?

https://github.com/highlightjs/highlight.js/blob/main/src/languages/cpp.js#L401

wkania:
> Do you think we could match function dispatch similar to how we do in CPP so we just catch all invocations?
> 
> https://github.com/highlightjs/highlight.js/blob/main/src/languages/cpp.js#L401

Looks good. Invocations of?

joshgoebel:
> Invocations of?

Any function... if it's a function we can highlight it as `title.function`...

wkania:
> > Invocations of?
> 
> Any function... if it's a function we can highlight it as `title.function`...

That what I thought. Currently, the highlighter does not support function invocation, only declarations. Do you want this to be part of this MR?

joshgoebel:
Well you had "function names (case-insensitive)" on your list... so not that's the first thing I jumped too... overall we're trying to move away from complex parsing an do simpler pattern matching... so for many languages the same rule can be used to detect function definition and function invocation.

```
function blah() ...
blah(32)
```

The same `[ident](` rule hits both.

joshgoebel:
If 99% of the time people use lower case keywords I'm ok with just trying that and seeing how it works (with the exception of the literals)... would that complete `keywords (both)` then?

wkania:
> If 99% of the time people use lower case keywords I'm ok with just trying that and seeing how it works (with the exception of the literals)... would that complete `keywords (both)` then?

This should be fine. Will look at it tomorrow.

wkania:
@joshgoebel Ready for review. 
Keywords match|0 and Error|0 were problematic.

joshgoebel:
> Keywords match|0 and Error|0 were problematic.

How so?

src/languages/php.js
```diff
@@ -263,13 +262,65 @@ export default function(hljs) {
...
+      }
+    });
+    return result;
+  };
+
   const KEYWORDS = {
-    keyword: KWS,
-    literal: LITERALS,
-    built_in: BUILT_INS
+    keyword: KWS.concat([ "match|0" ]),
```
joshgoebel, line 286: No, this is not the right solution.  The keyword lists should be what they are.  If the `|0` tagging causes a problem for your code on line 312 (or elsewhere) then you need to write a filter function that returns just raw keywords by stripping the `|0`...

src/languages/php.js
```diff
@@ -263,13 +262,65 @@ export default function(hljs) {
...
+        },
+      }
+    ]
+  };
+
+  const FUNCTION_INVOKE = {
+    relevance: 0,
+    match: [
+      /(?:->|::|\s|\(|\\)/,
+      regex.concat("(?!fn\\b|function\\b|match\\b|Error\\b|", KWS.join("\\b|"), "|", BUILT_INS.join("\\b|"), "\\b)"),
```
joshgoebel, line 312: How many of the builtins and keywords are truly used like functions?  Is it a majority?  In most languages it's only a few... if there was a list of 5-10 in PHP I'd rather be explicit rather than just negative match the ENTIRE world of keywords (which does have a cost).

joshgoebel: Or if we just added an explicit rule for `[BUILT_IN](` (if many are used like functions) that would also remove the need for all the negative matching of built-ins here...

src/languages/php.js
joshgoebel: But as I mention elsewhere I also question the need to use the full lists as a negative filter...

src/languages/php.js
joshgoebel: Also if you are doing negative filters then  you need a few markup tests to test the negative case as well to show that your negative filter is working as intended (you can add comments to explain the purposes of the negative tests).

wkania: > How many of the builtins and keywords are truly used like functions? Is it a majority? In most languages it's only a few... if there was a list of 5-10 in PHP I'd rather be explicit rather than just negative match the ENTIRE world of keywords (which does have a cost).




```
  const MIGHT_BE_BEFORE_OPENING_BRACKET = [
    "array",
    "case",
    "catch",
    "die",
    "echo",
    "exit",
    "include",
    "include_once",
    "print",
    "require",
    "require_once",
    "do",
    "else",
    "elseif",
    "empty",
    "eval",
    "finally",
    "for",
    "foreach",
    "fn",
    "function",
    "if",
    "int",
    "isset",
    "list",
    "match",
    "return",
    "switch",
    "throw",
    "trait",
    "try",
    "unset",
    "use",
    "while",
  ];
```
This was my first approach ^.
There will be also some from `BUILT_INS`.



> Also if you are doing negative filters then you need a few markup tests to test the negative case as well to show that your negative filter is working as intended (you can add comments to explain the purposes of the negative tests).

I reused old tests. New code didn't break them, but some comment would be in place. I will add some extra markup tests

joshgoebel:
> This was my first approach ^.

Ugh, ok - that's a lot... :-) We can keep it as-is and see if it works ok I suppose.

src/languages/php.js
wkania: > Also if you are doing negative filters then you need a few markup tests to test the negative case as well to show that your negative filter is working as intended (you can add comments to explain the purposes of the negative tests).

 @joshgoebel Added extra tests and some comments.

src/languages/php.js
```diff
@@ -263,13 +264,80 @@ export default function(hljs) {
...
+  /**
+   * @param {string[]} items */
+  const normalizeKeywords = (items) => {
+    /** @type string[] */
+    const result = [];
+    items.forEach(item => {
+      item.replace(/\|0$/, "");
+      result.push(item);
+    });
+    return result;
```
joshgoebel, line 302: Use `map` for transforms?

joshgoebel: Also this needs to match `\d+` not just `0`... so it's future proof.

joshgoebel: I think this is all I see.  I might step in after and make a negative look-ahead list helper... but this looks pretty clean right now.

wkania: @joshgoebel done

> I think this is all I see. I might step in after and make a negative look-ahead list helper... but this looks pretty clean right now.

Good idea

src/languages/php.js
```diff
@@ -263,13 +264,76 @@ export default function(hljs) {
...
+          4: "title.class",
+        },
+      }
+    ]
   };
+
+  const FUNCTION_INVOKE = {
+    relevance: 0,
+    match: [
+      /(?:->|::|\s|\(|\\)/,
```
joshgoebel, line 323: Please provide commented examples of all the variants you're using here. Is just an attempt to eliminate false positives - if so what would an example or two be?   Why do we need these at all vs just relying on `[ident](`...?

wkania: > Is just an attempt to eliminate false positives - if so what would an example or two be? Why do we need these at all vs just relying on `[ident](`...?
> 
> If this is critical please provide commented code examples of all the variants you're using here.

All the correct cases are in the `Function invoke` tests. I also added a comment and test for `First-class Callable Syntax`,
which was working correctly before this PR (because there was no function invoke). Now it would be a false positive, so I added all the rules.

joshgoebel: I'm not following...   Isn't this just an issue with the non-function guard?  If I do this, all tests pass:

```
diff --git a/src/languages/php.js b/src/languages/php.js
index 9f45841a..297bd55b 100644
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -321,7 +321,8 @@ export default function(hljs) {
     relevance: 0,
     match: [
       // to prevent First-class Callable Syntax from being confused as the function title
-      /(?:->|::|\s|\(|\\)/,
+      // /(?:->|::|\s|\(|\\)/,
+      /\b/,
       // to prevent keywords from being confused as the function title
       regex.concat("(?!fn\\b|function\\b|", normalizeKeywords(KWS).join("\\b|"), "|", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
       /\w+/,
@@ -375,7 +376,8 @@ export default function(hljs) {
       FUNCTION_INVOKE,
       {
-        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/
+        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?!\s*\()(?![a-zA-Z0-9_\x7f-\xff])/,
+        // scope:"wrong"
       },
       CONSTRUCTOR_CALL,
       {
```

IE, fixing the guard and using only a word boundary to match the start of function invokes... thoughts?  This using multi-match as "look-behind" is dangerous and I really don't want to encourage it at all if it can be avoided.

wkania: @joshgoebel Strange because I applied your fix and they fall.
```
1 failing

  1) php
       should not cause polynomial backtracking:
     Error: php/[10]/begin: Polynomial backtracking. By repeating any character that matches /\xa0/i, an attack string can be created.

    [a-zA-Z0-9_\x7f-\xff]*(?!\s*
    ^~~~~~~~~~~~~~~~~~~~~~   ^~~

Full pattern:
/(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?!\s*\()(?![a-zA-Z0-9_\x7f-\xff])/m

```

After changing `\s*` to `\s` I have a below error and a few offers.

```
-DateTimeImmutable::createFromMutable(<span class="hljs-keyword">new</span> <span class="hljs-title class_">\DateTime</span>(<span class="hljs-string">&#x27;now&#x27;</span>));
      +DateTimeImmutable::<span class="hljs-function title_ invoke__">createFromMutable</span>(<span class="hljs-keyword">new</span> <span class="hljs-title class_">\DateTime</span>(<span class="hljs-string">&#x27;now&#x27;</span>));
```
This regex do not detect static function invoke:
`DateTimeImmutable::createFromMutable();`
or method invoke
`$date->format();`
and we do not want to higlight First-class Callable Syntax as a function
`$test();`

P.S. I'm run tests inside Docker.

joshgoebel: > should not cause polynomial backtracking:

That can be ignored [for now] and fixed.  **I didn't have any markup test errors.**  Your'e saying you're seeing the same result right - all tests pass , other than the security issue?

So if we can fix that, then can this be simplified as I'm suggesting?

joshgoebel: Can names TRULY contain non-breaking spaces (0xA0)? That sounds like a bug to me...

wkania: > > should not cause polynomial backtracking:
> 
> That can be ignored [for now] and fixed. **I didn't have any markup test errors.** Your'e saying you're seeing the same result right - all tests pass with my patch , other than the security issue?
> 
> So if we can fix that, then can this be simplified as I'm suggesting?

It will be annoying



> Can names TRULY contain non-breaking spaces (0xA0)? That sounds like a bug to me...

[Yes](https://mnapoli.fr/using-non-breakable-spaces-in-test-method-names/), but I never saw anyone using it. Looks bad for me.

wkania: @joshgoebel fixed it with ` *` because only space is valid before the name of the function and opening bracket.

src/languages/php.js
```diff
@@ -374,7 +374,8 @@ export default function(hljs) {
       FUNCTION_INVOKE,
       {
         // swallow composed identifiers to avoid parsing them as keywords
-        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/
+        begin: /(::|->)+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?! *\()(?![a-zA-Z0-9_\x7f-\xff])/,
```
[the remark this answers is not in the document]
wkania: > _I'm not sure they all needed to change_, unless the `[ ]` is more precise? (it may likely be) IIRC the issue was with this line alone because `\xA0*` overlaps with `\s*`.... which is fixed easily enough by changing the look-ahead \s to a space, if that works for our use cases. :-) The character groups in the regex here also could have been fixed. (I might just do this)
> 
> So, is this more precise? (in which case I'll likely change them to `[ ]` for readability.
> 
> Any final thoughts? Thanks for all the hard work on this. I think the finish line is just ahead. :)

```
new    
\DateTime   
();
```
This is also allowed so I will introduce const SEPARATOR `[ \t\n]`. This should cover all the cases. Probably this will be better in all those places, if non-breaking space may be part of the name and `\s` select it.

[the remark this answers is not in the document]
joshgoebel: > This is also allowed 

I imagined it might be. :-)

> so I will introduce const SEPARATOR

I think "WHITESPACE" might be better (and add a comment saying it purposely excludes `0xA0`), etc...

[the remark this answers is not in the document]
wkania: > I think "WHITESPACE" might be better (and add a comment saying it purposely excludes `0xA0`), etc...

Done @joshgoebel

joshgoebel (approved):

joshgoebel:
@wkania Thanks for all the effort.  I might release 11.4 soon just to see how this case change plays out.

PR #3430, opened 2021-12-17
 (Typescript) highlight obvious user defined type names (interface, etc) #3269 

### LINKED ISSUE #3269: (Typescript) not highlight user defined types
See this example:
image
as you can see in boudingRect params the `Rect` type has not been highlighted. also interface name. 
I think these one should be highlighted

nbeck415:
### Changes

Starting an implementation for recognizing user-defined titles and type names for Typescript interfaces. Typescript.js should recognize any valid identifier (previously worked based solely on the assumption identifier was in camelcase) and highlight it as a "title." 

### Checklist
- [x] Implemented for interface identifiers
- [ ] Implemented for other user-defined types
- [x] Added markup tests

### DIFF
--- a/src/languages/typescript.js
+++ b/src/languages/typescript.js
@@ -45,16 +45,12 @@ export default function(hljs) {
       1: "keyword",
       3: "title"
     },
-    end: /\{/,
     end: /\{|\s/,
     excludeEnd: true,
     keywords: {
       keyword: 'interface extends',
       built_in: TYPES
-    },
-    contains: [
-      tsLanguage.exports.CLASS_REFERENCE
-    ]
+    }
   };
 
   const USE_STRICT = {
--- a/src/languages/typescript.js
+++ b/src/languages/typescript.js
@@ -15,6 +15,7 @@ export default function(hljs) {
   const tsLanguage = javascript(hljs);
 
   const IDENT_RE = ECMAScript.IDENT_RE;
+  const IDENT_GEN = /[a-zA-Z]\w*/;
   const TYPES = [
     "any",
     "void",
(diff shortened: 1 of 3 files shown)


src/languages/typescript.js
```diff
@@ -33,18 +34,25 @@ export default function(hljs) {
...
+    begin: [
+      /interface/,
+      /\s+/,
+      IDENT_GEN
+    ],
+    beginScope: {
+      1: "keyword",
+      3: "title"
+    },
+    end: /\{|\s/,
```
joshgoebel, line 48: Why do we need to end it at all vs just match it?

joshgoebel: It'll end automatically after the title...

nbeck415:
Is the fix here to simply to remove the end case?

joshgoebel:
Yes, and then begin can change to match, see `wren.js` and other grammars for simple examples of this same pattern.

joshgoebel:
You got it!

src/languages/typescript.js
```diff
@@ -36,17 +36,15 @@ export default function(hljs) {
...
     ],
-    beginScope: {
+    scope: {
       1: "keyword",
       3: "title"
     },
-    end: /\{|\s/,
-    excludeEnd: true,
     keywords: {
       keyword: 'interface extends',
```
joshgoebel, line 49: Oh, n/m this isn't going to get the "extends" though... we'd need a second match variant to handle `interface xyz extends abc`  That is the current preference over the more ambiguous "look for a {" which can have false positives in other languages.  Again you should have examples to refer to in `ruby` or `wren`... you just need `variants` to house two multi-matchers... 

And of course we need some markup tests to catch things like this for us. :)

nbeck415:
It looks like it is working! I added the variant and wrote some markdown tests for a few different interface definition scenarios

src/languages/typescript.js
```diff
@@ -33,18 +34,52 @@ export default function(hljs) {
...
       built_in: TYPES
-    },
-    contains: [
-      tsLanguage.exports.CLASS_REFERENCE
-    ]
+    }
   };
+
+  const IMPORT = {
```
joshgoebel, line 71: So we define this, but do we use it anywhere?

joshgoebel:
Ping.

PR #3442, opened 2021-12-30
Remove arcade from core to its own repo

jf990:
ArcGIS Arcade should not be in the core repo as it is a specialty scripting language. We added this before we decided to have grammars in separate repos. This PR removes it from the core repo and maintenance will move to https://github.com/jf990/highlightjs-arcade. There I updated to v11 and arcade v1.16.

- https://github.com/jf990/highlightjs-arcade/issues/16

### Changes
- remove grammar `arcade` from core
- relocated to https://github.com/jf990/highlightjs-arcade
- Arcade grammar updated to v1.16

### Checklist
- [ ] Moved markup tests to https://github.com/jf990/highlightjs-arcade


joshgoebel:
> ArcGIS Arcade should not be in the core repo as it is a specialty scripting language. We added this before we decided to have grammars in separate repos. This PR removes it from the core repo and maintenance will move to https://github.com/jf990/highlightjs-arcade.

Certainly not opposed to this with speciality languages, but technically I worry just ripping it out may constitute a breaking change.  We may need to instead deprecate it and list the new URL in supported languages... while it lives on in the repository until v12. 

@highlightjs/core Any thoughts on whether ripping out a grammar NOT a part of the default distribution should be considered breaking?  I'm leaning towards yes... someone who build the library yesterday should be able to build it with the same arguments tomorrow (until v12) without it break.

So if that's the case, what immediate action can we take on this in the interim?  My todo list for v12 is still pretty short.

jf990:
@joshgoebel after thinking about it I agree, this is a bad idea because it would be a breaking change and there are devs that use this syntax highlighter. I am going to close this PR and do the update to the existing files in this repo and keep them in sync with the 3rd party repo. thanks.

joshgoebel:
Want to revisit when v12 is closer?

jf990:
@joshgoebel we will revisit this at another time, v12 sounds good. thanks.

PR #3446, opened 2021-12-30
Update Arcade to v1.16

jf990:
Update Arcade grammar to ArcGIS Arcade version 1.16: new keywords, code clean up, relevance update, v11 update.

### Changes
languages arcade.js updated

### Checklist
- [x] Verified markup tests

### DIFF
--- a/src/languages/arcade.js
+++ b/src/languages/arcade.js
@@ -1,12 +1,14 @@
-/*
+/**
  Language: ArcGIS Arcade
  Category: scripting
  Author: John Foster <jfoster@esri.com>
  Website: https://developers.arcgis.com/arcade/
  Description: ArcGIS Arcade is an expression language used in many Esri ArcGIS products such as Pro, Online, Server, Runtime, JavaScript, and Python
+ @param {HLJSApi} hljs
+ @returns {LanguageDetail}
+ @type LanguageFn
 */
 
-/** @type LanguageFn */
 export default function(hljs) {
   const IDENT_RE = '[A-Za-z_][0-9A-Za-z_]*';
   const KEYWORDS = {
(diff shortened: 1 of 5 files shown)


src/languages/arcade.js
```diff
@@ -75,7 +249,8 @@ export default function(hljs) {
   ]);
 
   return {
-    name: 'ArcGIS Arcade',
+    aliases: [ 'arcade' ],
```
joshgoebel, line 252: Where did name go?  `alias` isn't necessary if it just matches the filename.

src/languages/arcade.js
```diff
@@ -75,7 +249,8 @@ export default function(hljs) {
...
+    case_insensitive: true,
```
joshgoebel, line 253: This quite possibly could be why your relevance broke (if this is the language you were referring to)... that changes a lot of things (makes them less precise, more likely to false positive, etc).

joshgoebel, line 253: The big changes here were keywords and case insensitive, right?    That and relevance?  If I back out all the tiny changes (that I asked you to change back) and just copied over the keywords and case sensitivity and took a look at the relevance issues myself, would that achieve the goals here?

I know you wanted to pull this out so you were probably tuning this a bit more to your liking, but now that it's staying...

jf990: @joshgoebel that's correct the only real changes are the addition of new keywords in the recent language update and case insensitive. The relevance was changed only to get the tests to pass so it has no functional significance to the parsing. So I'm fine with working out what you think is best so we can get this in the next update.

src/languages/arcade.js
jf990: not sure, i think this was unintentional. it should be `arcade`, I'll revert `name`.

src/languages/arcade.js
jf990: @joshgoebel I just restored `name`

not sure what you want to do about `relevance`, do you want me to revert all of those changes?

joshgoebel: You don't need to do anything further, I'll take a pass from scratch and tag you on the keyword/sesnativity change.

PR #3448, opened 2021-12-31
Add GraphQL

jf990:
Add grammar GraphQL

### Changes
languages/graphql.js, detect and markup tests

### Checklist
- [X] Added markup tests


src/languages/graphql.js
```diff
@@ -0,0 +1,79 @@
...
+      },
+      {
+        className: "meta",
+        begin: "@",
+        end: "\\W",
+        excludeEnd: true
+      },
+      {
+        className: "symbol",
+        begin: GQL_NAME + '\\s*:',
```
joshgoebel, line 64: Looks good, but ALL the strings need to be regex literals (unless that's impossible like a very complex expression join - but all of these looks doable to me)... but for simple cases like here you can use  the regex helpers: `hljs.regex.concat(GQL_NAME, /\s*:/)`, etc...

src/languages/graphql.js
```diff
@@ -0,0 +1,79 @@
...
+        end: "\\.",
+        relevance: 0
+      },
+      {
+        className: "punctuation",
+        begin: GQL_PUNCTUATION,
+        relevance: 0
+      },
+      {
+        className: "variable",
```
joshgoebel, line 50: Please use the new `scope` key.

src/languages/graphql.js
```diff
@@ -0,0 +1,79 @@
...
+      hljs.NUMBER_MODE,
+      {
+        className: "punctuation",
+        begin: "[.]{2}",
+        end: "\\.",
+        relevance: 0
+      },
+      {
+        className: "punctuation",
+        begin: GQL_PUNCTUATION,
```
joshgoebel, line 46:
```suggestion
        match: GQL_PUNCTUATION,
```

Prefer match if there is no `end`.

src/languages/graphql.js
```diff
@@ -0,0 +1,79 @@
...
+      literal: "true false null"
+    },
+    contains: [
+      hljs.HASH_COMMENT_MODE,
+      hljs.QUOTE_STRING_MODE,
+      hljs.NUMBER_MODE,
+      {
+        className: "punctuation",
+        begin: "[.]{2}",
+        end: "\\.",
```
joshgoebel, line 41:
```suggestion
        match: "[.]{3}",
```

Why not the much simpler?  Is this mEant to match `..` STUFF `.`??  If not we need a single regex.

src/languages/graphql.js
```diff
@@ -0,0 +1,79 @@
...
+        "schema",
+        "directive",
+        "interface",
+        "union",
+        "scalar",
+        "fragment|10",
+        "enum",
+        "on"
+      ],
+      literal: "true false null"
```
joshgoebel, line 32: Array please.

src/languages/graphql.js
```diff
@@ -0,0 +1,79 @@
...
+ Category: scripting, protocols, web
+ Author: John Foster [jf990][], and others
+ Description: Highlight GraphQL queries with highlight.js.
+*/
+/** @type LanguageFn */
+export default function(hljs) {
+  const GQL_NAME = '[_A-Za-z][_0-9A-Za-z]*';
+  const GQL_PUNCTUATION = '[\\!\\(\\)\\:\\=\\[\\]\\{\\|\\}]{1}';
+  return {
+    name: "graphql",
```
joshgoebel, line 12:
```suggestion
    name: "GraphQL",
```

This should be a "pretty name"/branded name, etc...

joshgoebel (changes requested):
Lots of tiny things, but overall looks good - shouldn't be super hard to finish up.

joshgoebel:
I plan to push a release in next few days just heads up, if this is ready it'll land in that.

src/languages/graphql.js
jf990: was concerned spread operator to not match `....` and other combos of more than 3 dots, so won't `{3}` match `....`?

jf990: I changed as you suggest and tested a variety of scenarios and it seems to work as expected. good catch, thank you!

jf990:
@joshgoebel suggestions applied and tested, looks good to me. thanks for your review.

joshgoebel:
Ping.

joshgoebel (changes requested):
Can't tell what you changed, I'm not seeing changes on the comments I posted...?

jf990:
@joshgoebel not sure how I missed your comments but I did, looking at it now.

src/languages/graphql.js
```diff
@@ -0,0 +1,81 @@
...
+      },
+      {
+        scope: "meta",
+        begin: /@/,
+        end: /\W/,
+        excludeEnd: true
+      },
+      {
+        scope: "symbol",
+        begin: hljs.regex.concat(GQL_NAME, /\s*:/),
```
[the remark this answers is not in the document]
jf990: i think i addressed this and it cleaned up some extra markup, so I think this turned out to be an improvement

src/languages/graphql.js
```diff
@@ -0,0 +1,81 @@
...
+        "mutation",
+        "subscription|10",
+        "type",
+        "input",
+        "schema",
+        "directive",
+        "interface",
+        "union",
+        "scalar",
+        "fragment|10",
```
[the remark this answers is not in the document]
jf990: @joshgoebel if I take the relevance away I absolutely cannot get the detection test to pass. I'm at a loss on this one and I was forced to put it back in. take a look and let me know if you have any suggestions

src/languages/graphql.js
```diff
@@ -62,16 +62,10 @@ export default function(hljs) {
...
-            scope: "symbol",
-            begin: GQL_NAME,
-            end: /\s*:/,
-            excludeEnd: true
-          }
-        ],
+        begin: hljs.regex.lookahead(hljs.regex.concat(GQL_NAME, /\s*:/)),
+        includeBegin: true,
+        end: /\s*:/,
+        excludeEnd: true,
```
joshgoebel, line 68:
```suggestion
        begin: regex.concat(GQL_NAME,regex.lookahead(/\s*:/)),
        end: regex.lookahead(/\s*:/),
```

excludeEnd is casually deprecated.  There is no `includeBegin`, that's just the default, but we shouldn't need it.  I think you want to match the name but not the delimiter, right?  And we should set a `const regex` to avoid repeating hljs. over and over since these are already long enough.

src/languages/graphql.js
```diff
@@ -15,16 +15,16 @@ export default function(hljs) {
     keywords: {
       keyword: [
         "query",
-        "mutation",
-        "subscription|10",
+        "mutation|10",
```
joshgoebel, line 18:
```suggestion
        "mutation",
```

If you're having relevance issues, just throw those to me at the end to solve.  The correct fix isn't to boost random words that we include in `default`...

jf990: ok, but the tests are going to fail

joshgoebel (changes requested):
Almost there. :-)  Thanks for all the effort.

src/languages/graphql.js
joshgoebel: Let me deal with the tests...

joshgoebel: `graphql` *might* be reasonable to consider for the default bundle even, and if so it needs careful review in regard to relevance vs all the others, etc...

src/languages/graphql.js
jf990: I must be doing something wrong, if I did not have `excludeEnd` then the `:` gets picked up in the scope. if I didn't include it in the `end` matcher then it gets too greedy and picks up everything until it finds a `:`

joshgoebel: > then the : gets picked up in the scope

That's what look-ahead guards against, but once it's in look-ahead it doesn't need to be in end anymore... because begin is already matching it -just not consuming it.

All tests are green... any final thoughts? :)

joshgoebel:
Long term (if we consider this for adding to default) we might need to add some combo rules such as:

`mutation [IDENT_RE] {`... and of course the mutation name would also be a `title.class`, etc... So then that would only match `mutation Book {` but not `mutation {` from some other language... and then we could perhaps bump the relevance for those keywords (in context) a little higher... but I think this is a good start for now.

src/languages/graphql.js
jf990: I'd like to go forward with this version and I can consider any updates for a future version, if you are OK with that

joshgoebel:
Cool, I'll let this simmer another day or two then circle back for a final review.  I find that often catches things I miss, but I I can probably fix up anything else I find. Thanks for all the help!

src/languages/graphql.js
```diff
@@ -0,0 +1,75 @@
...
+      hljs.QUOTE_STRING_MODE,
+      hljs.NUMBER_MODE,
+      {
+        scope: "punctuation",
+        match: /[.]{3}/,
+        relevance: 0
+      },
+      {
+        scope: "punctuation",
+        begin: /[\!\(\)\:\=\[\]\{\|\}]{1}/,
```
joshgoebel, line 49: Should some of these not be `operator`?

jf990: @joshgoebel I was trying to mimic the prism highlighting done on [graphql.org](https://graphql.org/learn/), if you inspect a sample on their doc pages the punctuation chars are given their own scope

```xml
<span class="token keyword">function</span>
<span class="token plain"> </span>
<span class="token function">Query_me</span>
<span class="token punctuation">(</span>
<span class="token plain">request</span>
<span class="token punctuation">)</span>
...
```

so I was trying to recreate that, at least initially, and it seems OK in my testing. I'm now leaning to leaving it in.

As for graphql operators, they are tagged as `and:` or `or:`, I did not apply special highlighting for those because they clash with regular properties and I was having difficulty getting them correct.

joshgoebel: Ok, we'll leave as-is for now and see how it goes.

joshgoebel:
@jf990 Thanks for all the work on this!

joshgoebel:
**Size Change:** +1.15 kB (0%) 
**Total Size:** 715 kB

| Filename | Size | Change |  |
| :--- | :---: | :---: | :---: |
| `./build/es/languages/stata.min.js` | 6.65 kB | +10 B (0%) |  |
| `./build/languages/pf.min.js` | 849 B | +6 B (+1%) |  |
| `./build/languages/stata.min.js` | 6.66 kB | +13 B (0%) |  |
| `./build/es/languages/graphql.min.js` | 543 B | +543 B (new file) | 🆕 |
| `./build/languages/graphql.min.js` | 550 B | +550 B (new file) | 🆕 |

jf990:
@joshgoebel thanks for your help, definitely much better thanks to your review feedback.

PR #3449, opened 2021-12-31
tools: check-theme: consider background-color too

felipec:
This is the precise property we want to check.

### DIFF
--- a/tools/checkTheme.js
+++ b/tools/checkTheme.js
@@ -193,9 +193,9 @@ class CSSRule {
   constructor(rule, body) {
     this.rule = rule;
     if (rule.declarations) {
-      this.bg = rule.declarations.find(x => x.property =="background-color")?.value;
+      this.bg = rule.declarations.find(x => x.property == "background-color")?.value;
       if (!this.bg) {
-        this.bg = rule.declarations.find(x => x.property =="background")?.value;
+        this.bg = rule.declarations.find(x => x.property == "background")?.value;
       }
       this.fg = rule.declarations.find(x => x.property =="color")?.value;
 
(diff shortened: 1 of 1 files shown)


tools/checkTheme.js
```diff
@@ -193,7 +193,7 @@ class CSSRule {
   constructor(rule, body) {
     this.rule = rule;
     if (rule.declarations) {
-      this.bg = rule.declarations.find(x => x.property =="background")?.value;
+      this.bg = rule.declarations.find(x => x.property =="background" || x.property =="background-color")?.value;
```
joshgoebel, line 196: Nice, but I think actually we'd need two finds to be "correct", no?  It would be possible to have:

```css
background:  content-box no-repeat something;
background-color: specific color;
```

So you'd want to first scan ALL the declarations for `background-color`, and then failing that scan a second pass for `background`?

joshgoebel (changes requested):

tools/checkTheme.js
felipec: OK. Done.

tools/checkTheme.js
```diff
@@ -193,7 +193,10 @@ class CSSRule {
   constructor(rule, body) {
     this.rule = rule;
     if (rule.declarations) {
-      this.bg = rule.declarations.find(x => x.property =="background")?.value;
+      this.bg = rule.declarations.find(x => x.property =="background-color")?.value;
+      if (!this.bg) {
+        this.bg = rule.declarations.find(x => x.property =="background")?.value;
```
joshgoebel, line 198:
```suggestion
      this.bg = rule.declarations.find(x => x.property == "background-color")?.value;
      if (!this.bg) {
        this.bg = rule.declarations.find(x => x.property == "background")?.value;
```

joshgoebel (approved):

joshgoebel:
Thanks!

PR #3452, opened 2022-01-05
arcade v1.16

joshgoebel:
- chore(arcade) eslint --fix, explode keywords
- enh(arcade) Add missing keywords for Arcade v1.16

### Checklist


joshgoebel:
@jf990 boom done.  I zeroed out overly common words like left/right/top which were mucky with the auto-detection tests.  No need to change the rule relevance.

jf990:
@joshgoebel thank you for your help with this.

PR #3453, opened 2022-01-05
update twig keyword list

MrYamous:
### Changes
Update keywords list for Twig with elements for Symfony according to [documentation](https://twig.symfony.com/doc/3.x/)

### Checklist
- [x] Added markup tests, or they don't apply here because it's about updating keywords list


joshgoebel:
Are functions always followed by `()` or is that optional?

MrYamous:
> Are functions always followed by `()` or is that optional?

Twig functions are always followed by `()`

joshgoebel:
So like: 

```
{% set user = json_encode(user) %}
{{ logout_path() }}
```

The latter is a function?  And what is `json_encode`?  Currently we don't highlight it because it's not preceded by a `|` but it seems any filter can be used as a function in some contexts.  Could you help clarify?

MrYamous:
`logout_path()` is a function, always followed by `()`

`json_encode()` is a filter, it should be preceded by `|`.
On top of that, 
- some filters can't have parameters, like `abs`
- for some others, parameters are optional, `[1,2,3]|join` or `[1,2,3]|join('-')` are valid (with different output)

According to that,  `{% set user = json_encode(user) %}` isn't valid, it would be something like `{% set encodedUser = user|json_encode %}` or `{% set encodedUser = user|json_encode(constant('JSON_PRETTY_PRINT')) %}`

src/languages/twig.js
```diff
@@ -98,13 +135,18 @@ export default function(hljs) {
...
     "if",
     "import",
     "include",
     "macro",
     "sandbox",
     "set",
+    "stopwatch",
+    "trans",
+    "trans_default_domain",
+    "transchoice",
```
joshgoebel, line 149: What does it mean for something to be both a tag name and a filter?  Referring here to lines 147/149 vs 116/117.

joshgoebel: Of you're just copying the lists, hmmmm.

joshgoebel:
@allejo You know more about the PHP world than I... do we really want to consider Twig/Symphony as the same thing?  I'd say in the past we have not since none of these keywords are currently a part of the grammar - so this would be an expansion of the scope of the grammar... I know technically one can use Twig on it's own... I'm trying to decide if we want to expand the Twig grammar like this or if we should prefer to have a second Twig/Symphony grammar instead.

Any thoughts?

src/languages/twig.js
allejo: In Twig, if something can be used both as filter and a tag, it'd look like this,

```twig
{% trans %}
  I have some text here that I'm writing manually
{% endtrans %}
```

and

```twig
{{ some_variable | trans }}

^ shorthand for:

{% trans %}{{ some_variable }}{% endtrans %}
```

allejo:
I am really torn on this one. Twig is probably most commonly used alongside Symfony but as you mentioned, it can be used outside the Symfony context and it is in large projects (e.g. phpBB, Drupal, October CMS, WordPress via a plug-in).

My biggest concern is that the Twig grammar is often used as a holdover for other similar languages such as Liquid and Jinja (both of which hljs don't support). The syntax itself is almost identical but filters, functions, and tags all vary across the different languages.

I think the best approach would be for the Twig grammar to be less dependent on the giant keyword list and be smarter about detecting patterns for filters, functions, and tags but that's outside the scope of this PR.

---

tl;dr: adding these new keywords will benefit Symfony developers who use hljs to highlight Twig. It doesn't negatively affect anyone else who uses Twig outside of Symfony besides increasing the grammar size slightly.

joshgoebel:
> I think the best approach would be for the Twig grammar to be less dependent on the giant keyword list

This makes more sense if we give up on keywords/relevance in the future.  But there are challenges since for many languages without the keywords you have to lean more heavily on parsing.... but in many cases where it's just `blah(` or `|blah` then patterns vs keywords could work well - and produce much smaller grammars.

joshgoebel:
@MrYamous Thanks so much for the work on this.

src/languages/twig.js
```diff
@@ -10,26 +10,51 @@ Category: template
...
+    "path",
     "random",
     "range",
+    "relative_path",
+    "render",
+    "render_esi",
     "source",
-    "template_from_string"
+    "template_from_string",
+    "url"
```
joshgoebel, line 51:
```suggestion
    "url|0"
```

src/languages/twig.js
```diff
@@ -10,26 +10,51 @@ Category: template
...
+    "form_widget",
     "html_classes",
     "include",
+    "is_granted",
+    "logout_path",
+    "logout_url",
     "max",
     "min",
     "parent",
+    "path",
```
joshgoebel, line 43:
```suggestion
    "path|0",
```

src/languages/twig.js
```diff
@@ -10,26 +10,51 @@ Category: template
...
     "block",
     "constant",
+    "controller",
     "country_timezones",
+    "csrf_token",
     "cycle",
     "date",
     "dump",
+    "expression",
+    "form",
```
joshgoebel, line 26:
```suggestion
    "form|0",
```

src/languages/twig.js
```diff
@@ -10,26 +10,51 @@ Category: template
 export default function(hljs) {
   const regex = hljs.regex;
   const FUNCTION_NAMES = [
+    "absolute_url",
+    "asset",
```
joshgoebel, line 14:
```suggestion
    "asset|0",
```

src/languages/twig.js
```diff
@@ -10,26 +10,51 @@ Category: template
...
+    "asset_version",
     "attribute",
     "block",
     "constant",
+    "controller",
```
joshgoebel, line 19:
```suggestion
    "controller|0",
```

Issue #3454, opened 2022-01-06
(Java) braceless else statement highlights code differently


Aster89:
**Describe the issue**
```html
<pre><code class="language-java">int a = l;
if ( O == l )
  l = O1;
else {
  a = 01;
}</code></pre>
```
is shown as
![with-braces](https://user-images.githubusercontent.com/20521900/148339298-20a5fb18-28d7-4d46-b3d3-bd192273b806.png)
whereas if I remove the `{` and `}`, it is shown as
![without-braces](https://user-images.githubusercontent.com/20521900/148339349-b9af2068-3e39-4a5b-b686-30e7f22977dd.png)

Honestly, I don't if also the coloring of the `a` on the first line is a mistake. Probably it is.

**Which language seems to have the issue?**
Java

**Are you using `highlight` or `highlightAuto`?**

Not sure if it's an answer, but I've downloaded it (yesterday) form [here](https://highlightjs.org/download/), so I think I'm using `highlight`.

**Sample Code to Reproduce**
See snippet above and [this](https://jsfiddle.net/Enrico89/wk6atfdx/2/).

**Expected behavior**
See screenshot above

**Additional context**
Not sure what else is needed.

joshgoebel:
Without the { it looks like a possible variable assignment:

```
int a = 0;
else a = 0;
```

So we can filter that out.  Are there other keywords that would lead to this some type of confusion we should also filter out?

PR #3455, opened 2022-01-06
fix(java) prevent false variable init on `else`

joshgoebel:
- fix(java) prevent false variable init on `else`
- changelog

Resoles #3454 

### Checklist


allejo (approved):
👍 looks sane to me

image

PR #3458, opened 2022-01-09
Fix highlighting of + and - in brainfuck grammar

LyricLy:
Previously, the first 2 `+` or `-` characters in a sequence of increments/decrements (e.g. `+++++++`) would not be highlighted correctly, due to the special relevance counting rule not having the correct class name set.

### Changes
Resolve the issue by setting the class name to `literal`.


src/languages/brainfuck.js
```diff
@@ -35,6 +35,7 @@ export default function(hljs) {
       },
       {
         // this mode works as the only relevance counter
+        className: 'literal',
         begin: /(?:\+\+|--)/,
         contains: [LITERAL]
```
joshgoebel, line 40: I think it's supposed to be a look-ahead... does this also work?
```suggestion
        begin: /(?=\+\+|--)/,
        contains: [LITERAL]
```

LyricLy: Yes, that also works.

joshgoebel:
Please add a changelog entry.

joshgoebel (approved):

joshgoebel:
**Size Change:** +39 B (0%) 
**Total Size:** 714 kB

| Filename | Size | Change |  |
| :--- | :---: | :---: | :---: |
| `./build/es/languages/brainfuck.min.js` | 330 B | +19 B (+6%) | 🔍 |
| `./build/highlight.min.js` | 8.04 kB | +2 B (0%) |  |
| `./build/languages/brainfuck.min.js` | 335 B | +18 B (+6%) | 🔍 |

PR #3459, opened 2022-01-10
enh(php) named arguments and fix php constants

wkania:
We have a new release so I added a new version.

### Changes
[Named Arguments](https://wiki.php.net/rfc/named_params)
There was no test for the most basic type of constant reference and it didn't work :(, so I fixed it.

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -393,16 +393,17 @@ export default function(hljs) {
     ]
   };
 
+  const NAMED_ARGUMENT = {
+    scope: 'attr',
+    match: regex.concat(IDENT_RE, regex.lookahead(':'), regex.lookahead(/(?!::)/)),
+  };
   const PARAMS_MODE = {
     relevance: 0,
     begin: /\(/,
     end: /\)/,
     keywords: KEYWORDS,
     contains: [
-      {
-        scope: 'attr',
-        match: regex.concat(IDENT_RE, regex.lookahead(':'), regex.lookahead(/(?!::)/)),
-      },
+      NAMED_ARGUMENT,
       VARIABLE,
       LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
       hljs.C_BLOCK_COMMENT_MODE,
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -328,21 +328,6 @@ export default function(hljs) {
     ]
   };
 
-  const FUNCTION_INVOKE = {
-    relevance: 0,
-    match: [
-      /\b/,
-      // to prevent keywords from being confused as the function title
-      regex.concat("(?!fn\\b|function\\b|", normalizeKeywords(KWS).join("\\b|"), "|", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
-      IDENT_RE,
-      regex.concat(WHITESPACE, "*"),
-      regex.lookahead(/(?=\()/)
-    ],
-    scope: {
-      3: "title.function.invoke",
-    }
-  };
-
   const CONSTANT_REFERENCE = regex.concat(IDENT_RE, "\\b(?!\\()");
 
   const LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON = {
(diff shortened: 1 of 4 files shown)


src/languages/php.js
```diff
@@ -394,6 +393,42 @@ export default function(hljs) {
...
+      regex.concat("(?!fn\\b|function\\b|", normalizeKeywords(KWS).join("\\b|"), "|", normalizeKeywords(BUILT_INS).join("\\b|"), "\\b)"),
+      IDENT_RE,
+      regex.concat(WHITESPACE, "*"),
+      regex.lookahead(/(?=\()/)
+    ],
+    scope: {
+      3: "title.function.invoke",
+    }
+  };
+  FUNCTION_INVOKE.contains = [
```
joshgoebel, line 410: Why isn't this just inside the original declaration?

src/languages/php.js
```diff
@@ -394,6 +393,42 @@ export default function(hljs) {
...
+    {
```
joshgoebel, line 411: Please extract into `PARAMS_MODE` as we do in so many other grammars.

src/languages/php.js
```diff
@@ -394,6 +393,42 @@ export default function(hljs) {
...
+      relevance: 0,
+      begin: /\(/,
+      end: /\)/,
+      keywords: KEYWORDS,
+      contains: [
+        {
+          scope: 'attr',
+          match: IDENT_RE + regex.lookahead(':') + regex.lookahead(/(?!::)/),
```
joshgoebel, line 419: These may not always return strings, please use `regex.concat`.

src/languages/php.js
```diff
@@ -368,6 +353,20 @@ export default function(hljs) {
           2: "variable.language",
         },
       },
+      {
```
joshgoebel, line 356: If you add anymore you may have to start naming them and breaking them out into their own constants. :)

src/languages/php.js
wkania: Its self-reference FUNCTION_INVOKE

joshgoebel: I see, but just add the `contains` key along with all the others when you create the object literal.

joshgoebel, line 410:
```suggestion
    contains:  [
```

joshgoebel: or some such.

src/languages/php.js
wkania: This time that's all :). Back then I didn't notice they break each over.

src/languages/php.js
wkania: TS2448: Block-scoped variable 'FUNCTION_INVOKE' used before its declaration.

src/languages/php.js
wkania: Something like this with the common part?
```
const PARAMS_MODE = [
              VARIABLE,
              LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
              hljs.C_BLOCK_COMMENT_MODE,
              STRING,
              NUMBER
            ]

Comment (joshgoebel) [reply src/languages/php.js:411]: No, move the whole begin: `(` end: `)` rule into named constant `PARAMS_MODE`...

File: src/languages/php.js (hunk 1 above)
Comment (joshgoebel) [reply src/languages/php.js:410]: Oh.  Define what you can, only do the modification after.

```
PARAMS_MODE.contains.push(FUNCTION_INVOKE);
```

File: src/languages/php.js (hunk 3 above)
Comment (wkania) [reply src/languages/php.js]: Done

File: src/languages/php.js (hunk 2 above)
Comment (wkania) [reply src/languages/php.js]: @joshgoebel Done

File: src/languages/php.js [hunk 5]
```diff
@@ -394,6 +393,43 @@ export default function(hljs) {
...
+  const PARAMS_MODE = {
+    relevance: 0,
+    begin: /\(/,
+    end: /\)/,
+    keywords: KEYWORDS,
+    contains: [
+      {
+        scope: 'attr',
+        match: regex.concat(IDENT_RE, regex.lookahead(':'), regex.lookahead(/(?!::)/)),
+      },
```
joshgoebel, line 405:
```suggestion
      NAMED_ARGUMENT,
```

abstract this, like the others please.

joshgoebel (changes requested):
Looking good I think! Almost there!  Thanks for all the effort on PHP!

wkania:
> Looking good I think! Almost there! Thanks for all the effort on PHP!

Thx, but still a few things missing

src/languages/php.js
wkania: Done

src/languages/php.js
```diff
@@ -368,6 +353,20 @@ export default function(hljs) {
           2: "variable.language",
         },
       },
+      {
+        match: [
+          PASCAL_CASE_CLASS_NAME_RE,
+          regex.concat(
+            /::/,
+            regex.lookahead(/(?!class\b)/)
```
joshgoebel, line 361: Can you think of a nice short name for `negative lookahead`?  Really this doesn't need the wrapper, and it's being double wrapped now:

```
/(?=(?!class\b))/
```

I'm not asking you to change it, just if you have any ideas... this comes up a lot and we don't currently have a super elegant way to handling the negative case... Perhaps `lookaheadNot`?

```js
regex.lookaheadNot(/class\b/);
```

wkania: Why just not `regex.negativeLookahead(/(class\b)/)`? After all, that's its name.

joshgoebel (approved):

src/languages/php.js
joshgoebel: maybe.

PR #3462, opened 2022-01-17
Add grep to bash keywords

splch:
### Changes

grep is a very common bash command, so I'm adding to the list.

### Checklist
- [ ] Markup tests don't apply since it's a single word being added
- [x ] Updated the changelog at `CHANGES.md`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,11 @@
 
 Grammars:
 
+- enh(bash) Match `grep` keyword [Spencer Churchill][]
 - enh(php) named arguments [Wojciech Kania][]
 - fix(php) PHP constants [Wojciech Kania][]
 
+[Spencer Churchill]: https://github.com/splch
 [Wojciech Kania]: https://github.com/wkania
 
 ## Version 11.4.0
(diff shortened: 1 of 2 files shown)


joshgoebel:
Is it a GNU core utility though? If not we need a new list...

splch:
> Is it a GNU core utility though? If not we need a new list...

It's referenced in the [coreutils](https://www.gnu.org/software/coreutils/manual/coreutils.html) but I don't know if it's included. In any case grep is part of [every *nix OS](https://wikipedia.org/wiki/Grep) so it's either included there or in the kernel.

joshgoebel:
Doesn't seem to be.

https://github.com/coreutils/coreutils

So lets create a new `POPULAR_UTILS` collection to hold such things...

splch:
> Doesn't seem to be.
> 
> https://github.com/coreutils/coreutils
> 
> So lets create a new `POPULAR_UTILS` collection to hold such things...

Oh thanks for checking, I wasn't sure where to look 😅

That sounds like a great idea! I'll look into it in a little.

joshgoebel:
Ping?

joshgoebel:
Closing for inactivity.  Whenever you have time to work on this again it can be reopened.  Thanks.

PR #3464, opened 2022-01-19
Fix highlighting of int8, int16, int32, int64 in Angelscript

codecat:
### Changes
This fixes types such as `int64` and `uint64` only coloring the `int` and `uint` parts of the type in Angelscript.

### Checklist
- [x] Added markup tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Grammars:
 
 - enh(php) named arguments [Wojciech Kania][]
 - fix(php) PHP constants [Wojciech Kania][]
+- fix(angelscript) incomplete int8, int16, int32, int64 highlighting [Melissa Geels][]
 
 [Wojciech Kania]: https://github.com/wkania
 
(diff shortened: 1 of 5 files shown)


test/markup/angelscript/default.txt
```diff
@@ -15,7 +15,7 @@ namespace MyApplication
         array<SomeClass@> m_children;
         array<array<SomeClass@>> m_subChildren; // Nested templates
 
-        int m_thing;
+        uint64 m_thing;
```
joshgoebel, line 18: You can leave detect alone, but please add tests for BOTH in `markup`. Thanks!

CHANGES.md
```diff
@@ -4,6 +4,7 @@ Grammars:
 
 - enh(php) named arguments [Wojciech Kania][]
 - fix(php) PHP constants [Wojciech Kania][]
+- fix(angelscript) incomplete int8, int16, int32, int64 highlighting [Melissa Geels][]
```
joshgoebel, line 7: Also need a footnote on line ~9.

joshgoebel:
@codecat thanks!

PR #3468, opened 2022-01-23
http-server as dev dependency

Vanderscycle:
added a script so that anyone can have a one quick run and see their changes as is.  Nodemon prob better in the long run.

Just added this changes after being confused as to how to view my changes in the browser.

### Changes

added a script and one dev dependency to package.json so that there's a one stop quick npm command to view the files as is in the browser

### Checklist


package.json
```diff
@@ -49,7 +49,7 @@
     "test-detect": "mocha test/detect",
     "test-browser": "mocha test/browser",
     "test-parser": "mocha test/parser",
-    "view-browser": "npm run build_and_test_browser && http-server ./build/"
```
joshgoebel: Why wouldn't it be rooted into build?

Vanderscycle: I was a bit confused as to which build I had to use. I can make the change.

Vanderscycle: sorry, I was a bit confused. I changed the script a bit so that it focuses on the online tools created by the highlight.js team. The whole purpose of this pr, is so that new contributors can easily use a one script run command to access the online dev tools.

package.json
```diff
@@ -48,7 +48,8 @@
     "test-markup": "mocha test/markup",
     "test-detect": "mocha test/detect",
     "test-browser": "mocha test/browser",
-    "test-parser": "mocha test/parser"
+    "test-parser": "mocha test/parser",
+    "use-browser-tools": "rm -rf ./build && npm run build_and_test_browser && http-server ./tools"
```
joshgoebel, line 52: Wait, this is for tools? Why?  You can just `open tools/developer.html` locally, it doesn't require a web server at all to work...

joshgoebel:
I originally thought this was about demo... in `build`... which I'm not sure just works on it's own (but maybe it also does)... but `tools` always has (I use it regularly)... if it's really about multiple things then we many need a nicer strategy for tooling entirely not just throw a web server over top of it.

What tooling were you having difficult using/seeing originally?

Vanderscycle:
> What tooling were you having difficult using/seeing originally?
Just running the test env, I'm still very junior so most of my website uses Vite which serves the files that I need ```npm run dev```. I know there's a multitude of methods that can be used to serve static files and it took me a while to make it work. 

I was on the fence with this pr as I have trouble explaining why adding helps the project and new contributors in general. I was simply hoping for a single npm/yarn/pnpm/etc. command to serve the static files for a quick view and refresh while you make changes to the new colorscheme.

Regardless, I do not want to create more work, and will keep tinkering with the idea on my own when I decide to add tokyo night storm and light in the not too distant future.

PR #3471, opened 2022-01-28
Improve llvm syntax highlighting

Flakebi:
### Changes
- Numbers are allowed to have a leading + sign
- Hex floats can have a format specified after the 'x'
- Strings cannot escape quotes with backslashes, so a `\"` ends a string

### Checklist
- [x] Updated the changelog at `CHANGES.md` (it’s a pretty small fix, not sure if this needs to be mentioned?)


src/languages/llvm.js
```diff
@@ -109,12 +109,11 @@ export default function(hljs) {
       // another language than an actual comment
       hljs.COMMENT(/;\s*$/, null, { relevance: 0 }),
       hljs.COMMENT(/;/, /$/),
-      hljs.QUOTE_STRING_MODE,
       {
         className: 'string',
         variants: [
           // Double-quoted string
-          { begin: /"/, end: /[^\\]"/ },
+          { begin: /"/, end: /"/ },
```
joshgoebel, line 116: If there is only one rules now then we no longer need the `variants` parent... and do you think while you're here perhaps we could add the actual escaping.  Is it not `\00` style, if I'm not mistaken?  IE, you'd just need a child inside `contains` that matched `\\\d\d` and scoped as `char.escape`.

Thoughts?

joshgoebel:
> (it’s a pretty small fix, not sure if this needs to be mentioned?)

We mention almost all fixes. :)

Flakebi:
Thanks! I fixed your comment and added a changelog entry. (diff of my changes: https://github.com/highlightjs/highlight.js/compare/be47761cb54d2bbeaf16006bb962df1b6f4eecd9..3ee1056de807099140def2e1e362cd6e2ae4853d)

joshgoebel (approved):

joshgoebel:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `./build/es/core.min.js` | 8 kB | +1 B (0%) |
| `./build/es/highlight.min.js` | 8 kB | +1 B (0%) |
| `./build/es/languages/llvm.min.js` | 1.44 kB | +7 B (0%) |
| `./build/highlight.min.js` | 8.04 kB | -1 B (0%) |
| `./build/languages/llvm.min.js` | 1.45 kB | +8 B (+1%) |

joshgoebel:
@Flakebi Thanks so much!

Issue #3470, opened 2022-01-28
[Enhancement request] Allow importing named language imports


webketje:
**Is your request related to a specific problem you're having?**
When using a bundler on the front-end it is important to only include the required languages to reduce bundle size,
but it feels a bit clunky to register languages like this: 
```js
import hljs from 'highlight.js/lib/core';
import js from 'highlight.js/lib/languages/javascript';
import json from 'highlight.js/lib/languages/json';
import yml from 'highlight.js/lib/languages/yaml';
import md from 'highlight.js/lib/languages/markdown';

hljs.registerLanguage('javascript', js);
hljs.registerLanguage('json', json);
hljs.registerLanguage('yaml', yml);
hljs.registerLanguage('markdown', md);
```

**The solution you'd prefer / feature you'd like to see added...**
It would be much better if it could just be done like this:

```js
import hljs from 'highlight.js/lib/core';
import { javascript, json, yaml, markdown } from 'highlight.js/lib/languages';

hljs.registerLanguage('javascript', javascript);
hljs.registerLanguage('json', json);
hljs.registerLanguage('yaml', yaml);
hljs.registerLanguage('markdown', markdown);
```
This only requires adding `lib/languages/index.js` which has lines like: 

```js
export { default as json } from './json';
export { default as javascript } from './javascript';
// aliases could be done like this:
export { default as js } from './javascript';
```
I could make a PR for this (please mention whether aliasing is desirable, I believe this could also be implemented in client code as `import { javascript as js } from 'highlight.js/languages'`)

joshgoebel:
Aren't dynamic imports the true solution to this problem?

webketje:
Ok for aliases, but..

> Aren't dynamic imports the true solution to this problem?

..misses the point (the issue is about DX). I'll try to rephrase: requiring a user (developer) of highlight.js to write 10 separate import lines is way less developer-friendly then allowing them to write a single import line with named imports, like `import {lang1, lang2, lang3, lang4 } from 'highlight.js/languages'` 

Static vs dynamic imports will work with both

joshgoebel:
> the issue is about DX)

Exactly, when I said dynamic I was imagining of something like:

```js
const supportedLanguages = ["basic", "cobol", "other"]
supportedLanguages.forEach(lang => {
  import(`${languages_dir}/${lang}`).then(def => hljs.registerLanguage(def)
});
```

No need to repeat anything.

webketje:
[...]

PR #3473, opened 2022-02-01
(enh) named lang exports index candidate impl for Node & bundler builds (#3470)

webketje:
This is an exploratory PR for issue #3470

### Changes
- Adds an index generation script under `tools/lib`
- Uses that script in `tools/build_node.js`
- Outputs 2 extra files `build/lib/languages/index.js` and `build/es/languages/index.js`

To test that this worked as intended for Node I created 2 files (locally):
`test-import.mjs`
```js
import { abnf, oneC, xml } from './build/es/languages/index.js';
import hljs from './build/es/index.js';

hljs.registerLanguage('abnf', abnf);
hljs.registerLanguage('1c', oneC);
hljs.registerLanguage('xml', xml);
```
`test-import.js`
```js
const { abnf, oneC, xml } = require('./build/lib/languages');
const hljs = require('./build/lib');

hljs.registerLanguage('abnf', abnf);
hljs.registerLanguage('1c', oneC);
hljs.registerLanguage('xml', xml);

console.log(hljs.getLanguage('abnf'));
```

### Checklist
- [x] Markup tests don't apply here
- [ ] TBD: name conversion for languages which are stored in a file with a `-` (dash) are camelCased, e.g `pythonRepl`
- [x] TBD: name conversion for languages which start with a number converts the first number to textual representation and camelCases the rest (eg `1c` becomes `oneC`). This is required for ES exports and has been kept for CJS exports to keep consistency

### DIFF
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -6,7 +6,7 @@ const { getLanguages } = require("./lib/language.js");
 const { install, mkdir, installCleanCSS } = require("./lib/makestuff.js");
 const { filter } = require("./lib/dependencies.js");
 const { rollupWrite } = require("./lib/bundling.js");
-const { generateCJSLanguageIndex, generateESLanguageIndex } = require('./lib/indexgen.js');
+const { generateCJSLanguageIndex, generateESLanguageIndex } = require('./lib/language_indexgen.js');
 const log = (...args) => console.log(...args);
 
 // https://nodejs.org/api/packages.html#packages_writing_dual_packages_while_avoiding_or_minimizing_hazards
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -6,6 +6,7 @@ const { getLanguages } = require("./lib/language.js");
 const { install, mkdir, installCleanCSS } = require("./lib/makestuff.js");
 const { filter } = require("./lib/dependencies.js");
 const { rollupWrite } = require("./lib/bundling.js");
+const { generateCJSLanguageIndex, generateESLanguageIndex } = require('./lib/indexgen.js');
 const log = (...args) => console.log(...args);
 
 // https://nodejs.org/api/packages.html#packages_writing_dual_packages_while_avoiding_or_minimizing_hazards
@@ -195,6 +196,8 @@ async function buildNode(options) {
   await buildCJSIndex("index", languages);
   await buildCJSIndex("common", common);
   await buildLanguages(languages, options);
+  await fs.writeFile(`${process.env.BUILD_DIR}/lib/languages/index.js`, generateCJSLanguageIndex());
+  await fs.writeFile(`${process.env.BUILD_DIR}/es/languages/index.js`, generateESLanguageIndex());
 
   log("Writing highlight.js");
   await buildNodeHighlightJS(options);
--- a/tools/lib/language_indexgen.js
+++ b/tools/lib/language_indexgen.js
@@ -1,0 +1,48 @@
+const { basename } = require('path');
+const glob = require('glob');
+
+// if a language name starts with a number, we convert it to a textual representation
+// as variables in JS cannot start with numbers:
+const numbers = ['zero', 'one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight', 'nine'];
+
+const allLangs = glob.sync("./src/languages/*.js");
+
+/**
+ * @param {string} str
+ */
+const camelcasify = (str) => {
+  while (str.match(/-/)) {
+    str = str.slice(0, str.indexOf('-')) + str[str.indexOf('-') + 1].toUpperCase() + str.slice(str.indexOf('-') + 2);
+  }
+  return str;
+};
+
+function generateESLanguageIndex() {
+  let indexContents = '';
+  allLangs.forEach(langDef => {
+    const langPath = `./${basename(langDef)}`;
+    let langExport = basename(langDef, '.js');
+    if (!isNaN(langExport[0])) langExport = `${numbers[parseInt(langExport[0])]}-${langExport.slice(1)}`;
+    const exportLine = `export { default as ${camelcasify(langExport)} } from '${langPath}';\n`;
+    indexContents += exportLine;
+  });
+  return indexContents;
+}
+
+function generateCJSLanguageIndex() {
+  let indexContents = 'module.exports = {\n';
+  allLangs.forEach((langDef, i) => {
+    const langPath = `./${basename(langDef)}`;
+    let langExport = basename(langDef, '.js');
+    if (!isNaN(langExport[0])) langExport = `${numbers[parseInt(langExport[0])]}-${langExport.slice(1)}`;
+    const exportLine = `  ${camelcasify(langExport)}: require('${langPath}')${i === allLangs.length - 1 ? '' : ','}\n`;
+    indexContents += exportLine;
+  });
+  indexContents += '};\n';
+  return indexContents;
+}
+
+module.exports = {
+  generateCJSLanguageIndex,
+  generateESLanguageIndex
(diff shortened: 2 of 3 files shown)


tools/lib/language_indexgen.js
```diff
@@ -0,0 +1,48 @@
+const { basename } = require('path');
+const glob = require('glob');
+
+// if a language name starts with a number, we convert it to a textual representation
+// as variables in JS cannot start with numbers:
+const numbers = ['zero', 'one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight', 'nine'];
```
joshgoebel, line 6: I don't love this complexity but see why you feel it's needed since `L_langname` or some such is a bit ugly for external consumption... is there a reason we couldn't just export an object literal with named keys?  I know `rollup` can do dead-code elimination of such things fairly simple.

```
import languages from 'highlight.js/lib/languages';
hljs.registerLanguage("xml", languages.xml)
hljs.registerLanguage("basic", languages.basic)
```

tools/build_node.js
```diff
@@ -195,6 +196,8 @@ async function buildNode(options) {
   await buildCJSIndex("index", languages);
   await buildCJSIndex("common", common);
   await buildLanguages(languages, options);
+  await fs.writeFile(`${process.env.BUILD_DIR}/lib/languages/index.js`, generateCJSLanguageIndex());
+  await fs.writeFile(`${process.env.BUILD_DIR}/es/languages/index.js`, generateESLanguageIndex());
```
joshgoebel, line 200:
```suggestion
  await fs.writeFile(`${process.env.BUILD_DIR}/lib/languages/index.js`, generateCJSLanguageIndex(languages));
  await fs.writeFile(`${process.env.BUILD_DIR}/es/languages/index.js`, generateESLanguageIndex(languages));
```

The build system languages are a build time choice, so index generation would need to take the build time language list as input.

tools/lib/language_indexgen.js
joshgoebel: Though you're still left with `languages["1c"]` for the outliers I guess vs `.` notation... so both of these are annoying in that we now have document this difference in behavior somewhere... and it's still going to trip someone up who's not expecting it.

webketje: @joshgoebel actually I wanted to use `_` but changed it due to existing project linting rules turning the whole editor red.

> L_langname or some such is a bit ugly for external consumption.

That would definitely solve the issue at hand (=having aligned, logically named CJS/EJS named exports) in a generic way, albeit an ugly one..

> is there a reason we couldn't just export an object literal with named keys?

For CJS that would work. But `import { 1c } from 'highlight.js/lib/languages'` would throw. And I think `import oneC from './1c'; export default { ['1c']: oneC }` will be handled differently from `export { default as oneC }`  (as in this will probably include the whole default export for EJS builds)

joshgoebel:
>  TBD: name conversion for languages which are stored in a file with a - (dash) are camelCased, e.g pythonRepl

I'd think it would be PascalCase or snake_case though... personally I'd lean towards snake case (to avoid someone having to rejigger the names in their head)... but if your rejoiner was PascalCase was a more idiomatic I could be won over easily enough.

> `oneC`

This is growing on me and we only have a single numeric starter... so... *shrugs*

tools/lib/language_indexgen.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+  return indexContents;
+}
+
+function generateCJSLanguageIndex() {
+  let indexContents = 'module.exports = {\n';
+  allLangs.forEach((langDef, i) => {
+    const langPath = `./${basename(langDef)}`;
+    let langExport = basename(langDef, '.js');
+    if (!isNaN(langExport[0])) langExport = `${numbers[parseInt(langExport[0])]}-${langExport.slice(1)}`;
```
joshgoebel, line 37: This and line 25 looks like it should be extracted into a function (or perhaps added to whatever the final name of `camel*` turns out to be)...

webketje:
> I'd think it would be PascalCase or snake_case though... personally I'd lean towards snake case (to avoid someone having to rejigger the names in their head)... but if your rejoiner was PascalCase was a more idiomatic I could be won over easily enough. Internally these are all anonymous functions, but if we were to "upgrade" them one day they'd probably be PascalCase named classes or something.

Idiomatically is PascalCase for classes, camelCase for instances/vars, and snake_case, eh.. for data & i18n vars ? So you could have a base class `Lang` and instances `myLang`, or you could have a base class `Lang` and extending classes `MyLang`. I don't mind either. I don't think it matters.. it's different for each lang too.. Feels more "right" to write PHP (in capitals) & javascript (in regular) & Python (capitalized). So I'll follow your pick. Exports can be renamed by users if they wish `import { MyLang as myLang } from`.

PS: is it desired for the index files to have `.js` appended to each line ? I found this weird, is it intentional?

joshgoebel:
> PascalCase for classes, camelCase for instances/vars,

Ah I guess in my head I think of them as "entities", it does seem you could go either way... Lets do `oneC`  then because it looks nicer than `OneC` (reminds me of Hi-C fruit punch drink or something)... and they aren't classes right now anyways... We also have `language.name` or the metadata in the headers we could rely on, but I'm not sure how that would help us in this case.

> PS: is it desired for the index files to have .js appended to each line ? I found this weird, is it intentional?

Aren't you building them from scratch, or are you talking about our other files?  

```
commit a010c151328b8944a7a03e5335956de8baf63295
Author: Antoine du Hamel <duhamelantoine1995@gmail.com>
Date:   Thu Jun 11 18:35:38 2020 -0400

    (chore) add `.js` extension to import statements (#2601)

    Adds file extensions to all import specifiers in ./src/ files.  This is useful to run the files straight from source with a web browser , Node.js ESM or Deno.

    - Also add eslint rules regarding extensions for imports
```

Relevant?

joshgoebel:
`Error: Cannot find module './lib/indexgen.js'
Require stack:`

webketje:
@joshgoebel Ah I must have missed that when I renamed it to `language_indexen.js`. Sorry have been caught up with maintenance on [metalsmith](https://github.com/metalsmith)

joshgoebel:
I just was reviewing this... doesn't this potentially result in SLOWER builds/more CPU/more memory usage for someone using these "meta" indexes vs just requiring/importing exactly what they need... ie, in the CJS case doesn't this now require requiring and parsing a full 1mb of code (all grammars) even if someone only intends to use 10 languages and a fraction of that?

Wouldn't this also hold true for ESM... I'm talking specifically of the Node library being used directly, not bundling for the client... I could imagine a bundler not walking all of those imports, though I'm not sure to what degree such optimizations are implemented in practice.

joshgoebel:
Also I think client-side ESM is likely the future (long-term)... so how does this perform there?   Wouldn't using these "convenience" indexes result in loading 1mb of JS even if only 50kb is necessary?

webketje:
@joshgoebel Interesting point! I created 4 variants of a script (in the lib/es build folders, one for direct import, another for index-linked import) and logged the RAM heap used. The scripts only do a single `require/import`. Below are the results (multiple runs yield the same results

```bash
▲ = node ./build/es/test-es-imports-index
The script uses approximately 7.31 MB
▲ = node ./build/es/test-es-imports
The script uses approximately 3.83 MB
▲ = node ./build/lib/test-cjs-imports-index
The script uses approximately 7 MB
▲ = node ./build/lib/test-cjs-imports
The script uses approximately 3.52 MB
```
So correct, it adds about 3.5MB of RAM for all languages

joshgoebel:
@webketje I seriously appreciate your time and research here just don't think it's a good direction at this time.

PR #3478, opened 2022-02-11
improve elixir language recognition and support

brainlid:
This improves Elixir language recognition and support.

Minor: Adds the keyword `defguard`. Pronounced "def guard".

Major: Improve highlight support for Elixir. Specifically when recognizing and coloring references to a module.

### Changes

The areas indicated in the graphic are colored yellow, the same color as the module name in the declaration. Prior to this change, those all appear as a plain term and are gray.

Screenshot_20220211_075221

### Checklist


src/languages/elixir.js
```diff
@@ -258,6 +259,10 @@ export default function(hljs) {
       begin: ELIXIR_IDENT_RE + ':(?!:)',
       relevance: 0
     },
+    { // Usage of a module, struct, etc. Customized and paired with atom-one-dark.css customization
+      className: 'title.class',
+      begin: '([A-Z][a-zA-z0-9_]+)'
```
joshgoebel, line 264:
```suggestion
      begin: '([A-Z][a-zA-Z0-9_]+)'
```

No?

CHANGES.md
```diff
@@ -14,6 +14,7 @@ Grammars:
 - enh(ts) modify TypeScript-specific keywords and types list [anydonym][]
 - fix(brainfuck) fix highlighting of initial ++/-- [Christina Hanson][]
 - fix(llvm) escaping in strings and number formats [Flakebi][]
+- fix(elixir) recognize references to modules [Mark Ericksen][]
```
joshgoebel, line 17:
```suggestion
- enh(elixir) recognize references to modules [Mark Ericksen][]
```

joshgoebel:
Looks reasonable, but we also need some markup tests to verify the behavior of the new module highlight rule.

brainlid:
> Looks reasonable, but we also need some markup tests to verify the behavior of the new module highlight rule.

Great! Now I just need to figure out how to do that. :)

brainlid:
Making progress

brainlid:
@joshgoebel added new test file and updated others for the added highlighting. 

I believe it is ready. Let me know if there's anything else I need to do. Thankd!

joshgoebel:
Does ALL_CAPS have any traditional semantic meaning in Elixir?

brainlid:
> Does ALL_CAPS have any traditional semantic meaning in Elixir?

I can't think of anything. It's valid code, but it's not used for things like constants as in other languages. Is there something specific you're asking about?

joshgoebel:
Well some languages make a distinction and right now your regex matches both so had to ask. So if there is no real convention then I assume assuming it's a module would be fine?

brainlid:
> Well some languages make a distinction and right now your regex matches both so had to ask. So if there is no real convention then I assume assuming it's a module would be fine?

Ah, I see. Thanks for clarifying. Yes. In Elixir it's really just modules, functions, and data. A module starts with an uppercase letter. The rest can also be upper but Pascal case is the convention.

Thanks for following through. I appreciate the work that's been done on the library.

joshgoebel (changes requested):

src/languages/elixir.js
```diff
@@ -258,6 +259,11 @@ export default function(hljs) {
       begin: ELIXIR_IDENT_RE + ':(?!:)',
       relevance: 0
     },
+    { // Usage of a module, struct, etc.
+      className: 'title.class',
+      begin: '([A-Z][a-zA-Z0-9_]+)',
```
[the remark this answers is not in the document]
brainlid: I added a word break condition and updated the tests to verify. All passing.

brainlid:
Thanks for the feedback. How's it look now?

src/languages/elixir.js
```diff
@@ -258,6 +259,11 @@ export default function(hljs) {
       begin: ELIXIR_IDENT_RE + ':(?!:)',
       relevance: 0
     },
+    { // Usage of a module, struct, etc.
+      className: 'title.class',
+      begin: '(\\b[A-Z][a-zA-Z0-9_]+)',
```
joshgoebel, line 264:
```suggestion
      begin: /(\b[A-Z][a-zA-Z0-9_]+)/,
```

joshgoebel:
@brainlid Thanks so much!

PR #3482, opened 2022-02-15
(themes) Add ```tokyo-night-light```

Vanderscycle:
### Changes

added ```tokyo-night-light``` and made some corrections to comments on my previous pr. Original [repo](https://github.com/enkia/tokyo-night-vscode-theme)
image

checkTheme.js resultss
image

### Checklist


joshgoebel:
How much of this is a "direct port" vs a HLJS imagining of the theme?  I was going to ask you to tune that 1.96 ratio to at least a 3, but now I'm wondering how we should treat "externally" ported themes from an accessibility POV.  IE, should baseline accessibility or "fidelity to the original" be more important drivers?

@highlightjs/core  Any thoughts?

Vanderscycle:
Its an accurate port as far as my adapting process went, but I did have to squint my eyes at certain parts of the code. I should have created this pr as a draft first.

allejo:
> I think perhaps for accessibility that a pass over ALL the `comment` styles in **all** (1st party, not Base16) themes and pulling them up to a `3` or `4.5` contrast and then recommend that those who want the common "i can barely see the comments" look/feel should apply `opacity` of their desired choosing to `hljs-comment`. That way we equally support both accessibility and user choice.

I _really_ like the `opacity` idea! i.e. be accessible by default, and you have to opt-in to make it less accessible. You got my vote on this approach.

However, I agree with your statement of splitting this off into a separate thread. I feel like it'd be appropriate to have some discussion/thoughts regarding how we define an "accessible theme." [WCAG is flawed in some cases](https://blog.datawrapper.de/color-contrast-check-data-vis-wcag-apca/) and there's a new player in town: [APCA (it even comes with its own npm package we can use!)](https://github.com/w3c/wcag/issues/695#issuecomment-986154459). So I'd want to explore these options before we make a decision.

But for my immediate response as to what to do right now for new themes: default to accessible colors, opt-into less accessible with `opacity`. Yes, we're porting a theme from another location and ideally, we'd want to be loyal to the original, but that doesn't mean we should shrug our shoulders and be the provider of something inaccessible if it's being introduced into our product.

tools/developer.html
```diff
@@ -1,22 +1,27 @@
...
   <meta charset="utf-8">
   <title>highlight.js developer</title>
 
   <link rel="stylesheet" href="../src/styles/default.css">
 
   <style>
-    body { padding: 20px;
-    font-family: sans-serif;
-     }
+    body {
```
joshgoebel, line 11: Please undo all the linting type changes to this file and only include the 2-3 line changes that are relevant to this PR.

joshgoebel:
Please see review comments.  Hope to push out a release soon, if this is ready it can be part of 11.5 :-)

Vanderscycle:
@joshgoebel I will work on it today, sorry that my pr is taking forever.

tools/developer.html
Vanderscycle: undid all of my linting changes, sry

joshgoebel:
@Vanderscycle  No problem, glad to see it getting finished up!

PR #3492, opened 2022-02-25
[draft] XML processing directives

### LINKED ISSUE #3479: XML mode doesn't highlight processing instructions
`highlight.js` does not highlight processing instructions in XML mode (except for the pseudo processing instruction `<?xml ...?>`)

I am using `highlight.js` 11.4.0 and the following HTML code:

```html
<pre class="language-xml">
&lt;?ul4 langs(*names)?&gt;
&lt;?whitespace strip?&gt;
&lt;?doc Languages?>
&lt;ul>
	&lt;?for n in names?>
		&lt;li>
			&lt;?printx n?>
		&lt;/li>
	&lt;?end for?>
&lt;/ul> 
</pre>
```

I include `highlight.js` like this (in the `<head>`):

```html
<script src="/static/highlight.js/11.4.0/highlight.min.js"></script>
```

I activate `highlight.js` by the following `<script>` at the end of the `<body>`:

```html
<script>
	document.addEventListener('DOMContentLoaded', (event) => {
		document.querySelectorAll('pre, code').forEach(el => {
			hljs.highlightElement(el);
		});
	});
</script>
```

The only highlighted parts in the result however are the start and end tags. None of the processing instructions get hightlighted:

Bildschirmfoto 2022-02-14 um 11 55 36

For information about XML processing instructions see: https://www.w3.org/TR/REC-xml/#sec-pi

joshgoebel:
Closes #3479.

### Checklist


test/markup/index.js
```diff
@@ -31,7 +31,7 @@ function testLanguage(language, {testDir}) {
 
           // Uncomment this for major changes that rewrite the test expectations
           // which will then need to be manually compared by hand of course
-          // require('fs').writeFileSync(filename, actual);
+          require('fs').writeFileSync(filename, actual);
```
allejo, line 34: Should be re-commented again?

allejo (approved):
These changes look fine to me and fix the problem at hand. I'm not sure how I feel about just highlighting strings in directives and not their respective attributes (we highlight "attributes" in the meta `<!` sections).

image

Taking a look at two separate editors, it looks like they highlight the attributes in directives.

image
image

test/markup/index.js
joshgoebel: Yeah.

joshgoebel:
We also have the problem of things inside `meta` not being styled super well (or consistently) historically other than `meta string` and `meta keyword`...   so that's a great point but I think I'm happy to circle back and solve it another day. :)

allejo:
> ...I think I'm happy to circle back and solve it another day

I'm fine with that, it's nothing blocking this PR imo

Issue #3489, opened 2022-02-24
(bash) keywords with numbers are not recognized


mmattel:
**Describe the issue**
I have added some code to add custom keywords post the last register entry which is:
`hljs.registerLanguage('yaml', require('highlight.js/lib/languages/yaml'))`
```
const kwds = ['apt', 'apt-get', 'pecl',
'phpenmod', 'phpdismod', 'a2enmod', 'a2dismod', 'a2ensite', 'a2dissite',
'systemctl', 'service', 'sudo', 'wget', 'tar', 'mysql', 'php', 'grep'];

var built_in = hljs.getLanguage('bash').keywords.built_in;
hljs.getLanguage('bash').keywords.built_in = [...new Set([...built_in ,...kwds])];
```

This works great and tested with many scripts to highlight keywords - with one exception:

```
echo "Enabling Apache Modules"
a2enmod dir env headers mime rewrite setenvif
service apache2 reload
```

`echo` and `service` (added) gets correctly highlighted, but not `a2enmod`. After some testes, I renamed in both the text to highlight and the constant  `a2enmod` -->  `aenmod` and this works. The same is true for other a2xxx keywords.

**Which language seems to have the issue?**
`bash` but I guess this is not dependent to a language

**Are you using `highlight` or `highlightAuto`?**
`hljs.highlightElement(node)`

**Sample Code to Reproduce**
See above

**Expected behavior**

Keywords containing numbers should be recognized normally

**Additional context**

joshgoebel:
> `$pattern: /\b[a-z._-]+\b/,`

Our `keywords.$pattern` here is too narrow... we probably need to match `[a-z]` first then also open it up for numerics... technically you could probably start with a number as well but I'd be worried about false positives (performance issue) and I'm also not sure that's a super common thing... is it?

None of the commands/keywords we have so far seem to start with a number.

PR would be welcome.

mmattel:
Regarding a PR, I think you are much faster than I am as you know where to change. The following tests have been made with [regex101](https://regex101.com) in ECMAScript mode and with the [Java Regular Expression Tester](https://www.freeformatter.com/java-regex-tester.html) - both showing the same results.

**I have found a possible regex:** (note it is a bit faster than the original 😃 )
`\b[a-z][a-z0-9._-]+\b`

![image](https://user-images.githubusercontent.com/3321281/155738816-f1972d5d-7d7b-4976-9a1c-cf662ae3ce4d.png)

[...]

PR #3494, opened 2022-02-26
enh(bash) `$pattern`: allow numbers after the first character

mmattel:
Fixes: https://github.com/highlightjs/highlight.js/issues/3489 ((bash) keywords with numbers are not recognized)

With the new regex, also numbers INSIDE the bash pattern is allowed.
Before, eg. `a2enmod` would not have been recognized. Now it gets properly recognized.

For the details of the test results see: https://github.com/highlightjs/highlight.js/issues/3489#issuecomment-1050938466

@joshgoebel

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ New Grammars:
 
 Grammars:
 
+- enh(bash) improved pattern regex [Martin Mattel][]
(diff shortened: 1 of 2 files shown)


joshgoebel (approved):

joshgoebel:
@mmattel Thanks so much!

PR #3499, opened 2022-03-01
enh(objectivec) add scope `types`

joshgoebel:
### Checklist
- [x] Added markup tests, or they don't apply here because we don't need markup tests for basic keywords support

### DIFF
--- a/src/languages/objectivec.js
+++ b/src/languages/objectivec.js
@@ -12,11 +12,21 @@ export default function(hljs) {
     begin: '\\b(AV|CA|CF|CG|CI|CL|CM|CN|CT|MK|MP|MTK|MTL|NS|SCN|SK|UI|WK|XC)\\w+'
   };
   const IDENTIFIER_RE = /[a-zA-Z@][a-zA-Z0-9_]*/;
-  const KWS = [
+  const TYPES = [
     "int",
     "float",
-    "while",
     "char",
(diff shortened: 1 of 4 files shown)


ElijahPepe:
LGTM.

src/languages/objectivec.js
```diff
@@ -12,46 +12,51 @@ export default function(hljs) {
...
+    "signed",
+    "short",
+    "long",
+    "double",
+    "wchar_t",
+    "unichar",
+    "void",
+    "bool",
+    "BOOL",
+    "id|0",
```
allejo, line 29: What's this? I can't find any documentation on what `id|0` means in obj-c.

joshgoebel: <https://www.google.com/search?client=safari&rls=en&q=objective+c+types+%22id%22&ie=UTF-8&oe=UTF-8>

Just what google told me...

allejo: Ah I was looking for the `id|0` as a whole and couldn't find anything. What does the `|0` mean?

joshgoebel: 0 relevance.

allejo: 🤦  I didn't realize that was shorthand for relevance

allejo (approved):

PR #3511, opened 2022-03-28
explicitly set sideEffect for css and scss files, fixes #3504

### LINKED ISSUE #3504: sideEffect: false, stops webpack importing css

In 11.5.0, `sideEffects: false` was added to the `package.json`. This has stopped the styles being imported by webpack which now sees it as an unused zero side effect file. See https://github.com/webpack-contrib/mini-css-extract-plugin/issues/102#issuecomment-384947142.

dirkjf:
make it possible to import css or scss using ES6 `import 'file.css'`

 fixes #3504

### Changes
`"sideEffects": false,` to `"sideEffects": ["*.css", "*.scss"],`  

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "type": "commonjs",
   "main": "./lib/index.js",
   "types": "./types/index.d.ts",
-  "sideEffects": false,
+  "sideEffects": ["*.css", "*.scss"],
   "scripts": {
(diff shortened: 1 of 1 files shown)


joshgoebel (approved):

joshgoebel:
@dirkjf Thanks!

PR #3513, opened 2022-03-29
Assign authorship to Josh Goebel

isagalaev:
I'm getting contacted occasionally by people who look up the
package's author from npm's registry. Package.json doesn't have a
dedicated 'maintainer' field, so 'author' should work, I think.

(Don't think this change deserves to be in CHANGES.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,10 +7,7 @@
   ],
   "homepage": "https://highlightjs.org/",
   "version": "11.5.0",
-  "author": {
-    "name": "Ivan Sagalaev",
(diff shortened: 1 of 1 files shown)


joshgoebel:
Oh fun!  I wasn't going to presume to change this since author is different from maintainer, but since you're on board... :-). 

Hopefully you'll get bothered less in the future.

joshgoebel (approved):

isagalaev:
Yeah, it's one of those where practicality beats purity :-)

PR #3520, opened 2022-04-12
fix(markdown) strong/emphasis requires whitespace after

### LINKED ISSUE #3519: (markdown) List items inside blockquotes/HTML comments aren't highlighted correctly
Consider this markdown:

```
> * One (this point is italic)
> * Two
> * Three

This is italic, and because of a lack of termination (#21), everything after is italic

<!-- * but a comment terminates the italic -->

Regular text
```

It gets highlighted like this:

incorrect markdown highlighting

----

It is, however, valid markdown:

> * One (this point is italic)
> * Two
> * Three

This should not be italic

Regular text

---

Compare with the following markdown:

```
> - One
> - Two
> - Three

some text
```

that gets highlighted like this:

correct markdown highlighting

or this:

```
* One
* Two
* Three

some text
```

which gets highlighted like this:

correct markdown highlighting

joshgoebel:
Resolves #3519.  Prevents false positive of emphasis when bullets are nested inside block quotes.  This doesn't detect the bullets, it only fixes the false positive.

### Checklist

### DIFF
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -138,11 +138,11 @@ export default function(hljs) {
     contains: [], // defined later
     variants: [
       {
-        begin: /_{2}/,
+        begin: /_{2}(?!\s)/,
         end: /_{2}/
       },
       {
-        begin: /\*{2}(?![ ])/,
+        begin: /\*{2}(?!\s)/,
         end: /\*{2}/
       }
     ]
@@ -152,11 +152,11 @@ export default function(hljs) {
     contains: [], // defined later
     variants: [
       {
-        begin: /\*(?![* ])/,
+        begin: /\*(?![*\s])/,
         end: /\*/
       },
       {
-        begin: /_(?!_)/,
+        begin: /_(?![_\s])/,
         end: /_/,
         relevance: 0
       }
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -142,7 +142,7 @@ export default function(hljs) {
         end: /_{2}/
       },
       {
-        begin: /\*{2}/,
+        begin: /\*{2}(?![ ])/,
         end: /\*{2}/
       }
     ]
@@ -152,7 +152,7 @@ export default function(hljs) {
     contains: [], // defined later
     variants: [
       {
-        begin: /\*(?!\*)/,
+        begin: /\*(?![* ])/,
         end: /\*/
       },
       {
--- a/test/markup/markdown/bold_italics.expect.txt
+++ b/test/markup/markdown/bold_italics.expect.txt
@@ -19,3 +19,15 @@
 
 <span class="hljs-strong">**<span class="hljs-emphasis">*This is bold and italic*</span>**</span>
 <span class="hljs-strong">__<span class="hljs-emphasis">_This is bold and italic_</span>__</span>
+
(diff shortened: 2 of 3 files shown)


allejo:
Gonna start a quick discussion on this one. I did some light reading on the commonmark spec and learned two things:

- newlines count as terminating the given bold/italic sequence
- we'd get false positives with `__` too if there's a space after it

Given that any whitespace (including newlines) terminate the sequence, maybe use `\s` instead of `[ ]`. So I've got this snippet:

```js
const BOLD = {
  className: 'strong',
  contains: [], // defined later
  variants: [
    {
      begin: /_{2}(?!\s)/, // <- changed
      end: /_{2}/
    },
    {
      begin: /\*{2}(?!\s)/, // <- changed
      end: /\*{2}/
    }
  ]
};
const ITALIC = {
  className: 'emphasis',
  contains: [], // defined later
  variants: [
    {
      begin: /\*(?![*\s])/, // <- changed
      end: /\*/
    },
    {
      begin: /_(?![_\s])/, // <- changed
      end: /_/,
      relevance: 0
    }
  ]
};
```

_(I'd submit it as a review but GitHub doesn't let me make changes on non-touched lines)_

image

This PR as-is would result in the following,

image

allejo:
The markdown I'm using:

```markdown
** i shouldn't be italic**

* and I shouldn't be italic either*

__ not really bold__

_ not italic_

Hello

**actually bold**

__again, bold__

*italic*
_oh i'm italic_
```

How GitHub renders this

---

** i shouldn't be italic**

* and I shouldn't be italic either*

__ not really bold__

_ not italic_

Hello

**actually bold**

__again, bold__

*italic*
_oh i'm italic_

double-beep:
> - newlines count as terminating the given bold/italic sequence

Not *completely* accurate:

```
**not bold
**

*not italic
*

**should be
bold**

*should be
italic*
```

---

**not bold
**

*not italic
*

**should be
bold**

*should be
italic*

joshgoebel:
> So I've got this snippet:

Did you try and see if all the tests still passed with that snippet?

allejo:
> > So I've got this snippet:
> 
> Did you try and see if all the tests still passed with that snippet?

Ran it against `npm run build_and_test` and just got an auto-detect failure (but I think that's due to my Node version; this happens on `main` for me too).

```
  1720 passing (14s)
  3 pending
  1 failing

  1) hljs.highlightAuto()
       should be detected as sfz:

      AssertionError: sample.txt should be detected as sfz, but was gradle
      + expected - actual

      -gradle
      +sfz
```

---

> Not _completely_ accurate:
> 
> ```
> **not bold
> **
> 
> *not italic
> *
> 
> **should be
> bold**
> 
> *should be
> italic*
> ```

Ahhh good catch on this, you're right. Anything that's not "punctuation" before the closing sequence keeps it bold/italic. But if it's just a newline and closing, then it's not valid since the newline is considered "punctuation"

allejo:
I don't mean to overstep and push to your branch, @joshgoebel! I'd like to vote that we fix the immediate false positives by disallowing newlines inside of bolds and italics. And in the future, come up with a more robust expression that matches the markdown spec and can handle newlines.

allejo (approved):

PR #3523, opened 2022-04-15
Support `distributed` keyword for Swift

mportiz08:
This new syntax is in support of the ["Distributed Actor Isolation"][1] feature that has been accepted for Swift 5.7.

[1]: https://github.com/apple/swift-evolution/blob/main/proposals/0336-distributed-actor-isolation.md

### Changes

New Swift language keyword: `distributed`

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Grammars:
 - enh(swift) add support for `distributed` keyword [Marcus Ortiz][]
 
 [Bradley Mackey]: https://github.com/bradleymackey
+[Marcus Ortiz]: https://github.com/mportiz08
 
 ## Version 11.5.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 Grammars:
 
 - enh(swift) add SE-0335 existential `any` keyword (#3515) [Bradley Mackey][]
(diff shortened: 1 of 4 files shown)


mportiz08:
@joshgoebel I noticed that the linter check is failing, but the issues seem to stem from files not touched by this PR, and I believe the same check is failing on `main`.

Let me know if I missed anything there though, thanks!

CHANGES.md
```diff
@@ -3,6 +3,7 @@
 Grammars:
 
 - enh(swift) add SE-0335 existential `any` keyword (#3515) [Bradley Mackey][]
+- enh(swift) add support for `distributed` keyword [Marcus Ortiz][]
 
 [Bradley Mackey]: https://github.com/bradleymackey
```
joshgoebel, line 8: You need to add a link ref for your name here as well, it doesn't matter if it's way down below somewhere, each release needs their own link references.

joshgoebel (changes requested):
Otherwise looks good.

CHANGES.md
mportiz08: Got it, thanks. [Updated](https://github.com/highlightjs/highlight.js/pull/3523/commits/69eb3f1ddbcdf85f55477db1542ee7a9f4340b5f)

joshgoebel (approved):

PR #3529, opened 2022-04-26
Introducing Unicode letter support for XML element names/tags and attribute names

### LINKED ISSUE #3256: (XML) Unicode letters should be recognized in element and attribute names

martin-honnen:
Resolves #3256 

### Changes
Fixed the xml regular expressions to use the Unicode "letter" class/category `L` and Unicode based regular expressions with e.g. `/.../u` so that, instead of only ASCII characters, any Unicode letter in XML element/tag names and XML attribute names is recognized and the markup is properly highlighted.

### Checklist
- [x] Added two markup tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,9 +4,11 @@ Grammars:
 
 - enh(swift) add SE-0335 existential `any` keyword (#3515) [Bradley Mackey][]
 - enh(swift) add support for `distributed` keyword [Marcus Ortiz][]
+- enh(xml) recognize Unicode letters instead of only ASCII letters in XML element and attribute names (#3256)[Martin Honnen][]
 
 [Bradley Mackey]: https://github.com/bradleymackey
 [Marcus Ortiz]: https://github.com/mportiz08
+[Martin Honnen]: https://github.com/martin-honnen
 
 ## Version 11.5.0
 
--- a/src/languages/xml.js
+++ b/src/languages/xml.js
@@ -8,9 +8,15 @@ Audit: 2020
 /** @type LanguageFn */
 export default function(hljs) {
   const regex = hljs.regex;
-  // Element names can contain letters, digits, hyphens, underscores, and periods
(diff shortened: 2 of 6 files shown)


joshgoebel (approved):
Awesome.

PR #3534, opened 2022-05-11
3rd-party support for Oak

smithtim:
Oak is a proof checker focused on simplicity, readability, and ease of use: https://oakproof.org/

### Changes
Added Oak to supported languages (3rd party).

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,11 +6,13 @@ Grammars:
 - enh(swift) add support for `distributed` keyword [Marcus Ortiz][]
 - enh(xml) recognize Unicode letters instead of only ASCII letters in XML element and attribute names (#3256)[Martin Honnen][]
 - Added 3rd party Toit grammar to SUPPORTED_LANGUAGES [Serzhan Nasredin][]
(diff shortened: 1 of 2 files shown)


joshgoebel:
This looks good.  For your grammer I'd suggest:

- Use `scope`, not `className`
- Use `match` not begin
- Use our actual `keyword` support  for best forward compatibility without inventing your own custom modes

joshgoebel (approved):

smithtim:
Thanks for the feedback.  I have made the first two changes.  For the third, my keywords use regular expressions rather than strings, and I don't know how to do that with `keywords`.

joshgoebel:
A lot of your keyboards are simple and could use the regular engine, for the others you can use a mix, using the regex to find the combos but then still using the keyword engine to do the highlighting of each individual word.

```
      {
        match: regex.concat(/\b/, regex.either(...MULTI_WORDS), /\b/).
        keywords: { "built_in" : INDIVIDUAL_KEYWORD_ARRAY }
      },
```

smithtim:
I see, thanks for looking into it.  I would prefer to keep it as is if that's okay — that will be easier for me.

PR #3535, opened 2022-05-11
substr is deprecated, use substring

nuxodin:
Hi
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

### Changes

Replace substr width substring, using with one argument only should be compatible.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring

### Checklist


joshgoebel (approved):

joshgoebel:
Thanks!

PR #3541, opened 2022-05-22
fix(rust) recognize `include_bytes!` macro

Serial-ATA:
### Changes
Change builtin `include_bin` to `include_bytes`, as the former doesn't exist.

### Checklist
- [X] Added markup tests, or they don't apply here because it is a simple word correction.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 Grammars:
 
+- fix(rust) recognize `include_bytes!` macro (#3541) [Serial-ATA][]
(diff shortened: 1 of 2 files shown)


joshgoebel:
Was the former deprecated then?

Serial-ATA:
I've never known it to be `bin`, and `bytes` has been around since 1.0 so it's probably safe to assume that even if that was the name in the past, no one is using it anymore.

joshgoebel (approved):

joshgoebel:
@Serial-ATA  Thanks so much!

PR #3547, opened 2022-06-02
Fix: The END mode should be object rather than array

MegatronKing:
The END mode should be object rather than array.

### Checklist


joshgoebel (approved):

joshgoebel:
@MegatronKing Thanks so much!

PR #3572, opened 2022-07-01
Match "::" as comment

herrenP:
Batch files may have have "::" instead of "rem" to mark a line as a comment

### DIFF
--- a/src/languages/dos.js
+++ b/src/languages/dos.js
@@ -8,7 +8,7 @@ Website: https://en.wikipedia.org/wiki/Batch_file
 /** @type LanguageFn */
 export default function(hljs) {
   const COMMENT = hljs.COMMENT(
-    /^\s*@?rem\b/, /$/,
+    /^\s*(@?rem\b|::)/, /$/,
     { relevance: 10 }
   );
   const LABEL = {


joshgoebel:
This would need further work (and tests)... `::` is far too broad a match and sure to lead to false positives with many languages... we'd likely need to split this into multiple comment rules with a separate rule for  `::` as it's certainly not worthy of 10 relevance, perhaps 0.

- can these only start at the beginning of a line, or can they start anywhere?

herrenP:
> This would need further work (and tests)... `::` is far too broad a match and sure to lead to false positives with many languages... we'd likely need to split this into multiple comment rules with a separate rule for `::` as it's certainly not worthy of 10 relevance, perhaps 0.
> 
> * can these only start at the beginning of a line, or can they start anywhere?

I see. They can start anywhere on a line given that they are only preceeded by spaces.

joshgoebel:
Yep, need a separate rule then that's not pinned to `^`.

joshgoebel:
Ping.

joshgoebel:
Closing due to lack of response.  This can be reopened in the future when you have time to work on pushing it forward again.

PR #3575, opened 2022-07-06
enh(gml) Added additional keywords

Bluecoreg:
### Changes

Added GM 2022 compatibility functions.

### Checklist


joshgoebel:
Please split our your large commit here so that this can be properly reviewed - right now it's impossible to tell what has changed cause it's all jumbled together.

- First commit, re-order just the *existing* list(s).
- Second commit, add new keywords.

That way the second commit can be easily reviewed to see what has actually changed.  Thanks.

Issue #3536, opened 2022-05-12
(JSON) null/booleans are keywords but should be literals


kevinkucharczyk:
You can see this in the examples on the [public demo page](https://highlightjs.org/static/demo/).
![CleanShot 2022-05-12 at 14 19 11](https://user-images.githubusercontent.com/3127419/167991382-11244654-dbc4-46c2-a78a-fb41d86f9105.png)

`null`/`true`/`false` should be **literals**, but instead are keywords:
![CleanShot 2022-05-12 at 14 20 01](https://user-images.githubusercontent.com/3127419/167991505-bb5349ce-7485-40d6-b1c9-3d19a6b6c248.png)

This is based on the definitions provided in highlight.js's [scope reference](https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html), where those values are clearly marked as literals:
![CleanShot 2022-05-12 at 14 22 40](https://user-images.githubusercontent.com/3127419/167991746-a958d5a0-b1c3-4d01-ab2f-0de88a668044.png)

joshgoebel:
A PR would be accepted for this.

shikhar13012001:
Can I work on this issue? @joshgoebel

joshgoebel:
Sure should be trivial. Will require updating the markup tests.

shikhar13012001:
I will try my best
Could you tell me the target files for this ? @joshgoebel

joshgoebel:
Look for `javascript` files in the source tree...

joshgoebel:
@shikhar13012001 ping.

shikhar13012001:
@joshgoebel  sorry sir, but I am unable to find the given files :(

joshgoebel:
Er, I mean `json`, but all the files are in the same `src` folder.

joshgoebel:
Resolved via #3576

PR #3576, opened 2022-07-06
Fixes issue #3536

shikhar13012001:
(JSON) null/booleans are keywords but should be literals

[issue#3536](https://github.com/highlightjs/highlight.js/issues/3536)

### Changes

Removed literals object 
### Checklist


joshgoebel:
You said they should be literals, why did you removed them?

shikhar13012001:
@joshgoebel  I think I have done some mistake please let do a fix in this. Also I am having trouble as to what to do in particular? can you please help me a bit?

joshgoebel:
You need to change/add the `scope` to the existing `LITERAL` mode, not remove the mode entirely - and after that you'll need to update the test suite as well.  You may need to read the docs: https://highlightjs.readthedocs.io/en/latest/index.html  Perhaps read some of the other grammars (some are quite simple) to get an idea how they work.

I really can't really help further.  It would take longer to provide step-by-step instructions than to just fix it myself.    Good luck! 👍🏼

joshgoebel:
Please push what you have so I can review it with Github's tooling.

shikhar13012001:
> Please push what you have so I can review it with Github's tooling.

I have pushed the code

joshgoebel:
The comment above explains why this cannot be changed to use `keywords`.  I have fixed.

joshgoebel (approved):

joshgoebel:
Can you please add changelog entry and I'll still give you credit for the fix since you put in the time...

shikhar13012001:
> Can you please add changelog entry and I'll still give you credit for the fix since you put in the time...

Yes sure, thanks ! 🎉

PR #3581, opened 2022-07-13
Add COBOL to supported languages

KTSnowy:
GitHub repo: https://github.com/otterkit/highlightjs-cobol
NPM: https://www.npmjs.com/package/highlightjs-cobol

CDN links like unpkg.com are also working and I made a codepen to test syntax highlighting:
https://codepen.io/ktsnowy/pen/vYRXxov

### Changes
Added COBOL language and link for the highlightjs-cobol package in the [SUPPORTED_LANGUAGES.md](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md)

### Checklist

> Added markup tests

Not yet due to COBOL having a quite large keyword list, but I've been testing it on codepen.

### DIFF
--- a/SUPPORTED_LANGUAGES.md
+++ b/SUPPORTED_LANGUAGES.md
@@ -41,6 +41,7 @@ The table below shows the full list of languages (and corresponding classes/alia
 | C/AL                    | cal                    |         |
 | Cache Object Script     | cos, cls               |         |
 | CMake                   | cmake, cmake.in        |         |
+| COBOL                   | cobol, standard-cobol   | [highlightjs-cobol](https://github.com/otterkit/highlightjs-cobol) |
 | Coq                     | coq                    |         |
 | CSP                     | csp                    |         |
 | CSS                     | css                    |         |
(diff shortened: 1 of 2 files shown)


SUPPORTED_LANGUAGES.md
```diff
@@ -41,6 +41,7 @@ The table below shows the full list of languages (and corresponding classes/alia
 | C/AL                    | cal                    |         |
 | Cache Object Script     | cos, cls               |         |
 | CMake                   | cmake, cmake.in        |         |
+| COBOL                   | cobol, standard-cobol   | [highlightjs-cobol](https://github.com/otterkit/highlightjs-cobol) |
```
joshgoebel, line 44: Aliases like this only exist for user convenience... is there a reason a user would choose to [prefer and] write out the much longer `standard-cobol` when they could just write `cobol`?

joshgoebel (approved):
Help me understand the alias... if there is a good reason we can keep it, but it seems strange.  Otherwise this is approved.

SUPPORTED_LANGUAGES.md
KTSnowy: This could be useful if a COBOL vendor decides to make their own highlightjs package later on.

COBOL has many variants and dialects, and it could be useful in the future to differentiate between Standard COBOL, Enterprise COBOL, Visual COBOL, etc.

KTSnowy: A user might prefer to use the `standard-cobol` alias when using multiple dialects on the same page. This is due to incompatibilities between variants.

joshgoebel: Aliases as future proofing... interesting.  I'm not sure anyone else is doing that, but ok. :)

PR #3582, opened 2022-07-14
fix: typeerror with unicode-regex

AlbertAZ1992:
Babel uses regexpu to transpile unicode regex to ES5. 
But it has a bug if enclosed in square brackets: https://github.com/mathiasbynens/regexpu-core/issues/13,  then throw error:

`TypeError: symbol.charCodeAt is not a function` 

### Changes

Use regex with `new RegExp`

### Checklist


joshgoebel:
It seems this has since been fixed by https://github.com/thematters/matters-web/commit/ef837eb764c32e0819a63dd61f1a37db1a99228d... I would suggest upgrading.

AlbertAZ1992:
Ok, i find the bug was fixed the bug https://github.com/babel/babel/pull/13767 
My build tool did make a version lock for babel. 
Thanks a lot!

PR #3587, opened 2022-07-20
fix(js) do not flag `import()` as a function, rather a keyword 

nathnolt:
SEE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import

No issue made for this.

### Changes
Added the import operator function statement to the list of BUILT_IN_VARIABLES
and to the noneOf inside of the FUNCTION_CALL object, (although I'm not 100% sure if this is required)

### Checklist

Does such a small edit need an edit of changes.md?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,10 @@
 ## Version 11.7.0
 
-Grammars:
-- enh(js) added `import` operator 
+New Grammars:
 - added 3rd party LookML grammar to SUPPORTED_LANGUAGES [Josh Temple][]
-- added `import` operator to the JavaScript language definition [nathnolt][]
+
+Grammars:
+- fix(js) do not flag `import()` as a function, rather a keyword [nathnolt][]
 - fix(bash) recognize the `((` keyword [Nick Chambers][]
 
 [Josh Temple]: https://github.com/joshtemple
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## Version 11.6.1
+
+Grammars:
+- added `import` operator to the JavaScript language definition
+
 ## Version 11.6.0
 
 Supported Node.js versions:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,9 +4,11 @@ Grammars:
 - enh(js) added `import` operator 
 - added 3rd party LookML grammar to SUPPORTED_LANGUAGES [Josh Temple][]
 - added `import` operator to the JavaScript language definition [nathnolt][]
+- fix(bash) recognize the `((` keyword [Nick Chambers][]
 
 [Josh Temple]: https://github.com/joshtemple
 [nathnolt]: https://github.com/nathnolt
+[Nick Chambers]: https://github.com/uplime
 
 ## Version 11.6.0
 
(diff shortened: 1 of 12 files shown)


joshgoebel:
> this.buildDefinitions(data["palette"] ?? {}),

Yes.

I wonder though if this shouldn't be a built-in global or even a keyword?  variable seems wrong OTTOMH.

nathnolt:
I don't know man, I just changed it for myself and thought, I guess I could make a pull request.

nathnolt:
Also: how do I test out the highlighting, like have it spit out a demo page where I can insert code into, so I can test out if the changes I've made work or not.

joshgoebel:
> so I can test out if the changes I've made work or not.

See the Building and Testing documentation.  Build a browser release. Then fire up `./tools/developer.html`, that's the easiest way.

CHANGES.md
```diff
@@ -1,10 +1,12 @@
 ## Version 11.7.0
 
 Grammars:
-- Added 3rd party LookML grammar to SUPPORTED_LANGUAGES [Josh Temple][]
+- enh(js) added `import` operator
```
joshgoebel, line 4: Please add credit here and a link below (line ~7)...

src/languages/lib/ecmascript.js
```diff
@@ -141,6 +141,7 @@ export const BUILT_IN_VARIABLES = [
   "arguments",
   "this",
   "super",
+  "import",
```
joshgoebel, line 144: We already consider this a `keyword`, is there a reason you think it's a built-in variable?  Typically this is used for "objects"...

src/languages/javascript.js
```diff
@@ -360,7 +360,8 @@ export default function(hljs) {
       /\b/,
       noneOf([
         ...ECMAScript.BUILT_IN_GLOBALS,
-        "super"
+        "super",
+        "import"
```
joshgoebel, line 364: When the functional version of import is used `import()` should we not highlight it as a function?  Thoughts?

joshgoebel:
> I just changed it for myself and thought, I guess I could make a pull request.

What problem were you trying to solve with this patch?

CHANGES.md
nathnolt: Like this?

src/languages/lib/ecmascript.js
joshgoebel: This addition can be removed since we already highlight it as a keyword.

nathnolt: Okay, but wouldn't that make it not have that same color when used as a function, aka `import X from 'y.js'` being colored but `const x = import('y.js')` would just become the default function color?

joshgoebel: No, that is what `noneOf` resolves... you should be able to test both and see the results.

joshgoebel:
Need to close this up soon if you'd like it in 11.7... that may be the last release before 12.

src/languages/javascript.js
joshgoebel, line 364:
```suggestion
        "super"
```

joshgoebel (approved):

joshgoebel:
@nathnolt Thanks so much, I went ahead and pushed this all the way home.

nathnolt:
@joshgoebel thank you

PR #3589, opened 2022-07-22
Update sql.js

### LINKED ISSUE #3588: SQL - Create role, alter role, drop role - only the first word is highlighted instead of both
**Describe the issue**
Create role, alter role, drop role, drop table - only the first word is highlighted instead of both.

SQL

...

**Sample Code to Reproduce**

CREATE ROLE user1 WITH LOGIN PASSWORD 'password1';

**Expected behavior**
The both word - CREATE ROLE should be highlighted.

**Additional context**
"CREATE TABLE",  " ALTER TABLE" SQL-commands  are highlighted correctly. Commands with ROLE should be highlighted similarly.

EugeniaSimonova:
`Resolves #3588`

### Changes
"create role",    "alter role",   "drop role",   "drop table" have been added to the "const COMBOS" section.

### Checklist

### DIFF
--- a/src/languages/sql.js
+++ b/src/languages/sql.js
@@ -593,7 +593,11 @@ export default function(hljs) {
     "nulls first",
     "nulls last",
     "depth first",
-    "breadth first"
+    "breadth first",
+    "create role",
+    "alter role",
+    "drop role",
+    "drop table"
   ];
 
   const FUNCTIONS = RESERVED_FUNCTIONS;


joshgoebel:
Our SQL (`sql`) support is official **standard** SQL spec only.  Can you point to somewhere showing `ROLE` is a keyword in standard SQL and not just a DB vendor add-on?  For DB specific highlighting it's recommend that one use DB specific grammars, not the `sql` grammar.

https://stackoverflow.com/questions/21813895/where-can-i-find-sql-language-specification

joshgoebel:
Closing for inactivity.

PR #3604, opened 2022-08-29
fix(typescript) patterns like `<T = ` are not valid JSX

### LINKED ISSUE #3603: (TypeScript) Generic is not picked up properly
**Describe the issue**

typescript, no auto-detection

I don't know what this means.

**Sample Code to Reproduce**

I tried to reproduce the issue using your jsfiddle, but the typescript generic is not even displaying there ¯\\_(ツ)_/¯.  
~~https://jsfiddle.net/x3nzp2rf/~~  
updated: https://jsfiddle.net/sm706vyd/2/

I have this code on StackOverflow.

meta: https://meta.stackoverflow.com/questions/420116/typescript-syntax-highlight-with-generic-not-picked-up-properly

first observed issue here: https://stackoverflow.com/questions/72794594/typescript-interface-merging-to-make-an-external-libs-interface-property-more-s/73527814#73527814

    ```lang-ts
    import { useWeb3React as useWeb3React_ } from '@web3-react/core'
    
    export const useWeb3React: <T = any>(key?: string) => Modify<
      ReturnType<typeof useWeb3React_<T>>,
      { chainId: SupportedChainIds }
    > = useWeb3React_ as any
    
    declare global {
      type SupportedChainIds = 1 | 4
    }
    ```

Which produces incorrect syntax highlight

``` lang-ts
import { useWeb3React as useWeb3React_ } from '@web3-react/core'

export const useWeb3React: <T = any>(key?: string) => Modify<
  ReturnType<typeof useWeb3React_<T>>,
  { chainId: SupportedChainIds }
> = useWeb3React_ as any

declare global {
  type SupportedChainIds = 1 | 4
}
```

After a tweak I achieve correct syntax highlight

``` lang-ts
import { useWeb3React as useWeb3React_ } from '@web3-react/core'

export const useWeb3React: < T = any >(key?: string) => Modify<
  ReturnType<typeof useWeb3React_<T>>,
  { chainId: SupportedChainIds }
> = useWeb3React_ as any

declare global {
  type SupportedChainIds = 1 | 4
}
```

This is the tweak:
```
- export const useWeb3React: <T = any>(key?: string) => Modify<
+ export const useWeb3React: < T = any >(key?: string) => Modify<
```

**Expected behavior**

(without the spaces around `< T = any >` - correct is: `<T = any>`)

**Actual behavior**

**Additional context**

joshgoebel:
Resolves #3603.

### Changes

Adds `=` as another character that isn't not allowed to follow a potential HTML/JSX:

- `<tag x=` potentially valid JSX
- `<tag =` invalid JSX

### Checklist


src/languages/javascript.js
```diff
@@ -65,10 +66,18 @@ export default function(hljs) {
...
       // handled by simpleSelfClosing rule
 
-      // `<From extends string>`
-      // technically this could be HTML, but it smells like a type
       let m;
       const afterMatch = match.input.substring(afterMatchIndex);
+
+      // some more template typing stuff
+      //  <T = any>(key?: string) => Modify<
+      if ((m = afterMatch.match(/^\s+=/))) {
```
allejo, line 74:
```suggestion
      if ((m = afterMatch.match(/^\s*=/))) {
```

Personally, I've never seen anyone syntactically leave out spaces in generics but it is valid TS to not have spaces in between the `=` and also invalid JSX to not have an attribute on the LHS of an `=`.

The above change still keeps highlighting and unit tests pass.

image

allejo (commented):
Fixes the problem for me! Just a minor thought in a comment

image

allejo (approved):

PR #3608, opened 2022-09-04
(chore) fix type notation

pd4d10:
### Changes

fix type notation

### Checklist
- [ ] Added markup tests, or they don't apply here because... N/A
- [ ] Updated the changelog at `CHANGES.md` N/A


joshgoebel:
What is `/private`?   What editor are you using or why did you need to add this?   There are many places in the source we "magically" reference types - and I just trust the editor to figure out it - as that is the only time I use this annotations - for listing inside the editor.

pd4d10:
image

It's in the VSCode editor

I guess it is because this type is defined in `types/index.d.ts` with scope 'highlight.js/private', and should be explicitly imported

joshgoebel:
Thanks!

joshgoebel (approved):

PR #3617, opened 2022-09-12
enh(Ruby) misc improvements

joshgoebel:
Misc improvements to the Ruby grammar.

### Changes

- adds more psuedo-keywords
- highlights class names more oten

### Checklist


src/languages/ruby.js
```diff
@@ -18,10 +18,24 @@ export default function(hljs) {
   )
   ;
   const CLASS_NAME_WITH_NAMESPACE_RE = regex.concat(CLASS_NAME_RE, /(::\w+)*/)
+  // very popular ruby built-ins that one might even assume
+  // are actual keywords (despite that not being the case)
+  const PSEUDO_KWS = [
```
SleeplessByte, line 23: `module_function`

and

`class << self` (that exact version)

also come to mind.

joshgoebel: `class << self`

That should be handled since `class` is a keyword and self is a `language.variable`, unless you're trying to suggest it needs special treatment somehow?

allejo (approved):
Tested locally and the improvements seem fine to me

PR #3618, opened 2022-09-13
Fix typo

iwalfy:
### Changes

Fixed typo (Microtik instead of MikroTik)
### Checklist


joshgoebel:
Thanks.

Issue #3619, opened 2022-09-13
Support Prism.js grammar parsing but use the Highligh.js HTML/theme pipeline


joshgoebel:
Related #2212.

I'm not sure this belongs in core yet, and I don't necessarily want a hard or soft Prism dependency.  I suppose perhaps you could also pass `Prism` into the function itself though so it's late binding?

### Caveats

- Using custom parsers is still private API since the `Emitter` API is public yet.  Hence the `__emitTokens` name of the property.
- use the `emit_tokens` branch

### What needs to be done

- Figure out where this goes, how to package it, etc.
- Write a comprehensive `prismTypeToScope` function

### Example Usage

To replace our JavaScript support with Prism's:

```js
const prismJS = fromPrism(Prism.languages.javascript)
hljs.registerLanguage("javascript", prismJS);
```

### Code

```js
function prismTypeToScope(name) {
  return name;
}

function prismTokensToEmitter(tokens, emitter) {
  tokens.forEach(token => {
    if (typeof(token) === "string") {
      emitter.addText(token)
    } else if (token.type) {
      let scope = prismTypeToScope(token.type)
      if (typeof(token.content) === "string") {
        emitter.addKeyword(token.content, scope)
      } else { // array of tokens
        emitter.openNode(scope)
        prismTokensToEmitter(token.content, emitter)
        emitter.closeNode()
      }
    }
  })
}

function prismParserWrapper(code, emitter) {
  const tokens = Prism.tokenize(code, this._prism)
  prismTokensToEmitter(tokens, emitter)
}

function fromPrism(prismGrammar) {
  return function(hljs) {
    let self = {}

    Object.assign(self, {
      _prism: prismGrammar,
      _emitTokens: prismParserWrapper.bind(self)
    })
    return self
  }
}
```

joshgoebel:
@RunDevelopment If I wanted to package this as a small stand-alone ESM module, would I just `import Prism` and let people's bundlers/client-side figure it out or would it be simpler to just pass in the Prism object that the user is responsible to import on their own?  I'm spoiled by the fact that Highlight.js has zero real dependencies.

Or perhaps `Prism.languages.javascript` has some backreference to `Prism` itself?  That'd be useful here.

Thoughts?

RunDevelopment:
> Actually I don't even know if Prism is ESM client-side yet

We are going to be. ESM is very much planned as the only module system. We are likely also going to have monolithic files for compatibility, but that's about it.

[...]

PR #3620, opened 2022-09-13
private __emitTokens API

joshgoebel:
Needed for #3619.  Plan is to release this with 10.7.

### Changes

- Allows grammars to provide their own parsers that output to the private HLJS `Emitter` API
- A grammar may supply a `__emitTokens(code, emitter)` function as part of the language data it returns when called

### Checklist
- [x] Add unit tests for `__emitTokens`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 Improvements:
 
 - added a function to default export to generate a fresh highlighter instance to be used by extensions [WisamMechano][]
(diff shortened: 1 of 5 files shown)


joshgoebel:
@wooorm Relevant to you because once this goes public the Emitter API will become part of our public API since custom grammars will need to trust this API to be stable.  No rush, but it'll happen eventually I think.  I'd love to see what people could do with this.

PR #3622, opened 2022-09-13
Fix arithmetic detection in Bash highlighter

uplime:
### Changes
This adds support for Bash's arithmetic keyword `((`. The following example currently gets treated as a heredoc, due to the `<<` token:

```bash
(( 19 << 42 ))
printf 'Line one\n'
printf 'Line two\n'
```

Screenshot of the (slightly modified) demo:

image

Throwing in what would be the heredoc boundary, it highlights properly again:

```bash
(( 19 << 42 ))
42
printf 'Line one\n'
printf 'Line two\n'
```

Screenshot of the (slightly modified) demo:

image

With this change in place, it will correctly highlight the original example:

image

### Checklist

### DIFF
--- a/src/lib/compile_keywords.js
+++ b/src/lib/compile_keywords.js
@@ -22,7 +22,7 @@ const DEFAULT_KEYWORD_SCOPE = "keyword";
  * @param {boolean} caseInsensitive
  */
 export function compileKeywords(rawKeywords, caseInsensitive, scopeName = DEFAULT_KEYWORD_SCOPE) {
-  /** @type KeywordDict */
+  /** @type {import("highlight.js/private").KeywordDict} */
   const compiledKeywords = Object.create(null);
 
   // input can be a string of keywords, an array of keywords, or a object with


joshgoebel:
So what's the effective difference between `(( ))` and `$(( ))`?

uplime:
@joshgoebel `$(( ))` is standard and expands to the result of the express.

```
echo "$(( 3 + 4 ))"
```

effectively becomes

```
echo "7"
```

`((` on the other hand is a keyword (a command with special parsing rules) that has an exit status like a command. There's a number of usecases, but I commonly use it for tests:

```
if (( x < 7 )); then
  # ...
fi
```

joshgoebel:
**AWesome.**

I'd say we need to add a small test to bash's `arithmetic` markups tests for this.  And will of course need a changelog entry, etc.

uplime:
Good call! I'm assuming those changes need to be attached to the PR as well? I just made a couple more commits with some tests and an entry in the change log.

test/markup/bash/arithmetic.txt
```diff
@@ -6,3 +6,10 @@ echo $(( 16#$yumi / 1003 ))
 B=20
 yumi=deadbeef
 echo $(( $B#$yumi / 1003 ))
+
+(( 19 << 42 ))
```
joshgoebel, line 10:
```suggestion

# test that << within an expression doesn't false positively as a heredoc...
(( 19 << 42 ))
```

joshgoebel: THis context is helpful later.  And the expect will need to be updated.  Otherwise looks great!

joshgoebel:
Still needs the matching expect test since we added the comment.

uplime:
Oh sorry, I didn't realize the comment was a part of the actual test. I'll fix that right now!

joshgoebel (approved):

joshgoebel:
@uplime Thanks so much!

PR #3636, opened 2022-10-14
fix(nix) Support escaped dollar sign in strings

### LINKED ISSUE #3635: (nix) Escaped dollar sign not highlighted correctly
**Describe the issue**

Nix allows escaping antiquotes like this:

```nix
{
  a = ''
    Antiquote: ${x}
    Escaped dollar sign: ''${y}
  '';
}
```

nix

highlight

**Sample Code to Reproduce**

```nix
{
  a = ''
    Antiquote: ${x}
    Escaped dollar sign: ''${y}
  '';
}
```

**Expected behavior**

See the github highlighting

**Additional context**

h7x4:
Added a rule that escapes `''$` in strings.

Resolves #3635

### Checklist

- [X] Added markup tests

### DIFF
--- a/src/languages/nix.js
+++ b/src/languages/nix.js
@@ -45,6 +45,10 @@ export default function(hljs) {
     end: /\}/,
     keywords: KEYWORDS
   };
+  const ESCAPED_DOLLAR = {
+    className: 'char.escape',
+    begin: /''\$\{/,
+  };
   const ATTRS = {
     begin: /[a-zA-Z0-9-_]+(\s*=)/,
     returnBegin: true,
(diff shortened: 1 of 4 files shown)


src/languages/nix.js
```diff
@@ -45,6 +45,10 @@ export default function(hljs) {
     end: /\}/,
     keywords: KEYWORDS
   };
+  const ESCAPED_DOLLAR = {
+    className: 'char.escape',
+    begin: /''\$\{/,
```
joshgoebel, line 50: Is `''` escaping the `$` or `${`... your issue seems to indicate it's escaping `$`, so I'm a little confused at seeing the { here.

h7x4: I think I've confused myself here a little. The only reason you would ever want to escape the dollar, is to escape an antiquote. But it is indeed the dollar itself that should be escaped. Updating now.

joshgoebel (approved):

PR #3643, opened 2022-10-22
Improve Gherkin support

### LINKED ISSUE #3638: (gherkin) highlight wrongly for unclosed '<'
**Describe the issue**

When there is a '<' in someline, but without a '>', the line and following lines are highlighted wrongly

gherkin

highlight

**Sample Code to Reproduce**

```gherkin
Feature: foo
  Scenario: test scenario <foo>
  Given a < b
  Then a is less than b
```

**Expected behavior**

**Additional context**

image

### LINKED ISSUE #3639: (gherkin) Rule keyword is not highlighted
**Describe the issue**

Rule keyword is not highlighted

gherkin

highlight

**Sample Code to Reproduce**

```gherkin
Feature: foo
  Rule: rule 1
    Scenario: scenario 1
```

**Expected behavior**

Rule keyword is highlighted.

**Additional context**

image

Hirse:
Resolves #3638
Resolves #3639

image

### Changes
- Move keywords to mode to support keywords with spaces and those ending with a colon
- Update list of keywords (see below)
- Change `*` from `symbol` to `keyword` since that is how it's described in the reference
- Tighten variable definition to disallow whitespace
- Only recognize keywords, comments, tags, and docstrings at the beginning of a line
- Support docstrings starting with triple backticks
- Remove (non-docstring) string mode 

#### Reference
* https://cucumber.io/docs/gherkin/reference
  image
* https://cucumber.io/docs/gherkin/languages/#gherkin-dialect-en-content
  image

### Checklist
- [X] Added markup tests


pkufranky:
thanks very much

src/languages/gherkin.js
```diff
@@ -9,10 +9,24 @@ export default function(hljs) {
   return {
     name: 'Gherkin',
     aliases: [ 'feature' ],
-    keywords: 'Feature Background Ability Business\ Need Scenario Scenarios Scenario\ Outline Scenario\ Template Examples Given And Then But When',
+    keywords: {
+      // Custom pattern to support keywords with spaces and those ending with a colon
+      $pattern: /[A-Z][a-z]+(?: [A-Z][a-z]+)?:?/,
```
joshgoebel, line 14: This is going to break and have weird edge cases like `And When`... which would match BOTH words, and then not highlight any keywords because all we do is scan the text with this regex looking for POTENTIAL keywords.

Multi-word keywords are better handled with modes I think.

joshgoebel:
> Technically all features (except variables) are only valid at the beginning of a line.

Can't we do this with multi-match and `^` (start line)?  Just use the first part of the match to grab the whitespace...

joshgoebel:
> and those ending with a colon

Is the colon part of the keyword though?

Hirse:
Updated with a new version that removes the custom keyword pattern and fixes false positives.

I couldn't figure out a good way to match for "beginning of line, skipping the whitespace" as `/^[ \t]*/` triggers the "zero-width regex" error.

src/languages/gherkin.js
```diff
@@ -5,66 +5,146 @@
...
             end: /```/
           }
+        ],
+      },
+      {
+        begin: /^\|.*\|$/,
+        scope: 'string'
+      },
+      {
+        begin: /^[ \t]+/,
```
joshgoebel, line 80: This repetition is frustrating.  Could a single top-level rule to eat tabs at the start of lines accomplish this same thing?  If not why not add it to each multi-match rule as an optional first component:

```
        begin: [
          LINE_BEGINS_WS,
          /(Feature|Rule|Examples?|Scenario(?:s| Outline| Template)?|Background)/,
          /:/
        ],
        beginScope: {
          2: 'keyword',
          3: 'punctuation',
        },
```

Hirse: Yes, eating whitespace of the beginning of the line with an empty mode seems to work. Is that what you had in mind? 😄

joshgoebel: The multi-match method I show above is what I had in mind, yes.  That's the best way I know to handle this kind of thing (without putting the whitespace inside the highlight zone).

pkufranky:
Any progress for this pr?

joshgoebel:
@Hirse is this ready to go? sorry for long wait.  coud you update changes.md?

joshgoebel (approved):

PR #3648, opened 2022-10-23
Update number pattern to catch variables containing numbers

### LINKED ISSUE #3645: (javascript, typescript) Incorrect handling of digits after `$` in identifiers
**Describe the issue**

`$` isn't treated as a character valid in identifiers, even though it is allowed in identifiers, and is treated like alphabetic letters and `_`. This causes e.g. `1` in `$1` to be highlighted like a numeric literal.

`javascript`, `typescript`

I assume `highlightElement()` uses `highlightAuto()`.

**Sample Code to Reproduce**

The test code is:
```js
const $1 = "buck-one";
const $this = `${$1}`;
const bu$ = "bus";
const $$ = "buck-squared"
console.log(`${new Date()}, ${90 ^ 34}`);
```

https://jsfiddle.net/Keyacom/0ezu3jbd/9/

**Expected behavior**

GitHub syntax highlighting correctly doesn't highlight any digits that follow `$`. Highlight.js must do the same thing. The refrence code block is in the section above.

**Additional context**

lachieh:
Currently, numbers are matched based on the word boundary (`\b`) but that means that a valid variable identifier like `$0` will have the `0` highlighted as a number.

Resolves #3645

### Changes
* Updated the regex patterns for numbers to lookbehind for a `$` symbol.

### Checklist


joshgoebel:
We can't use look-behinds, they aren't yet supported universal by all browsers (ie: Safari).

Typically for things like this you'd instead try to match the offending item... ie, add an earlier rule to match/capture `$2323` patterns, and just ignore them (vs highlight)... this will prevent the number rule from firing and should also be another easy fix.

taufik-nurrohman:
Adding `[$][a-z\d]*` as variable pattern before the number patterns should fix the issue. As the variable name that contains number has been captured before.

joshgoebel:
> ` [$][a-z\d]*`

We don't need a-z though since all we're trying to solve is the false-positive with `$[number]`, yes?

taufik-nurrohman:
OK.

lachieh:
> **Awesome contribution!  Great tests.**
> 
> Sadly,  we can't use look-behinds, they aren't yet supported universal by all browsers (ie: Safari).

I thought about this, but followed sit since there are two other tests that also use lol behind. The hidden solution before will work though so I'll look at that.

lachieh:
Nevermind, did a cursory search and found this, but didn't notice it was a comment.

https://github.com/highlightjs/highlight.js/blob/abcb53deafd07aaa4ef787c8a9285acb25912051/src/languages/r.js#L173-L176

I will update with the suggested pattern!

lachieh:
Ok, this has been updated. I did have to leave the `a-z` in because it was failing on the test below. Adding the `a-z` in makes all tests pass.

```
 1) typescript
       should markup identifiers_that_include_keywords:
```
```diff
- actual
+ expected

-<span class="hljs-keyword">const</span> result = $<span class="hljs-title function_">class</span>();
+<span class="hljs-keyword">const</span> result = $class();
```

If you don't mind, would you consider adding the https://github.com/highlightjs/highlight.js/labels/hacktoberfest-accepted label?

joshgoebel:
> I did have to leave the a-z in because it was failing on the test below.

Isn't that because you are using * when it should be +?

`/[$][a-z\d]+/`

We don't care about just plain $ unless it's followed by variable like stuff... All you're really trying to skip is $\d so that should be a sufficient regex IF you require at least a single digit after the $, which `+` will do for you.

lachieh:
Yes, fair point. Thanks for clarifying. Updated pattern and merged upstream.

src/languages/javascript.js
```diff
@@ -105,6 +105,8 @@ export default function(hljs) {
   const NUMBER = {
     className: 'number',
     variants: [
+      // Skip numbers when they are part of a variable name
+      { begin: /\$\d+/, className: "", skip: true },
```
joshgoebel, line 109:
```suggestion
      // Skip numbers when they are part of a variable name
      { match: /\$\d+/, skip: true },
```

joshgoebel:
Oh, you made it a variant..... :-). Interesting.  I'll fix.

lachieh:
Yeah, took reference from here: https://github.com/highlightjs/highlight.js/blob/e0f75779513fd6fd112cc08e326785a3e04d0635/src/languages/python.js#L332-L338

joshgoebel:
> Yeah, took reference from here:

Yeah, but that example sort of IS still an instance of params, just "empty params"... where as $123 isn't really an instance of a numeric, it's a variable. :-). I'm being pedantic now.

But what you're doing is one easy way to get that rule in all the same spots you'd want the NUMBER rule (which is convenient)... so that's an argument for it... so I might just revert and commit your last revision after all. :-)

lachieh:
Gotcha, I understand. I've moved it to before the both uses of `NUMBER` and added a few more tests for the same pattern in template strings. I think this is fine since it is only used twice, but if you'd prefer it inside the `NUMBER` I can make that change (or you can).

Thanks for the time and input!

joshgoebel (approved):
Perfect!

PR #3649, opened 2022-10-25
Add Java 17 permits keyword to

### LINKED ISSUE #3646: (java) Sealed Classes keywords
**Describe the issue**
[JEP 409: Sealed Classes](https://openjdk.org/jeps/409) introduces ´permits´ as new contextual Keywords to the **Java** Language [JLS 3.9. Keywords](https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-ContextualKeyword). The contextual keyword ´permits´ is not rendered as keyword within a class definition.

Java, Language is set strict.

I use hightlight via [ascidoctor revealJS](https://docs.asciidoctor.org/reveal.js-converter/latest/converter/syntax/syntax-highlighting/)

...

**Sample Code to Reproduce**
Can be reproduced with the developer.html and seen in the Tests _test/markup/java/titles.txt_.
```java
sealed interface Command permits LoginCommand {
    void run();
}
```

**Expected behavior**
´permits´ should be hightlighted as keyword.
```html
<span class="hljs-keyword">sealed</span> <span class="hljs-keyword">interface</span> <span class="hljs-title class_">Command</span> <span class="hljs-keyword">permits</span> LoginCommand {
    <span class="hljs-keyword">void</span> <span class="hljs-title function_">run</span><span class="hljs-params">()</span>;
}
```
**Additional context**
´permits´ is a contextual keyword in java, this means it is only a keyword wihtin a defined scope (here class definition).
Is it possible to implement this with hightlight.js without to much efford?

MBoegers:
Java 17 introduced sealed classes/interfaces with [EP 409: Sealed Classes](https://openjdk.org/jeps/409). The new keywords sealed and non-sealed were present but permits was missing.

Resolves #3646

### Changes
Permits is now a general keyword and the present tests are fixed acordingly.

### Checklist
- [ ] Added markup tests, see test/markup/java/titles.expect.txt

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ Grammars:
 - fix(bash) recognize the `((` keyword [Nick Chambers][]
 - fix(nix) support escaped dollar signs in strings [h7x4][]
 - enh(cmake) support bracket comments [Hirse][]
+- enh(java) add permits keyword to java [MBoegers][]
 
 [Josh Goebel]: https://github.com/joshgoebel
 [Josh Temple]: https://github.com/joshtemple
@@ -21,6 +22,7 @@ Grammars:
 [Nick Chambers]: https://github.com/uplime
 [h7x4]: https://github.com/h7x4
 [Hirse]: https://github.com/Hirse
+[MBoegers]: https://github.com/MBoegers
 
 
 ## Version 11.6.0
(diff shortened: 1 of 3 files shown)


ffontecha080 (approved):
good

joshgoebel (approved):
Thanks so much!

PR #3650, opened 2022-10-25
Add Java 14 switch expression keyword yield

### LINKED ISSUE #3647: (java) switch expression keywords
**Describe the issue**
[JEP 361: Switch Expressions](https://openjdk.org/jeps/361) intruduces new keywords ´yield´ to *Java* in context of ´switch´ but ´yield´ is not rendered as such.

Java, language is set fixed.

I use hightlight via [asciidoc reveal.js](https://docs.asciidoctor.org/reveal.js-converter/latest/converter/syntax/syntax-highlighting/)

**Sample Code to Reproduce**
```java
switch(a) {
 case Object -> yield 1;
}
```

**Expected behavior**
```html
<span class="hljs-keyword">switch</span>(a) {
 <span class="hljs-keyword">case</span> Object -&gt; <span class="hljs-keyword">yield</span> <span class="hljs-number">1</span>;
}
```

**Additional context**
´yield´ is a contextual keyword, this means it can be a normal identifier in Java. It is a keyword within the switch expression, see [JLS 3.9. Keywords](https://docs.oracle.com/javase/specs/jls/se14/html/jls-3.html#jls-3.9)

MBoegers:
Java 14 introduced with [JEP 361: Switch Expressions](https://openjdk.org/jeps/361) a new keyword yield.

Resolves #3647

### Changes
yield is now a keyword, a new test for switches are added

### Checklist
- [x] Added markup tests, see test/markup/java/switch.expect.txt

### DIFF
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -72,7 +72,8 @@ export default function(hljs) {
     'requires',
     'exports',
     'do',
-    'sealed'
+    'sealed',
+    'yield'
   ];
 
   const BUILT_INS = [
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -73,7 +73,7 @@ export default function(hljs) {
     'exports',
     'do',
     'sealed',
-    'yield'
+    'yield',
     'permits'
   ];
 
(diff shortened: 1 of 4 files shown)


src/languages/java.js
```diff
@@ -73,6 +73,7 @@ export default function(hljs) {
     'exports',
     'do',
     'sealed',
+    'yield'
     'permits'
```
joshgoebel, line 77:
```suggestion
    'yield',
    'permits'
```

joshgoebel (approved):

PR #3653, opened 2022-11-02
Make our language metadata parsing stricter

allejo:
Reported by @razetime on Discord, when a language's metadata contains an empty "Requires", it will cause this cryptic error to surface:

```
$ node tools/build.js -t node
Starting build.
Finished build.
Writing styles.
/home/razetime/Documents/Code/gen/highlight.js/tools/build_node.js:184
  const common = languages.filter(l => l.categories.includes("common"));
                                         ^

TypeError: Cannot read properties of undefined (reading 'categories')
    at /home/razetime/Documents/Code/gen/highlight.js/tools/build_node.js:184:42
    at Array.filter (<anonymous>)
    at Object.buildNode [as build] (/home/razetime/Documents/Code/gen/highlight.js/tools/build_node.js:184:28)
    at async doTarget (/home/razetime/Documents/Code/gen/highlight.js/tools/build.js:90:3)
```

The cause of this problem is that our regex for parsing the metadata end up with incorrect data at times. For example, when given `Requires: ` it will end up parsing the body as `""` which later ends up evaluating to "undefined" when being mapped as a language.

### Changes

I changed the regex to be a bit more specific when looking at metadata fields and expecting there to be a value there and not just default to an empty string.

### Checklist

- [ ] ~~Added markup tests~~ No functional or grammar changes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,10 @@ Grammars:
 - enh(java) add yield keyword to java [MBoegers][]
 - enh(java) add permits keyword to java [MBoegers][]
 
+Developer Tools:
+
+- (fix) make metadata parsing less fragile and prone to error [Vladimir Jimenez]
+
 [Josh Goebel]: https://github.com/joshgoebel
 [Josh Temple]: https://github.com/joshtemple
 [nathnolt]: https://github.com/nathnolt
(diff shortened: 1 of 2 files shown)


joshgoebel:
I think this is best fixed at the place in the code where we get "" confused... not the regex... the regex purposely check for the existence of comments and those meta-tags are not required to start at the beginning of lines either.

joshgoebel:
IE, the fix probably belongs around here: https://github.com/highlightjs/highlight.js/pull/3653/files#diff-b68005823c82aa5e25478f89f7e7c3cd0ac64068a814a262d08fdd8d96542d35R54

joshgoebel:
Resolved by 36915434

PR #3657, opened 2022-11-11
More Scala 3 soft keywords

guersam:
### Changes
- add Scala 3 `derives` soft keyword
- add missing Scala 3 soft modifiers (`infix`, `opaque`, `open`)

### Checklist

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -137,6 +137,25 @@ export default function(hljs) {
     }
   };
 
+  const IMPORT_EXPORT = {
+    beginKeywords: 'import export',
+    end: /[;\n]/,
+    excludeEnd: true,
+    contains: [
+      { beginKeywords: 'as' },
+      {
+        begin: '{',
+        end: '}',
+        excludeBegin: true,
+        excludeEnd: true,
+        contains: [
+          TYPE,
+          { beginKeywords: 'as' },
+        ]
+      }
+    ],
+  };
+
   // TODO: use negative look-behind in future
   //       /(?<!\.)\binline(?=\s)/
   const INLINE_MODES = [
@@ -184,6 +203,7 @@ export default function(hljs) {
       hljs.C_NUMBER_MODE,
       EXTENSION,
       END,
+      IMPORT_EXPORT,
       ...INLINE_MODES,
       USING_PARAM_CLAUSE,
       ANNOTATION,
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -72,7 +72,7 @@ export default function(hljs) {
 
   const CLASS = {
     className: 'class',
-    beginKeywords: 'class object trait type',
+    beginKeywords: 'class object trait type enum',
     end: /[:={\[\n;]/,
     excludeEnd: true,
     contains: [
@@ -80,6 +80,11 @@ export default function(hljs) {
       hljs.C_BLOCK_COMMENT_MODE,
       {
         beginKeywords: 'extends with',
+        relevance: 10
+      },
+      {
+        beginKeywords: 'derives',
+        contains: [ TYPE ],
         relevance: 10
       },
       {
@@ -99,7 +104,7 @@ export default function(hljs) {
         relevance: 0,
         contains: [ TYPE ]
       },
-      NAME
+      NAME,
     ]
   };
 
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -156,11 +156,23 @@ export default function(hljs) {
     beginScope: { 2: "keyword", }
   };
 
+  const SOFT_MODIFIER = {
+    begin: [
+      /\b(?<!\.)(infix|inline|opaque|open|transparent)\b/,
+      /\s+/,
+      /(?=object|def|val|var|type|given|class|trait|object|enum|case class|case object|infix|inline|opaque|open|transparent)/,
+    ],
+    beginScope: {
+      1: 'keyword',
+    },
+    contains: [ 'self' ],
+  };
+
   return {
     name: 'Scala',
     keywords: {
       literal: 'true false null',
-      keyword: 'type yield lazy override def with val var sealed abstract private trait object if then forSome for while do throw finally protected extends import final return else break new catch super class case package default try this match continue throws implicit export enum given transparent'
+      keyword: 'type yield lazy override def with val var sealed abstract private trait object if then forSome for while do throw finally protected extends import final return else break new catch super class case package default try this match continue throws implicit export enum given'
     },
     contains: [
       hljs.C_LINE_COMMENT_MODE,
@@ -174,7 +186,8 @@ export default function(hljs) {
       END,
       ...INLINE_MODES,
       USING_PARAM_CLAUSE,
-      ANNOTATION
+      ANNOTATION,
+      SOFT_MODIFIER,
     ]
   };
 }
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -84,7 +84,6 @@ export default function(hljs) {
       },
       {
         beginKeywords: 'derives',
-        contains: [ TYPE ],
       },
       {
         begin: /\[/,
@@ -139,14 +138,11 @@ export default function(hljs) {
   const IMPORT_EXPORT = {
     beginKeywords: 'import export',
     end: /[;\n]/,
-    excludeEnd: true,
     contains: [
       { beginKeywords: 'as' },
       {
         begin: '{',
         end: '}',
-        excludeBegin: true,
-        excludeEnd: true,
         contains: [
           TYPE,
           { beginKeywords: 'as' },
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -174,17 +174,18 @@ export default function(hljs) {
     beginScope: { 2: "keyword", }
   };
 
-  const SOFT_MODIFIER = {
-    begin: [
-      /(?=^|;)\s*/,
-      /\b(infix|inline|opaque|open|transparent)\b/,
+  const SOFT_MODIFIERS = [ 'infix', 'inline', 'opaque', 'open', 'transparent' ];
+  const SOFT_MODIFIER_RE = regex.either(...SOFT_MODIFIERS);
+
+  const SOFT_MODIFIER_MODE = {
+    begin: regex.concat(
+      /(^|;)\s*/,
+      SOFT_MODIFIER_RE,
       /\s+/,
-      /(?=object|def|val|var|type|given|class|trait|object|enum|case class|case object|infix|inline|opaque|open|transparent)/,
-    ],
-    beginScope: {
-      2: 'keyword',
-    },
-    contains: [ 'self' ],
+      regex.anyNumberOfTimes(regex.concat(SOFT_MODIFIER_RE, /\s+/)),
+      regex.lookahead(/(object|def|val|var|type|given|class|trait|object|enum|case class|case object)/),
+    ),
+    keywords: SOFT_MODIFIERS,
   };
 
   return {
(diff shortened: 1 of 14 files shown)


guersam:
Wanted to add `as` as well, but couldn't find out how to to write `import`/`export` mode properly.

reference: https://docs.scala-lang.org/scala3/reference/soft-modifier.html

guersam:
Add `as` as well.

src/languages/scala.js
```diff
@@ -137,6 +137,25 @@ export default function(hljs) {
...
+    beginKeywords: 'import export',
+    end: /[;\n]/,
+    excludeEnd: true,
+    contains: [
+      { beginKeywords: 'as' },
+      {
+        begin: '{',
+        end: '}',
+        excludeBegin: true,
+        excludeEnd: true,
```
joshgoebel, line 150: Since there is no active scope here you shouldn't need these... the `{}` aren't part of any named highlighted region... unless I'm missing something?

src/languages/scala.js
```diff
@@ -137,6 +137,25 @@ export default function(hljs) {
     }
   };
 
+  const IMPORT_EXPORT = {
+    beginKeywords: 'import export',
+    end: /[;\n]/,
+    excludeEnd: true,
```
joshgoebel, line 143: Ditto for this.

src/languages/scala.js
```diff
@@ -82,6 +82,11 @@ export default function(hljs) {
         beginKeywords: 'extends with',
         relevance: 10
       },
+      {
+        beginKeywords: 'derives',
+        contains: [ TYPE ],
+        relevance: 10
```
joshgoebel, line 88: Was this needed to resolve a auto-detect conflict?  Otherwise we're really not handing out 10 anymore (relevance is due to change a lot soon with v12 anyways).  So if not, I'd remove it and give it the default relevance.

src/languages/scala.js
```diff
@@ -151,11 +175,23 @@ export default function(hljs) {
     beginScope: { 2: "keyword", }
   };
 
+  const SOFT_MODIFIER = {
+    begin: [
+      /\b(?<!\.)(infix|inline|opaque|open|transparent)\b/,
```
joshgoebel, line 180: We can't use look-behind because it's not supported in Safari, sorry.

joshgoebel: Is the "followed by" part of the rule on 182 not enough to prevent false positives?

src/languages/scala.js
```diff
@@ -151,11 +175,23 @@ export default function(hljs) {
...
+      1: 'keyword',
+    },
+    contains: [ 'self' ],
+  };
+
   return {
     name: 'Scala',
     keywords: {
       literal: 'true false null',
-      keyword: 'type yield lazy override def with val var sealed abstract private trait object if then forSome for while do throw finally protected extends import final return else break new catch super class case package default try this match continue throws implicit export enum given transparent'
```
joshgoebel, line 158: No longer a keyword?

src/languages/scala.js
guersam: Removed.

src/languages/scala.js
guersam: A false positive case without look-behind is as below:

```
a.infix == b.infix

opaque type MyInt = Int
```

expected:
```html
a.infix == b.infix

<span class="hljs-keyword">opaque</span> <span class="hljs-class"><span class="hljs-keyword">type</span> <span class="hljs-title">MyInt</span> </span>= <span class="hljs-type">Int</span>
```

actual:
```html
a.infix == <span class="hljs-keyword">b.infix</span>

<span class="hljs-keyword">opaque</span> <span class="hljs-class"><span class="hljs-keyword">type</span> <span class="hljs-title">MyInt</span> </span>= <span class="hljs-type">Int</span>
```


but in the next case, both `transparent` and `trait` should be keywords.
```
transparent
trait Foo
```

Would you give me some advice to handle this case?

src/languages/scala.js
guersam: [soft keywords](https://docs.scala-lang.org/scala3/reference/soft-modifier.html) like `transparent` are treated as a normal identifier except some special cases.

src/languages/scala.js
guersam: Removed look-behind.

guersam: Looking for another corner case...

joshgoebel: For `.blah` I'd add a "property" rule that would catch those and highlight them as properties so they don't trigger false positives as keywords.

guersam: Thanks. `a.infix` can also be written as `a infix` in a certain condition, so I'v just revised the matching rule in the lower level.

guersam:
Merge conflict resolved.

src/languages/scala.js
joshgoebel: You resolved this, but it's unchanged.  Ditto on the above as well. Unresolving.

src/languages/scala.js
```diff
@@ -151,11 +174,25 @@ export default function(hljs) {
     beginScope: { 2: "keyword", }
   };
 
+  const SOFT_MODIFIERS = [ 'infix', 'inline', 'opaque', 'open', 'transparent' ];
+  const SOFT_MODIFIER_RE = regex.either(...SOFT_MODIFIERS);
+
+  const SOFT_MODIFIER_MODE = {
+    begin: regex.concat(
+      /(^|;)\s*/,
```
joshgoebel, line 182: Is this truly necessary with all the context provided by the look-ahead? Do you have an example of a false positive if this was removed? 

Generally we don't try and match on "prior" content as we may want to match that in the future such as the `;` should long-term be flagged as punctuation, etc... and it's harder to do that with rules like this.

src/languages/scala.js
```diff
@@ -82,6 +82,10 @@ export default function(hljs) {
         beginKeywords: 'extends with',
         relevance: 10
       },
+      {
+        beginKeywords: 'derives',
+        contains: [ TYPE ],
```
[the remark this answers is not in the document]
guersam:
```
case class Foo derives Eq,
  JsonCodec
```

In an example like the above, I want to things:
1. (necessary) highlight soft keyword `derives` 
2. (optional) highlight `Eq`, `JsonCodec` as `TYPE` 

But as you noted 2 doesn't works with the `contains` parameter. I just removed it for now and it still looks good to me.

src/languages/scala.js
guersam: Sure, false positive cases without `/(^|;)\s*/`:

### case 1
`notinline given Int = 9`

```
      -not<span class="hljs-keyword">inline</span> <span class="hljs-keyword">given</span> <span class="hljs-type">Int</span> = <span class="hljs-number">9</span>
      +notinline <span class="hljs-keyword">given</span> <span class="hljs-type">Int</span> = <span class="hljs-number">9</span>
```

### case 2

`a.infix == b.infix`
```
      -a.infix == b.<span class="hljs-keyword">infix</span>
      +a.infix == b.infix
```

I understand what you're worried about and will be happy to get some advice to improve this.

guersam: BTW, thanks for the thorough review so far!

src/languages/scala.js
guersam: Removed this too.

src/languages/scala.js
joshgoebel: Typically case one you can use a `\b` at the beginning of the expression.

Case two what we usually do is match `.thing` separately with another rule (these are `property` accesses) and that avoids them being mismatched as keywords.

guersam: Thanks for the comment.

For case 1, if I replace it with `\b` the following false positives arise:
1. `val transparent = 43` 
```
      -<span class="hljs-keyword">val</span> <span class="hljs-keyword">transparent</span> = <span class="hljs-number">43</span>
      +<span class="hljs-keyword">val</span> transparent = <span class="hljs-number">43</span>
```

2. `c opaque d`
```
      -c <span class="hljs-keyword">opaque</span> d
      +c opaque d
```

For case 2, what I'm afraid of is that in Scala `a.infix` can also be written as `a infix` [in a certain condition](https://www.scala-lang.org/api/3.2.1/scala/language$.html#postfixOps-0). 

---

If prior contents matching is such a big concern, it seems a reasonable option to me to remove it and embrace some false positives because it's still better than the status quo. What do you think?

joshgoebel: What is `notinline`, it looks like a keyword?

joshgoebel: > embrace some false positives because it's still better than the status quo.

I've often thought if we can't get the highlighting right (say of soft modifiers) that we shouldn't highlight them at all, but there are good arguments for both directions I'd say.

> val transparent = 43

Not sure I understand this example and how you're saying it breaks if you use \b?

guersam: > What is notinline, it looks like a keyword?

`notinline` is just an arbitrary string that contains `inline`. Probably `\b` will work for this case.

> Not sure I understand this example and how you're saying it breaks if you use \b?

`transparent` is a soft keyword, that means it has special meaning only if it's placed in a certain syntactic position. In this `val transparent = 43` example, `transparent` is just an identifier that is bound to `43`, therefore it should not be highlighted.

vt58474956:
Added as as well.

bishabosha:
hey, what needs fixing in this PR? anything we can help with (Scala Center)

src/languages/scala.js
joshgoebel: >  In this `val transparent = 43` example, transparent is just an identifier that is bound to 43, therefore it should not be highlighted.

And since the look-ahead doesn't match there, it would not be highlighted.  So it seems lookahead already handles that case, unless I'm missing something?

src/languages/scala.js
joshgoebel: > Thanks. a.infix can also be written as a infix in a certain condition

Well if you didn't want to be inconsistent (between those two cases) you could match on `.thing` but NOT highlight it (just to avoid the false positive).  This is a very common strategy used in many grammars to handle easy false positives.

src/languages/scala.js
```diff
@@ -132,6 +135,22 @@ export default function(hljs) {
...
+      {
+        begin: '{',
+        end: '}',
+        contains: [
+          TYPE,
+          { beginKeywords: 'as' },
+        ]
+      }
+    ],
+  };
```
joshgoebel, line 152: From looking at your example woud'n't this work just as well:

```suggestion
  const IMPORT_EXPORT = {
    beginKeywords: 'import export',
    end: /[;\n]/,
    contains: [
      { beginKeywords: 'as' },
      TYPE
    ],
  };
```

We try to be a simple pattern matcher (rather than a full parser) whenever we can - as it makes the grammars simpler, smaller, and easier to maintain.

guersam:
@bishabosha Thanks for your partitication! I've invited you as a collaborator of [my fork](https://github.com/guersam/highlight.js/).

src/languages/scala.js
joshgoebel: Ping.

bishabosha:
> @bishabosha Thanks for your partitication! I've invited you as a collaborator of [my fork](https://github.com/guersam/highlight.js/).

@guersam Hey sorry I think I missed your comment (and the invitation to your fork expired)

guersam:
@bishabosha Invited you again :)

bishabosha:
> @bishabosha Invited you again :)

thanks, I've accepted it!

joshgoebel:
Closing for lack of activity.  This can be reopened anytime you have time to come back to it.  I've left some comments above that have never been addressed.

PR #3665, opened 2022-11-17
add subLanguage support to markdown

stutrek:
In addition to the subLanguage attribute, there are also a significant number of markdown tokens without a relevance value. Should that also be changed?

### Changes

This adds subLanguage configurations to \`\`\` and \~\~\~ in the markdown language config, allowing markdown files with a significant portion of content in code blocks to be recognized as markdown.

### Checklist
- [x] Added markup tests, or they don't apply here because... tests for markup are already covered

### DIFF
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -21,17 +21,19 @@ export default function(hljs) {
   const CODE = {
     className: 'code',
     variants: [
-      // TODO: fix to allow these to work with sublanguage also
-      { begin: '(`{3,})[^`](.|\\n)*?\\1`*[ ]*' },
-      { begin: '(~{3,})[^~](.|\\n)*?\\1~*[ ]*' },
-      // needed to allow markdown as a sublanguage to work
       {
-        begin: '```',
-        end: '```+[ ]*$'
+        begin: '```[a-z0-9]*',
+        end: '```+[ ]*$',
+        excludeBegin: true,
+        excludeEnd: true,
+        subLanguage: []
       },
       {
-        begin: '~~~',
-        end: '~~~+[ ]*$'
+        begin: '~~~[a-z0-9]*',
+        end: '~~~+[ ]*$',
+        excludeBegin: true,
+        excludeEnd: true,
+        subLanguage: []
       },
       { begin: '`.+?`' },
       {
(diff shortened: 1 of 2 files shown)


joshgoebel:
> there are also a significant number of markdown tokens without a relevance value. Should that also be changed?

No, hand-tuned relevance is going to be deprecated in the future as it's of questionable value and impossible to get right for a large set of languages without a ton of research and infrastructure we do not have.

>  allowing markdown files with a significant portion of content in code blocks to be recognized as markdown.

Did something prevent them from being recognized as markdown previously?

I'm not sure what to do here as `subLanguage: []` has serious performance implications and this change would make highlighting markdown files potentially MUCH slower.  Despite the TODO I also think this starts to get more into rendering of the markdown than highlighting it (though I know some disagree).  I might potentially come around on that if it weren't for the performance implications as well.  

_I'm wondering if this wouldn't make a very short plugin that people could simply opt-into if they wanted this behavior._

@allejo Any thoughts?

stutrek:
[The test file I added](https://github.com/highlightjs/highlight.js/pull/3665/files#diff-bef9733607d84b7ad0f9fcc2cd7dd44c9b645a3204ba32a373386a7954500348) is a good example of how easily markdown can be interpreted as a different language. There aren't near enough markdown tokens to override the huge number of tokens in a short code block. Adding `subLanguage` resolves this because it adds that relevance to markdown's score.

It could be a plugin, it could also be added as an alternate language file. For our purposes, I am using this file (with a few additions for relevance) in place of the default markdown implementation.

joshgoebel:
I was pretty sure I'd typed up another response, but I don't see it here.  I don't think we want to merge this as is... the "autodetect all" is just too much... if we wanted to discuss a way to **dynamically** only support named languages, such as:

````
```ruby
def add(a,b) 
end
```
````

Such that any language (or alias) loaded was usable... but that gets into issues with the load order, etc - that would need to be discussed/worked out.  I'd be open to that discussion though.  

If someone wants the exact behavior of this PR as you said it's easy to just add/define a custom variant of markdown that does that.

Closing this, but happen to continue the discussion if you're interested in the dynamic route idea.

PR #3679, opened 2022-12-17
PHP NOWDOC Support

doiftrue:
Support [NOWDOC syntax](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc) for php.js.

NOTE: NOWDOC don't parse php variables inside the string.

HEREDOC improvements: quote syntax support `<<<"END"`.

### Checklist
- [x] Added markup tests.


src/languages/php.js
```diff
@@ -56,14 +56,19 @@ export default function(hljs) {
     end: /[ \t]*(\w+)\b/,
     contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
   });
+  const NOWDOC = hljs.END_SAME_AS_BEGIN({
+    begin: /<<<[ \t']*(\w+)'?\n/,
+    end: /[ \t]*(\w+)\b/,
+  });
```
joshgoebel, line 62: Should this not also support normal escapes and variable substitution, same as HEREDOC?

src/languages/php.js
```diff
@@ -56,14 +56,19 @@ export default function(hljs) {
     end: /[ \t]*(\w+)\b/,
     contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
   });
+  const NOWDOC = hljs.END_SAME_AS_BEGIN({
+    begin: /<<<[ \t']*(\w+)'?\n/,
```
joshgoebel, line 60: The `'` here doesn't look correct.  I think you want something like:

```
(\w+|'\w+')
```

What you have now would match `''''''''''WORD` which is not correct.

src/languages/php.js
```diff
@@ -52,12 +52,12 @@ export default function(hljs) {
...
   });
   const HEREDOC = hljs.END_SAME_AS_BEGIN({
-    begin: /<<<[ \t]*(\w+)\n/,
+    begin: /<<<[ \t]*"?(\w+)"?\n/,
     end: /[ \t]*(\w+)\b/,
     contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
   });
   const NOWDOC = hljs.END_SAME_AS_BEGIN({
-    begin: /<<<[ \t']*(\w+)'?\n/,
+    begin: /<<<[ \t]*'(\w+)'\n/,
```
joshgoebel, line 60: It must always be quoted?

src/languages/php.js
```diff
@@ -52,12 +52,12 @@ export default function(hljs) {
     contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
   });
   const HEREDOC = hljs.END_SAME_AS_BEGIN({
-    begin: /<<<[ \t]*(\w+)\n/,
+    begin: /<<<[ \t]*"?(\w+)"?\n/,
```
joshgoebel, line 55: This allows:

`"BOO`

`BOO"`

You need to use an OR (like I showed you before) to make sure they are balanced.

src/languages/php.js
doiftrue: Good catch! Fixed! Made nowdoc more strict. 

Also, quote (") wrapper support added for HEREDOC >>> "END".

PHP allows write `<<<END` OR `<<<"END"` for heredoc. Now `<<<"END"` supported and tests improved.

src/languages/php.js
doiftrue: Fixed!

src/languages/php.js
doiftrue: `"` itself must not be included in match. For now I don't know how to do that. `(\w+|"\w+")` this is not work.

I don't think it's a problem. It's just highlighter but not PHP code parser.

doiftrue: `<<<[ \t]*(?:(\w+)|"(\w+)")` don't works as well because hljs.END_SAME_AS_BEGIN works with first group only...

doiftrue: Any ideas how to solve this?

joshgoebel: Make the outside group capturing and the two inside ones non-capturing - though technically you don't need the inside groups at all the `|` and outside group should be fine by themselves.

doiftrue: But then an outside group takes over the `"` sign and we don't need it - it breaks everything.

image

https://regex101.com/r/I0FG3H/1

doiftrue: And even more we have one more problem here 
image

doiftrue: The only solution I found - is rewrite `hljs.END_SAME_AS_BEGIN` and then `(?:(\w+)|"(\w+)")` works good:

```js
  const HEREDOC = {
    begin: /<<<[ \t]*(?:(\w+)|"(\w+)")\n/,
    end: /[ \t]*(\w+)\b/,
    contains: hljs.QUOTE_STRING_MODE.contains.concat(SUBST),
    'on:begin': (m, resp) => { resp.data._beginMatch = m[1] || m[2]; },
    'on:end': (m, resp) => { if (resp.data._beginMatch !== m[1]) resp.ignoreMatch(); },
  };
```

joshgoebel: Yeah, I was just about to mention it's just JS under the covers and you could tweak it like that... :-) Looks good enough to me.

doiftrue: But I think it's better to leave `begin: /<<<[ \t]*"?(\w+)"?\n/,`. 

That's okay about `"BOO` & `BOO"`. Because we, in other hand, don't check the full workability of the HEREdoc construct, which implies that the indentation of the text will be exactly the same as the last END.

joshgoebel: When it's easy to avoid false positives with a strong regex, we go that route... which is this case means matching quoted OR unquoted, not half quoted.

src/languages/php.js
joshgoebel: But the end of now doc doesn't have to be quoted?

joshgoebel (approved):
Looks good! Thanks!

PR #3686, opened 2023-01-07
Update sqf.js to match latest changes in Arma 3 v2.11

Leopard20:
### Changes

- Added new commands
- Removed invalid commands
- Removed `#` as an illegal character (legal as of Arma 3 v1.82)

This is a follow up to #3193. It was quite outdated so I opened a new one. It also applies the requested changes to make it not fail the detection asserts.

P.S: I think GitHub's diff is broken because it shows the entire file has changed. It looks correct in VSCode though.


src/languages/sqf.js
```diff
@@ -1,2529 +1,2630 @@
...
+      keyword:
+        'define undef ifdef ifndef else endif include'
+    },
+    contains: [
+      {
+        begin: /\\\n/,
+        relevance: 0
+      },
+      hljs.inherit(STRINGS, {
+        className: 'string'
```
joshgoebel, line 2587: Why is this necessary?

src/languages/sqf.js
```diff
@@ -1,2529 +1,2630 @@
...
+    className: 'meta',
+    begin: /#\s*[a-z]+\b/,
+    end: /$/,
+    keywords: {
+      keyword:
+        'define undef ifdef ifndef else endif include'
+    },
+    contains: [
+      {
+        begin: /\\\n/,
```
joshgoebel, line 2583: For line continuation, yes?

src/languages/sqf.js
```diff
@@ -1,2529 +1,2630 @@
...
+    'worldToScreen'
+    ];
+
+  // list of keywords from:
+  // https://community.bistudio.com/wiki/PreProcessor_Commands
+  const PREPROCESSOR = {
+    className: 'meta',
+    begin: /#\s*[a-z]+\b/,
+    end: /$/,
+    keywords: {
```
joshgoebel, line 2577: Technically this could collapse to just `keywords` a nested hash isn't necessary if one isn't specified `keyword` is the default scope used.

src/languages/sqf.js
```diff
@@ -1,2529 +1,2630 @@
...
+    'worldToModel',
+    'worldToModelVisual',
+    'worldToScreen'
+    ];
+
+  // list of keywords from:
+  // https://community.bistudio.com/wiki/PreProcessor_Commands
+  const PREPROCESSOR = {
+    className: 'meta',
+    begin: /#\s*[a-z]+\b/,
```
joshgoebel, line 2575: Does this need to start at the beginning of a line?

src/languages/sqf.js
Leopard20: > For line continuation, yes?

Dunno. It was there from previous contributions. But I guess yeah

src/languages/sqf.js
Leopard20: > Does this need to start at the beginning of a line?

Whitespaces are allowed too. Will update this too.

src/languages/sqf.js
```diff
@@ -1,2529 +1,2630 @@
...
+      hljs.C_BLOCK_COMMENT_MODE,
+      hljs.NUMBER_MODE,
+      VARIABLE,
+      FUNCTION,
+      STRINGS,
+      PREPROCESSOR
+    ],
+      illegal: [
+      //$ is only valid when used with Hex numbers (e.g. $FF) or at the beginning of localized strings (e.g. "$STR_HELLO")
+      /\$[^Sa-fA-F0-9]/,
```
joshgoebel, line 2619: What is the `S` here?

src/languages/sqf.js
```diff
@@ -1,2529 +1,2630 @@
...
+  // In SQF, a local variable starts with _
+  const VARIABLE = {
+    className: 'variable',
+    begin: /\b_+[a-zA-Z]\w*/
+  };
+
+  // In SQF, a function should fit myTag_fnc_myFunction pattern
+  // https://community.bistudio.com/wiki/Functions_Library_(Arma_3)#Adding_a_Function
+  const FUNCTION = {
+    className: 'title',
```
joshgoebel, line 51:
```suggestion
    className: 'title.function',

```

src/languages/sqf.js
Leopard20: > What is the `S` here?

I wasn't sure if it would flag a string like `"$STR_bla"` as illegal. (which are localized strings)

src/languages/sqf.js
```diff
@@ -1,2529 +1,2630 @@
...
+    'for', 'forEach', 'from',
+    'if',
+    'local',
+    'private',
+    'switch', 'step',
+    'then', 'throw', 'to', 'try',
+    'waitUntil', 'while', 'with'
+  ];
+
+  const LITERAL = [
```
joshgoebel, line 93: This is a LOT of literals.  **Consider** `variable.constant` and `variable.language` perhaps... `pi` (if it's math) seems like a reference to a literal (a constant), not a literal itself.

I won't force any changes here since I dunno the language, but just making you aware we have more scopes to differentiate these sorts of things.

Leopard20: > This is a LOT of literals. **Consider** `variable.constant` and `variable.language` perhaps... `pi` (if it's math) seems like a reference to a literal (a constant), not a literal itself.
> 
> I won't force any changes here since I dunno the language, but just making you aware we have more scopes to differentiate these sorts of things.

Technically they're all commands (so they should be part of built-in array)
Again this is from previous contributions so I didn't want to change them.

Could you point me to the list of scopes you guys have?

src/languages/sqf.js
joshgoebel: I'm not sure I yet follow how one knows it's a localized string - they alway have a $STR prefix?  If it's actually quoted though then it's already handled by STRINGS rule and doesn't need to be guarded against here since its inside the `string` scope already, unless I'm missing something?

Leopard20: > I'm not sure I yet follow how one knows it's a localized string - they alway have a $STR prefix? If it's actually quoted though then it's already handled by STRINGS rule and doesn't need to be guarded against here since its inside the `string` scope already, unless I'm missing something?

Yeah if the string rule prevents it I should remove it. I didn't actually test this to be sure.

Dahlgren:
> P.S: I think GitHub's diff is broken because it shows the entire file has changed. It looks correct in VSCode though.
Your new file has carriage return on most/all lines, resave with just newline like the previous version

src/languages/sqf.js
```diff
@@ -1079,24 +1165,32 @@ export default function(hljs) {
...
     'getUnitLoadout',
     'getUnitTrait',
+    'getUnloadInCombat',
+    'getUserInfo',
     'getUserMFDText',
     'getUserMFDValue',
     'getVariable',
     'getVehicleCargo',
-    'getVehicleTIPars',
+    'getVehicleTiPars',
```
Dahlgren, line 1189:
```suggestion
    'getVehicleTIPars',
```
https://community.bistudio.com/wiki/getVehicleTIPars

src/languages/sqf.js
```diff
@@ -1402,14 +1507,15 @@ export default function(hljs) {
...
+    'lockedCameraTo',
     'lockedCargo',
     'lockedDriver',
     'lockedInventory',
     'lockedTurret',
     'lockIdentity',
     'lockInventory',
     'lockTurret',
-    'lockWP',
+    'lockWp',
```
Dahlgren, line 1518:
```suggestion
    'lockWP',
```
https://community.bistudio.com/wiki/lockWP

src/languages/sqf.js
```diff
@@ -1854,28 +1972,25 @@ export default function(hljs) {
...
     'setBehaviour',
     'setBehaviourStrong',
     'setBleedingRemaining',
     'setBrakesRTD',
-    'setCameraEffect',
     'setCameraInterest',
     'setCamShakeDefParams',
     'setCamShakeParams',
-    'setCamUseTI',
+    'setCamUseTi',
```
Dahlgren, line 1985:
```suggestion
    'setCamUseTI',
```
https://community.bistudio.com/wiki/setCamUseTI

src/languages/sqf.js
```diff
@@ -1884,7 +1999,7 @@ export default function(hljs) {
     'setCurrentTask',
     'setCurrentWaypoint',
     'setCustomAimCoef',
-    'setCustomMissionData',
+    'SetCustomMissionData',
```
Dahlgren, line 2002:
```suggestion
    'setCustomMissionData',
```
https://community.bistudio.com/wiki/setCustomMissionData

src/languages/sqf.js
```diff
@@ -1907,10 +2022,9 @@ export default function(hljs) {
     'setEditorObjectScope',
     'setEffectCondition',
     'setEffectiveCommander',
-    'setEngineRPMRTD',
     'setEngineRpmRTD',
     'setFace',
-    'setFaceAnimation',
+    'setFaceanimation',
```
Dahlgren, line 2027:
```suggestion
    'setFaceAnimation',
```
https://community.bistudio.com/wiki/setFaceAnimation

src/languages/sqf.js
```diff
@@ -1931,7 +2045,7 @@ export default function(hljs) {
     'setGroupIconParams',
     'setGroupIconsSelectable',
     'setGroupIconsVisible',
-    'setGroupId',
+    'setGroupid',
```
Dahlgren, line 2048:
```suggestion
    'setGroupId',
```
https://community.bistudio.com/wiki/setGroupId

src/languages/sqf.js
```diff
@@ -2059,12 +2179,12 @@ export default function(hljs) {
     'setTaskResult',
     'setTaskState',
     'setTerrainGrid',
+    'setTerrainHeight',
     'setText',
-    'setThrottleRTD',
     'setTimeMultiplier',
+    'setTiParameter',
```
Dahlgren, line 2185:
```suggestion
    'setTIParameter',
```
https://community.bistudio.com/wiki/setTIParameter

src/languages/sqf.js
```diff
@@ -2099,14 +2222,13 @@ export default function(hljs) {
...
     'setVehicleId',
-    'setVehicleInit',
     'setVehicleLock',
     'setVehiclePosition',
     'setVehicleRadar',
     'setVehicleReceiveRemoteTargets',
     'setVehicleReportOwnPosition',
     'setVehicleReportRemoteTargets',
-    'setVehicleTIPars',
+    'setVehicleTiPars',
```
Dahlgren, line 2231:
```suggestion
    'setVehicleTIPars',
```
https://community.bistudio.com/wiki/setVehicleTIPars

src/languages/sqf.js
```diff
@@ -2147,7 +2269,7 @@ export default function(hljs) {
     'showCommandingMenu',
     'showCompass',
     'showCuratorCompass',
-    'showGPS',
+    'showGps',
```
Dahlgren, line 2272:
```suggestion
    'showGPS',
```
https://community.bistudio.com/wiki/showGPS

src/languages/sqf.js
```diff
@@ -2156,12 +2278,13 @@ export default function(hljs) {
     'shownCompass',
     'shownCuratorCompass',
     'showNewEditorObject',
-    'shownGPS',
+    'shownGps',
```
Dahlgren, line 2281:
```suggestion
    'shownGPS',
```
https://community.bistudio.com/wiki/shownGPS

src/languages/sqf.js
```diff
@@ -96,6 +135,8 @@ export default function(hljs) {
     'sideEmpty',
     'sideLogic',
     'sideUnknown',
+    'sideEnemy',
+    'sideFriendly',
```
Dahlgren, line 139: Not in alphabetic order

Leopard20:
I'm not gonna update it beyond this point. The point was to update the SQF commands list. Further changes to other stuff can be made in future contributions.

Regarding the spelling changes, like I said they come from the game itself, not the wiki. I prefer to keep it as the game says, which is more correct.

src/languages/sqf.js
joshgoebel: https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html

> Technically they're all commands (so they should be part of built-in array).

Let's not do that. :-). For first party grammars we take a "wider perspective" view of such things.  Literal concepts that are non-reducible like true, false, null, nil, undefined... these are considered literals.  We prefer consistency in this across our grammars - rather than adhering to any particular grammars technicalities. 

Since `pi` is simply a label for 3.1459 etc it would be better thought of as a `variable.constant`... and the fact that what provides that is a `function pi()` doesn't really change that "wider perspective" categorization.

src/languages/sqf.js
```diff
@@ -1,15 +1,45 @@
...
+_KEYWORDS = ['if'];                                                //Array of all KEYWORDS
+_LITERALS = ['west'];                                              //Array of all LITERALS
+_allCommands = createHashMap;
+{
+	_type = _x select [0,1];
+	if (_type != "t") then {
+		_command_lowercase = ((_x select [2]) splitString " ")#(((["n", "u", "b"] find _type) - 1) max 0);
+		_command_uppercase = supportInfo ("i:" + _command_lowercase) # 0 # 2;
+		_allCommands set [_command_lowercase, _command_uppercase];
+	};
```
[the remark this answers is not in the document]
joshgoebel: Our source uses spaces, not tabs if that's what you're asking.

src/languages/sqf.js
joshgoebel: Is the language perhaps case insensitive?  That would explain a lot.

joshgoebel:
Indeed, `case_insensitive: true,` so the exact casing doesn't matter - using what the language provides for easier maintenance on our end makes the most sense.

joshgoebel:
Whatever you did with line endings made this almost impossible to rebase cleanly.  I finally just gave up and went with `git diff x..y | patch` to squash the whole history - then made another pass at converting any remaining tabs to spaces.  I tried to make sure you were still noted as the committer of record.

joshgoebel (approved):

PR #3695, opened 2023-01-21
enh(http) Allow arbitrary protocol versions

rijenkii:
### Changes
According to [RFC 2616 § 3.1](https://www.rfc-editor.org/rfc/rfc2616#section-3.1), `HTTP-Version` is defined as
```abnf
HTTP-Version = "HTTP" "/" 1*DIGIT "." 1*DIGIT
```
So, at least one digit in major and minor version.
And, according to [RFC 9114 § 4.3.1-4.3.2](https://www.rfc-editor.org/rfc/rfc9114.html#name-request-pseudo-header-field), HTTP/3 requests and responses implicitly have a protocol version of "3.0".

However, `curl` for instance in addition to `HTTP/1.0` and `HTTP/1.1` outputs versions `HTTP/2` and `HTTP/3`, without the minor version.

This pull request makes all current and future versions of HTTP protocol compatible, with and without the minor version.

This also allows to write `HTTP/0`, e.g. if I really do not care about the protocol version in my code block.

### Checklist
- [x] Added markup tests

### PS

Why is there a 11.9.0 in the `CHANGES.md`? Latest released version is 11.7.0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,8 +11,10 @@ New Grammars:
 Grammars:
 
 - enh(javascript) add sessionStorage to list of built-in variables [Jeroen van Vianen][]
+- enh(http) Add support for HTTP/3 [Rijenkii][]
 
 [Jeroen van Vianen]: https://github.com/morinel
+[Rijenkii]: https://github.com/rijenkii
 
 ## Version 11.7.0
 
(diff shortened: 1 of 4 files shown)


joshgoebel:
_Thanks for the contribution!_

I'd say we care a lot more about practicality (highlighting actual REAL HTTP requests) than fully RFC compliance (at least as long as we still have auto-detect)... so lets scope this back to known HTTP versions for the present.  I think that would mean simply adding a `3` to the regex would be sufficient?

rijenkii:
Yeah, I get the point.
I have force-pushed a commit which only adds support for HTTP/3 and a test for it.

joshgoebel (approved):
Awesome!

PR #3715, opened 2023-02-13
Add Motoko and Candid to supported languages

rvanasa:
This PR includes links to [highlightjs-motoko](https://github.com/rvanasa/highlightjs-motoko), which provides syntax highlighting for [Motoko](https://github.com/dfinity/motoko#readme) as well as [Candid](https://github.com/dfinity/candid#readme) (an [IDL](https://en.wikipedia.org/wiki/Interface_description_language) commonly used with Motoko).


rvanasa:
Updated; thanks!

joshgoebel:
The correct startup function is now `hljs.highlightAll();`... your examples are using the old API.

joshgoebel:
Also could you please update the changelog to announce the two new supported languages?  After that I think we're good to go.

joshgoebel (approved):

PR #3740, opened 2023-03-21
enh(protobuf) adds `proto` alias for Protobuf

dimitropoulos:
`proto` is the extension for protobuf files, and is used in [other](https://github.com/shikijs/shiki/blob/44b9b0c458744267d824a8283eac4d6393f65013/packages/shiki/src/languages.ts#L104) syntax highlighters as the language id.

### Changes

Adds `proto` alias for Protobuf

### Checklist
- [x] Added markup tests, or they don't apply here because... this is just an update to an alias

### DIFF
--- a/src/languages/protobuf.js
+++ b/src/languages/protobuf.js
@@ -47,7 +47,7 @@ export default function(hljs) {
 
   return {
     name: 'Protocol Buffers',
-    aliases: ['proto', 'protobuf'],
+    aliases: ['proto'],
     keywords: {
       keyword: KEYWORDS,
       type: TYPES,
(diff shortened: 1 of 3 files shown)


src/languages/protobuf.js
```diff
@@ -47,6 +47,7 @@ export default function(hljs) {
 
   return {
     name: 'Protocol Buffers',
+    aliases: ['proto', 'protobuf'],
```
joshgoebel, line 50:
```suggestion
    aliases: ['proto'],
```

joshgoebel (approved):

PR #3754, opened 2023-04-14
(Fix) Haskell added 'where' inside data block. Resolves #3753

### LINKED ISSUE #3753: (Haskell) GADT syntax highlighted incorrectly
**Describe the issue**

The `where` keyword in the Haskell GADT syntax is not highlighted correctly

`haskell`

`highlightElement`

**Sample Code to Reproduce**

https://jsfiddle.net/8j5zb4ny/1/

**Expected behavior**

The `where` keyword at the end of the first line should get a different highlighting to show that it is not a type variable

Somarfagell:
### Changes
The where keyword in the Haskell GADT syntax wasn't highlighted correctly, adding 'where' keyword inside the data block correct the issue.

Added 'where' keyword inside data block inside *haskell.js*
### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/haskell.js
+++ b/src/languages/haskell.js
@@ -129,7 +129,7 @@ export default function(hljs) {
         className: 'class',
         begin: '\\b(data|(new)?type)\\b',
         end: '$',
-        keywords: 'data family type newtype deriving',
+        keywords: 'data family type newtype deriving where',
         contains: [
           PRAGMA,
           CONSTRUCTOR,


joshgoebel:
Nice, but is this also correct for `type` and `newtype` blocks also?

joshgoebel:
Nice, but is this also correct for `type` and `newtype` blocks also?

Somarfagell:
> Nice, but is this also correct for `type` and `newtype` blocks also?

I haven't checked, but I suppose it would work, Do you want me to check it? 👀

joshgoebel:
It should work [as in be highlighted], but is that correct and valid Haskell?  I don't know when `where` is and isn't supposed to be a keyword.

joshgoebel:
Ping.

joshgoebel:
Closing older PRs without activity.  When you have time to come back to this we can reopen.

PR #3756, opened 2023-04-16
Add "until" and "select" as Bash keywords

verhovsky:
See https://github.com/tree-sitter/tree-sitter-bash/issues/133


joshgoebel:
Add changelog entry please?  Both parts (link and footnote).

joshgoebel (approved):

PR #3766, opened 2023-04-25
fix language detection regex - Issue 3762

### LINKED ISSUE #3762: language aliases c++ and h++ are not treated as C++ but C

When I specify a piece of C++ code using the language name "c++" or "h++", it treats it as a piece of C code. 

Here's a piece of html:
```html
<pre><code class="language-c++">using namespace std;</code></pre>
```
After executing `hljs.highlightAll();`, it becomes:
```html
<pre><code class="language-c++ hljs language-c">using namespace <span class="hljs-built_in">std</span>;</code></pre>
```
Same thing happens with h++

**Expected behavior**

It should become:
```html
<code class="language-c++ hljs language-cpp"><span class="hljs-keyword">using</span> <span class="hljs-keyword">namespace</span> std;</code>
```

**Additional context**

santi4o:
Updated language detection regex to match "+" characters

Resolves #3762 

### Changes
Hello! I updated the regex to `/\blang(?:uage)?-(\w[\w+-]*)/i` where `(\w[\w+-]*)` will match a word character followed by zero or more occurrences of word characters, hyphens, or plus symbols. I also removed the last boundary matcher (`\b`) because it would not match languages ending with `+` or `-` like `c++`

### Checklist
- [ ] Markup tests (I think they don't apply here)

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -70,7 +70,7 @@ const HLJS = function(hljs) {
     ignoreUnescapedHTML: false,
     throwUnescapedHTML: false,
     noHighlightRe: /^(no-?highlight)$/i,
-    languageDetectRe: /\blang(?:uage)?-([\w-]+)\b/i,
+    languageDetectRe: /\blang(?:uage)?-(\w[\w+-]*)/i,
     classPrefix: 'hljs-',
     cssSelector: 'pre code',
     languages: null,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,7 @@ Parser:
 
 - add removePlugin api [faga295][]
 - fix typo in language name of JavaScript [Cyrus Kao][]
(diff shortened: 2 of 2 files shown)


src/highlight.js
```diff
@@ -70,7 +70,7 @@ const HLJS = function(hljs) {
     ignoreUnescapedHTML: false,
     throwUnescapedHTML: false,
     noHighlightRe: /^(no-?highlight)$/i,
-    languageDetectRe: /\blang(?:uage)?-([\w-]+)\b/i,
+    languageDetectRe: /\blang(?:uage)?-(\w[\w+-]*)/i,
```
joshgoebel, line 73: I think we need to replace the `\b` with something to avoid partial matches of a longer string... perhaps something like `(?:$|(?=\s))` ie end of string or a trailing space... since we execute this against the entire css class name string...

joshgoebel:
Ping.

Issue #3769, opened 2023-04-28
Make name required for Language


alexeyinkin:
Currently 3 languages miss `name`:
- dsconfig
- php
- python-repl

If the project is converted to TypeScript, this could be verified statically.

joshgoebel:
Been thinking about this one.  Not sure I like forcing this on all 3rd party developers with v12.... What is the big win here again vs just the auto-registration?  

And no we have no interest currently in TypeScript proper.

joshgoebel:
Closing because of lack of activity.

PR #3770, opened 2023-04-28
Add names to dsconfig, php, python-repl (#3769)

alexeyinkin:
- Addresses #3769 

### Changes
Adds names to the three languages missing it.
Removes the implicit name setting for a language if it misses one.
Updates the `LanguageDetails` type to reflect that there is always a name.

### Checklist

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -853,8 +853,6 @@ const HLJS = function(hljs) {
       // entire highlighter
       lang = PLAINTEXT_LANGUAGE;
     }
-    // give it a temporary name if it doesn't have one in the meta-data
-    if (!lang.name) lang.name = languageName;
     languages[languageName] = lang;
     lang.rawDefinition = languageDefinition.bind(null, hljs);
 
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -477,6 +477,7 @@ export default function(hljs) {
   };
 
   return {
+    name: 'PHP',
     case_insensitive: false,
     keywords: KEYWORDS,
     contains: [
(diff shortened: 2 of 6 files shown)


src/highlight.js
```diff
@@ -853,8 +853,6 @@ const HLJS = function(hljs) {
       // entire highlighter
       lang = PLAINTEXT_LANGUAGE;
     }
-    // give it a temporary name if it doesn't have one in the meta-data
-    if (!lang.name) lang.name = languageName;
```
joshgoebel, line 857: We can't assume all 3rd party grammars are type safe, so this would still be necessary.

src/languages/php.js
```diff
@@ -477,6 +477,7 @@ export default function(hljs) {
   };
 
   return {
+    name: 'PHP',
```
joshgoebel, line 480:
```suggestion
    name: 'PHP (snippet)',
```

vs `template`... PHP alone is ambiguous.

types/index.d.ts
```diff
@@ -162,7 +162,7 @@ declare module 'highlight.js' {
     export interface Mode extends ModeCallbacks, ModeDetails {}
 
     export interface LanguageDetail {
-        name?: string
+        name: string
```
joshgoebel, line 165: We'd need to hold this until v12 as this is a breaking change.

joshgoebel:
Ping.

joshgoebel:
Ping. Did you want to finish work on this?

joshgoebel:
Closing for inactivity, if you return this can always be reopened.

PR #3774, opened 2023-04-29
(core) fix up types of new Emitter API

joshgoebel:
Just some forgotten type cleanups for the new emitter API


joshgoebel:
@woorm does this look correct?

joshgoebel:
@wooorm does this look correct?

wooorm:
Yep, works for me, I can remove https://github.com/wooorm/lowlight/blob/0ab194d3ce67d12ef0b2f00b6c665791e1d77187/lib/core.js#L61 and everything works.

PR #3779, opened 2023-05-04
Fix LanguageDetail.keywords and ModeDetail.keywords types (#3778)

### LINKED ISSUE #3778: Incorrect type of `LanguageDetail.keywords`
`LanguageDetail.keywords` are currently defined as

https://github.com/highlightjs/highlight.js/blob/cc93b0e6d9209924e2ec2d55648a1e74a757a40c/types/index.d.ts#L172

But we have at least 3 formats:

1. Whitespace-separated string as in [Cos](https://github.com/highlightjs/highlight.js/blob/cc93b0e6d9209924e2ec2d55648a1e74a757a40c/src/languages/cos.js#L32).
2. `string[]` as in [Dockerfile](https://github.com/highlightjs/highlight.js/blob/cc93b0e6d9209924e2ec2d55648a1e74a757a40c/src/languages/dockerfile.js#L12).
3. `Record<string, string | string[]>` (plus optional `$pattern`) as in [Dart](https://github.com/highlightjs/highlight.js/blob/cc93b0e6d9209924e2ec2d55648a1e74a757a40c/src/languages/dart.js#L195).

The 2nd one is illegal in that definition. And the 3rd one is not as specific as it can be.

Should the correct one be `string | string[] | Record<string, string | string[]>` ?

Same goes for `ModeDetail`.

Since we are here, must we really support that variety of definitions? Is there a performance difference in loading that? Or is there no reason to prefer whitespace-separated string over `string[]`? Could we change to something like `Record<string, string[]>` in version 12? (plus some provision for `$pattern: string`, not sure how to code that correctly).

The reason I care is that I am porting your package to Dart, and the porting script for diverse definitions is harder to maintain.

alexeyinkin:
- Resolves #3778

### Checklist

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -169,7 +169,7 @@ declare module 'highlight.js' {
         disableAutodetect?: boolean
         contains: (Mode)[]
         case_insensitive?: boolean
-        keywords?: Record<string, any> | string
+        keywords?: string | string[] | Record<string, string | string[]>
         isCompiled?: boolean,
         exports?: any,
         classNameAliases?: Record<string, string>
@@ -250,7 +250,7 @@ declare module 'highlight.js' {
         parent?: Mode
         starts?:Mode
         lexemes?: string | RegExp
-        keywords?: Record<string, any> | string
+        keywords?: string | string[] | Record<string, string | string[]>
         beginKeywords?: string
         relevance?: number
         illegal?: string | RegExp | Array<string | RegExp>
@@ -272,5 +272,3 @@ declare module 'highlight.js/lib/languages/*' {
     const defineLanguage: LanguageFn;
     export default defineLanguage;
 }
-
-


types/index.d.ts
```diff
@@ -169,7 +169,7 @@ declare module 'highlight.js' {
         disableAutodetect?: boolean
         contains: (Mode)[]
         case_insensitive?: boolean
-        keywords?: Record<string, any> | string
+        keywords?: string | string[] | Record<string, string | string[]>
```
joshgoebel, line 172: We typically use `Array<type>` for arrays not `[]`.  Can we do that here also for consistency?

alexeyinkin: In this file I can see one `Array<` and 24 instances of `[]`.
That only instance is `Array<string | RegExp>` so it must use this syntax.
I suggest keeping `[]` where they are possible in this file.

joshgoebel (approved):

PR #3788, opened 2023-05-19
Added RDF/XML as an alias for the "HTML, XML" language

csnyulas:
Added RDF/XML as an alias for the "HTML, XML" language. RDF/XML is an important file format (and also a serialization of Web Ontology Language models, and an official DOCTYPE). It is a W3C recommendation, and it is widely used in the Semantic Web community to represent (OWL) ontologies and RDF graphs.

### Changes

Just added an additional alias "RDF/XML" for the "HTML, XML" language in the `xml.js` and `SUPPORTED_LANGUAGES.md` files, and logged these changes in the `CHANGES.md`. 

### Checklist
- [ ] Added markup tests, or they don't apply here because... (I am not sure if aliases need to be tested individually, and if yes how)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,9 +10,16 @@ Dev tool:
 - (chore) Update dev tool to use the new `highlight` API. [Shah Shabbir Ahmmed][]
 - (enh) Auto-update the highlighted output when the language dropdown changes. [Shah Shabbir Ahmmed][]
 
+Core Grammars:
+
+- Added "RDF/XML" as an alias for the "HTML, XML" grammar, and listed it also in SUPPORTED_LANGUAGES [csnyulas][]
+
+
 [Shah Shabbir Ahmmed]: https://github.com/shabbir23ah
 [Josh Goebel]: https://github.com/joshgoebel
 [Checconio]: https://github.com/Checconio
+[csnyulas]: https://github.com/csnyulas
+
 
 
 ## Version 11.8.0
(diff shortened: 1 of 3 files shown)


joshgoebel:
Thanks.

I should be added as simply as `rdf`, not `RDF/XML`

csnyulas:
Re: https://github.com/highlightjs/highlight.js/pull/3788#issuecomment-1555052440

RDF graphs can be serialized in many formats. **RDF/XML is just one of them** (which, BTW, is the primary exchange syntax for [OWL 2](https://en.wikipedia.org/wiki/OWL_2), and must be supported by all OWL 2 tools): https://en.wikipedia.org/wiki/Resource_Description_Framework#Serialization_formats

In fact, there are many other more popular serialization formats for RDF, like **Turtle** and **JSON-LD**, which would be also very nice to be supported by highlight.js.

So, writing just "rdf" would be not sufficient, and would be confusing. It would be also misleading, if it would be displayed as such (i.e. simply "rdf").
Why is "RDF/XML" not good? Is there a problem with including a "/" in the alias?

joshgoebel:
> RDF graphs can be serialized in many formats. RDF/XML is just one of them 

Say no more :)  I understand the problem now.

> Is there a problem with including a "/" in the alias?

Yeah, often these are pulled from CSS class names and I don't think that's valid... 

How about `rdf_xml`?

csnyulas:
I guess `rdf_xml` would do it. It is definitely OK to be used as an internal "identifier", but it would be even better if it could be rendered nicely as "**RDF/XML**" in the generated output. However, I don't know how is that controlled.

image

Thank you!

joshgoebel:
We don't render language names in our output... pretty names as part of some UI would be an implementer's responsibility.   We do provide `grammar.name`, but that isn't helpful here.

>  It is definitely OK to be used as an internal "identifier",

That is what aliases are, internal identifiers.

joshgoebel:
So are we going back to `rdf_xml` then?

joshgoebel:
Ping.

PR #3789, opened 2023-05-19
Update Haxe grammar

RobertBorghese:
### Changes
Updated `languages/haxe.js` to properly highlight some missing Haxe keywords and features.

### Checklist
- [x] [Macro reification](https://haxe.org/manual/macro-reification-expression.html) (macro keyword + `$identifiers`)
- [x] `final` keyword ([added in Haxe 4.0](https://github.com/HaxeFoundation/haxe/wiki/What's-new-in-Haxe-4#final))
- [x] `is` keyword
- [x] Support numeric literal suffix ([added in Haxe 4.3](https://haxe.org/blog/evolution-meeting-2021))
- [x] Support numeric separators ([added in Haxe 4.3](https://haxe.org/blog/evolution-meeting-2021))
- [x] Fixed metadata arguments being colored unnecessarily.
- [x] Support non-colon metadata.
- [x] Fixed `abstract` module highlighting so doesn't conflict with abstract [class](https://haxe.org/manual/types-abstract-class.html), function, and [new keyword](https://haxe.org/blog/evolution-meeting-2021/#self-access-for-abstracts).

### DIFF
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -20,7 +20,7 @@ export default function(hljs) {
     aliases: [ 'hx' ],
     keywords: {
       keyword: 'abstract break case cast catch continue default do dynamic else enum extern '
-               + 'final for function here if import in inline is never new override package private get set '
+               + 'final for function here if import in inline is macro never new override package private get set '
                + 'public return static super switch this throw trace try typedef untyped using var while '
                + HAXE_BASIC_TYPES,
       built_in:
@@ -54,6 +54,10 @@ export default function(hljs) {
         className: 'number',
         begin: HAXE_NUMBER_RE,
         relevance: 0
+      },
+      {
+        className: 'built_in',
+        begin: "\\$" + IDENT_RE,
       },
       {
         className: 'meta', // compiler meta
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -16,7 +16,7 @@ export default function(hljs) {
     name: 'Haxe',
     aliases: [ 'hx' ],
     keywords: {
-      keyword: 'break case cast catch continue default do dynamic else enum extern '
+      keyword: 'abstract break case cast catch continue default do dynamic else enum extern '
                + 'final for function here if import in inline is never new override package private get set '
                + 'public return static super switch this throw trace try typedef untyped using var while '
                + HAXE_BASIC_TYPES,
@@ -90,7 +90,7 @@ export default function(hljs) {
       },
       {
         className: 'class', // abstracts
-        beginKeywords: 'abstract',
+        begin: '\\babstract\\b(?=\\s*' + hljs.IDENT_RE + '\\s*\\()',
         end: '[\\{$]',
         contains: [
           {
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -56,7 +56,7 @@ export default function(hljs) {
         relevance: 0
       },
       {
-        className: 'built_in',
+        className: 'variable',
         begin: "\\$" + IDENT_RE,
       },
       {
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -8,10 +8,10 @@ Website: https://haxe.org
 
 export default function(hljs) {
   const IDENT_RE = '[a-zA-Z_$][a-zA-Z0-9_$]*';
-  const IDENT_FUNC_RETURN_TYPE_RE = '([*]|[a-zA-Z_$][a-zA-Z0-9_$]*)';
+  const IDENT_FUNC_RETURN_TYPE_RE = /([*]|[a-zA-Z_$][a-zA-Z0-9_$]*)/;
 
   // C_NUMBER_RE with underscores and literal suffixes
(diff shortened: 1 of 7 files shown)


src/languages/haxe.js
```diff
@@ -47,11 +50,20 @@ export default function(hljs) {
...
       hljs.C_BLOCK_COMMENT_MODE,
-      hljs.C_NUMBER_MODE,
+      {
+        className: 'number',
+        begin: HAXE_NUMBER_RE,
+        relevance: 0
+      },
+      {
+        className: 'built_in',
+        begin: "\\$" + IDENT_RE,
```
joshgoebel, line 60: What kind of identifier is this exactly?

src/languages/haxe.js
```diff
@@ -10,14 +10,17 @@ export default function(hljs) {
   const IDENT_RE = '[a-zA-Z_$][a-zA-Z0-9_$]*';
   const IDENT_FUNC_RETURN_TYPE_RE = '([*]|[a-zA-Z_$][a-zA-Z0-9_$]*)';
 
+  // C_NUMBER_RE with underscores and literal suffixes
+  const HAXE_NUMBER_RE = '(-?)(\\b0[xX][a-fA-F0-9_]+|(\\b\\d+(\\.[\\d_]*)?|\\.[\\d_]+)(([eE][-+]?\\d+)|i32|u32|i64|f64)?)'
```
joshgoebel, line 14: Please write full regex as a regex literal, not a string.

src/languages/haxe.js
```diff
@@ -47,11 +50,20 @@ export default function(hljs) {
...
+      },
+      {
+        className: 'built_in',
+        begin: "\\$" + IDENT_RE,
+      },
       {
         className: 'meta', // compiler meta
-        begin: '@:',
-        end: '$'
+        begin: '@:?',
```
joshgoebel, line 64: Since we're changing, please write full regex as a regex literal, not a string.

src/languages/haxe.js
RobertBorghese: The `$` is used to inject a variable into a macro template.

Without going into too much detail, the best way I can describe it is it's for a system similar to JavaScript template strings. Injects a variable into the template:
```js
`the value of xyz is: $xyz`
```

But instead of strings, it's for generating expressions:
```haxe
var fieldName = "bar";

// generate programmatic structure representation of "foo.bar" expression
var expressionData = macro foo.$fieldName;
```

As you can see, Github (and other Haxe grammars) give it a distinct yellow? So that's what I was going for, but sorry if I used wrong className.

src/languages/haxe.js
RobertBorghese: 👍

src/languages/haxe.js
```diff
@@ -89,7 +101,7 @@ export default function(hljs) {
       },
       {
         className: 'class', // abstracts
-        beginKeywords: 'abstract',
+        begin: '\\babstract\\b(?=\\s*' + hljs.IDENT_RE + '\\s*\\()',
```
joshgoebel, line 104: Is all this necessary?  Is `abstract` a soft/contextual keyboard in Haxe?

src/languages/haxe.js
RobertBorghese: Would it be preferred if I switched everything to regex literals? I wouldn't mind.

src/languages/haxe.js
RobertBorghese: Early in Haxe, the `abstract` keyword was only for this specific declaration:
```haxe
abstract SomeName(TYPE) from TYPE to TYPE
```

But now it's a way more common keyword that's used as an special identifier and a descriptor for `class` and `function`. 

Preferably, it should be treated just like an arbitrary keyword at this point,. This code is more meant to setup the correct context for coloring "from" and "to". Those are not reserved keywords and can be used as normal variable names, so should  only be colored in the specific context above.

Maybe there's a better way to do it? Or would it be okay just to have "from" and "to" colored as keywords?

src/languages/haxe.js
joshgoebel: Looks like a variable, no?  I think if it's a variable we'd want to use `variable`.

src/languages/haxe.js
joshgoebel: Sure, just make it a separate commit from the other fixes for easier review.  You could also update `className` to the new `scope` naming.

src/languages/haxe.js
joshgoebel: This all makes sense given from and to are contextual.

joshgoebel:
Could you rebase on main now that I fixed the test suite?

RobertBorghese:
Done! 👍

joshgoebel:
Also will need to update CHANGELOG

joshgoebel (approved):
Looks good.  Just needs changelog.

RobertBorghese:
Hope that looks good!

joshgoebel (approved):

joshgoebel:
Thanks so much!

Issue #3791, opened 2023-05-21
(bash) escaped apostrophe is highlighted wrong way


hancar:
**Describe the issue**

**Which language seems to have the issue?**
bash

**Are you using `highlight` or `highlightAuto`?**
not sure, probably highlight

**Sample Code to Reproduce**
A minimal example is probably:
<code>echo \\'not-a-highlighted-string\\'</code>
highlight.js colors it from apostrophe to apostrophe.

It starts to be really annoying, if one wants to highlight a complicated grep with apostrophes inside a regex.
For example: <code>grep -oP '(?<=^name: ['\\''"])[^'\''"]*'</code>

**Expected behavior**
<code>echo \\'not-a-highlighted-string\\'</code>
Should not be highlighted at all, the same way as:
<code>echo \\"not-a-highlighted-string\\"</code>

(And concerning the complicated example: the highlighted strings should not start or end on the escaped apostrophes)

PR #3792, opened 2023-05-21
 fix(bash): do not delimit a string by \'

hancar:
Resolves https://github.com/highlightjs/highlight.js/issues/3791

### Changes
Escaped apostrophe in bash is not a start or an end of a string any more.

### Checklist
- [✔] Added markup tests, or they don't apply here because...
- [✔] Updated the changelog at `CHANGES.md`

### DIFF
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -60,9 +60,7 @@ export default function(hljs) {
   };
   SUBST.contains.push(QUOTE_STRING);
   const ESCAPED_QUOTE = {
-    className: '',
-    begin: /\\"/
-
+    match: /\\"/
   };
   const APOS_STRING = {
     className: 'string',
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -70,9 +70,7 @@ export default function(hljs) {
     end: /'/
   };
   const ESCAPED_APOS = {
-    className: '',
-    begin: /\\'/
-
+    match: /\\'/
   };
   const ARITHMETIC = {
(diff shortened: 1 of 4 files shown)


src/languages/bash.js
```diff
@@ -69,6 +69,11 @@ export default function(hljs) {
     begin: /'/,
     end: /'/
   };
+  const ESCAPED_APOS = {
+    className: '',
```
joshgoebel, line 73: This key shouldn't be necessary at all if you don't want to assign a scope/class to the match.

src/languages/bash.js
```diff
@@ -69,6 +69,11 @@ export default function(hljs) {
...
+    begin: /\\'/
```
joshgoebel, line 74: Use `match` not begin for a single regex match.

src/languages/bash.js
```diff
@@ -69,6 +69,11 @@ export default function(hljs) {
...
+
```
joshgoebel, line 75: Remove extra blank line please.

hancar: Thank you for the review. I just want to say, that the code is a remake of `const ESCAPED_QUOTE` above. The `ESCAPED_QUOTE` also has the  `className: ''`, the `begin` and even the blank line. If you want I can refactor it, in this pull request.

joshgoebel: > If you want I can refactor it, in this pull request.

Sure.

joshgoebel (approved):

PR #3794, opened 2023-05-26
Update leaf syntax to leaf 4

dannflor:
### Changes
Overhauled the leaf syntax to match the new leaf 4 grammar
https://docs.vapor.codes/leaf/overview

### Checklist

### DIFF
--- a/src/languages/leaf.js
+++ b/src/languages/leaf.js
@@ -6,7 +6,7 @@ Website: https://docs.vapor.codes/leaf/overview
 Category: template
 */
 
-export default function leaf(hljs) {
+export default function(hljs) {
   return {
     name: 'Leaf',
     contains: [
--- a/src/languages/leaf.js
+++ b/src/languages/leaf.js
@@ -7,75 +7,77 @@ Category: template
 */
 
 export default function(hljs) {
+  const IDENT = /([A-Za-z_][A-Za-z_0-9]*)?/;
+  const LITERALS = [
+    'true',
+    'false',
+    'in'
+  ];
   return {
     name: 'Leaf',
     contains: [
+      // #ident():
       {
-        className: 'function',
-        begin: '#' + '[A-Za-z_0-9]*' + '\\(',
-        returnBegin: true,
+        match: [
+          /#+/,
+          IDENT,
+          /(?=\()/,
+        ],
+        scope: {
+          1: "punctuation",
+          2: "keyword"
+        },
+        // will start up after the ending `)` match from line ~44
+        // just to grab the trailing `:` if we can match it
+        starts: {
+          contains: [
+            {
+              match: /\:/,
+              scope: "punctuation"
+            }
+          ]
+        },
         contains: [
           {
-            className: 'keyword',
-            begin: '#'
-          },
-          {
-            className: 'title',
-            begin: '[A-Za-z_][A-Za-z_0-9]*'
-          },
-          {
-            className: 'params',
-            begin: '\\(',
-            end: '\\)\\:?',
+            scope: 'params',
+            begin: /\(/,
+            end: /\)(?=\:?)/,
             endsParent: true,
             relevance: 7,
             contains: [
               {
-                className: 'string',
+                scope: 'string',
                 begin: '"',
                 end: '"'
               },
               {
-                className: 'keyword',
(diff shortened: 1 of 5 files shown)


src/languages/leaf.js
```diff
@@ -1,23 +1,23 @@
...
-Author: Hale Chan <halechan@qq.com>
-Description: Based on the Leaf reference from https://vapor.github.io/documentation/guide/leaf.html.
+Author: Samuel Bishop <sambobjr@gmail.com>
+Description: A Swift-based templating language created for the Vapor project.
+Website: https://docs.vapor.codes/leaf/overview
+Category: template
 */
 
-export default function(hljs) {
+export default function leaf(hljs) {
```
joshgoebel, line 9: Revert please. This is purposely unnamed.

src/languages/leaf.js
```diff
@@ -26,22 +26,57 @@ export default function(hljs) {
...
+                className: 'keyword',
+                begin: 'true|false|in',
+              },
               {
                 className: 'variable',
                 begin: '[A-Za-z_][A-Za-z_0-9]*'
-              }
+              },
+              {
+                className: 'operator',
```
joshgoebel, line 47: Operators can only exist within params?

dannflor: Correct. Leaf is a templating language so there's no use for them outside of a pair of parentheses

src/languages/leaf.js
```diff
@@ -26,22 +26,57 @@ export default function(hljs) {
...
             endsParent: true,
+            relevance: 7,
             contains: [
               {
                 className: 'string',
                 begin: '"',
                 end: '"'
               },
+              {
+                className: 'keyword',
```
joshgoebel, line 39: Ditto, are these only valid within params?

dannflor: See above

joshgoebel:
Thanks, but it looks like this whole grammar is super old and using very old conventions - we have nicer ways to do a lot of this now.  I'll try to pull this down and make a pass at cleaning it up then toss it back.

Is there a list of official helpers/keywords in Leaf (loop, for, etc)?  If so we should build them in so they can be highlighted as `built-in` perhaps...

dannflor:
https://docs.vapor.codes/leaf/overview is basically all you have to work off of. Anyone can define any tag though, so it doesn't really make sense to distinguish between built-in and user defined. The only built in tag that has any different syntax from an arbitrarily user defined one is `for` requiring an `in` in the parameter body. But you could also mark every one of the tags that is on that page as built-in.

joshgoebel:
Screen Shot 2023-05-25 at 10 00 55 PM

Pretty sure if/else/etc should all match though... ie, they are a single scope...

dannflor:
Sounds good to me. We've been living without syntax highlighting on those specific keywords for a couple of years now so I don't know if there are any strong opinions lol

dannflor:
I updated the expected markup to match what we want to generate. There's only one thing missing, which is tag functions can nest inside each other (nested functions omit the `#` at the beginning). I believe they can nest to an arbitrary depth, but I know regular expressions can't do that (though maybe you have some way around that). I'm not sure how to implement that in your current system for defining the grammar

joshgoebel:
> which is tag functions can nest inside each other (nested functions omit the # at the beginning)

Give me a few actual examples please.

dannflor:
Sorry, I put it in the markup test. It's basically something like:
`#if(count(arbitaryFunction(variable))) == 0):`

joshgoebel:
Regex doesn't support infinite nesting, but the parser does... for something like one rule immediately inside another we have a shortcut "self"... for less-direct looping you have to create the objects separate then link them together as the grammar is executed. IE:

- a
  - b
    - c

Where a contains b which contains c, etc... but there is nothing to stop `C` from containing `A`... and it can go as deep as necessary....

dannflor:
Hey, it works! LGTM

dannflor:
Is there something remaining to be done here holding up merging?

joshgoebel:
Yeah, just need to update the changelog :)

dannflor:
Done

joshgoebel (approved):

joshgoebel:
Thanks!

PR #3795, opened 2023-05-26
Simplify ReasonML syntax and align it with OCaml

jchavarri:
No related issue.

The current ReasonML syntax is very sophisticated but unfortunately the complexity leads to a lot of bugs in highlighting in comments and other blocks.

Besides the above, the way it is defined does not align with OCaml syntax, that ReasonML builds upon, which leads to changes in highlighting for websites that have snippets with both languages ([example](https://melange.re/communicate-with-javascript/)).

### Changes

- To solve the above issues, this PR simplifies ReasonML syntax by reusing most of the configuration from the OCaml one.
- It also adds and extends some markup tests to check additional use cases.
  - The first commit in the PR just adds the tests so the fixes can be visualized in the second one.
- Updates ReasonML keywords to remove some unused (e.g. `match`, `method`, `private`) and include missing ones (e.g. `pri`, `switch`). The list is taken directly from the [Reason parser code](https://github.com/reasonml/reason/blob/95485d5c4527641903d9bec81e80aa3c5dfb3d30/src/reason-parser/reason_declarative_lexer.mll#L73).

Here is a comparison on how the markup tests render, before and after this PR:

| Before  | After |
| ------------- | ------------- |
| reason-previous | reason-latest |

### Checklist
- [x] Added markup tests


joshgoebel:
Simplification can be good, but is there a good reason we lost support for operators?

src/languages/reasonml.js
```diff
@@ -6,301 +6,63 @@ Author: Gidi Meir Morris <oss@gidi.io>
...
+        'array bool bytes char exn|5 float int int32 int64 list lazy_t|5 nativeint|5 ref string unit ',
+      literal:
+        'true false'
+    },
     illegal: '(:-|:=|\\$\\{|\\+=)',
     contains: [
-      hljs.COMMENT('/\\*', '\\*/', { illegal: '^(#,\\/\\/)' }),
-      {
-        className: 'character',
-        begin: '\'(\\\\[^\']+|[^\'])\'',
```
joshgoebel, line 247: Please use real literal regex whenever you can, not double-escaped strings, which makes it much harder to read the actual regex.  I realize that might have been how it was before, but if we're pretty much starting over we need to follow the conventions we use for our more modern grammars - which is to use real regex whenever possible.

src/languages/reasonml.js
```diff
@@ -6,301 +6,63 @@ Author: Gidi Meir Morris <oss@gidi.io>
...
-  PARAMS_CONTENTS.push(MODULE_ACCESS_MODE);
-
   return {
     name: 'ReasonML',
     aliases: [ 're' ],
-    keywords: KEYWORDS,
+    keywords: {
+      $pattern: '[a-z_]\\w*!?',
+      keyword:
+        'and as asr assert begin class constraint do done downto else end esfun '
```
joshgoebel, line 15: Should be an array of strings, one keyword per line.  See many of the newer grammars that follow this pattern for easier maintenance.

src/languages/reasonml.js
```diff
@@ -6,301 +6,63 @@ Author: Gidi Meir Morris <oss@gidi.io>
...
+        + 'exception external for fun function functor if in include inherit '
+        + 'initializer land lazy let lor lsl lsr lxor mod module mutable new '
+        + 'nonrec object of open or pri pub rec sig struct switch then to try '
+        + 'type val virtual when while with',
+      built_in:
```
joshgoebel, line 20: These look like `type` to me.

src/languages/reasonml.js
```diff
@@ -6,301 +6,63 @@ Author: Gidi Meir Morris <oss@gidi.io>
...
+        className: 'type',
+        begin: '`[A-Z][\\w\']*'
       },
-      CONSTRUCTOR_MODE,
-      {
-        className: 'operator',
-        begin: RE_OPERATOR_SPACED,
-        illegal: '-->',
+      { /* module or constructor */
+        className: 'type',
```
joshgoebel, line 44: This is now named `scope`. `className` is deprecated.

src/languages/reasonml.js
```diff
@@ -6,301 +6,63 @@ Author: Gidi Meir Morris <oss@gidi.io>
...
+        begin: '\\b[A-Z][\\w\']*',
```
joshgoebel, line 45: If there is no `end` matcher then you should use `match` instead of `begin` for clarity.

src/languages/reasonml.js
```diff
@@ -6,301 +6,63 @@ Author: Gidi Meir Morris <oss@gidi.io>
...
-          {
-            begin: /\{/,
-            end: /\}/,
-            relevance: 0,
-            skip: true
-          }
-        ].concat(MODULE_ACCESS_CONTENTS)
+        className: 'number',
+        begin:
+          '\\b(0[xX][a-fA-F0-9_]+[Lln]?|'
```
joshgoebel, line 60: Lets use the regex tools (`hljs.regex`) here... either `regex.either` or this could also be done with `variants` (see docs).

src/languages/reasonml.js
jchavarri: https://github.com/highlightjs/highlight.js/pull/3795/commits/ec0846191fbfeb48f1c57877fbd1e8fd57b3aab0

src/languages/reasonml.js
jchavarri: https://github.com/highlightjs/highlight.js/pull/3795/commits/70e83b3ee7703499bc5f98ae3493b0ec6d261577

src/languages/reasonml.js
jchavarri: These are built-in types. I would like to keep compatibility with OCaml syntax, which also defines them as built-in:

https://github.com/highlightjs/highlight.js/blob/73a398ddca4339273fd1203f1ecabab86d75de5c/src/languages/ocaml.js#L26-L27

There are also other languages that do something similar, like Dart:

https://github.com/highlightjs/highlight.js/blob/73a398ddca4339273fd1203f1ecabab86d75de5c/src/languages/dart.js#L197-L198

I made the definition more explicit in https://github.com/highlightjs/highlight.js/pull/3795/commits/d1c3e5833ab7ea627cfdfa0fb275813e69540dcd. Wdyt?

src/languages/reasonml.js
jchavarri: https://github.com/highlightjs/highlight.js/pull/3795/commits/a72d1f9137e8e7aa5ab51d58349d64c8cb41f706

src/languages/reasonml.js
jchavarri: https://github.com/highlightjs/highlight.js/pull/3795/commits/7f04ca8b37f494d5b6999b3b77a815f59880671d

src/languages/reasonml.js
jchavarri: https://github.com/highlightjs/highlight.js/pull/3795/commits/355e5003525e5ef24799a6167372606e29c1c5f5

jchavarri:
@joshgoebel thanks for the review, I think I tackled all the comments.

> but is there a good reason we lost support for operators?

Mostly compatibility with OCaml syntax, which I would like to keep for the reasons mentioned in the description. If you feel strongly about re-adding them, I can do so in a separate PR where operators get added to both syntaxes?

src/languages/reasonml.js
joshgoebel: >  to keep compatibility with OCaml syntax,

Perhaps Ocaml should also change?  What is it that we do consider a type in ReasonML/OCaml?

joshgoebel:
> If you feel strongly about re-adding them

A major simplification could perhaps accept tradeoffs, but since we already have #2500 and the future of the project is higher fidelity I don't think we should start going the opposite direction.  Why not put operators back and then just add them to OCaml also if the idea is equal fidelity?

jchavarri:
> Why not put operators back and then just add them to OCaml also if the idea is equal fidelity?

Added operators back in https://github.com/highlightjs/highlight.js/pull/3795/commits/33617405afc19640cac1a4f233f17d9e57612d61. I'd rather add OCaml ones in a separate PR to not increase the scope of this one.

src/languages/reasonml.js
jchavarri: A type in ReasonML/OCaml is an abstraction that represents a set of values, same as in any other typed language. There are [built-in types](https://ocaml.org/docs/data-types#built-in-compound-types) and [user-defined types](https://ocaml.org/docs/data-types#a-simple-custom-type).

jchavarri: I think it is useful to store built-in types in a separate scope, for the sake of having higher fidelity precisely. Note this example:

imagen

joshgoebel: Do other highlighters consider these to be separate concepts worthy of separate highlighiting?

joshgoebel:
>  I'd rather add OCaml ones in a separate PR

Of course that totally makes sense.

src/languages/reasonml.js
jchavarri: The vscode-ocaml grammar has a separate highlight for built-in types:

https://github.com/ocamllabs/vscode-ocaml-platform/blob/a170e84a20078a816184ae705b7ee4b6896c4a20/syntaxes/ocaml.json#L617-L619

joshgoebel:
Did we tackle everything here?

jchavarri:
> Did we tackle everything here?

@joshgoebel Yes, I think so.

joshgoebel (approved):

PR #3799, opened 2023-05-30
fix(haskell) Do not treat dashes inside infix operators as comments

zlondrej:
This fixes cases where operators like `$--`, `-->` or similar are treated as comment.

### Changes
Only detects line comments according to lexical structure described in Haskell 98 report: https://www.haskell.org/onlinereport/lexemes.html

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/haskell.js
+++ b/src/languages/haskell.js
@@ -7,8 +7,26 @@ Category: functional
 */
 
 export default function(hljs) {
+
+  /* See:
+     - https://www.haskell.org/onlinereport/lexemes.html
+     - https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/binary_literals.html
+     - https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/numeric_underscores.html
+     - https://downloads.haskell.org/ghc/9.0.1/docs/html/users_guide/exts/hex_float_literals.html
+  */
+  const decimalDigits = '([0-9]_*)+';
+  const hexDigits = '([0-9a-fA-F]_*)+';
+  const binaryDigits = '([01]_*)+';
+  const octalDigits = '([0-7]_*)+';
+  const ascSymbol = '[!#$%&*+.\\/<=>?@\\\\^~-]';
+  const uniSymbol = '(\\p{S}|\\p{P})' // Symbol or Punctuation
+  const special = '[(),;\\[\\]`|{}]';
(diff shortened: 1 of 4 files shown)


joshgoebel (approved):
Thanks!

src/languages/haskell.js
```diff
@@ -7,8 +7,26 @@ Category: functional
...
+  const ascSymbol = '[!#$%&*+.\\/<=>?@\\\\^~-]';
+  const uniSymbol = '(\\p{S}|\\p{P})' // Symbol or Punctuation
+  const special = '[(),;\\[\\]`|{}]';
+  const symbol = `(${ascSymbol}|(?!(${special}|[_:"']))${uniSymbol})`;
+
   const COMMENT = { variants: [
-    hljs.COMMENT('--', '$'),
+    // Double dash forms a valid comment only if it's not part of legal lexeme.
+    // See: Haskell 98 report: https://www.haskell.org/onlinereport/lexemes.html
+    hljs.COMMENT(`(?<!${symbol})--+(?!${symbol})`, '$'),
```
joshgoebel, line 29: Can't have negative look-behind, not well supported by Safari yet.  Sorry.

joshgoebel: If we don't need the first character (to be highlighted) the simplest way to do this is include it in the match, but not higlight it... multi-match is good for such things:

Where `yyyy` is a keyword only when NOT preceeded by `x`.

```
{
begin: [
  /[^x]/,
  /yyyy/
],
beginScope: {
  2: "keyword"
},
}
```

joshgoebel:
Ping, can we look into removing that negative look-behind?

zlondrej:
Unfortunately, this won't do. Character class subtraction (need for it stems from unicode part of 'symbols' rule in Haskell report) is not supported by JS regexp and the lookarounds are the only way to simulate those. But that won't work within character range.

I guess the only way is to produce those ranges manually.

joshgoebel (approved):

joshgoebel:
> Otherwise I would need hljs rules to accept function to be able to have custom parsers that regex doesn't handle (in fully compatible way).

There are `on:begin and `on:end` per mode callbacks that allow you to "qualify" a match... and a grammar itself may include a whole custom parsing engine - but not individual rules. (https://highlightjs.readthedocs.io/en/latest/mode-reference.html#emittokens).  We use callbacks sparingly though because of the expense and an opportunity to use a custom parser in core hasn't come up yet.  

Your fix seems fine and is a common workaround for these types of issues.

PR #3803, opened 2023-06-11
enh(swift) add parameter pack keywords

bradleymackey:
Parameter packs are supported in Swift 5.9

Their usage can be seen in [SE-0393](https://github.com/apple/swift-evolution/blob/main/proposals/0393-parameter-packs.md). This essentially just requires a promotion of `each` to a keyword and matching `repeat` and `each` as keywords in generic parameters.

### Changes
- Support `repeat` and `each` parameter pack keywords.

### Checklist


src/languages/swift.js
```diff
@@ -302,7 +302,11 @@ export default function(hljs) {
     end: />/,
     contains: [
       ...COMMENTS,
-      TYPE
+      TYPE,
+      {
+        match: /(repeat|each)/,
+        className: "keyword"
```
joshgoebel, line 308: Lets start using the new `scope` naming... `className` is on the way out.

joshgoebel (approved):

joshgoebel:
So what is repeat when used outside a generic?

joshgoebel:
Is it not a keyword in those other contexts?

bradleymackey:
It is a keyword in those contexts as well, it's already in the "keywords" list though because it's already used for loops.

```swift
repeat {
    print(number)
    number += 1
} while number <= 20
```

joshgoebel:
Ah that's what I was missing. 😀

src/languages/swift.js
```diff
@@ -302,7 +302,11 @@ export default function(hljs) {
     end: />/,
     contains: [
       ...COMMENTS,
-      TYPE
+      TYPE,
+      {
```
joshgoebel, line 306: Is there a reason we aren't just using `keywords` here vs a custom match rule?

PR #3804, opened 2023-06-12
enh(swift) regex literal support

bradleymackey:
Swift 5.7 introduced regex literals, specified in [SE-0354](https://github.com/apple/swift-evolution/blob/main/proposals/0354-regex-literals.md).

This includes support for bare-slash (`/example/`) and extended regex literals (`#/example/#`, `##/example/##`, etc).
- Bare slash literals are not enabled by default in Swift 5, but will be in Swift 6.
- Bare slash literals cannot start with whitespace, but extended literals can.
- Only extended regex literals support newlines.
- Multiline extended literals support comments starting with `#`, continuing until end of line.
- Follows convention of raw strings, supporting up to 3 levels of delimiting for extended literals, ensuring `#` delimiters match.

### Changes
- Add regex literal support

### Checklist

### DIFF
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -208,12 +208,6 @@ export default function(hljs) {
     ],
   };
 
-  const EXTENDED_REGEXP_COMMENT = {
-    scope: 'comment',
-    begin: /#/,
-    end: /$/,
-  };
-
   // Adapted from hljs.REGEXP_MODE
   // https://docs.swift.org/swift-book/documentation/the-swift-programming-language/lexicalstructure/#Regular-Expression-Literals
   const EXTENDED_REGEXP_LITERAL = (rawDelimiter) => ({
@@ -221,7 +215,11 @@ export default function(hljs) {
     end: concat(/\//, rawDelimiter),
     contains: [
       ...REGEXP_CONTENTS,
-      EXTENDED_REGEXP_COMMENT
+      {
+        scope: 'comment',
+        begin: /#/,
+        end: /$/,
+      }
     ]
   });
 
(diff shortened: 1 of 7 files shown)


bradleymackey:
Closing because of edge cases with bare-slash literals and operators. I'll reopen once resolved.

src/languages/swift.js
```diff
@@ -180,6 +180,54 @@ export default function(hljs) {
...
+    //
+    //     3 / something
+    //
+    // (which will then blow up when regex's `illegal` sees the newline)
+    begin: /(?=\/[^\s][^/\n]*\/)/,
+    contains: [
+      {
+        begin: /\/[^\s]/,
+        end: /\//,
+        illegal: /\n/,
```
joshgoebel, line 207: Is this needed with the "lookahead" you're doing on line 202 which should guarantee there will be no newlines?

bradleymackey: You're right, this works as intended without this additional rule. This was heavily inspired by `REGEXP_MODE` which uses the [same `illegal` rule](https://github.com/highlightjs/highlight.js/blob/98be55bf6520f62c581a498509baf9387461dce0/src/lib/modes.js#L164). The suite also passes when this `illegal` rule is removed as well, should I include that removal in this patch as well for consistency?

bradleymackey:
Added comment support for extended literals, starting with `#` ending with a newline.

src/languages/swift.js
joshgoebel: Sure, I can't see what purpose it's serving.  I wonder if the whole rule also couldn't be flattened by just lifting the rule out of the contains and adding the lookahead to it's begin?

bradleymackey: Good suggestion! Yes did the trick much more elegantly

bradleymackey:
Fixed comments in extended literals, only multiline literals support comments.

PR #3807, opened 2023-06-14
Build non-minified CSS files for node & CDN

### LINKED ISSUE #3797: Publish non-minified styles to CDN
Please build both minified/non-minified styles and publish them on npm/`@highlightjs/cdn-assets`.

We use the npm package to bundle some external libraries - https://github.com/atk4/ui/tree/4.0.0/public/external - and non-minified files are preferred for debugging and diffing changes.

Thank you.

mvorisek:
fix #3797

### Changes

Beatified `.css` files are newly created in node/cdn build. `.min.css` files are build as before.

### Checklist

### DIFF
--- a/tools/build_browser.js
+++ b/tools/build_browser.js
@@ -169,9 +169,11 @@ function installDemoStyles() {
 
   glob.sync("**", { cwd: "./src/styles" }).forEach((file) => {
     const stat = fss.statSync(`./src/styles/${file}`);
+    if (stat.isDirectory()) return;
+
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`);
-    } else if (!stat.isDirectory) {
+      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`, false);
+    } else {
       // images, backgrounds, etc
       install(`./src/styles/${file}`, `demo/styles/${file}`);
     }
--- a/tools/build_browser.js
+++ b/tools/build_browser.js
@@ -172,7 +172,7 @@ function installDemoStyles() {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`, { minify: false });
+      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`);
     } else {
       // images, backgrounds, etc
       install(`./src/styles/${file}`, `demo/styles/${file}`);
--- a/tools/build_browser.js
+++ b/tools/build_browser.js
@@ -172,7 +172,7 @@ function installDemoStyles() {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`, false);
+      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`, { minify: false });
     } else {
       // images, backgrounds, etc
       install(`./src/styles/${file}`, `demo/styles/${file}`);
--- a/tools/build_cdn.js
+++ b/tools/build_cdn.js
@@ -141,7 +141,8 @@ function installStyles() {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file}`, false);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`, true);
     } else {
       // images, backgrounds, etc
       install(`./src/styles/${file}`, `styles/${file}`);
--- a/tools/build_cdn.js
+++ b/tools/build_cdn.js
@@ -141,8 +141,8 @@ function installStyles() {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `styles/${file}`, false);
-      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`, true);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file}`, { minify: false });
+      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`, { minify: true });
     } else {
       // images, backgrounds, etc
       install(`./src/styles/${file}`, `styles/${file}`);
--- a/tools/build_node.js
+++ b/tools/build_node.js
@@ -173,8 +173,9 @@ async function buildNode(options) {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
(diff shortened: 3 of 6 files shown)


tools/build_browser.js
```diff
@@ -169,9 +169,11 @@ function installDemoStyles() {
 
   glob.sync("**", { cwd: "./src/styles" }).forEach((file) => {
     const stat = fss.statSync(`./src/styles/${file}`);
+    if (stat.isDirectory()) return;
+
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`);
-    } else if (!stat.isDirectory) {
+      installCleanCSS(`./src/styles/${file}`, `demo/styles/${file}`, false);
```
joshgoebel, line 175: Isn't this a change in behavior?  Those files are currently minified and this would make them not, right?

tools/build_node.js
```diff
@@ -173,8 +173,9 @@ async function buildNode(options) {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `styles/${file}`);
-      installCleanCSS(`./src/styles/${file}`, `scss/${file.replace(".css", ".scss")}`);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file}`, false);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`, true);
```
joshgoebel, line 177: I'm confused, isn't passing false the same as not passing the argument?  Both `false` and `undefined` (which not passed) are falsy values...

tools/build_browser.js
mvorisek: Data change yes, functional no. I have choosed non-minified style for demo styles.

tools/build_node.js
mvorisek: The first false/undefined is correct. About the second point - please amend my PR directly to wanted style/function.

tools/lib/makestuff.js
```diff
@@ -26,10 +26,10 @@ code.hljs {
 }
 `.trim();
 
-function installCleanCSS(file, dest) {
+function installCleanCSS(file, dest, minify) {
   const theme = fs.readFileSync(file, { encoding: "utf8" });
   const content = DEFAULT_CSS + "\n" + theme;
```
joshgoebel, line 31:
```suggestion
function installCleanCSS(file, dest, opts = {}) {
  const minify = opts.minify ?? true; // default is to minify
  
  const theme = fs.readFileSync(file, { encoding: "utf8" });
  const content = DEFAULT_CSS + "\n" + theme;
```

tools/build_browser.js
joshgoebel: > Non-minified style for demo styles is probably preferred

_Oh... I wasn't even thinking about the demo site..._What is the thinking that non-minified should be preferred?  I don't think anyone is looking at these files - and [if they are] don't modern browser make it easy enough to review CSS in play regardless of whether the source files are minified or not?

mvorisek: Without source map they don't show the source nicely.

tools/lib/makestuff.js
mvorisek: Done, please enable the CI to double check the CI is passing.

tools/build_browser.js
joshgoebel: Let's focus on the CDN and Node.js builds as originally discussed.  Every modern browser I'm aware of allows you to (with built-in dev tools) easily not only edit the CSS live - but also un-minifying it as well to easily read it.  **I'm not seeing any need to change the Demo site.**  If someone wants the original files they can always fetch the repository from GitHub itself and start from there.

mvorisek: Done.

tools/build_node.js
```diff
@@ -173,8 +173,9 @@ async function buildNode(options) {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `styles/${file}`);
-      installCleanCSS(`./src/styles/${file}`, `scss/${file.replace(".css", ".scss")}`);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file}`, { minify: false });
+      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`, { minify: true });
+      installCleanCSS(`./src/styles/${file}`, `scss/${file.replace(".css", ".scss")}`, { minify: false });
```
joshgoebel, line 178: Doesn't this also need both minified and non-minified versions?

joshgoebel: Ah, good point.  I really don't know what these files are intended for (they predate me) - they aren't real SCSS (obviously)... they will be removed with v12 (as a breaking chance)... but until then this makes sense, I think we can un-minify them.

tools/build_cdn.js
```diff
@@ -141,7 +141,8 @@ function installStyles() {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file}`, { minify: false });
+      installCleanCSS(`./src/styles/${file}`, `styles/${file.replace(".css", ".min.css")}`, { minify: true });
```
joshgoebel, line 145: Ok, this is non-breaking because we're adding a NEW file, not changing behavior.

tools/build_node.js
```diff
@@ -173,8 +173,9 @@ async function buildNode(options) {
     if (stat.isDirectory()) return;
 
     if (file.endsWith(".css")) {
-      installCleanCSS(`./src/styles/${file}`, `styles/${file}`);
-      installCleanCSS(`./src/styles/${file}`, `scss/${file.replace(".css", ".scss")}`);
+      installCleanCSS(`./src/styles/${file}`, `styles/${file}`, { minify: false });
```
joshgoebel, line 176: This is slightly breaking though, as the content of `.css` here is changing... (from minified to un-minified). I think this is small enough break that it could be addressed in changelog with something like:

```
CAVEATS
- In the `node` build `styles/*.css` files now ship un-minified
- Minified counterparts can be found at `styles/*.min.css` (consistent with our CDN packaging)
```

Does that read clearly to you?  If so I think we may be ready to ship this one. :)

mvorisek: Yes, maybe mention "\`node\`/\`cdn\` builds" (`node`/`cdn` builds are basically the same).

Would you land these docs changes or do you want me to to push them?

PS: also `??` (from https://github.com/highlightjs/highlight.js/pull/3807#discussion_r1233529800) needs to be changed according to the linter

joshgoebel: I'll handle the docs and lint issue, I'll try to get to this in next few days.

mvorisek: @joshgoebel didn't you forget?

mvorisek: @joshgoebel any news?

joshgoebel (approved):

mvorisek:
Thank you ♥

joshgoebel:
Sorry took so long.

mvorisek:
Would it be possible to tag a new release?

PR #3810, opened 2023-06-23
Scala - add using directives support

bishabosha:
Add support for highlighting using directives

e.g. 

```scala
//> using scala 3.3.0
//> using options -source:future -Xfatal-warnings
//> using test.resourceDir ./resources
```

for reference, see the new directives syntax at https://scala-cli.virtuslab.org/docs/reference/directives

originally based on https://github.com/scala/docs.scala-lang/pull/2845

### Changes

Add a new comment kind for using directives, highlighting keys as a `type`, values as a `string`, and `using` as a keyword

### Checklist

### DIFF
--- a/src/languages/scala.js
+++ b/src/languages/scala.js
@@ -151,6 +151,34 @@ export default function(hljs) {
     beginScope: { 2: "keyword", }
   };
 
+  // glob all non-whitespace characters as a "string"
+  // sourced from https://github.com/scala/docs.scala-lang/pull/2845
+  const DIRECTIVE_VALUE = {
+    className: 'string',
+    begin: /\S+/,
+  }
+
+  // directives
+  // sourced from https://github.com/scala/docs.scala-lang/pull/2845
+  const USING_DIRECTIVE = {
+    begin: [
+      '//>',
+      /\s+/,
+      /using/,
+      /\s+/,
+      /\S+/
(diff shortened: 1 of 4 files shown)


src/languages/scala.js
```diff
@@ -151,13 +151,44 @@ export default function(hljs) {
...
+  // sourced from https://github.com/scala/docs.scala-lang/pull/2845
+  const DIRECTIVE_KEY = {
+    className: 'type',
+    begin: /\S+/,
+  }
+
+  // directives
+  // sourced from https://github.com/scala/docs.scala-lang/pull/2845
+  const USING_DIRECTIVE = hljs.COMMENT('//>', '\n', {
+    contains: [
```
joshgoebel, line 171: I think this would be simpler using a multi-mother for the first part, then contains for just the DIRECTIVE_VALUES... example of multi-match:

https://github.com/highlightjs/highlight.js/blob/main/src/languages/wren.js#L109

joshgoebel: Either starting from comment or perhaps starting from `using` if truly the whole line is still ALSO considered a comment...

bishabosha: thanks for the tip! I used begin as a match instead

joshgoebel (approved):

PR #3813, opened 2023-07-02
enh(nsis) Add !assert compiler flag

idleberg:
### Changes
- adds `!assert` command, as introduced by NSIS v3.09 (see [changelog](https://nsis.sourceforge.io/Docs/AppendixF.html#v3.09))

### Checklist


joshgoebel (approved):
Looks good! Thanks!

PR #3818, opened 2023-07-10
enh(sql) `limit`, `ilike` keyword support

erikn69:
### Changes

`limit`, `ilike` keyword support added

### Checklist
- [ ] Added markup tests, or they don't apply here because reserved word added


joshgoebel:
These are NOT valid sources of reserved works for `sql`:

- [MySQL 3.23.x](http://www.educat.hu-berlin.de/doc/mysql-3.23/manual_Reference.html#Reserved_words)
- [MySQL 4.x](http://dev.mysql.com/doc/refman/4.1/en/reserved-words.html)
- [MySQL 5.x](http://dev.mysql.com/doc/refman/5.7/en/keywords.html)
- [MySQL 8.x](https://dev.mysql.com/doc/refman/8.0/en/keywords.html)
- [PostGreSQL 8.1](http://www.postgresql.org/docs/8.1/static/sql-keywords-appendix.html)
- [MS SQL Server 2000](http://msdn2.microsoft.com/en-us/library/aa238507)
- [MS ODBC](http://msdn2.microsoft.com/en-us/library/aa238507)
- [Oracle 10.2](http://download-west.oracle.com/docs/cd/B19306_01/server.102/b14200/ap_keywd.htm)

It was decided that SQL is only keywords from the official SQL spec... if you can show that ILIKE and LIMIT are included in the official SQL spec (rather than just commonly used) we can add them, otherwise not.

For better fidelity one should likely use a MySQL or PostrgreSQL specific grammar, not `sql`...  The grammar we ship is not meant to be comprehensive.

joshgoebel:
Some quick googling makes it appear these are not standard ANSI SQL reserved words.  If you find something showing the opposite please drop a link and this can be re-opened.

Issue #3821, opened 2023-07-16
(nsis) Incorrect highlighting of keywords


idleberg:
**Describe the issue**

I have identified several cases where the highlighting for NSIS keywords is incorrect:

- keywords should only highlight when at the beginning of the line (with or without leading whitespace)
- keyword blocks should highlight both, the opening and closing keywords 

**Which language seems to have the issue?**

NSIS

**Are you using `highlight` or `highlightAuto`?**

Neither. I'm using the Mocha tests.

**Expected behavior**

Keywords should be highlighted correctly.

joshgoebel:
@idleberg Any interest in coming back and finishing this one?  Otherwise may close this issue for age.  This seems to be an obscure language (at least based the activity level around it) and I try and limit our open issues to something reasonable, those we might expect someone to show up and be able to jump into.  

If you'd like to flesh out your two items a bit further in the issue (so that someone not familiar with NSIS could maybe pick this up) with actual examples, failing cases, etc... we might be able to leave this around a while longer.

Thoughts?

idleberg:
@joshgoebel Thanks for the reminder, I will give this another go

idleberg:
These issues were addressed in https://github.com/highlightjs/highlight.js/pull/4414

PR #3822, opened 2023-07-16
enh(nsis) improve highlighter rules and tests

idleberg:
I have identified several highlighting issues and reported them https://github.com/highlightjs/highlight.js/issues/3821. In my *memory*, these inconsistencies did not exist in earlier versions, but I could be wrong.

This PR adds some tests to identify the problems Obviously, the highlighting rules need to be updated before a merge, so I consider this PR *work in progress*.

Specifically, I tried adding block-keywords such as `!macro` / `!macroend` to the [beginKeywords](https://github.com/highlightjs/highlight.js/blob/700626f10a6acb4f4eb6bdcbe3bf618d4b8e0fa7/src/languages/nsis.js#L544) block, but this did not solve the problem. Could the exclamation mark, as a non-word character cause this problem? Also, defining `!macroend` *before* the `!macro` keyword did not have an effect. I'm looking for some direction on how to tackle these issues!

### Changes

- ✅ add test for macro definitions
- ✅ add test for runtime condition
- ✅ fix expected markup for !ifdef keyword
- ✅ fix issues caused by manual editing
- ✅ remove unused import

### Todo

- identify and write missing test cases
- fix specific highlighter issues

### Checklist

### DIFF
--- a/src/languages/nsis.js
+++ b/src/languages/nsis.js
@@ -107,7 +107,8 @@ export default function(hljs) {
     "READONLY",
     "SHCTX",
     "SHELL_CONTEXT",
-    "SYSTEM|TEMPORARY",
+    "SYSTEM",
+    "TEMPORARY",
   ];
 
   const COMPILER_FLAGS = [
@@ -348,7 +349,7 @@ export default function(hljs) {
     "ManifestSupportedOS",
     "MessageBox",
     "MiscButtonText",
-    "Name|0",
+    "Name",
     "Nop",
     "OutFile",
     "Page",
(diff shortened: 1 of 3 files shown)


joshgoebel:
> Could the exclamation mark, as a non-word character cause this problem?

Yes, if using `keywords` you need to redefine the keyword pattern, see `keywords.$pattern` in the docs (or source). Such as:

```js
    keywords: {
      $pattern: '[a-zA-Z][a-zA-Z0-9_$%!#]*',
      keyword: KEYWORDS
    },
```

joshgoebel:
Closing for inactivity, but we can reopen anytime you have time to revisit this work.

PR #3828, opened 2023-07-26
(enh) add lib/common.js to sideEffects

### LINKED ISSUE #3827: vite/rollup tree shakes lib/common.js

I investigated a few issues (e.g. https://github.com/highlightjs/vue-plugin/issues/44) reported in the `vue-plugin` repo where production builds are stripping out `import 'highlight.js/lib/common';` side effect imports 

Turns out vite (which uses rollup internally) tree shakes side effect imports if the files themselves are not marked as side effect in `package.json`

```
npm create vite@latest
npm install highlight.js
```

Then inside `src/main.ts`, add:

```
import 'highlight.js/lib/common'
import hljs from 'highlight.js/lib/core'
console.log(hljs.listLanguages())
```

In debug builds `npm run dev`, it should list all languages registered inside `highlight.js/lib/common`

However in production builds `npm run build && npm run preview`, it will print an empty array.

**Expected behavior**

Should print all languages in both debug/production builds

**Additional context**

This package.json should be updated to:

```
  "sideEffects": [
    "./es/common.js",
    "./lib/common.js",
    "*.css",
    "*.scss"
  ],
```

Trinovantes:
vite/rollup will treeshake side effect imports (whereas webpack will not) if the files themselves are not marked in package.json

Resolves #3827

### Changes

Mark `./lib/common.js` and `./es/common.js` as files with side effects

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,8 @@
   "main": "./lib/index.js",
   "types": "./types/index.d.ts",
   "sideEffects": [
+    "./es/common.js",
+    "./lib/common.js",
     "*.css",
     "*.scss"
   ],


joshgoebel:
The ES version does a direct import... that's not a side effect, is it?

Trinovantes:
Honestly I have no idea how rollup's tree shaking algorithm works. It always strips the import unless both lines are marked as side effects.

joshgoebel (approved):

Issue #1274, opened 2016-08-18
(sql) remove vendor specific SQL keywords (they can live in their own grammars)


andrestaht:
Currently [sql.js](https://github.com/isagalaev/highlight.js/blob/master/src/languages/sql.js) keywords consist of different types of SQL keywords (SQL, [PostgreSQL](https://github.com/isagalaev/highlight.js/commit/42d6ba4b133b1b5f2c87115e8f0e1bddcc735b07#diff-73a3218e3423e943035e7c4f6aa04122), [Oracle](https://github.com/isagalaev/highlight.js/commit/c915a19a00c83901f4d741d072d946b7a9c435a2#diff-73a3218e3423e943035e7c4f6aa04122) etc).
Lets extract them (PostgreSQL, Oracle) into separate languages because it is causing issues when using standard SQL.

Sannis:
@andrestaht can you post some code examples?

Kirill:
Example MS SQL. After PRINT highlight break
![2017-02-26 20_39_20](https://cloud.githubusercontent.com/assets/302940/23342485/c6637cb6-fc63-11e6-8a9a-5b1daf6001a8.png)

```sql
USE db_main;
GO
SET NOCOUNT ON;

-- Rebuild or reorganize indexes based on their fragmentation levels
DECLARE @work_to_do TABLE (
    objectid int
    , indexid int
    , pagedensity float
    , fragmentation float
    , numrows int
)

DECLARE @objectid int;
DECLARE @indexid int;
DECLARE @schemaname nvarchar(130); 
DECLARE @objectname nvarchar(130); 
DECLARE @indexname nvarchar(130); 
DECLARE @numrows int
DECLARE @density float;
DECLARE @fragmentation float;
DECLARE @command nvarchar(4000); 
DECLARE @fillfactorset bit
DECLARE @numpages int

-- Select indexes that need to be defragmented based on the following
-- * Page density is low
-- * External fragmentation is high in relation to index size
PRINT 'Estimating fragmentation: Begin. ' + convert(nvarchar, getdate(), 121) 
INSERT @work_to_do
SELECT
    f.object_id
    , index_id
    , avg_page_space_used_in_percent
    , avg_fragmentation_in_percent
    , record_count
FROM 
    sys.dm_db_index_physical_stats (DB_ID(), NULL, NULL , NULL, 'SAMPLED') AS f
WHERE
[...]

PR #3830, opened 2023-07-27
enh(sql) add limit keyword

CoffeelessProgrammer:
### Changes
Added `limit` keyword to sql syntax highlighting. Limit is commonly in SQL flavors to specify number of records/rows to be returned by a query.

### Checklist
- [X] Added markup tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ Core Grammars:
 - fix(haskell) do not treat double dashes inside infix operators as comments [Zlondrej][]
 - enh(rust) added `eprintln!` macro [qoheniac][]
 - enh(leaf) update syntax to 4.0 [Samuel Bishop][]
+- enh(sql) add `limit` keyword [CoffeelessProgrammer][]
 
 Dev tool:
 
@@ -38,6 +39,7 @@ Dev tool:
 [Bradley Mackey]: https://github.com/bradleymackey
 [qoheniac]: https://github.com/qoheniac
 [Samuel Bishop]: https://github.com/dannflor
+[CoffeelessProgrammer]: https://github.com/CoffeelessProgrammer
 
 
 ## Version 11.8.0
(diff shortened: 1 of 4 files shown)


CoffeelessProgrammer:
Feel free to squash the commits, leaving CoffeelessProgrammer as author.

joshgoebel:
Is LIMIT an ANSI SQL keyword?  I don't believe so.  If not, this will be closed as a #wontfix.

Please see #1274.

Our `sql` grammar is for ANSI SQL only, not a catch all for "common" SQL or any particular engine... if one wishes great MySQL or PostgreSQL support they should use a grammar specific to those engines, not `sql`.

CoffeelessProgrammer:
Not part of ANSI SQL, thanks for the distinction. I assumed it was for general purpose SQL.

LIMIT is part of a few popular flavors incl. SQLite, Postgres, MySQL. But as you mentioned, this probably isn't enough reason to include it.

This is subjective, but I find `LIMIT 10, 5` more intuitive, clear, concise, and flexible than equivalent expressions in other flavors, e.g. `OFFSET 10 ROWS FETCH NEXT 5 ROWS ONLY`, `WHERE ROWNUM <= 15`, `TOP 5 ROWS START AT 11`, `FROM BOOK ROWS 11 TO 15`, etc.

Related links below for future reference:

- [Is there an ANSI SQL alternative to the MYSQL LIMIT keyword?](https://stackoverflow.com/questions/595123/is-there-an-ansi-sql-alternative-to-the-mysql-limit-keyword)
- [How universal is the LIMIT statement in SQL?](https://stackoverflow.com/questions/1528604/how-universal-is-the-limit-statement-in-sql)

joshgoebel:
> but I find LIMIT 10, 5 more intuitive, clear, concise, and flexible than equivalent expressions

I'd wholly agree there, but that's not the criteria here. :-). We made this decision a while back because everyone was just cramming everything into `sql` and it's hard to know where to draw the line, so ANSI is what we decided... making it a bit of a toy grammar perhaps, but there are good alternatives.

PR #3839, opened 2023-08-12
Add 32 and 64 bit ARMv8 Registers

NAThompson:
ARMv8 provides 31 64 bit registers labelled `x0-x30`. In addition, their 32 bit low bits are labelled `w0-w30`.

Add them to the known register set in armasm.js.

See [here](https://developer.arm.com/documentation/den0024/a/ARMv8-Registers) for details.

### Checklist
- [ ] Added markup tests (need to learn how to do this!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ New Grammars:
 
 Core Grammars:
 
+- enh(armasm) added `x0-x30` and `r0-r30` ARMv8 registers [Nicholas Thompson][]
 - enh(haxe) added `final`, `is`, `macro` keywords and `$` identifiers [Robert Borghese][]
 - enh(haxe) support numeric separators and suffixes [Robert Borghese][]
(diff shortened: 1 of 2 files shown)


joshgoebel:
Seems like a pretty odd omission for us to have in the library and never notice until now... Are these registers only in newer chips?

joshgoebel:
>  I believe that github uses this library, 

It does not so far as I am aware.

joshgoebel (approved):

CHANGES.md
```diff
@@ -18,6 +18,7 @@ New Grammars:
 
 Core Grammars:
 
+- enh(armasm) added `x0-x30` and `r0-r30` ARMv8 registers [Nicholas Thompson][]
```
joshgoebel, line 21:
```suggestion
- enh(armasm) added `x0-x30` and `w0-w30` ARMv8 registers [Nicholas Thompson][]
```

joshgoebel (approved):

PR #3853, opened 2023-09-02
fix(css) fix overly greedy pseudo class matching

### LINKED ISSUE #3815: (css) Wrong highlight :first-of-type :first-child
**Describe the issue**
Wrong highlight :first-of-type :first-child, "-of-type" "-child" is not highlighted

CSS

highlightElement

**Sample Code to Reproduce** 
chatGPT
hljs

**Expected behavior**
Correct highlighting :)

**Additional context**

bradleymackey:
Resolves #3815 

### Changes
- Ensure CSS attributes/elements are both sorted, then reversed to ensure matching does not greedily match shorter strings.
- `ATTRIBUTES` already used this trick, but we should sort as well to ensure that an ordering issue does not affect this behaviour.

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/lib/css-shared.js
+++ b/src/languages/lib/css-shared.js
@@ -118,6 +118,9 @@ export const TAGS = [
   'video'
 ];
 
+// Sorting, then reversing makes sure longer attributes/elements like
+// `font-weight` are matched fully instead of getting false positives on say `font`
+
 export const MEDIA_FEATURES = [
   'any-hover',
   'any-pointer',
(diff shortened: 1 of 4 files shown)


src/languages/lib/css-shared.js
```diff
@@ -153,7 +156,7 @@ export const MEDIA_FEATURES = [
   'max-width',
   'min-height',
   'max-height'
-];
+].sort().reverse();
```
joshgoebel, line 159: SHouldn't our lists already be sorted, or is this just a failsafe?

bradleymackey: It's a failsafe: ideally this should be a custom lint or something, but this works for now.

joshgoebel (approved):

PR #3856, opened 2023-09-04
enh(perl) add support for the new class system

### LINKED ISSUE #3855: (perl) add support for the new class system `Cor`.
A new class system (`Cor`) was introduced in Perl 5.38 and additional keywords/logic were added with it.
With that, an enhancement to the Perl language highlight is needed.

An example of the new system is as follows:
```perl
use v5.38;

use Object::Pad;

class Example::Subclass :isa(Example::Base 2.345) {
    field $_config :param :reader :writer;
    field $_name : param = 'Test';

    method test() {
        $_name == 'Test' ? 'y' : 'n';
    }
}
```
The keywords `class`, `field` and `method` are new ones and these accept attributes in the form `:<attr>` and also versioning by adding a number in front of the class title/name. All of these are unsupported by highlight.js yet.

bmeneg:
This PR brings the highlight support for the new Perl class system (`Cor`) by adding the keywords `class`, `method` and `field` and their supported additional features, like attributes and version number.

For getting it done, the regex for matching numbers were reworked because floating points were not fully supported alongside the use of `_` in between number digits.

Resolves #3855


src/languages/perl.js
```diff
@@ -258,19 +261,44 @@ export default function(hljs) {
...
+    ],
+    contains: [ ATTRS ],
+  };
+  const NUMBER = {
+    className: 'number',
+    variants: [
+      // in Perl numbers can have '_' in the middle, but not at the end,
+      // so we lookbehind the '\b' for a non-trailing '_'.
+      // decimal numbers
+      { match: /0?\.[0-9][0-9_]+(?<!_)\b/ },
```
joshgoebel, line 292: We can't use look-behind yet, it's too newly supported in Safari. Sorry.

src/languages/perl.js
```diff
@@ -258,19 +261,44 @@ export default function(hljs) {
...
-      // negative look-ahead tries to avoid matching patterns that are not
-      // Perl at all like $ident$, @ident@, etc.
-      `(?![A-Za-z])(?![@$%])`
-    ) },
-    {
-      begin: /[$%@][^\s\w{]/,
-      relevance: 0
-    }
-  ] };
+  const ATTRS = {
```
joshgoebel, line 264: For consistency wouldn't this be `ATTR` (singular)?

src/languages/perl.js
joshgoebel: Would these work mostly OK if we just removed this check?

bmeneg: Hmm, that's not considered an error at runtime, just a warning, so removing the lookbehind should be fine, or as you said "mostly ok" for now :) I could change the regex to `0?\.[0-9][0-9_]+[^_\W]\b` to something similar, but lets just remove it, as it's just a warning in the linter.

bmeneg: Ah, but if you're suggesting to completely remove the `match`, it gets the specific case of a number starting with a dot, eg. `.32` is translated to `0.32` into the language. Keeping the match separated guarantees the period will be highlighted as a single number, like `4.23` would.

I could certainly include it in the second match, but it would get quite confusing to understand. I'll improve the comments there for future ref.

src/languages/perl.js
bmeneg: New changes pushed.

src/languages/perl.js
```diff
@@ -258,19 +261,44 @@ export default function(hljs) {
...
-    ) },
-    {
-      begin: /[$%@][^\s\w{]/,
-      relevance: 0
-    }
-  ] };
+  const ATTR = {
+    className: 'attribute',
+    begin: /\s+:\s*\w+/,
+    end: /(\\s*\\(.*?\\))?/,
```
joshgoebel, line 267: What is this trying to match?  Is an end even necessary here if the `\w+` will match the whole attribute?

src/languages/perl.js
```diff
@@ -258,19 +261,44 @@ export default function(hljs) {
...
+  };
+  const NUMBER = {
+    className: 'number',
+    variants: [
+      // decimal numbers:
+      // include the case where a number starts with a dot (eg. .9), and
+      // the leading 0? avoids mixing the first and second match on 0.x cases
+      { match: /0?\.[0-9][0-9_]+\b/ },
+      // include the special versioned number (eg. v5.38)
+      { match: /\bv?(0|[1-9][0-9_]*(\.[0-9_]+)?|[1-9][0-9_]*)\b/ },
```
joshgoebel, line 294: Does the `v` variant only work after `use`?

src/languages/perl.js
bmeneg: First, the current version of the regex is already bogus, I changed from using single quotation mark to slashes and didn't change the escaping char from within the regex.
But, the idea here is that I was expecting to have the `attribute` class to span the entire attribute and its possible parameters, for instance, if we want a class to inherit a versioned class we can have:
```perl
class Example::Subclass :isa(Example::Base 2.345) { ... }
```
so the attribute would be the whole `:isa(Example::Base 2.345)` and not just `:isa`. So I was expecting to have something like:
```html
<span class="hljs-attribute"> :isa(Example::Base 2.345)</span> 
```
Instead of the attribute span get only `:isa` than the others be a normal "title" and a normal "number" (like I wrongly put in the markup test file :facepalm:)

src/languages/perl.js
bmeneg: Well observed. Yes, that's specific to `use`. However, thinking a bit more on that, it would be better to remove any matching with number, as it's not actually a number, but a package name. I want to remove the current behavior of highlighting half the number (`.38` is highlighted) and have no highlighting at all:
```perl
use v5.38;
```
But I'm not sure how to do it.

bmeneg: At first, I thought so, but later I realized classes can also be versioned with the `v` in front of the number, like `class MyClass v1.0.0 {..}` and, consequently, these versioned classes can be inherited with `:isa(MyClass v1.0.0)` (`:isa` is an attribute).

test/markup/perl/class.expect.txt
```diff
@@ -0,0 +1,10 @@
+<span class="hljs-keyword">use</span> <span class="hljs-number">v5.38</span>;
+
+<span class="hljs-class"><span class="hljs-keyword">class</span> <span class="hljs-title">Example</span>::<span class="hljs-title">Subclass</span><span class="hljs-attribute"> :isa(Example::Base 2.345)</span> </span>{
+    <span class="hljs-keyword">field</span> $_config<span class="hljs-attribute"> :param</span><span class="hljs-attribute"> :param</span><span class="hljs-attribute"> :param</span>;
+    <span class="hljs-keyword">field</span> $_name<span class="hljs-attribute"> : param</span> = <span class="hljs-string">&#x27;Test&#x27;</span>;
+
+    <span class="hljs-function"><span class="hljs-keyword">method</span> <span class="hljs-title">test</span>() </span>{
+        $_name == <span class="hljs-string">&#x27;Test&#x27;</span> ? <span class="hljs-string">&#x27;y&#x27;</span> : <span class="hljs-string">&#x27;n&#x27;</span>;
```
joshgoebel, line 8: Would we want to tag the identifier portion of these as variables?  Ie, `$_name` ?

src/languages/perl.js
joshgoebel: We can leave it and see how it works.

test/markup/perl/class.expect.txt
bmeneg: Oh! I completely missed it. Yes! It should be tagged. I'll get it done.

bmeneg: So, it turns out that all tests were ignoring the presence of variables due to the missing `variable` scope and changing it also required to change Mojolicious (`src/languages/mojolicious.js`) test case, which also had a conflict with the `%=` tag. I pushed 2 new commits to get variables to work.

bmeneg:
I was not able to reproduce the "Node.js CI/build" failure locally, but looking at the logs I'm not sure it's something related to the changes. At the same time, the Lint error also occurs in the `main` branch, meaning it's out-of-scope of this PR.

What am I missing?

joshgoebel:
Yeah, something isn't happy... I'll try to find some time to look at it soon.

joshgoebel (approved):

joshgoebel:
Thanks for the help!

Issue #3823, opened 2023-07-17
(CSS) custom property `--_margin-top: 1rem`, `top` gets wrongly highlighted


miragecraft:
**Describe the issue**

In CSS,  custom property such as `--_margin-top: 1rem`, `top` gets wrongly highlighted

It seems to be caused by some kind of interaction between the underscore and hyphen.

**Which language seems to have the issue?**

CSS, I'm using `language="css"` and NOT using auto-detect.

**Sample Code to Reproduce**

```
.example {
    --_margin-top: 1rem;
}
```

**Expected behavior**

`--_margin-top` should be treated as an atomic unit of text for highlighting purposes.

joshgoebel:
`-` is a word boundary, so the rule matches:

```
      {
        className: 'attribute',
        begin: '\\b(' + css.ATTRIBUTES.join('|') + ')\\b'
      },
```

One fix here might be to use a multi-matcher. (example: https://highlightjs.readthedocs.io/en/latest/mode-reference.html#beginscope). You'd first match `\s` (with no scope) and then match the attribute... that's the closest we can currently get to a negative look-behind, which we don't support yet because Safari is still slow to that party.

A PR would be welcome here if you hav time.

akhtarmdsaad:
I am Workin on it

> `-` is a word boundary (`\b`), so the rule matches:
> 
> ```
>       {
>         scope: 'attribute',
>         match: '\\b(' + css.ATTRIBUTES.join('|') + ')\\b'
>       },
> ```
> 
> One fix here might be to use a multi-matcher. (example: https://highlightjs.readthedocs.io/en/latest/mode-reference.html#beginscope). You'd first match `\s` (with no scope) and then match the attribute... that's the closest we can currently get to a look-behind, which we don't support yet because Safari is still slow to that party.
> 
> Once we could use lookbehind you'd just do a lookbehind for "not space" instead of `\b`...
> 
> A PR would be welcome here if you hav time.

I am working in it

akhtarmdsaad:
I am working on it

akhtarmdsaad:
There is a bug in src/languages/lib/css-shared.js file. adding an underscore( _ ) fixes the problem.

Harsh3363:
Hi @joshgoebel I would like to solve this issue please assign this to me

joshgoebel:
We don't usually do assignments, but I don't think anyone is rushing to complete this before you do.  Happy to see a PR if you contribute one.

[...]

PR #3863, opened 2023-09-19
Fix/issue 3823

akhtarmdsaad:
# Fixed CSS variable detection Syntax

It resolves the issue #3823  

### Changes

There is a bug in src/languages/lib/css-shared.js file. 
adding an underscore( _ ) in CSS_VARIABLES fixes the problem.
Code Before:
<code>
CSS_VARIABLE: {
      className: "attr",
      begin: /--[A-Za-z][A-Za-z0-9_-]*/  
    }
</code>
image

Code After:
<code>
CSS_VARIABLE: {
      className: "attr",
      begin: /--[A-Za-z_][A-Za-z0-9_-]*/  
      //Look change in ^ here
    }
</code>
after

### Checklist

### DIFF
--- a/src/languages/lib/css-shared.js
+++ b/src/languages/lib/css-shared.js
@@ -38,7 +38,7 @@ export const MODES = (hljs) => {
     },
     CSS_VARIABLE: {
       className: "attr",
-      begin: /--[A-Za-z][A-Za-z0-9_-]*/
+      begin: /--[A-Za-z_][A-Za-z0-9_-]*/
     }
   };
 };
--- a/test/markup/css/variables.txt
+++ b/test/markup/css/variables.txt
@@ -7,3 +7,9 @@ body {
  --textBlue: blue;
  color: var(--textBlue);
 }
+
+body {
+ --_margin-top: 1rem;
+ margin-top: var(--_margin-top);
+}
+
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,6 +41,7 @@ Core Grammars:
 - fix(swift) `warn_unqualified_access` is an attribute [Bradley Mackey][]
 - enh(swift) macro attributes are highlighted as keywords [Bradley Mackey][]
 - enh(stan) updated for version 2.33 (#3859) [Brian Ward][]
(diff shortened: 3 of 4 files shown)


test/markup/css/variables.txt
```diff
@@ -7,3 +7,9 @@ body {
  --textBlue: blue;
  color: var(--textBlue);
 }
+
+body {
+ --_margin-top: 1rem;
+ margin-top: var(--_margin-top);
+}
+
```
joshgoebel, line 15: You'll also need to update the matching `variables.expect.txt` file with the correct output.

src/languages/cpp.js
```diff
@@ -62,7 +62,7 @@ export default function(hljs) {
   const PREPROCESSOR = {
     className: 'meta',
     begin: /#\s*[a-z]+\b/,
-    end: /$/,
+    end: /\s/,
```
joshgoebel, line 65: This cpp change will need to be removed from the CSS PR.

CONTRIBUTORS.md
```diff
@@ -25,3 +25,4 @@
 ### Individual Contributors
 
 Highlight.js has also been greatly improved over the years thanks to the help of [many other contributors](https://github.com/highlightjs/highlight.js/graphs/contributors).  A big thank you to everyone who has helped make this project what it is today!
+- Md Saad Akhtar <akhtarmdsaad@gmail.com>
```
joshgoebel, line 28: Please instead add yourself to the changelog around line ~60... the only contributors mentioned here are former maintainers, everyone else can be found in the GitHub contributor graph.  We no longer maintained a textual list since Github does this for us.

```suggestion
```

src/languages/lib/css-shared.js
```diff
@@ -38,7 +38,7 @@ export const MODES = (hljs) => {
     },
     CSS_VARIABLE: {
       className: "attr",
-      begin: /--[A-Za-z][A-Za-z0-9_-]*/
+      begin: /--[A-Za-z_][A-Za-z0-9_-]*/
```
joshgoebel, line 41: Looking good!

joshgoebel (approved):

Issue #3505, opened 2022-03-21
(C++) highlight incorrectly


zerolfx:
**Describe the issue**

![Screenshot_20220320_225211](https://user-images.githubusercontent.com/20219056/159199285-fd9e499e-7c90-476b-808b-d0f5f228522b.png)

Everything after the define line has no highlighting.

**Which language seems to have the issue?**

C++

**Are you using `highlight` or `highlightAuto`?**

highlight.js 11.5.0

**Sample Code to Reproduce**

```cpp
int a = 1;

#define what do { cout << ">"; } while (0)

int main() {
}
```

**Expected behavior**

Highlight correctly after the define line.

adobemomo:
I found this bug introduced at #3041. `<.*>` matches `<< ">` and leaves `"` alone. All preprocessor keywords are treated as meta-keywords. But only `#include` needs to be followed with angle bracket quoted string. `PREPROCESSOR` in `cpp.js/c.js` might be subdivided to deal with different scenarios. Or simply change `<.*>` into `<[^"]*>`. The latter would contradict the lexical rule mentioned in [http://eel.is/c++draft/lex.header](url), though I think people won't use `"` in their header file name...

joshgoebel:
You may be right that this needs to be scoped to `#include`...

joshgoebel:
Either of your willing to submit a PR for this?

PR #3865, opened 2023-09-20
Fix highlight problem in preprocessor

akhtarmdsaad:
The highlighting of preprocessor goes wrong. Everything after the define line has no highlighting.

This PR resolves the Issue #3505 

### Changes

There was a simple change in cpp.js file.
path: `src/languages/cpp.js`

<b>Before:</b>
<code>
const PREPROCESSOR = {
    className: 'meta',
    begin: /#\s*[a-z]+\b/,
    end: /$/, <----------------------------- see this
    keywords: { keyword:
        'if else elif endif define undef warning error line '
        + 'pragma _Pragma ifdef ifndef include' },
    contains: ......
</code>
image

<b>After</b>
<code>
const PREPROCESSOR = {
    className: 'meta',
    begin: /#\s*[a-z]+\b/,
    end: /\s/, <----------------------- the change
    keywords: { keyword:
        'if else elif endif define undef warning error line '
        + 'pragma _Pragma ifdef ifndef include' },
    contains: ......
</code>

image

<b>*The header files <....> are treated as html elements and hence not shown </b>

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,6 +38,7 @@ New Grammars:
 - added 3rd party Luau grammar to SUPPORTED_LANGUAGES [Robloxian Demo][]
 - added 3rd party ReScript grammar to SUPPORTED_LANGUAGES [Paul Tsnobiladzé][]
 - added 3rd party Zig grammar to SUPPORTED_LANGUAGES [Hyou BunKen][]
+- added 3rd party WGSL grammar to SUPPORTED_LANGUAGES [Arman Uguray][]
 
 Developer Tool:
 
@@ -69,6 +70,7 @@ Themes:
 [Lisa Ugray]: https://github.com/lugray
 [TaraLei]: https://github.com/TaraLei
 [Vasily Polovnyov]: https://github.com/vast
+[Arman Uguray]: https://github.com/armansito
 
 
 ## Version 11.9.0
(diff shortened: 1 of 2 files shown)


akhtarmdsaad:
Should I add the contains property or end the preprocessor after "\s" ? This ending doesn't seems to me a proper way so adding the contains might be a good approach. can you suggest me

joshgoebel:
If this does fix 3505, it's a very "round-about" fix... the stated issue there is that *lines* following a pre-processor line are not highlighted at all (because of how we handling templating), not that the content *on the same line* immediately following a pre-processor directive is not highlighted.

The current behavior seems an intentional choice.  True, it's not what VS Code does, but seems unclear to me that one way is superior.  The text that follows any random compiler directive is not guaranteed to be C code, and thus using those same rules to highlight it (as normal C code) seems quite wrong in some ways.

I'll leave this open for now to see if it draws any further discussion/opinions.

joshgoebel:
Github highlighting (just for curiosity):

```cpp
// this is a continued\
comment.
#include <iostream>
#include <iostream>
#define what do{ cout << "Hello"; } while(0);
#define PI 3.14;
#pragma warning(disable : 1234) 
#if !defined(CONFIG_OPTION)
#error "CONFIG_OPTION is not defined."
#endif
#line 42 "mycode.cpp"
#undef PI
```

PR #3871, opened 2023-09-22
fix: fixed #3870 Function names with a prefix shared with a built-in …

### LINKED ISSUE #3870: (JavaScript) Function names with a prefix shared with a built-in global are confused for a global
JavaScript and TypeScript

`highlight`

**Sample Code to Reproduce**

```js
eval();
neval();
eval2();
```
Screenshot 2023-09-22 at 07 25 08

**Expected behavior**

An identifier that starts with the prefix of a built-in global should not be recognized as that built in global. For example, the identifiers `eval2`, `evaluate`, `evaluation` are all confused with the built-in global `eval` and highlighted incorrectly.

The expected highlithing:

Screenshot 2023-09-22 at 07 25 49

arnog:
…global are confused for a global

Don't exclude a built-in global (e.g. `eval`) followed by an identifier , but exclude built-in global followed by  an open paren (i.e. `eval(`)

See #3870 for an example.

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -384,7 +384,7 @@ export default function(hljs) {
     match: regex.concat(
       /\b/,
       noneOf([
-        ...ECMAScript.BUILT_IN_GLOBALS,
+        ...ECMAScript.BUILT_IN_GLOBALS.map((x) => x + '\\('),
         "super",
         "import"
       ]),


src/languages/javascript.js
```diff
@@ -384,7 +384,7 @@ export default function(hljs) {
     match: regex.concat(
       /\b/,
       noneOf([
-        ...ECMAScript.BUILT_IN_GLOBALS,
+        ...ECMAScript.BUILT_IN_GLOBALS.map((x) => x + '\\('),
```
joshgoebel, line 387: Don't we need this same protection for super and import also?  And wouldn't just adding a `\b` on the other side of `noneOf` get the job done here without explicitly looking for a `(`?

arnog: yes, there might be more bugs in the way the parser handles keywords and other keywords affected. I haven't tested that and I'm not sure of what the logic behind the original code was, but your suggestion seems reasonable. 

Same thing for using `\b` instead of `(`, again I'm not sure of what the original logic was so I tried to find a minimal solution that would fix the test case I encountered.

joshgoebel: Happy to review again after you update.

arnog: That PR was just a suggestion. I'll let someone with a better understanding of the codebase to contribute a fix.

PR #3875, opened 2023-10-02
enh(llvm) Match additional types

wangtz0607:
Match the following types according to the [LLVM Language Reference Manual](https://llvm.org/docs/LangRef.html):

```
void
iN
half
bfloat
float
double
fp128
x86_fp80
ppc_fp128
x86_amx
x86_mmx
ptr
label
token
metadata
opaque
```

### Changes

`src/languages/llvm.js` is updated. Tests in `test/markup/llvm` are also updated accordingly.

### Checklist


src/languages/llvm.js
```diff
@@ -13,7 +13,7 @@ export default function(hljs) {
   const IDENT_RE = /([-a-zA-Z$._][\w$.-]*)/;
   const TYPE = {
     className: 'type',
-    begin: /\bi\d+(?=\s|\b)/
+    begin: /\b(?:void|i\d+|half|bfloat|float|double|fp128|x86_fp80|ppc_fp128|x86_amx|x86_mmx|ptr|label|token|metadata|opaque)(?=\s|\b)/
```
joshgoebel, line 16: Is there a reason we couldn't use keyword matching for most of these rather than a long regex?

wangtz0607: It would be great if types other than `i\d+` can be matched without using a long regex, but I'm not sure how to do it.

joshgoebel:
See the docs for `keyword`, you can pass a hash instead with multiple arrays of keywords for every key... so you'd just add a `type` key to `keywords`... lots of examples in the source code as well.

wangtz0607:
Done.

joshgoebel (approved):

joshgoebel (approved):

joshgoebel (approved):

PR #3888, opened 2023-10-15
add (markdown) emphasis escaping

Crater44:
## This feature adds the possibility of escaping italic on markdown

Before the only way to write a word with underscore (e.g. SYS_ADMIN) was placing a space after the underscore, now it's possible to escape it with a backslash so it won't trigger the italic feature.

## Change
The change consist on combining the actual negative look-ahead for a space with a negative look-behind for a backslash

### DIFF
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -152,11 +152,11 @@ export default function(hljs) {
     contains: [], // defined later
     variants: [
       {
-        begin: /\*(?![*\s])/,
+        begin: /(?<!\\)\*(?![*\s])/,
         end: /\*/
       },
       {
-        begin: /_(?![_\s])/,
+        begin: /(?<!\\)_(?![_\s])/,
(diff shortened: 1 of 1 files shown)


joshgoebel:
Sadly, we cannot use negative look behind yet as that feature is still too newly added to Safari.

joshgoebel:
Actually, hold that thought, perhaps it's time to change this policy: https://github.com/highlightjs/highlight.js/issues/3890

Crater44:
@joshgoebel great so if my change can be added that would be great to me because I have it patched in one of my projects and is not the best way to have it

joshgoebel:
No lookbehind for now as it would be a breaking change... why not just have a single (separate) rule that matches `\*` but then chooses not to highlight it?  Would that hurt anything else?

joshgoebel:
Ping.

Issue #3798, opened 2023-05-29
(Python) or"string" highlights the raw string instead of `or` operator


IamMusavaRibica:
**Describe the issue**
Incorrect highlighting of python `or` operator preceded by string without whitespace, highlight.js treats the string as a raw string (with `r` prefix)

**Which language seems to have the issue?**
Python

**Are you using `highlight` or `highlightAuto`?**
highlight

**Sample Code to Reproduce**
https://jsfiddle.net/beh3j1nw/

**Expected behavior**
This is valid Python code
```py
print(test()or'string')
```
![image](https://github.com/highlightjs/highlight.js/assets/65848142/c7c53741-7793-4478-8440-51e3518ce790)

joshgoebel:
Probably just a matter of adding a `\bor` rule prior to the string rule... or would it be simpler to add `\b` to the string prefixes perhaps, forcing them to stand on their own?

MohamedAli00949:
I have read the Python highlighting script and found the solution: string highlighting the prefixes such as (r, u, b, f) and I will fix it now.
After solving it, I will submit my PR.

joshgoebel:
Solved by #3897.

PR #3897, opened 2023-10-21
fix(python) fix `or` conflicts with string highlighting

MohamedAli00949:
I have fixed the error of highlighting the reserved words such as 'or' in the below code.
```
print(test()or'string')
```
That should be in the below image.
image
Not as this highlight.
image

That happens because of the string prefixes (f, r, u, and b) or (F, R, U, and B).

This issue has been described at #3798

### Changes
I have found that happens from string variants begin RegExes with the prefix `(\W|^)`.

### Checklist
- [x] Update the string variants begin RegExes with the prefix `(\W|^)`.

### DIFF
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -176,7 +176,7 @@ export default function(hljs) {
     contains: [ hljs.BACKSLASH_ESCAPE ],
     variants: [
       {
-        begin: /([uU]|[bB]|[rR]|[bB][rR]|[rR][bB])?'''/,
+        begin: /(\W|^)([uU]|[bB]|[rR]|[bB][rR]|[rR][bB])?'''/,
         end: /'''/,
         contains: [
           hljs.BACKSLASH_ESCAPE,
@@ -185,7 +185,7 @@ export default function(hljs) {
         relevance: 10
       },
       {
-        begin: /([uU]|[bB]|[rR]|[bB][rR]|[rR][bB])?"""/,
+        begin: /(\W|^)([uU]|[bB]|[rR]|[bB][rR]|[rR][bB])?"""/,
         end: /"""/,
         contains: [
           hljs.BACKSLASH_ESCAPE,
@@ -194,7 +194,7 @@ export default function(hljs) {
         relevance: 10
       },
       {
-        begin: /([fF][rR]|[rR][fF]|[fF])'''/,
+        begin: /(\W|^)([fF][rR]|[rR][fF]|[fF])'''/,
         end: /'''/,
         contains: [
           hljs.BACKSLASH_ESCAPE,
@@ -204,7 +204,7 @@ export default function(hljs) {
         ]
       },
       {
-        begin: /([fF][rR]|[rR][fF]|[fF])"""/,
+        begin: /(\W|^)([fF][rR]|[rR][fF]|[fF])"""/,
         end: /"""/,
         contains: [
           hljs.BACKSLASH_ESCAPE,
@@ -214,25 +214,25 @@ export default function(hljs) {
         ]
       },
       {
-        begin: /([uU]|[rR])'/,
+        begin: /(\W|^)([uU]|[rR])'/,
         end: /'/,
         relevance: 10
       },
       {
-        begin: /([uU]|[rR])"/,
+        begin: /(\W|^)([uU]|[rR])"/,
         end: /"/,
         relevance: 10
       },
       {
-        begin: /([bB]|[bB][rR]|[rR][bB])'/,
+        begin: /(\W|^)([bB]|[bB][rR]|[rR][bB])'/,
         end: /'/
       },
       {
-        begin: /([bB]|[bB][rR]|[rR][bB])"/,
+        begin: /(\W|^)([bB]|[bB][rR]|[rR][bB])"/,
         end: /"/
       },
       {
-        begin: /([fF][rR]|[rR][fF]|[fF])'/,
+        begin: /(\W|^)([fF][rR]|[rR][fF]|[fF])'/,
         end: /'/,
         contains: [
           hljs.BACKSLASH_ESCAPE,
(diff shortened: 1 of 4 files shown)


src/languages/python.js
```diff
@@ -176,7 +176,7 @@ export default function(hljs) {
     contains: [ hljs.BACKSLASH_ESCAPE ],
     variants: [
       {
-        begin: /([uU]|[bB]|[rR]|[bB][rR]|[rR][bB])?'''/,
+        begin: /(\W|^)([uU]|[bB]|[rR]|[bB][rR]|[rR][bB])?'''/,
```
joshgoebel, line 179: Aren't you really just expressing "word boundary" here... and can't that be done much simpler using `\b`?

MohamedAli00949: I think it's doing the same thing here. 
But it didn't come first to my mind. 
Thanks for your advice and I will replace `(\W|^)` with it.

MohamedAli00949: @joshgoebel please take another look at my changes.
Many thanks for considering my request.

joshgoebel:
We need to add some `markup` tests as well to prove this actually works as intended.

MohamedAli00949:
I have tested the code with `(\W|^)` and passed all checks 
But after I replaced it with `\b` it didn't pass the python and I didn't discover that at the time.
I have tried to fix the tests but I find it difficult to understand them now so I want any resource (eg: docs, or videos) to understand them easily and fix the tests.

MohamedAli00949:
Now, I have discovered the issue with this code.
The issue is that the word bounding expression selects everything from the beginning to the end as a string and the contains also.
In this sample, you will see that the prompt '...' has been selected as a string not as meta.
image

That's the right show
image

I tried to solve it until now it didn't solve it.
My suggested solution is to contact the contains expirations with the beginning expression or go back to use the `(\W|^)` expression.

MohamedAli00949:
Hey @joshgoebel, what is your opinion I understand the test mockups and I think we should put some of them such as `print(test()or'text')` and try that with every string prefix. 
I want your opinion on fixing and testing my changes bugs.

Thanks before any thing

joshgoebel:
An example of every string type could work... I assume we'd put that in a test called `strings` or some such if we don't have that already.  As far as the exact regex I'll take another look at the smyself once we have all the test cases.

MohamedAli00949:
Hi @joshgoebel, I have added some markup tests to test my changes and fixed selecting the char before the string prefix as a car such as highlighting the 'plus operator(+)' at the following example `print('kj'+f'kj')`.
Take a look and tell me if you require any other changes.

Many thanks for considering my request.

joshgoebel:
In your test you have:

`orbr"https://"`

Is this valid, an `or` followed by a string or invalid?

MohamedAli00949:
> In your test you have:
> 
> `orbr"https://"`
> 
> Is this valid, an `or` followed by a string or invalid?

I think it's invalid, and `orrb"https://"` also.
I want to test `or` with `br` or `rb` prefixes and I have tested them at `or rb"https://"` and `or br"https://"`.
Thanks for the good note

joshgoebel:
What is an `rrf` string?

MohamedAli00949:
> What is an `rrf` string?

i think its was for testing if the three letters prefer will be highlighted or highlighting the right string.

joshgoebel:
I meant is `rrf` valid python?

MohamedAli00949:
> I meant is `rrf` valid python? Orr `rff`

The `rrf` is not valid python and I have used it to test the regex of begin.
The result does not highlight the prefix and starts highlighting the string from the quotes only.

MohamedAli00949:
Hi @joshgoebel, 
I know that you have done hard work with my PRs and all project PRs and I appreciate that. 
But I don't know what I can do now to merge my PR, fix more issues, and do more on this nice project.

Thanks in advance

joshgoebel:
> The result does not highlight any of them and starts highlighting the string from the quotes only.

It's totally OK to include some false positives in the tests (with a comment to explain what purpose they are serving) - but any false positives should still be valid Python code.  We only care about how well we highlight actual real-life valid Python.

MohamedAli00949:
> > The result does not highlight any of them and starts highlighting the string from the quotes only.
> 
> It's totally OK to include some false positives in the tests (with a comment to explain what purpose they are serving) - but any false positives should still be valid Python code. We only care about how well we highlight actual real-life valid Python.

You convinced me, thanks ❤

MohamedAli00949:
> > The result does not highlight any of them and starts highlighting the string from the quotes only.
> 
> It's totally OK to include some false positives in the tests (with a comment to explain what purpose they are serving) - but any false positives should still be valid Python code. We only care about how well we highlight actual real-life valid Python.

I have committed the right tests now, feel free to review and give me your good advice.

MohamedAli00949:
Hi @joshgoebel, 
I know that I have talked more and more to you and I hope to answer my request.
If you have any time please look at the last changes at this request and the last 2 PRs I have pulled before Hecktoperfest finishing.

MohamedAli00949:
Hi @joshgoebel, Any news about reviewing the changes?

MohamedAli00949:
Hi @joshgoebel, If I have removed some of the test commits and collected the testing at one commit would that be best?

joshgoebel:
If we just match `or` first - all your tests still pass... do you have additional failing tests you'd like to come up with?

MohamedAli00949:
> If we just match `or` first - all your tests still pass... do you have additional failing tests you'd like to come up with?

Firstly, I am sorry for replying late, I didn't see your comment until now, and thanks for your changes.
Secondly, I think I didn't have any other tests.

src/languages/python.js
```diff
@@ -214,7 +214,7 @@ export default function(hljs) {
         ]
       },
       {
-        begin: /([uU]|[rR])'/,
+        begin: /(([uU]|[rR]))'/,
```
joshgoebel, line 217:
```suggestion
        begin: /([uU]|[rR])'/,
```

joshgoebel:
Please add a changelog entry then and we should be good to go.

joshgoebel (approved):

PR #3899, opened 2023-10-22
Yaml fix

### LINKED ISSUE #3894: (YAML) keys containing brackets highlighted wrongly
**Describe the issue**
Highlight.js library seems not to be able to color YAML keys properly, when there is brackets ( or ) inside YAML keys.

YAML

highlight

Calling it out like so in Vue 3 project:
`hljs.highlight(data.value, {language: 'yaml'}).value`

Also reproduces in demo app:
https://highlightjs.org/demo

...

**Sample Code to Reproduce**
1. Go to https://highlightjs.org/demo
2. language YAML, random theme
3. code:
field1: "1"
field2: "2"
field3(broken): "3"
"field4(broken)": "4"
field5: "5"

**Expected behavior**
field3(broken) and "field4(broken)" are colored same way as field1, field2, field5

**Additional context**
image

aneesh98:
Resolves #3894 
### Changes

This is related to yaml highlighting
1. Corrected highlighting for keys having brackets in them.
2. For Quoted Versions added 2 new variants.

### Checklist


src/languages/yaml.js
```diff
@@ -20,11 +20,16 @@ export default function(hljs) {
...
+      // added brackets support 
+      { begin: '\\w[\\w :\\(\\)\\/.-]*:(?=[ \t]|$)' },
       { // double quoted keys
         begin: '"\\w[\\w :\\/.-]*":(?=[ \t]|$)' },
       { // single quoted keys
-        begin: '\'\\w[\\w :\\/.-]*\':(?=[ \t]|$)' }
+        begin: '\'\\w[\\w :\\/.-]*\':(?=[ \t]|$)' },
+      { // double quoted keys - with brackets
+        begin: '"\\w[\\w :\\(\\)\\/.-]*":(?=[ \t]|$)' },
+      { // single quoted keys - with brackets
```
joshgoebel, line 31: Now that we have these, do we need the "without brackets" variants, doesn't this cover both cases?

aneesh98: Hi @joshgoebel , My bad sorry :sweat_smile: . Have changed it

joshgoebel: Nice.  Now can we convert them to proper regex with `//` not strings and simplify all that escaping?  I'm pretty sure the `\t` escaping it already broken since it's not double escaped....

aneesh98: Hey @joshgoebel

Just as a matter of interest, I checked if \t is broken or not and the \t matching is working fine I believe, even without double escaping. I added a key and put a tab after the colon and it was highlighted correctly. (Check the last key below). Let me know if I tested it correctly :sweat_smile: 
image

Also I have made the changes suggested.

Just out of curiosity I noticed there are several other places also where instead of JavaScript RegExp objects we are using Strings to represent regex.
for example
https://github.com/highlightjs/highlight.js/blob/17494c81dbade12834feedddf6fb208bcc9cbcce/src/languages/yaml.js#L78-L81

Is there any reason to prefer strings over regexp object? Just wanted to get a better understanding of the project. 
Thanks in advance.

joshgoebel: > Is there any reason to prefer strings over regexp object? 

Sometimes when we build larger regex with `+` you need strings... but we have `regex.join` also that can take regex... so before just changing everything you have to look at everywhere it's used to make sure you aren't breaking anything.

joshgoebel: >  I checked if \t is broken or not and the \t matching is working fine I believe, even without double escaping

Right it would, but in one case you have a _literal_ tab and the other you have a \t escape sequence... both will match a tab of course, hence the same behavior.

aneesh98: Alright! I got it. Thanks

aneesh98: Is there anything else that needs to be done?

joshgoebel (approved):

PR #3902, opened 2023-10-27
enh(sql) added `tables` keyword

ztomm:
### Changes
Change contains the missing keyword "TABLES" for the SQL language.

Syntax example:

```
SHOW TABLES;
```

### Checklist


joshgoebel:
Please provide documentation (a link is great) to show this is part of Standard SQL and not merely a popular extension that many DBs agree upon... the built-in SQL grammar is only for Standard SQL.  We do not add "missing" keywords just because they are used by modern DB engines.

We should probably rename this in v12 to `standard_sql.rb`.

joshgoebel:
Looks like TABLES (plural) is not a keyword in the standard.

https://en.wikipedia.org/wiki/List_of_SQL_reserved_words

Issue #3817, opened 2023-07-10
(Rust) Escaped double quotations in string at attribute are not highlighted properly


keiichiw:
**Describe the issue**

Escaped double quotations in string at attributes are not treated as a escaped character but a normal quotation.

**Which language seems to have the issue?**

Rust

**Sample Code to Reproduce**

```rust
#[derive(ThisError)]
enum MyError {
  #[error("\" appears in a string")]
  UnexpectedDoubleQuote
}

fn main() {}
```

[sample at jsfiddle](https://jsfiddle.net/1s274xfe/)

![image](https://github.com/highlightjs/highlight.js/assets/3708146/3cb74873-a66c-4e6d-a57a-487be6bb3768)

**Expected behavior**

The escaped double quotation `\"` shouldn't be treated as a normal double quotation.

joshgoebel:
Confirm.  Looks like a good first issue, probably need to reuse the full string rules inside `meta`... can ALL the syntactic features of strings be used in a `#[]` block?

mohiyaddeen7:
{
        className: 'string',
        variants: [
          { begin: /b?r(#*)"(.|\n)*?"\1(?!#)/ }, // Handles raw strings with escaped double quotes
          { begin: /b?'\\?(x\w{2}|u\w{4}|U\w{8}|.)'/ } // Handles character literals with escaped double quotes
  ]
        ]
      }

we should add (.|\\")*? to the first variant to match any character or an escaped double quote within the string.
we should add  (\\")  to the second variant to allow for escaped double quotation marks within character literals.

update code snippet : 

{
  className: 'string',
  variants: [
    { begin: /b?r(#*)"(.|\\")*?"\1(?!#)/ }, // Handles raw strings with escaped double quotes
    { begin: /b?'\\?(x\w{2}|u\w{4}|U\w{8}|.|\\")'/ } // Handles character literals with escaped double quotes
  ]
}

please correct me if i am wrong.

joshgoebel:
Very hard to review code here when just thrown in as regular text, but if you'd be game to make a PR I'd be happy to review.

PR #3904, opened 2023-10-29
Fix rust escape double quots in string

MohamedAli00949:
I have fixed unescaping the `"` at string attributes such as I have seen in the #3817 issue.

### Changes

The string scops didn't contain the backslash escape expression which is `BACKSLASH_ESCAPE` so I have put it at every string scope.

Before
image

After
rust-issue

### Checklist
- [x] Add `BACKSLASH_ESCAPE` to string scops to avoid this issue.
- [x] Add markup tests to test the validation of the solution.


MohamedAli00949:
hi @joshgoebel, this is a simple issue and feel free to review it and give me your pieces of advice.

src/languages/rust.js
```diff
@@ -193,9 +193,19 @@ export default function(hljs) {
...
+              hljs.BACKSLASH_ESCAPE
+            ]
+          },
+          {
+            begin: /b?'\\?(x\w{2}|u\w{4}|U\w{8}|.)'/,
+            contains: [
+              hljs.BACKSLASH_ESCAPE
+            ]
+          },
+        ],
```
joshgoebel, line 208: This doesn't work... you'd need `end` as well and need to rewrite any specific internals (like the unicode matching) to be sub rules inside the `contains`.  And you can't use match history (`\1`) across begin/end combos... how many `#` characters are used in practice? is there a max limit?

MohamedAli00949: I am sorry, I can't understand you.
Do you mean removing this change avoiding complexity or another thing?

joshgoebel: Please add test cases for these and you will see that they do not work they way they are currently written.

MohamedAli00949: You are right, this change didn't work and the contains at this these variants.
I will remove it and commit changes again.
Thanks in advance

joshgoebel:
Ok, I supposing fixing one of the variants s better than nothing.  Please add a changelog entry.

MohamedAli00949:
@joshgoebel, I have updated the changelogs at CHANGES.md Is that what you expect?

joshgoebel:
Please only include your single change in the log.

joshgoebel (approved):

MohamedAli00949:
@joshgoebel, Thanks for approving the changes, I hope to merge it if possible.

PR #3905, opened 2023-10-29
Fix horizontal line grammar for Markdown.

TaraLei:
According to commonmark a horizontal line starts with at most 3 spaces indentation, followed by three or more either dashes or asterisks, followed by spaces. Previously no indentation was recognized, and any combination of dash and asterisks was accepted.
See <https://spec.commonmark.org/0.12/#horizontal-rules> for more information.

### Changes
Used a more specific regular expression

### Checklist
Did not add markup test since horizontal lines are generally not highlighted.


TaraLei:
ah, I accidentally pushed another commit in this PR; I'll remove it and create a fresh PR with a single commit.

Issue #3868, opened 2023-09-20
(csharp) Raw String Literals are highlighted incorrectly


rosenzweigv:
**Describe the issue**

C# Raw String Literals are highlighted incorrectly. It seems that they are highlighted like normal strings, as instances of `"` in the middle of the strings ends the string coloring prematurely.

**Which language seems to have the issue?**

csharp

**Are you using `highlight` or `highlightAuto`?**

highlight

**Sample Code to Reproduce**
[https://jsfiddle.net/xae1v6wL/](https://jsfiddle.net/xae1v6wL/)

**Expected behavior**
See sample code -- The `"quoted text"` should look the same as the rest of the string 

**Additional context**
None

akhtarmdsaad:
is this issue resolved

joshgoebel:
Merged.

PR #3906, opened 2023-10-30
Add raw string literal highlighting for C# 11.

TaraLei:
Fixes issue #3868.

This fails the tests as the Vala default.txt is recognized now as C#. However, Vala is very close in syntax to C#, and the default.txt also seems to be valid C# so not sure what to do about this.

### Changes
Added raw string regex to the csharp definition.

### Checklist
- [x] Added markup tests


src/languages/csharp.js
```diff
@@ -164,6 +164,11 @@ export default function(hljs) {
     ],
     relevance: 0
   };
+  const RAW_STRING = {
+    className: 'string',
+    begin: /"""("*)(?!")(.|\n)*?"""\1/,
```
joshgoebel, line 169: Will `.` itself not already match `\n` here?

joshgoebel (approved):

src/languages/csharp.js
```diff
@@ -164,6 +164,11 @@ export default function(hljs) {
...
+    relevance: -1
```
joshgoebel, line 170:
```suggestion
    relevance: 1
```

Issue #3719, opened 2023-02-21
(markdown) Horizontal rule with 3 asterisks (***) causes bold text highlight


don-patterson:
**Describe the issue**
In Markdown, a line that contains only `***` is supposed to be a [horizontal rule](https://www.markdownguide.org/basic-syntax/#horizontal-rules), but highlightjs interprets it as the start of bold text.

**Which language seems to have the issue?**
markdown

**Are you using `highlight` or `highlightAuto`?**
`highlight`

**Sample Code to Reproduce**
see https://jsfiddle.net/4kwz8Ld9/ where this snippet causes the issue:
```markdown
# Hello
this should be a horizonal rule:
***
but it gets highlighed as bold text, followed by *
It's pretty minor since most people would use --- instead
***
That is another horizontal rule, but ends the bold block instead 
```

**Expected behavior**
It appears that GitHub renders it as expected (grey highlights to the `***`) instead of bolding the following characters.

**Additional context**
I only noticed this because I was fiddling with a specific format that I wanted to use in revealjs slides. Super minor issue, but figured it would be worth mentioning.

don-patterson:
just messing around locally and I think putting HORIZONTAL_RULE ahead of BOLD fixes it:
```diff
diff --git a/src/languages/markdown.js b/src/languages/markdown.js
index 7d5a1506..0a6dd182 100644
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -225,13 +225,13 @@ export default function(hljs) {
     ],
     contains: [
       HEADER,
+      HORIZONTAL_RULE,
       INLINE_HTML,
       LIST,
       BOLD,
       ITALIC,
       BLOCKQUOTE,
       CODE,
-      HORIZONTAL_RULE,
       LINK,
       LINK_REFERENCE
     ]
```

I can test it out a bit more and put up a PR

don-patterson:
ah, it's not that simple. There's kinda a race for priority between bold+italic and this horizontal rule. I might be able to untangle them some day

joshgoebel:
One could use variants and expose custom matchers for ruled lines that match say `^***$`, etc... so that you could catch the most common cases... I wonder if the current rule matcher isn't a little bit too loose though?

Mynk11:
Picking this up, since #3909 was closed for inactivity with @joshgoebel's review comments unanswered. Addressing all three of them:

[...]

PR #3909, opened 2023-10-31
Fix horizontal line grammar for Markdown

TaraLei:
According to commonmark a horizontal line starts with at most 3 spaces indentation,
followed by three or more either dashes or asterisks, followed by spaces.
Previously no indentation was recognized, and any combination of dash and asterisks
was accepted.
See <https://spec.commonmark.org/0.12/#horizontal-rules> for more information.

Resolves issue #3719 

### Changes

Created a more specific regex for horizontal rules.

### Checklist
- [x ] Added markup tests

### DIFF
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -15,8 +15,7 @@ export default function(hljs) {
     relevance: 0
   };
   const HORIZONTAL_RULE = {
-    begin: '^[-\\*]{3,}',
-    end: '$'
+    begin: '^[ ]{0,3}([-]{3,}|[\*]{3,})[ \t]*$'
   };
   const CODE = {
     className: 'code',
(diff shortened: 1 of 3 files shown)


src/languages/markdown.js
```diff
@@ -15,8 +15,7 @@ export default function(hljs) {
     relevance: 0
   };
   const HORIZONTAL_RULE = {
-    begin: '^[-\\*]{3,}',
-    end: '$'
+    begin: '^[ ]{0,3}([-]{3,}|[\*]{3,})[ \t]*$'
```
joshgoebel, line 18: - Don't we also need `_`?
- Does `\s*$` work here?
- Is the `\` needed for the `*` inside the `[]`?

joshgoebel:
Closing for inactivity.  Happy to reopen if you return to work on this further.

PR #3910, opened 2023-10-31
Add highlighting for hard line breaks in markdown.

TaraLei:
Highlights hard line breaks as an "addition" in markdown.

### Changes

Added a LINEBREAK regex case for hard line breaks.

### DIFF
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -216,6 +216,11 @@ export default function(hljs) {
     end: '$'
   };
 
+  const LINEBREAK = {
+    className: 'addition',
+    begin: /(  |\\)$/
+  };
+
   return {
     name: 'Markdown',
     aliases: [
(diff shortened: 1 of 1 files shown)


src/languages/markdown.js
```diff
@@ -216,6 +216,11 @@ export default function(hljs) {
     end: '$'
   };
 
+  const LINEBREAK = {
+    className: 'addition',
```
joshgoebel, line 220: We'd need a better, more semantic scope name... this is a weird one, so I'm not sure what to suggest.

Also I don't think we want to highlight empty space - too much of a chance for false positives in better Markdown formats that don't use this weird way of marking line breaks but where someone might have extra whitespace.  The `\` at least is less likely to suffer from false positives, so could make sense for some cases.

joshgoebel:
Closing for inactivity.  We can reopen if you come back and have time to push this forward.

PR #3918, opened 2023-11-02
fix(bash): Comments false positives

### LINKED ISSUE #3917: False positives for comments in bash
**Describe the issue**

bash

**Sample Code to Reproduce**
```sh
$ echo asdf#qwert
asdf#qwert
# compared to
$ echo asdf #qwert
asdf
```

You can see that github's highlighter gets it right, but hljs wraps everything following `#` in a comment span:

```
<span class="hljs-comment">#qwert</span>
```

This is incorrect according to [POSIX 2.3 "Token Recognition"](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03) (see rules 8 and 9 in particular), and from my testing sh, bash and zsh are consistent with this.

**Expected behavior**
If there is no space or start-of-line in front of a `#`, it should not start a comment span.

**Additional context**
I'm working on a fix for this myself right now, will submit a PR in a bit.
EDIT: PR is ready, see link below.

iFreilicht:
In POSIX shells, the `#` needs to be after a whitespace or start-of-line to delimit a comment. Compare:

```sh
$ echo asdf#qwert
asdf#qwert
$ echo asdf #qwert
asdf
```

This is fixed with this PR.

I checked a few other languages as well that use #-delimited comments, but it seems this behaviour is unique to bash and other POSIX shells.

Fixes #3917

### Changes
Inside the bash mode, `HASH_COMMENT_MODE` was wrapped in another mode that prevents a match if the `#` is not after the start of the line or a whitespace character.

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -38,6 +38,18 @@ export default function(hljs) {
     end: /\)/,
     contains: [ hljs.BACKSLASH_ESCAPE ]
   };
+  const COMMENT = hljs.inherit(
+    hljs.COMMENT(),
+    {
+      match: [
+        /(^|\s)/,
+        /#.*$/
+      ],
+      scope: {
+        2: 'comment'
+      }
+    }
+  );
   const HERE_DOC = {
     begin: /<<-?\s*(?=\w+)/,
     starts: { contains: [
@@ -376,7 +388,7 @@ export default function(hljs) {
       hljs.SHEBANG(), // to catch unknown shells but still highlight the shebang
       FUNCTION,
       ARITHMETIC,
-      hljs.HASH_COMMENT_MODE,
+      COMMENT,
       HERE_DOC,
       PATH_MODE,
       QUOTE_STRING,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Core Grammars:
 - fix(css) fix overly greedy pseudo class matching [Bradley Mackey][]
 - enh(arcade) updated to ArcGIS Arcade version 1.24 [Kristian Ekenes][]
 - fix(yaml) fix for yaml with keys having brackets highlighted incorrectly [Aneesh Kulkarni][]
+- fix(bash) fix # within token being detected as the start of a comment [Felix Uhl][]
 
 Developer Tool:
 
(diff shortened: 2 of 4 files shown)


joshgoebel:
We'd prefer to do this with a multi-matcher rather than nesting, see an example: 

https://github.com/highlightjs/highlight.js/blob/main/src/languages/ruby.js#L247

And you could just use inherit to update the existing rule...

```js
hljs.inherit(HASH_COMMENT_MODE, 
// the changes you want to merge in
{
  match: // ...
}
```

iFreilicht:
Okay, done! I had to set begin and end to `undefined`, they seems to be incompatible with match.

src/languages/bash.js
```diff
@@ -38,6 +38,18 @@ export default function(hljs) {
     end: /\)/,
     contains: [ hljs.BACKSLASH_ESCAPE ]
   };
+  const COMMENT = hljs.inherit(
+    hljs.COMMENT(undefined, undefined),
+    {
+      match: [
+        /(^|\s)(?=#)/,
```
joshgoebel, line 45:
```suggestion
        /(^|\s)/,
```

Multi-match handles this, you don't need the look-ahead anymore.

src/languages/bash.js
```diff
@@ -38,6 +38,18 @@ export default function(hljs) {
     end: /\)/,
     contains: [ hljs.BACKSLASH_ESCAPE ]
   };
+  const COMMENT = hljs.inherit(
+    hljs.COMMENT(undefined, undefined),
```
joshgoebel, line 42: Wouldn't `hljs.COMMENT()` be equivalent - since unpaused arguments default to undefined?

src/languages/bash.js
iFreilicht: Ahh, awesome. Fixed!

src/languages/bash.js
iFreilicht: Absolutely. I'm very used to strict typescript where that would be forbidden 😄
Fixed!

joshgoebel (approved):

joshgoebel (approved):

PR #3925, opened 2023-11-13
Add categories to a lot of uncategorized grammars

allejo:
With the changes introduced to address https://github.com/highlightjs/highlightjs.org/issues/9, it became apparent that there are quite a few languages that are uncategorized.

image

### Changes

I've done my best to categorize the languages. I left some uncategorized because they seemed unique and I did introduce some new categories as well.

Here's what is left uncategorized after this PR,

image

### Checklist

### DIFF
--- a/src/languages/julia-repl.js
+++ b/src/languages/julia-repl.js
@@ -4,6 +4,7 @@ Description: Julia REPL sessions
 Author: Morten Piibeleht <morten.piibeleht@gmail.com>
 Website: https://julialang.org
 Requires: julia.js
+Category: system
 
 The Julia REPL code blocks look something like the following:
 
--- a/src/languages/julia-repl.js
+++ b/src/languages/julia-repl.js
@@ -4,7 +4,7 @@ Description: Julia REPL sessions
 Author: Morten Piibeleht <morten.piibeleht@gmail.com>
 Website: https://julialang.org
 Requires: julia.js
-Category: system
+Category: scientific
 
 The Julia REPL code blocks look something like the following:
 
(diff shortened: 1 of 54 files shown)


src/languages/julia-repl.js
```diff
@@ -4,6 +4,7 @@ Description: Julia REPL sessions
 Author: Morten Piibeleht <morten.piibeleht@gmail.com>
 Website: https://julialang.org
 Requires: julia.js
+Category: system
```
joshgoebel, line 7: Does system just mean "compiled to machine code"?

src/languages/ruby.js
```diff
@@ -4,7 +4,7 @@ Description: Ruby is a dynamic, open source programming language with a focus on
 Website: https://www.ruby-lang.org/
 Author: Anton Kovalyov <anton@kovalyov.net>
 Contributors: Peter Leonov <gojpeg@yandex.ru>, Vasily Polovnyov <vast@whiteants.net>, Loren Segal <lsegal@soen.ca>, Pascal Hurni <phi@ruby-reactive.org>, Cedric Sohrauer <sohrauer@googlemail.com>
-Category: common
+Category: common, system
```
joshgoebel, line 7: Ruby is system, not scripting?

joshgoebel:
Otherwise looks pretty good!

src/languages/julia-repl.js
allejo: That's kinda what I was going when I was categorizing these. However, I've corrected Julia to be scientific since that's what it's mostly used for.

src/languages/ruby.js
allejo: corrected to scripting, good catch

joshgoebel (approved):

Issue #90, opened 2012-05-24
Python `def` breaks simple Ruby functions


sugar700:
If Ruby define doesn't have $, {, = or ; it's guessed as Python (this snippet has 21 of relevance for Python and 6 for Ruby). Obviously, it's not Python as `def` should end with `:` but highlight.js still accepts it and gives it very high relevance compared to Ruby.

``` ruby
def hello_world
  # puts? Well, this is Ruby
  puts 'Hello, world!'
end
```

PR #3932, opened 2023-11-21
enh(delphi) Improve handling of numeric literals

Cirras:
### Changes

* Add support for digit separators, which were [introduced in Delphi 11](https://docwiki.embarcadero.com/RADStudio/Alexandria/en/What%27s_New#Binary_Literals_and_Digit_Separator)
    * This is also [allowed in FreePascal](https://wiki.freepascal.org/Integer#basics) via `{$modeSwitch underscoreIsSeparator+}`
* Allow all digits to be omitted for hexadecimal and binary literals (i.e. `$` and `%`)
    * This is an odd one, but it's allowed in Delphi
    * It produces a value of zero
* Add support for character strings with non-decimal numeric literals
    * e.g. `#90` = `#$5A` = `#&132` = `#%01011010` = `'Z'` 

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/delphi.js
+++ b/src/languages/delphi.js
@@ -166,7 +166,7 @@ export default function(hljs) {
     variants: [
       {
         // Hexadecimal notation, e.g., $7F.
-        begin: '\\$[0-9A-Fa-f]+' },
+        begin: '\\$[\\dA-Fa-f]+' },
       {
         // Hexadecimal literal with no digits
         begin: '\\$',
--- a/src/languages/delphi.js
+++ b/src/languages/delphi.js
@@ -166,33 +166,33 @@ export default function(hljs) {
     variants: [
       {
         // Regular numbers, e.g., 123, 123.456.
-        begin: /\b\d[\d_]*(\.\d[\d_]*)?/ },
+        match: /\b\d[\d_]*(\.\d[\d_]*)?/ },
       {
         // Hexadecimal notation, e.g., $7F.
-        begin: /\$[\dA-Fa-f_]+/ },
+        match: /\$[\dA-Fa-f_]+/ },
       {
         // Hexadecimal literal with no digits
-        begin: /\$/,
+        match: /\$/,
         relevance: 0 },
       {
         // Octal notation, e.g., &42.
-        begin: /&[0-7][0-7_]*/ },
+        match: /&[0-7][0-7_]*/ },
       {
         // Binary notation, e.g., %1010.
-        begin: /%[01_]+/ },
+        match: /%[01_]+/ },
       {
         // Binary literal with no digits
-        begin: /%/,
+        match: /%/,
         relevance: 0 }
     ]
   };
   const CHAR_STRING = {
     className: 'string',
     variants: [
-      { begin: /#\d[\d_]*/ },
-      { begin: /#\$[\dA-Fa-f][\dA-Fa-f_]*/ },
-      { begin: /#&[0-7][0-7_]*/ },
-      { begin: /#%[01][01_]*/ }
+      { match: /#\d[\d_]*/ },
+      { match: /#\$[\dA-Fa-f][\dA-Fa-f_]*/ },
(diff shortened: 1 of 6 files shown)


src/languages/delphi.js
```diff
@@ -164,20 +164,28 @@ export default function(hljs) {
     relevance: 0,
     // Source: https://www.freepascal.org/docs-html/ref/refse6.html
     variants: [
+      {
+        // Regular numbers, e.g., 123, 123.456.
+        begin: /\b\d[\d_]*(\.\d[\d_]*)?/ },
       {
         // Hexadecimal notation, e.g., $7F.
-        begin: '\\$[0-9A-Fa-f]+' },
+        begin: /\$[\dA-Fa-f_]*/ },
```
joshgoebel, line 172: This is not comparable because it can match no digits at all.

src/languages/delphi.js
```diff
@@ -164,20 +164,28 @@ export default function(hljs) {
...
       {
         // Octal notation, e.g., &42.
-        begin: '&[0-7]+' },
+        begin: /&[0-7][0-7_]*/ },
       {
         // Binary notation, e.g., %1010.
-        begin: '%[01]+' }
+        begin: /%[01_]*/ }
```
joshgoebel, line 178: This is not comparable because it can match no digits at all.

src/languages/delphi.js
```diff
@@ -164,20 +164,28 @@ export default function(hljs) {
...
     ]
   };
   const CHAR_STRING = {
     className: 'string',
-    begin: /(#\d+)+/
+    variants: [
+      { begin: /#\d[\d_]*/ },
+      { begin: /#\$[\dA-Fa-f][\dA-Fa-f_]*/ },
+      { begin: /#&[0-7][0-7_]*/ },
+      { begin: /#%[01][01_]*/ }
```
joshgoebel, line 187: Please use `match` since there is no `end` regex here...

src/languages/delphi.js
Cirras: Hi @joshgoebel,
This was an intentional change in [5d3110c](https://github.com/highlightjs/highlight.js/pull/3932/commits/5d3110c91f0869825f584d6645383de8a34707b1).
Delphi allows all digits to be omitted in hexadecimal literals.

src/languages/delphi.js
Cirras: Hi @joshgoebel,
This was an intentional change in [5d3110c](https://github.com/highlightjs/highlight.js/pull/3932/commits/5d3110c91f0869825f584d6645383de8a34707b1).
Delphi allows all digits to be omitted in binaryliterals.

src/languages/delphi.js
Cirras: Sure thing! I'll change that.

src/languages/delphi.js
joshgoebel: What is the purpose and value of a hexadecimal literal with no digits?

Cirras: > What is the purpose

From a language design perspective, I agree that it's a pretty bizarre thing for Delphi to allow.
My change was purely in the interest of accuracy to Delphi's official parser.

> and value

The value of the expression (in both cases) is 0.

joshgoebel: I'm worried auto-detect and other languages... $ and % are weird things to match on their own (and things that occur often in many types of files)... if we're going to add them now I think we need the original rule PLUS a new rule that matches `*` (set to 0 relevance) - so as not to influence auto-detection...

Or we can hold this until v12 when we stop caring about such things.

Cirras: > if we're going to add them now I think we need the original rule PLUS a new rule that matches `*` (set to 0 relevance) - so as not to influence auto-detection...

Very sensible! I'm happy to make that change.

Cirras: Done.

joshgoebel (approved):

PR #3937, opened 2023-12-10
fix(swift) correctly highlight generics and conformances in type definitions

bradleymackey:
Fixes highlighting of type definitions, especially when the given type is generic.

### Changes
- Fix entire type definition being highlighted as `title.class`, only the newly defined type should be highlighted as this. Inheritance and protocol conformances should be indicated with the `title.class.inherited` class instead, as this is better modelled by this inheritance class.
- Support generic signatures for type declarations
- Add more test cases for type declarations

### Checklist

### DIFF
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -464,7 +464,7 @@ export default function(hljs) {
       Swift.identifier,
       /\s*/,
     ],
-    scope: {
+    beginScope: {
       1: "keyword",
       3: "title.class"
     },
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -457,6 +457,23 @@ export default function(hljs) {
     end: /}/
   };
 
+  const TYPE_DECLARATION = {
+    match: [
+      /(struct|protocol|class|extension|enum|actor)/,
+      /\s+/,
+      /[A-Za-z$_][\u00C0-\u02B80-9A-Za-z$_]*/
+    ],
+    scope: {
+      1: "keyword",
+      3: "title.class"
+    },
+    keywords: KEYWORDS,
+    contains: [
+      GENERIC_PARAMETERS,
+      ...KEYWORD_MODES,
+    ]
+  };
+
   // Add supported submodes to string interpolation.
   for (const variant of STRING.variants) {
     const interpolation = variant.contains.find(mode => mode.label === "interpol");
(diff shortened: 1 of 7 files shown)


joshgoebel:
I know it's not exactly the same thing, but might `title.class.inherited` be better than `type`?

bradleymackey:
@joshgoebel I agree, this better models the semantics here. Please see latest changes.

src/languages/swift.js
```diff
@@ -457,6 +457,37 @@ export default function(hljs) {
     end: /}/
   };
 
+  const TYPE_DECLARATION = {
+    match: [
```
joshgoebel, line 461: Please use `begin` since this is opening a scope (since you have `contains`), even if end is implied.

src/languages/swift.js
```diff
@@ -457,6 +457,37 @@ export default function(hljs) {
...
+      1: "keyword",
+      3: "title.class"
+    },
+    keywords: KEYWORDS,
+    contains: [
+      GENERIC_PARAMETERS,
+      ...KEYWORD_MODES,
+      {
+        begin: /:/,
+        end: /[{]/,
```
joshgoebel, line 477: Why the `[]`? Is this just trying to match `\{`?

src/languages/swift.js
```diff
@@ -457,6 +457,37 @@ export default function(hljs) {
     end: /}/
   };
 
+  const TYPE_DECLARATION = {
+    match: [
+      /(struct|protocol|class|extension|enum|actor)/,
+      /\s+/,
+      /[A-Za-z$_][\u00C0-\u02B80-9A-Za-z$_]*/,
```
joshgoebel, line 464: What is this?  I'm wondering if perhaps it should be a named constant that we pull in from `Swift` like we do with other major patterns?

joshgoebel (changes requested):
Looks pretty good, just a few thoughts.

src/languages/swift.js
bradleymackey: This was the existing regex used when a type name is declared. However, I've had another look at [the grammar](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/summaryofthegrammar/#Declarations) and it appears that a type declaration is actually the same as `Swift.identifier` in this case. I'll update this.

src/languages/swift.js
```diff
@@ -473,8 +473,8 @@ export default function(hljs) {
       GENERIC_PARAMETERS,
       ...KEYWORD_MODES,
       {
-        begin: /:/,
-        end: /[{]/,
+        begin: ":",
+        end: "{",
```
joshgoebel, line 477: We prefer regex everywhere we can rather than strings.  My original point was that I think we want `\{` to properly escape `{`, rather than a character group `[{]` which looks a little strange.

src/languages/swift.js
```diff
@@ -458,7 +458,7 @@ export default function(hljs) {
   };
 
   const TYPE_DECLARATION = {
-    match: [
+    begin: [
```
joshgoebel, line 461: You may also need `beginScope` or some such, I forget.

src/languages/swift.js
bradleymackey: Ah got it, wasn't aware regex was preferred everywhere!

joshgoebel (approved):

PR #3940, opened 2023-12-13
fix(javascript) incorrect function name highlighting

### LINKED ISSUE #3870: (JavaScript) Function names with a prefix shared with a built-in global are confused for a global

cyfung1031:
* fix #3870
* override #3871

### Before

Screen Shot 2023-12-13 at 20 46 11

### After

Screen Shot 2023-12-13 at 20 46 08

cc @joshgoebel @arnog

### DIFF
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -388,8 +388,8 @@ export default function(hljs) {
         ...ECMAScript.BUILT_IN_GLOBALS,
         "super",
         "import"
-      ]),
-      IDENT_RE, regex.lookahead(/\(/)),
+      ].map(x => `${x}\\s*\\(`)),
+      IDENT_RE, regex.lookahead(/\s*\(/)),
     className: "title.function",
     relevance: 0
   };
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Core Grammars:
 - enh(delphi) allow digits to be omitted for hex and binary literals [Jonah Jeleniewski][]
 - enh(delphi) add support for digit separators [Jonah Jeleniewski][]
 - enh(delphi) add support for character strings with non-decimal numerics [Jonah Jeleniewski][]
+- fix(javascript) incorrect function name highlighting [CY Fung][]
 
 New Grammars:
 
@@ -36,6 +37,7 @@ Developer Tool:
 [Robloxian Demo]: https://github.com/RobloxianDemo
 [Paul Tsnobiladzé]: https://github.com/tsnobip
 [Jonah Jeleniewski]: https://github.com/cirras
+[CY Fung]: https://github.com/cyfung1031
 
 
 ## Version 11.9.0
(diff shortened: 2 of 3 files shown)


src/languages/javascript.js
```diff
@@ -592,7 +592,7 @@ export default function(hljs) {
         relevance: 0
       },
       {
-        match: [ /\bconstructor(?=\s*\()/ ],
+        match: [ new RegExp(`\\b(constructor|${ECMAScript.BUILT_IN_GLOBALS.map(x=>`${x}\\w+`).join('|')}|${ECMAScript.BUILT_IN_VARIABLES.map(x=>`${x}\\w+`).join('|')})(?=\\s*\\()`) ],
```
joshgoebel, line 595: Why so limited though vs matching just any valid function name?

```
[IDENT]()
```

cyfung1031: This is against ```js
  const FUNCTION_CALL = {
    match: regex.concat(
      /\b/,
      noneOf([
        ...ECMAScript.BUILT_IN_GLOBALS,
        "super",
        "import"
      ]),
      IDENT_RE, regex.lookahead(/\(/)),
    className: "title.function",
    relevance: 0
  };
  ```

joshgoebel: Oh right, we already do this... so isn't this just a bug/problem with the `noneOf` guard?  This seems to fix it:

```js
noneOf([
        ...ECMAScript.BUILT_IN_GLOBALS,
        "super",
        "import"
      ].map(x => `${x}\\(`)),
```

Thoughts?

joshgoebel: More complete to deal with spaces also:

```
  const FUNCTION_CALL = {
    match: regex.concat(
      /\b/,
      noneOf([
        ...ECMAScript.BUILT_IN_GLOBALS,
        "super",
        "import"
      ].map(x=> `${x}\\s*\\(`)),
      IDENT_RE, regex.lookahead(/\s*\(/)),
    className: "title.function",
    relevance: 0
  };
```

cyfung1031: It looks good. I amended as what you suggested.

joshgoebel:
Awesome, so now lets just add a changelog entry! :)

cyfung1031:
> Awesome, so now lets just add a changelog entry! :)

done.

joshgoebel (approved):

joshgoebel:
Will have to take another look later, tests aren't all passing.

cyfung1031:
> Will have to take another look later, tests aren't all passing.

I am not quite sure what the error suggested.
Is it because of the change or the conflict with the latest commits?

btw, `regex.lookahead(/\(/))` also works. 
only `.map(x => `${x}\\s*\\(`))` is required.

joshgoebel (approved):

PR #3941, opened 2023-12-14
Bugfix1c

Diversus23:
### Changes

* fix(1c) fix escaped symbols "+-;():=,[]" literals
* Added `1c-light` theme a like in the IDE 1C:Enterprise 8 (for 1c)

In IDE 1C:Enterprise 8:
image

New theme `1c-light`:
image

### Checklist

- [x] Update markup tests

### DIFF
--- a/src/languages/1c.js
+++ b/src/languages/1c.js
@@ -443,6 +443,12 @@ export default function(hljs) {
     ]
   };
 
+  const PUNCTUATION = {
+    match: /[;|(|)|:|=|,]/,
+    className: "punctuation",
+    relevance: 0
+  };
+
   // comment : комментарии
   const COMMENTS = hljs.inherit(hljs.C_LINE_COMMENT_MODE);
 
@@ -529,7 +535,8 @@ export default function(hljs) {
       SYMBOL,
       NUMBERS,
       STRINGS,
-      DATE
+      DATE,
+      PUNCTUATION
     ]
   };
 }
--- a/src/languages/1c.js
+++ b/src/languages/1c.js
@@ -444,7 +444,7 @@ export default function(hljs) {
   };
 
   const PUNCTUATION = {
-    match: /[;|(|)|+|\-|:|=|,]/,
+    match: /[;()+\-:=,]/,
     className: "punctuation",
     relevance: 0
   };
--- a/src/languages/1c.js
+++ b/src/languages/1c.js
@@ -444,7 +444,7 @@ export default function(hljs) {
   };
 
   const PUNCTUATION = {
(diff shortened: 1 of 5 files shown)


src/styles/1c-light.css
```diff
@@ -0,0 +1,133 @@
...
+  Website: https://softonit.ru/
+  License: see project LICENSE
+  Touched: 2023
+*/
+
+pre code.hljs {
+    display: block;
+    overflow-x: auto;
+    padding: 1em;
+  }
```
joshgoebel, line 15: Removed, this is added by the build process.
```suggestion
```

src/styles/1c-light.css
```diff
@@ -0,0 +1,133 @@
...
+  
+  code.hljs {
+    padding: 3px 5px;
+  }
```
joshgoebel, line 19: Ditto.
```suggestion
```

src/styles/1c-light.css
```diff
@@ -0,0 +1,133 @@
...
+  /* end baseline CSS */
+  
+  .hljs {
+    font-family: "Courier New", monospace; /* This font is used in IDE the 1C:Enterprise 8 */
+    color: #0000ff;
+    background: #ffffff;
+    font-weight: normal;
+    font-style: normal;
+  }
```
joshgoebel, line 28: There is no need for these `normal` resets and our themes are not allowed to change the font.

src/styles/1c-light.css
Diversus23: Thanks for the review. Done

src/styles/1c-light.css
Diversus23: Done

src/languages/1c.js
```diff
@@ -443,6 +443,12 @@ export default function(hljs) {
     ]
   };
 
+  const PUNCTUATION = {
+    match: /[;|(|)|+|\-|:|=|,]/,
```
joshgoebel, line 447: What are all the `|` for... `|` is just another character [with no special meaning] inside a `[]` block... I think you want to just remove them all?  Or have it only a single time if `|` itself is also punctuation...

joshgoebel:
Looking better, almost there!

src/languages/1c.js
Diversus23: > What are all the `|` for... `|` is just another character [with no special meaning] inside a `[]` block... I think you want to just remove them all? Or have it only a single time if `|` itself is also punctuation...

You're right Josh. This is a mistake. Deleted it. Thanks!

joshgoebel (approved):

Issue #3924, opened 2023-11-11
(rust) raw keyword variable names are not highlighted correctly


JayAndJef:
**Describe the issue**
In rust, one can put `r#` before a keyword name to escape it. Highlight.js does not recognize when this happens and instead highlights the keyword name as a keyword.

**Which language seems to have the issue?**
Rust

**Are you using `highlight` or `highlightAuto`?**
`highlight`

**Sample Code to Reproduce**
```rs
pub fn main() {
	let r#use = "hello"; // r#use should be highlighted as a variable
}
```

**Expected behavior**
In the above code block, `r#use` should be highlighted as a variable name. Instead, `r` is highlighted as a variable name and `use` is highlighted as a keyword.

**Additional context**

joshgoebel:
A simple rule that matches `r#*` and applies no scope would fix the issue of false positives here.

ThinkingVincent:
rush.js에서 새로운 정규표현식을 적용하여 문제를 해결해봤습니다.
https://regexr.com/
이 정규표현식 사이트에서
(r#)?[a-zA-Z_]\w*
이 정규식을 활용하면 모든 경우에 r#키워드가 제대로 식별됩니다.
test-browser에서도 정상적으로 작동하는 것 같아요.

ThinkingVincent:
I tried applying a new regex in rust.js to solve the problem.
https://regexr.com/
In this regex site
`(r#)?[a-zA-Z_]\w*`
With this regex, the r# keyword is properly identified in all cases.
It seems to work fine in the test-browser as well.

PR #3947, opened 2023-12-24
fix(rust): add support for r# raw identifiers

ThinkingVincent:
https://github.com/highlightjs/highlight.js/issues/3924
### Changes

In order to support raw identifiers in Rust, the regular expressions for `UNDERSCORE_IDENT_RE` and `IDENT_RE` need to be modified to include the r# keyword. However, since modes.js is shared by other languages, we need to ensure that only rust supports the r# keyword.

**Original (modes.js)**
```js
const UNDERSCORE_IDENT_RE = '[a-zA-Z_]\\w*';
const IDENT_RE = '[a-zA-Z]\\w*';
```

**Change (rust.js)**

So, add the following code to rust.js to support the r# keyword only in rust.
``` js
const UNDERSCORE_IDENT_RE = '(r#)?[a-zA-Z_]\\w*';
const IDENT_RE = '(r#)?[a-zA-Z]\\w*';
```

### Checklist

### DIFF
--- a/src/languages/rust.js
+++ b/src/languages/rust.js
@@ -9,13 +9,13 @@ Category: common, system
 /** @type LanguageFn */
 
 export default function(hljs) {
+  const regex = hljs.regex;
   // ============================================
   // Added to support the r# keyword, which is a raw identifier in Rust.
-  const RAW_IDENTIFIER = '(r#)?';
-  const UNDERSCORE_IDENT_RE = RAW_IDENTIFIER + hljs.UNDERSCORE_IDENT_RE;
-  const IDENT_RE = RAW_IDENTIFIER + hljs.IDENT_RE;
+  const RAW_IDENTIFIER = /(r#)?/;
+  const UNDERSCORE_IDENT_RE = regex.concat(RAW_IDENTIFIER, hljs.UNDERSCORE_IDENT_RE);
+  const IDENT_RE = regex.concat(RAW_IDENTIFIER, hljs.IDENT_RE);
   // ============================================
-  const regex = hljs.regex;
   const FUNCTION_INVOKE = {
     className: "title.function.invoke",
     relevance: 0,
--- a/src/languages/rust.js
+++ b/src/languages/rust.js
@@ -7,6 +7,13 @@ Category: common, system
 */
 
 /** @type LanguageFn */
+
+// ============================================
+// Added to support the r# keyword, which is a raw identifier in Rust.
+const UNDERSCORE_IDENT_RE = '(r#)?[a-zA-Z_]\\w*';
+const IDENT_RE = '(r#)?[a-zA-Z]\\w*';
+// ============================================
+
 export default function(hljs) {
   const regex = hljs.regex;
   const FUNCTION_INVOKE = {
@@ -15,7 +22,7 @@ export default function(hljs) {
     begin: regex.concat(
       /\b/,
       /(?!let|for|while|if|else|match\b)/,
-      hljs.IDENT_RE,
+      IDENT_RE,
       regex.lookahead(/\s*\(/))
   };
   const NUMBER_SUFFIX = '([ui](8|16|32|64|128|size)|f(32|64))\?';
@@ -176,7 +183,7 @@ export default function(hljs) {
     name: 'Rust',
     aliases: [ 'rs' ],
     keywords: {
-      $pattern: hljs.IDENT_RE + '!?',
+      $pattern: IDENT_RE + '!?',
       type: TYPES,
       keyword: KEYWORDS,
       literal: LITERALS,
@@ -216,7 +223,7 @@ export default function(hljs) {
         begin: [
           /fn/,
           /\s+/,
-          hljs.UNDERSCORE_IDENT_RE
+          UNDERSCORE_IDENT_RE
         ],
         className: {
           1: "keyword",
@@ -243,7 +250,7 @@ export default function(hljs) {
           /let/,
           /\s+/,
           /(?:mut\s+)?/,
-          hljs.UNDERSCORE_IDENT_RE
+          UNDERSCORE_IDENT_RE
         ],
         className: {
           1: "keyword",
@@ -256,7 +263,7 @@ export default function(hljs) {
         begin: [
           /for/,
           /\s+/,
-          hljs.UNDERSCORE_IDENT_RE,
+          UNDERSCORE_IDENT_RE,
           /\s+/,
           /in/
         ],
@@ -270,7 +277,7 @@ export default function(hljs) {
         begin: [
           /type/,
           /\s+/,
-          hljs.UNDERSCORE_IDENT_RE
+          UNDERSCORE_IDENT_RE
         ],
         className: {
           1: "keyword",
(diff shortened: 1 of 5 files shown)


src/languages/rust.js
```diff
@@ -176,7 +183,7 @@ export default function(hljs) {
     name: 'Rust',
     aliases: [ 'rs' ],
     keywords: {
-      $pattern: hljs.IDENT_RE + '!?',
+      $pattern: IDENT_RE + '!?',
```
joshgoebel, line 186: I don't think any of our literal list of keywords start with `#r` (since I don't see that you've added any keywords) so if that's the case this should be reverted to reduce the work the keyword engine needs to do.

"overmatching" with $pattern for things that will never be found in the keyword lists only slows the engine down.

src/languages/rust.js
```diff
@@ -7,6 +7,13 @@ Category: common, system
 */
 
 /** @type LanguageFn */
+
+// ============================================
+// Added to support the r# keyword, which is a raw identifier in Rust.
+const UNDERSCORE_IDENT_RE = '(r#)?[a-zA-Z_]\\w*';
+const IDENT_RE = '(r#)?[a-zA-Z]\\w*';
```
joshgoebel, line 14: Raw identifiers can be used almost anywhere regular identifiers can?

injae-kim: image
(FYI, r# keyword can be used on rust's variable but not highlighted well now)

src/languages/rust.js
```diff
@@ -216,7 +223,7 @@ export default function(hljs) {
         begin: [
           /fn/,
           /\s+/,
-          hljs.UNDERSCORE_IDENT_RE
+          UNDERSCORE_IDENT_RE
```
injae-kim, line 226: Maybe should we apply `(r#)?` only on `className: variable`?

L226 here, className is title.function so user normally not use `r#` prefix here 🤔 WDYT?

src/languages/rust.js
injae-kim, line 14:
```suggestion
const UNDERSCORE_INDENT_RE = '(r#)?[a-zA-Z_]\\w*';
const INDENT_RE = '(r#)?[a-zA-Z]\\w*';
```
fix typo `IDENT` -> `INDENT`

```
const RAW_IDENTIFIER = '(r#)?'
const UNDERSCORE_INDENT_RE = RAW_IDENTIFIER  + hljs.UNDERSCORE_INDENT_RE;
const INDENT_RE = RAW_IDENTIFIER  + hljs.INDENT_RE
```
also, maybe above code looks better? (just suggestion 😄)

ThinkingVincent: In Rust, r# is used when you want to use a reserved keyword in rust (fn, use, true, etc...) as a custom identifier.
However, in general, even when the word after r# is not a reserved keyword, it will not cause compilation errors, etc. So if you use r#, it should be highlighted as one identifier, including r#.

src/languages/rust.js
ThinkingVincent: [doc.1](https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html)
[doc.2](https://doc.rust-lang.org/book/appendix-01-keywords.html#raw-identifiers)

According to the documentation, r# keywords can even be used as function names 😆

src/languages/rust.js
ThinkingVincent: Thanks for the comment!

You're right, there are no keywords starting with `r#`, I think I modified that object without clearly understanding it.
I'll reflect and re-commit!
Thanks.

src/languages/rust.js
ThinkingVincent: Thanks for the comment!

That's a great suggestion. I think it looks much better! I'll reflect and commit!

src/languages/rust.js
injae-kim: > According to the documentation, r# keywords can even be used as function names

oh~ you're right. I didn't know this rust spec. please just ignore my comment~!

src/languages/rust.js
```diff
@@ -7,15 +7,22 @@ Category: common, system
 */
 
 /** @type LanguageFn */
+
 export default function(hljs) {
+  // ============================================
+  // Added to support the r# keyword, which is a raw identifier in Rust.
+  const RAW_IDENTIFIER = '(r#)?';
```
joshgoebel, line 14: Lets keep using `//` regex here... and pull in our `regex` utils and use `concat` for this... see abnf for an example.

https://github.com/highlightjs/highlight.js/blob/main/src/languages/abnf.js#L57

joshgoebel (changes requested):
One last small thing and this is good to merge I think.

Thanks so much!

ThinkingVincent:
I've committed to reflect your request.
However, i found one more issue.

스크린샷 2023-12-26 035347
Looking at the above test, the `r#` when the identifier is declared is working as desired, but the
``` rust
println!("{}", r#try);
```
When the identifier is referenced again, try appears as a reserved keyword again in `r#try`.

This issue is addressed in
https://github.com/highlightjs/highlight.js/pull/3947#discussion_r1435886988
Here, I found that changing the code back to the one I had reverted earlier fixed it.

스크린샷 2023-12-26 033717

src/languages/rust.js
ThinkingVincent: I've committed to reflect your request.
However, i found one more issue.

스크린샷 2023-12-26 035347
Looking at the above test, the `r#` when the identifier is declared is working as desired, but the
``` rust
println!("{}", r#try);
```
When the identifier is referenced again, try appears as a reserved keyword again in `r#try`.

This issue is addressed in
https://github.com/highlightjs/highlight.js/pull/3947#discussion_r1435886988
Here, I found that changing the code back to the one I had reverted earlier fixed it.

스크린샷 2023-12-26 033717

### Code where the reference to the `r#try` identifier is not recognized

``` rust
    name: 'Rust',
    aliases: [ 'rs' ],
    keywords: {
      $pattern: hljs.IDENT_RE + '!?',
      type: TYPES,
      keyword: KEYWORDS,
      literal: LITERALS,
      built_in: BUILTINS
    },
```

### Code where the reference to the `r#try` identifier is recognized

``` rust
    name: 'Rust',
    aliases: [ 'rs' ],
    keywords: {
      $pattern: IDENT_RE + '!?',
      type: TYPES,
      keyword: KEYWORDS,
      literal: LITERALS,
      built_in: BUILTINS
    },
```

joshgoebel (approved):

joshgoebel:
> When the identifier is referenced again, try appears as a reserved keyword again in r#try.

Ok, now we're talking about false positives, which is a different situation entirely.  I'm not sure we should use the keyword engine to consume these (but not match, and hence not apply scoping)...  can we add a few failing tests for this case and I'll take another look?

ThinkingVincent:
> > When the identifier is referenced again, try appears as a reserved keyword again in r#try.
> 
> Ok, now we're talking about false positives, which is a different situation entirely. I'm not sure we should use the keyword engine to consume these (but not match, and hence not apply scoping)... can we add a few failing tests for this case and I'll take another look?

알겠습니다, 그럼 새로운 문제에 대해 새 이슈를 생성하고 해결방법을 고민해볼게요.

ThinkingVincent:
> > When the identifier is referenced again, try appears as a reserved keyword again in r#try.
> 
> Ok, now we're talking about false positives, which is a different situation entirely. I'm not sure we should use the keyword engine to consume these (but not match, and hence not apply scoping)... can we add a few failing tests for this case and I'll take another look?

Ok, I'll create a new issue for the new problem and figure out how to fix it!

Issue #3948, opened 2023-12-26
[Request] Updated union type keywords in Rust


ThinkingVincent:
**Is your request related to a specific problem you're having?**
Rust supports keywords of type `union`.
[doc.1](https://github.com/rust-lang/rfcs/blob/master/text/1444-union.md)
[doc.2](https://blog.rust-lang.org/2017/07/20/Rust-1.19.html)

But `union` hasn't been added to the **rust.js** keyword list in highlight.js yet - all the other keywords are included.
https://github.com/highlightjs/highlight.js/blob/8b88a7d4c3e88f157a87b8d50088af329d700eeb/src/languages/rust.js#L73C4-L77C15

**The solution you'd prefer / feature you'd like to see added...**
Add the `union` string to the list of keywords in **rust.js.**

**Any alternative solutions you considered...**
When you use union as a purposeful type in your code, it is designed to be properly highlighted. It's just not included in the KEYWORDS array. So maybe you don't need to modify anything. However, if the KEYWORDS array should contain all the keywords supported by rust, it seems like a good idea to add union as well.
https://github.com/highlightjs/highlight.js/blob/8b88a7d4c3e88f157a87b8d50088af329d700eeb/src/languages/rust.js#L287-L296

**Additional context...**

### as is
![image](https://github.com/highlightjs/highlight.js/assets/103738589/292a8d9f-6b2c-4c29-8a1b-93daa953f4a0)

### to be
![image](https://github.com/highlightjs/highlight.js/assets/103738589/5c90a3c7-49b6-4c7f-a501-dee44f8737df)

joshgoebel:
What's an example of where `union` is used in Rust outside of defining a new class/data type?

ThinkingVincent:
As far as I know, the `union` keyword is not used except when defining a new union. The same goes for enum and struct, but they are all included in KEYWORDS, so I'm just suggesting it.

joshgoebel:
> The same goes for enum and struct, but they are all included in KEYWORDS

Ha, well that's a good argument.

PR #3949, opened 2023-12-26
Adding union to be recognized as a keyword in Rust

ThinkingVincent:
Adding union to be recognized as a keyword in Rust.

https://github.com/highlightjs/highlight.js/issues/3948
### Changes

Added the union keyword to the KEYWORDS array in rust.js.
image

### Checklist
  Even without adding any markup tests, the following code ensures that the highlighting is applied correctly.
``` rust
        begin: [
          /(?:trait|enum|struct|union|impl|for)/,
          /\s+/,
          UNDERSCORE_IDENT_RE
        ],
        className: {
          1: "keyword",
          3: "title.class"
        }
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Core Grammars:
 - fix(typescript): params types [Mohamed Ali][]
 - fix(rust) fix escaped double quotes in string  [Mohamed Ali][]
 - fix(rust) fix for r# raw identifier not being highlighted correctly. [JaeBaek Lee][]
+- enh(rust) Adding union to be recognized as a keyword in Rust. [JaeBaek Lee][]
 - fix(yaml) fix for yaml with keys having brackets highlighted incorrectly [Aneesh Kulkarni][]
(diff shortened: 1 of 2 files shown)


joshgoebel:
Please also provide a test case of `union` used in context (outside of defining a union)... see `tests/markup`, etc.

ThinkingVincent:
As I commented in the issue https://github.com/highlightjs/highlight.js/issues/3948, union is not commonly used except when defining unions, so I didn't write a separate test case for it.
of course, It can be used outside of defining a union, such as let r#union, but that's not relevant to this issue, as it's a case of the regex not recognizing union as a keyword.

joshgoebel:
> union is not commonly used except when defining unions, so I didn't write a separate test case for it.

Can you provide at least one sample usage outside of a union declaration other than `r#union`?  The keywords list is not necessarily intended to be exhaustive if the keywords already have their own separate rules.

joshgoebel (approved):

PR #3952, opened 2023-12-31
Update sql.js to include the LIMIT command

colin99d:
Adds the limit command to the sql file.

### Checklist


joshgoebel:
Please use a database specific grammar, our SQL grammar is only intended for actual ANSI SQL standard.  I'm pretty sure this has come up before and `LIMIT` is not a ANSI SQL keyword.  If you can please point to some docs that it is (part of ANSI SQL), we can reopen this.

PR #3954, opened 2024-01-04
(chore) update Node.js test versions

bradleymackey:
### Changes
- Drop Node 16 support from test matrix, this is no longer supported or receiving security updates.
- Start testing "Current" versions as well, starting with v21. This will allow us to catch issues caused by new releases ASAP.

### Checklist
- ~~[ ] Added markup tests, or they don't apply here because..~~.

### DIFF
--- a/.github/workflows/tests.js.yml
+++ b/.github/workflows/tests.js.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        # All 'LTS' and 'Current' versions
+        # All 'LTS' and 'Current' versions (https://nodejs.org/en/about/previous-releases)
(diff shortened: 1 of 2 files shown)


joshgoebel:
While we're at it shouldn't we add 22 and 24 to CI?

bradleymackey:
I don't think there's releases available for these yet, current latest is v21, but that's not LTS.

joshgoebel:
> I don't think there's releases available for these yet. The current latest is v21,

Turns out I was looking at the roadmap... LOL... yes, we should test v21 also.

joshgoebel (approved):

PR #3959, opened 2024-01-07
enh(swift) highlight function and macro call usage

bradleymackey:
Improve fidelity by highlighting function calls in addition to definitions. This is already done by javascript, for example.

| Before | After |
| --- | --- |
| Screenshot 2024-01-07 at 14 03 24 | Screenshot 2024-01-07 at 14 05 49 |

### Changes
- Add `negativeLookahead` helper to `lib/regex`
- Highlight all function and method calls as `title.function` (e.g. `myFunction()`, `obj.myFunction()`).
- Highlight all freestanding macro calls as `title.function` (e.g. `#myMacro()`).
- There can be >=0 whitespace between the end of the function name and the parentheses, but newlines are not permitted. Encapsulate this in a new regex, `TRAILING_PAREN_REGEX`
- Built-ins still highlight as such but with exceptions for methods–where they are semantically no longer builtins.
- Most keywords can exist as both properties and methods (e.g. `obj.if()`). Highlight keyword method calls also.
  - Number sign keywords and attribute keywords are permitted as raw functions, as they start with leading `#` and `@` respectively. If they don't start with these, they're no longer considered reserved.
- Highlight escaped function name calls also (e.g. ```` `myFunc`() ````). This feature exists so that reserved words can be used as identifiers.
- Variable accesses (e.g. `obj.myVariable`) remain unhighlighted for now.
- Add markup tests for all the cases and edge cases described.

#### Next Steps (for a future PR)
- Support trailing closures, which are method calls: `exampleCall { print("action") }`

### Checklist
- [x] Added markup tests


joshgoebel:
We can't use negative lookahead until v12 though because it's a breaking change. And I don't have a timetable on that yet.

bradleymackey:
@joshgoebel This is an adapted version of Javascript's implementation which is already using negative lookahead ( [see here](https://github.com/highlightjs/highlight.js/blob/6a52185d9b855130b5acaccef143a7bd602e7885/src/languages/javascript.js#L381)). My understanding is that only lookbehind is not allowed at this stage.

bradleymackey:
@joshgoebel Is there a way forward for this PR? If the concern is a language-specific breaking change, the Swift language itself is already using negative lookahead as well, [here](https://github.com/highlightjs/highlight.js/blob/14a5dcf36c8ea993d646fc37f27fbaabd086c7d5/src/languages/swift.js#L209).

joshgoebel:
Sorry for delay.  I think `title.function` might not be correct though typically we consider anything after `.` hanging off of an identifier (likely an object) a property... if the function is "global" or standalone then it could get `function.title`. IE:

```rb
debug_it("hello world")  # debug_it is `title.function`
person.eat("sandwich") # `eat` is a property of person, that also happens to be a method.
```

Thoughts?

src/languages/swift.js
```diff
@@ -40,9 +46,9 @@ export default function(hljs) {
     ],
     className: { 2: "keyword" }
   };
-  const KEYWORD_GUARD = {
-    // Consume .keyword to prevent highlighting properties and methods as keywords.
-    match: concat(/\./, either(...Swift.keywords)),
+  const KEYWORD_PROP_GUARD = {
```
joshgoebel, line 49: Ok, so we don't want the keyword engine to grab these as keywords, true, but if they are essentially "properly access" why wouldnt we scope them as that?

src/languages/swift.js
```diff
@@ -491,14 +490,56 @@ export default function(hljs) {
...
+  }
+
+  const METHODS_ONLY = [...Swift.keywords, ...Swift.builtIns];
+  const FUNCTION_CALL = {
+    relevance: 0,
+    variants: [
+      {
+        // Functions and macro calls
+        scope: "title.function",
+        keywords: KEYWORDS,
```
joshgoebel, line 504: This seems weird, why would we have keywords popping up in the middle of a function name?  Or is this truly to cover "it's a function name that's also a keyword"?

src/languages/swift.js
```diff
@@ -491,14 +490,56 @@ export default function(hljs) {
...
+      {
+        // Keywords/built-ins that only can appear as a method call
+        // e.g. foo.if()
+        match: [
+          /\./,
+          either(...METHODS_ONLY),
+          TRAILING_PAREN_REGEX,
+        ],
+        scope: {
+          2: "title.function",
```
joshgoebel, line 521: We'd call anything that *appears* to be an object property access a `property` for consistent highlighting across different grammars.

src/lib/regex.js
```diff
@@ -25,6 +25,14 @@ export function lookahead(re) {
   return concat('(?=', re, ')');
 }
 
+/**
+ * @param {RegExp | string } re
+ * @returns {string}
+ */
+export function negativeLookahead(re) {
```
joshgoebel, line 32: Awesome.

joshgoebel:
Ping.

joshgoebel (approved):

PR #3961, opened 2024-01-07
Fix interface LanguageDetail > keywords

### LINKED ISSUE #3936: typescript type mismatch

In 11.9.0, use the following code, You will see that the keyword type returned by python is incompatible
```typescript
import hljs from 'highlight.js/lib/core';
import python from 'highlight.js/lib/languages/python';

hljs.registerLanguage('python', python);
```
I found that the same situation exists with dart and swift

patrick-kw-chiu:
Resolves #3936 (though the issue is closed as completed)

### Changes

Added `RegExp` to `keywords`, to reflect the `$pattern` field in `python` and `dart`.

https://github.com/highlightjs/highlight.js/blob/6a52185d9b855130b5acaccef143a7bd602e7885/src/languages/python.js#L148-L154

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ Core Grammars:
 - fix(swift) correctly highlight generics and conformances in type definitions [Bradley Mackey][]
 - enh(swift) add package keyword [Bradley Mackey][]
 - fix(swift) ensure keyword attributes highlight correctly [Bradley Mackey][]
+- fix(types) fix interface LanguageDetail > keywords [Patrick Chiu]
 
 New Grammars:
 
@@ -57,6 +58,7 @@ Themes:
 [Josh Goebel]: https://github.com/joshgoebel
 [CY Fung]: https://github.com/cyfung1031
 [Vitaly Barilko]: https://github.com/Diversus23
+[Patrick Chiu]: https://github.com/patrick-kw-chiu
 
 
 ## Version 11.9.0
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -169,7 +169,7 @@ declare module 'highlight.js' {
         disableAutodetect?: boolean
         contains: (Mode)[]
         case_insensitive?: boolean
-        keywords?: string | string[] | Record<string, string | string[]>
+        keywords?: string | string[] | Record<string, string | string[] | RegExp>
         isCompiled?: boolean,
         exports?: any,
         classNameAliases?: Record<string, string>


joshgoebel:
Hmmm, this type is a bit broad... doesn't TypeScript have a union or combo type or something so we could have the Record but also merge in the very specific `{ $pattern: RegExp }` on top of that?  That would be a lot more accurate.

patrick-kw-chiu:
Make sense.. I guess we can't do something like: _**"All keys have a `string` type, but only a specific key (`$pattern`) have `string | RegExp`"**_. [Source](https://www.reddit.com/r/typescript/comments/114bk9m/how_can_i_specify_a_type_for_1_specific_key_in_a/) and [TypeScript playground 1](https://www.typescriptlang.org/play?#code/C4TwDgpgBA0hIGcoF4oCIDW8DuB7ATgCZpQA+6ARgK4CWANsAPo0B2aA3ALABQPoksHAUIIACgENgwCPhYooAbx5QV6ACRhJ02WgBcUAEoQA5gFEAHmB4BfHn3DQ4IPESSonLkRKky5AMkVlVQBtLBB9BGB8VmMAXQiomLIoSOiWY2DYmztuAGNcFkioMM8EfQ9hN0DuVSgNLV99AHpxJoAaIJUS4X0AcnEUxPTejprVFlwmQggAM1YIQj6B1Jje7O4gA)

The best we can do is to define `{ $pattern: RegExp }`. Then extend it by specifying an exclusive list of keys - `"keyword"` and `"built_in"` to have the `string | string[]` type. [TypeScript playground 2](https://www.typescriptlang.org/play?ssl=1&ssc=35&pln=1&pc=13#code/C4TwDgpgBA0hIGcoF4oCIDW8DuB7ATgCZpQA+6ARgK4CWANsAPo0B2aA3ALABQPoksHAUIIACgENgwCPhYooAbx5QV6ACRhJ02WgD8ALigJg+VgHMyRk+YDaAXUsAlCGYCiADzA8Avjz7hoOBA8IiRUIJCRCSkZOQAyRWVVGywQKFZBRDtDY1MWC3Jc2zsfP24AY1wWYyhUyIQARkMI4TDE7lUoDS1YwwB6cT6AGiSVOuFDGwBycSs8sym7EY7VanomVn7B0t4Kqprx0IAmZqFQ+SUVlW6Y2UmZufNF5c7DwkMHovypl9UWXCYhAgADNWBBCIxUh9Zl8FjsgA). (P.S. it doesn't seem to be able to define in 1 shot. It will yield the `A mapped type may not declare properties or methods` error)

Does it sound more accurate to you? If so, you can let me know the exclusive list of keys in the `keywords` field. I can make the changes accordingly.

joshgoebel:
> The best we can do 

Don't love it... really people should be free to define their own keyword types (in 3rd party grammars) and then provide matching CSS for that... so I think we'll merge this as-is... thanks for the additional research though.

Perhaps this is a good reason to consider lifting that key out of keywords into a top level key such as `keywords_pattern:` or some such... that would at least be simpler WRT to TypeScript I think.

joshgoebel (approved):

patrick-kw-chiu:
Ah, I see what you mean. I orginally thought the fields in `keyword` are exclusive 🙈

If we lift `keywords_patrern` out of `keywords`, it seems to give me an impression that, the behaviour of `keywords_pattern` is different than `keywords`. Not sure is it the case? Other than that, I don't have much comments 👍

joshgoebel:
The pattern is the regex used to identify *possible* keywords before they are matched against the literal list... I thought it was kind of "cool" to prefix it with `$` and make it part of the keywords hash itself - to distinguish it as a config element... I

PR #3964, opened 2024-01-13
Add operator detecting for c/cpp

TOMWT-qwq:
- Add operator detecting for c/cpp.
- Fix `FUNCTION_DECLARATION` not matching overloading operator.

### Changes

- Defined `OPERATOR` constant and added it somewhere.
- Modified something related to `FUNCTION_DECLARATION`.


joshgoebel:
When you say symbol do you mean `operator`?  If so we'd do this by first putting all the operators (including multi-character ones like `!=`) into an array then using our own `regex.either` function to build the actual regex.

joshgoebel:
I cleaned up the PR a bit to better match our coding standards... but I think the reason this hasn't been done before if the issue with `<`... sometimes it's an operator, sometimes it's the start of a template... so that case would need to be properly handled.

Right now there are a ton of failing markup tests... you'll need to go thru them and see which are operator related (which will simply need to be updated) and which are template related.

There is a line (you can uncomment) in the test/markup test code that will just "replace" the test result files with whatever we generate - it can be helpful for updating result files but if you do this please proof all the changes by hand to guarantee they are as expected.

TOMWT-qwq:
Nothing is matched according to the tests. Is `regex.escape(x)` working properly? I can't find its defination from the package downloaded [here](https://highlightjs.org/download)

joshgoebel:
So sorry I forgot to push the main file that does the linking... if pull and build it should work now.

TOMWT-qwq:
ok I'll deal with those tests

tomwt-awa:
Those tests are strong and I've found some bugs :)

Some of the failed tests are not related to cpp. I suppose they are related to the link of `regex.escape`. Do you think it's a good idea to separate these changes into two PRs?

tomwt-awa:
I set its relevance to 0 to make aotodetect working.

Finally I've dealt with most of them

TOMWT-qwq:
Is `->` an operator?

src/languages/cpp.js
```diff
@@ -94,6 +94,52 @@ export default function(hljs) {
...
+    '^',
+    '&',
+    '*',
+    '<',
+    '(?<!-)>',
+    '/',
+    '-(?!>)',
+    '+',
+    '?',
+    '(?<!:):(?!:)',
```
joshgoebel, line 131: Do we have a test case for these?  I'm pretty sure they won't work a these are string literals... `regex.escape` is going to escape all the regex stuff... we'd need to make these raw regex and then teach `escape` to just return raw regex as-is...

We should probably need a new `markup/cpp/operators.txt` test case with very minimal examples of every operator, just stuff like:

```
a == b
b > c
d * f
```

Or i'd even settle for just the full list of operators, one per line.

If there are false positives our regex are working around we'd want those included also - to make sure the behavior is correct.

---

Sadly we can't use negative look-behind until version 12 as it would be a breaking change.  What you can often do in these cases is match the offending pattern with a mode, just not assign it a scope... this effectively "skips over" the content and prevents it from being matches by other rules.

tomwt-awa:
I don't think it's a good idea to deal with these two cases in this PR. They are not related with operator.

- `::` and `->`(and `:` in some special case, I found one in the tests) I think they should be detected as punctuation.
- template `<>`. I think they should be detected as quotes or smth else.

Anyway let's finish this stage of my work. I'll deal with them in another two PRs.

src/languages/cpp.js
```diff
@@ -429,7 +480,28 @@ export default function(hljs) {
       /(?!switch)/,
       /(?!while)/,
       hljs.IDENT_RE,
-      regex.lookahead(/(<[^<>]+>|)\s*\(/))
+      regex.lookahead(/(<[^<>]+>|)\s*\(/)),
```
joshgoebel, line 483: Is this related to the operator detcetion?

src/languages/cpp.js
```diff
@@ -448,6 +521,7 @@ export default function(hljs) {
     // `return some()`, `else if()`, `(x*sum(1, 2))`
     variants: [
       {
+        returnBegin: true,
```
joshgoebel, line 524: We're trying to get away from these, please use a look-ahead instead to avoid capturing the content rather than `returnBegin`.

joshgoebel:
@allejo Any thoughts on whether it might be OK to just accept that `<>` as part of templates are going to be highlighted as operators (because it's too hard to distinguish) - for the increased fidelity of getting operators highlighted in general?  I've long resisted this, but maybe in cases where it wouldn't break highlighting in general we should just embrace it?

>  I'll deal with them in another two PRs.

Let's see what allejo says.  My concern is the template stuff may be very hard or even impossible to fix - so to accept this PR (separately) we'd have to be willing to accept that risk.  So we need to decide we're open to that type of brokeness long-term or if we even want to still consider it broken.

src/languages/cpp.js
TOMWT-qwq: I see.

src/languages/cpp.js
```diff
@@ -122,7 +171,10 @@ export default function(hljs) {
     relevance: 0
   };
 
-  const FUNCTION_TITLE = regex.optional(NAMESPACE_RE) + hljs.IDENT_RE + '\\s*\\(';
+  const FUNCTION_TITLE = regex.either(
+    regex.optional(NAMESPACE_RE) + hljs.IDENT_RE + '\\s*\\(',
+    'operator\\s*' + OPERATOR_RE + '\\s*\\(',
```
joshgoebel, line 176: The string "operator" is not part of a function title... and if you are trying to highlight the operator itself as a `title.function` in these cases it'd be better to use a multi-matcher to do that... see the documentation for `beginMatch`... so we'd match:

`"operator" [ws] [operator]` and scope each segment differently.

src/languages/cpp.js
```diff
@@ -429,7 +480,28 @@ export default function(hljs) {
       /(?!switch)/,
       /(?!while)/,
       hljs.IDENT_RE,
-      regex.lookahead(/(<[^<>]+>|)\s*\(/))
+      regex.lookahead(/(<[^<>]+>|)\s*\(/)),
+    returnBegin: 1,
+    contains:[
```
joshgoebel, line 485: I see now you're rewinding then parsing again, but why did you find this necessary?

test/markup/cpp/function-title.expect.txt
```diff
@@ -1,10 +1,10 @@
-<span class="hljs-function"><span class="hljs-type">int</span> <span class="hljs-title">main</span><span class="hljs-params">(<span class="hljs-type">int</span> argc, <span class="hljs-type">char</span>** argv)</span> </span>{
-  <span class="hljs-keyword">auto</span> a = std::<span class="hljs-built_in">make_unique</span>&lt;A&gt;();
-  B *b = <span class="hljs-keyword">new</span> <span class="hljs-built_in">B</span>();
-  <span class="hljs-type">int</span> c = c * <span class="hljs-built_in">sum</span>(<span class="hljs-number">1</span>, <span class="hljs-number">2</span>);
-  <span class="hljs-keyword">if</span> (a-&gt;<span class="hljs-built_in">check1</span>())
+<span class="hljs-function"><span class="hljs-type">int</span> <span class="hljs-title">main</span><span class="hljs-params">(<span class="hljs-type">int</span> argc, <span class="hljs-type">char</span><span class="hljs-operator">*</span><span class="hljs-operator">*</span> argv)</span> </span>{
+  <span class="hljs-keyword">auto</span> a <span class="hljs-operator">=</span> std<span class="hljs-operator">::</span><span class="hljs-built_in">make_unique</span>&lt;A&gt;();
```
TOMWT-qwq, line 2: I was trying not to match `<>` in this case with that and now I'm changing my mind and will revert that change.

src/languages/cpp.js
TOMWT-qwq:
```cpp
struct test{
	int a,b;
	inline bool operator+(const test&kkk)const{return a+kkk.b;}
};
```

I mean to highlight "operator" because I've noticed this case. I think it's a part of function title.

joshgoebel: Did you try the tools/developer.html tool?  This already works without any additional 
changes as I can tell.

src/languages/cpp.js
joshgoebel: With v12 I'll add a note that this is deprecated in favor of literal lookahead or our `regex.lookahead` helper.

src/languages/cpp.js
TOMWT-qwq: `operator` is detected as a keyword instead of title.

joshgoebel: OH... I see what you're expecting now... hmmm... let me noodle on that...

joshgoebel: I see the intent, but how is it you decided `operator` is part of the 'function name' in this case?  Is that a CPP way of thinking?  Does CPP not officially declare `operator` a keyword?

I think you can make the argument the operator (in this context) is indeed `title.function` scope since it's effectively the "name" of the function, but what leads you to suggest that "operator" is also part of the function name/title? (vs just being a keyword).

TOMWT-qwq: What's your opinion? You mean "operator" is used to decorate the real operator?

TOMWT-qwq: Oh I think you are right after reading smth on cppreference.

TOMWT-qwq: So how to deal with the overloaded operator? `title` or `operator`?

tomwt-awa:
I'm going to ask you what are we going to do next btw.

- I'm planning to add punctuation detect for `;` and `::`.
- I'm planning to add namespace `::` and template `<>` to the defination of a type and a function. I hope it can work for a structure of `a::A<b::B<c::C>>>`.

Any suggestion?

src/languages/cpp.js
tomwt-awa: But there is no `<span class="hljs-function"></span>` now.

tomwt-awa: Currently nothing is working properly :(

src/languages/cpp.js
```diff
@@ -488,10 +537,9 @@ export default function(hljs) {
         relevance: 0
       },
       {
-        begin: FUNCTION_TITLE,
-        returnBegin: true,
-        contains: [ TITLE_MODE ],
-        relevance: 0
+        match: FUNCTION_TITLE,
+        scope: 'title',
+        keywords: "operator",
```
TOMWT-qwq, line 542: 图片
It is working now.

TOMWT-qwq:
You know, a powerful part of hljs is `contains: ["self"]`. However, a regex is not that powerful. I don't think we can recognize everything clearly like a complier now.

PR #3970, opened 2024-01-18
Highlight Erlang char literals

nixxquality:
I've noticed that Erlang's character literals have been ignored and caused havoc in highlighting before, see example fiddle: https://jsfiddle.net/k5xm69vu/1/

~~Strangely enough, the best documentation I could find for this syntax feature was the Erlang module that parses its own syntax... but anyway, here it is: https://www.erlang.org/doc/man/erl_syntax#char-1~~
I was reading another part of the documentation and happened to find it: https://www.erlang.org/doc/reference_manual/data_types#number (the scratched out link above is also helpful as it mentioned the octal syntax)
It's similar to the char literal in C, but the syntax is a $ followed by either (a character) or (a backslash and (a character or up to 3 digits)).
The character can be anything, from alphanumerics to whitespace. If there's a backslash that means you've escaped some value and the 3 digits let you specify an exact octal character number.
There were a few edge cases I had to work out but I believe this is now the final version.

You can compare the markup tests to the following REPL output:
```erlang-repl
Eshell V14.0 (press Ctrl+G to abort, type help(). for help)
1> $a.
97
2> $\t.
9
3> $\011.
9
4> $a85.
* 1:3: syntax error before: 85
4> $aa.
* 1:3: syntax error before: a
4> $123.
* 1:3: syntax error before: 23
4> $\1234.
* 1:6: syntax error before: 4
4> $".
34
5> $\\.
92
6> $[.
91
7> $].
93
8> $ .
32
```

I followed the guide on https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html to assign this new check the "string" class because it's supposed to apply to "literal string, character".

### Checklist
- [x] Added markup tests


nixxquality:
I noticed nesting issues once I highlighted some real code (d'oh) and I believe I've fixed it now by merging together a bit with the earlier work. I added more test cases to cover this.

test/markup/erlang/char_literal.expect.txt
```diff
@@ -0,0 +1,16 @@
+Character = <span class="hljs-string">$a</span>
+TabCharacter = <span class="hljs-string">$\t</span>
+TabCharacterDecimal = <span class="hljs-string">$\011</span>
+IncorrectMixed = <span class="hljs-string">$a</span>85
+IncorrectTooLongChar = <span class="hljs-string">$a</span>a
+IncorrectTooLongNumber = <span class="hljs-string">$1</span>23
+IncorrectTooLongEscapeNumber = <span class="hljs-string">$\123</span>4
```
joshgoebel, line 7: Probably should try not to highlight these at all if they are invalid... could a look-ahead be used to look for following whitespace, etc?

src/languages/erlang.js
```diff
@@ -76,6 +76,10 @@ export default function(hljs) {
       }
     ]
   };
+  const CHAR_LITERAL = {
+    className: 'string',
+    begin: '\\$(\\\\([^0-9]|[0-9]{1,3}|)|.)',
```
joshgoebel, line 81: Please write this as a literal regex using `//` rather than as a string.

src/languages/erlang.js
```diff
@@ -76,6 +76,10 @@ export default function(hljs) {
       }
     ]
   };
+  const CHAR_LITERAL = {
+    className: 'string',
```
joshgoebel, line 80:
```suggestion
    scope: 'string',
```

src/languages/erlang.js
joshgoebel, line 81:
```suggestion
    match: '\\$(\\\\([^0-9]|[0-9]{1,3}|)|.)',
```

test/markup/erlang/char_literal.expect.txt
nixxquality: The reason I'm doing this is because it's a syntax error in the language to write that anyway.
I don't know what your policy is on highlighting syntax errors, but technically the error doesn't occur until *after* the character literal.

If you insist, I will add a look-ahead.

joshgoebel: > I don't know what your policy is on highlighting syntax errors

Generally we don't consider them at all... thanks for the reminder.  That said I don't think we'd need or want these 4 tests cases then - even if we left them with an explanatory comment I'm not sure what purpose they would service - since they are literally syntax error examples.

So I think just remove those 4 examples and we're good.

nixxquality:
Sorry about the delay. I removed the examples that had broken syntax per your request and I also added some more real code that acts as a test case for leaving CHAR_LITERAL out of the BLOCK_STATEMENT array.

joshgoebel (approved):
Looks good, thanks so much!

PR #3975, opened 2024-01-22
Enh cpp add numbers to preprocessor

TOMWT-qwq:
### Checklist


joshgoebel:
I think we need to make a new rule or a variant so that only `define` highlights numbers... not every compiler directive should have this behavior, right?  Trying to think of false positive cases...

PR #3981, opened 2024-01-31
Fix go number highlighting

lugray:
Go number literals are considerably more complex than previously expressed. Most notable, `_` is allowed in several contexts to allow digit grouping, hex numbers can take `p` exponents, and a number with a radix (`.`) requires digits before or after but not both. The additional test cases were pulled from https://go.dev/ref/spec.

### Changes
Add test cases from https://go.dev/ref/spec, and make the tests pass.

### Checklist

### DIFF
--- a/src/languages/go.js
+++ b/src/languages/go.js
@@ -109,23 +109,23 @@ export default function(hljs) {
         className: 'number',
         variants: [
           {
-            match: '-?\\b0[xX]\\.[a-fA-F0-9](_?[a-fA-F0-9])*[pP][+-]?\\d(_?\\d)*i?', // hex without a present digit before . (making a digit afterwards required)
+            match: /-?\b0[xX]\.[a-fA-F0-9](_?[a-fA-F0-9])*[pP][+-]?\d(_?\d)*i?/, // hex without a present digit before . (making a digit afterwards required)
             relevance: 0
           },
           {
-            match: '-?\\b0[xX](_?[a-fA-F0-9])+((\\.([a-fA-F0-9](_?[a-fA-F0-9])*)?)?[pP][+-]?\\d(_?\\d)*)?i?', // hex with a present digit before . (making a digit afterwards optional)
+            match: /-?\b0[xX](_?[a-fA-F0-9])+((\.([a-fA-F0-9](_?[a-fA-F0-9])*)?)?[pP][+-]?\d(_?\d)*)?i?/, // hex with a present digit before . (making a digit afterwards optional)
             relevance: 0
           },
           {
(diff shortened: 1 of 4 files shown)


src/languages/go.js
```diff
@@ -109,10 +109,29 @@ export default function(hljs) {
...
         variants: [
           {
-            begin: hljs.C_NUMBER_RE + '[i]',
-            relevance: 1
+            begin: '-?\\b0[xX]\\.[a-fA-F0-9](_?[a-fA-F0-9])*[pP][+-]?\\d(_?\\d)*i?', // hex without a present digit before . (making a digit afterwards required)
+            relevance: 0
           },
-          hljs.C_NUMBER_MODE
+          {
+            begin: '-?\\b0[xX](_?[a-fA-F0-9])+((\\.([a-fA-F0-9](_?[a-fA-F0-9])*)?)?[pP][+-]?\\d(_?\\d)*)?i?', // hex with a present digit before . (making a digit afterwards optional)
```
joshgoebel, line 116: Please use the newer `match` (instead of `begin`)... and please rewrite these to be literal regex, rather than escaped strings.

Also any reason we can't group the two octals together?

lugray:
> Also any reason we can't group the two octal rules together?

Defining the two together _before_ defining decimals caused us to match the beginnings of what should be matched as decimal/complex floats such as `0.`, `0.15e+0_2`, and `0.i`, and excludes as a number the decimal component.

Defining the two together _after_ defining decimals caused 0o style octals to get matched as a `0` decimal number, leaving the `o...` unhighlighted.

But writing this out made me realize that we don't ever _actually_ match an octal without the `[oO]` as an octal. To the regex, it looks like a decimal. So we can drop the second octal rule.

src/languages/go.js
```diff
@@ -109,10 +109,25 @@ export default function(hljs) {
...
+            match: '-?\\b0[xX]\\.[a-fA-F0-9](_?[a-fA-F0-9])*[pP][+-]?\\d(_?\\d)*i?', // hex without a present digit before . (making a digit afterwards required)
+            relevance: 0
           },
-          hljs.C_NUMBER_MODE
+          {
+            match: '-?\\b0[xX](_?[a-fA-F0-9])+((\\.([a-fA-F0-9](_?[a-fA-F0-9])*)?)?[pP][+-]?\\d(_?\\d)*)?i?', // hex with a present digit before . (making a digit afterwards optional)
+            relevance: 0
+          },
+          {
+            match: '-?\\b0[oO](_?[0-7])*i?', // leading 0o octal
```
joshgoebel, line 120: Need to change all these to literal regex, not escaped stings...

joshgoebel (approved):

PR #3989, opened 2024-02-18
Fix/issue 3972

### LINKED ISSUE #3972: (typescript) Using `namespace` as a variable name breaks highlighter state
**Describe the issue**
When I use `namespace` as the name of the variable, everything that follows is highlighted incorrectly.  While `namespace` is a TS keyword it's fine to use it as a variable name too.

`typescript`

`highlight`

**Sample Code to Reproduce**
```ts
const message = 'foo';
const namespace = 'bar';
function baz() {}
```

https://jsfiddle.net/g0fu1zmw/1/
https://www.typescriptlang.org/play?#code/MYewdgzgLgBAtgUwhAhgcwTAvDA5AMxBFwG4BYAKFEljBUQgAcVhMdcAjFAJ1MvwCuYYFACW4GFwBeACgCUMAN4BfSkA

**Expected behavior**
See GitHub or TS playground renderings above, for example.  It's fine (though not ideal) if `namespace` is highlighted as a keyword in this context, as long as it doesn't break highlighting of everything that follows.

**Additional information**
Original problem report in https://github.com/Reviewable/Reviewable/issues/1134.

akhtarmdsaad:
Problem: **When I use namespace as the name of the variable, everything that follows is highlighted incorrectly. While namespace is a TS keyword it's fine to use it as a variable name too.**
Solution: *Used a multimatcher to solve it*

### Before: 
image
  

### After: 
image

Resolves #3972 

### Changes

Before: 
```
const NAMESPACE = {
    beginKeywords: 'namespace',
    end: /\{/,
    excludeEnd: true,
    contains: [ tsLanguage.exports.CLASS_REFERENCE ]
  };
```

After: 
```
const NAMESPACE = {
    begin: [
      /namespace/,
      /\s+/,
      hljs.IDENT_RE
    ],
    beginScope: {
      1: "keyword",
      3: "variable"
    },
    end: /\{/,
    excludeEnd: true,
    contains: [tsLanguage.exports.CLASS_REFERENCE]
  };
```

### Checklist


src/languages/typescript.js
```diff
@@ -28,10 +28,18 @@ export default function(hljs) {
...
   const NAMESPACE = {
-    beginKeywords: 'namespace',
+    begin: [
+      /namespace/,
+      /\s+/,
+      hljs.IDENT_RE
+    ],
+    beginScope: {
+      1: "keyword",
+      3: "variable"
```
joshgoebel, line 38: From the examples (https://www.typescriptlang.org/docs/handbook/namespaces.html) it looks like namespaces are class-like and CamelCase in their naming... so I think `title.class` would be more appropriate here than variable.

src/languages/typescript.js
```diff
@@ -28,10 +28,18 @@ export default function(hljs) {
...
+    },
     end: /\{/,
     excludeEnd: true,
-    contains: [ tsLanguage.exports.CLASS_REFERENCE ]
+    contains: [tsLanguage.exports.CLASS_REFERENCE]
```
joshgoebel, line 42: Not sure we need any of this anymore - the match rule should be enough to highlight the keyword and namespace name... but should IDENT_RE used above to recognize namespace perhaps be `CLASS_REFERENCE[:match]` instead - if the same names rules apply.

akhtarmdsaad:
```
const NAMESPACE = {
    begin: [
      /namespace/,
      /\s+/,
      hljs.IDENT_RE
    ],
    beginScope: {
      1: "keyword",
      3: "title.class"
    }
  };
```

Is the above change enough or i should change something else.

test/markup/typescript/namespace_as_identifier.expect.txt
```diff
@@ -0,0 +1,4 @@
+
+            <span class="hljs-keyword">const</span> message = <span class="hljs-string">&#x27;foo&#x27;</span>;
+            <span class="hljs-keyword">const</span> <span class="hljs-keyword">namespace</span> = <span class="hljs-string">&#x27;bar&#x27;</span>;
```
joshgoebel, line 3: This isn't the desire behavior... We'll also need to remove it from `TS_SPECIFIC_KEYWORDS` (lets just comment it out - and mention it has a separate rule below).

joshgoebel (approved):

PR #3993, opened 2024-02-19
fix(haxe) fix new keyword false positives

### LINKED ISSUE #3992: (Haxe) using `new` at the start of a variable name highlights it separately from the rest of the variable name
**Describe the issue**

When naming a variable with `new` at the start of the variable name (ex: `newAngle`), it highlights `new` separately from the rest of the variable (highlights `new` one color, then highlights `Angle` another color)

[Haxe](https://haxe.org)

I believe I am using `highlight` / `highlightAll`

**Sample Code to Reproduce**
(Sample code is from the [haxeflixel.com documentation](https://haxeflixel.com/documentation/groundwork/), the hosted documentation is using a bit of custom highlighting, but it's going to move to highlightjs, which you can test on [dev.haxeflixel.com](https://dev.haxeflixel.com/documentation/groundwork/))
```hx
var newAngle:Float = 0;
 if (up)
 {
     newAngle = -90;
     if (left)
         newAngle -= 45;
     else if (right)
         newAngle += 45;
 }
 else if (down)
 {
     newAngle = 90;
     if (left)
         newAngle += 45;
     else if (right)
         newAngle -= 45;
 }
 else if (left)
     newAngle = 180;
 else if (right)
     newAngle = 0;
```

**Expected behavior**

Usually we'd want `new` to be highlighted differently (ex: creating a new variable `state = new FlxState()`), but when it's used as a variable, it should be highlighted as a variable rather than whatever the current implementation is for the `new` keyword.

**Additional context**
This error can be easily tested from the highlightjs online demo with nearly any theme: https://highlightjs.org/demo

ninjamuffin99:
Changes the regex in the haxe.js highlighting to only highlight `new` as a keyword if it's not used within a variable/class name, using `beginKeywords: 'new'` instead of `new *`

Resolves #3992 
### Changes

Chnages the regex to use `beginKeywords: 'new'` instead of *, so that it can capture `new` as a keyword, but not when it's used within a variable name
### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -89,7 +89,7 @@ export default function(hljs) {
       },
       {
         className: 'type', // instantiation
-        begin: /\bnew\b/,
+        beginKeywords: 'new',
         end: /\W/,
         excludeBegin: true,
         excludeEnd: true
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -89,7 +89,7 @@ export default function(hljs) {
       },
       {
         className: 'type', // instantiation
-        begin: /new */,
+        begin: /\bnew\b/,
         end: /\W/,
         excludeBegin: true,
         excludeEnd: true
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,6 @@ CAVEATS / POTENTIALLY BREAKING CHANGES
 - Drops support for Node 16.x, which is no longer supported by Node.js.
 
 Core Grammars:
-
 - enh(perl) fix false-positive variable match at end of string [Josh Goebel][]
 - fix(cpp) not all kinds of number literals are highlighted correctly [Lê Duy Quang][]
 - fix(css) fix overly greedy pseudo class matching [Bradley Mackey][]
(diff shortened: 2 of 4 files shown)


ninjamuffin99:
I will look into the markup tests sometime soon / hopefully later today

src/languages/haxe.js
```diff
@@ -89,7 +89,7 @@ export default function(hljs) {
       },
       {
         className: 'type', // instantiation
-        begin: /new */,
+        begin: /\bnew\b/,
```
joshgoebel, line 92: Would this also work:

```
beginKeywords: 'new',
```

joshgoebel: Depending on what the goal is here a multi-matcher might also be useful... see the docs on `match` for an example... also some of the newer grammers like Ruby or Wren have examples of this in action.  The Haxe grammar looks to be quite old and dusty.

Let me know if you have any questions.

ninjamuffin99: I've updated it to use beginKeywords and works good 🤝

ninjamuffin99:
added markup tests, and changed to use `beginKeywords: new`

joshgoebel (approved):

PR #3998, opened 2024-02-26
Added missing query keyword 'USE' to sql.js

FightRay:
The "USE" keyword should also be highlighted, considering it is used in a query to select a database. For example, "USE DATABASE_NAME;".

### Checklist

### DIFF
--- a/src/languages/sql.js
+++ b/src/languages/sql.js
@@ -445,6 +445,7 @@ export default function(hljs) {
     "upper",
     "user",
     "using",
+    "use",
     "value",
     "values",
     "value_of",


joshgoebel:
Please show some evidence this is part of the official SQL spec itself and not merely just modern dbs.  For modern DBs please use a DB specific grammar, not the included `sql`x.

FightRay:
> Please show some evidence this is part of the official SQL spec itself and not merely just modern dbs. For modern DBs please use a DB specific grammar, not the included `sql`.

There's no specific grammar for MySQL, but here's what I can find: https://www.geeksforgeeks.org/sql-use-database-statement/

joshgoebel:
Not that ChatGPT is always correct, but it agrees with my suspicions here

> The USE statement, commonly found in SQL databases, is used to select a specific database within a database system, making it the current working database for subsequent SQL commands. However, the USE statement is not part of the SQL standard defined by the American National Standards Institute (ANSI) or the International Organization for Standardization (ISO). Instead, it is a vendor-specific extension provided by some database management systems, such as MySQL and Microsoft SQL Server, to facilitate multi-database support within a single server instance.

So that means, we won't add it to our simple SQL grammar.  Someone needs to step up and contribute a 3rd party MySQL grammar... we already have MS SQL and PostgreSQL grammars.

Closing this as wont-fix.

FightRay:
> Not that ChatGPT is always correct, but it agrees with my suspicions here:
> 
> > The USE statement, commonly found in SQL databases, is used to select a specific database within a database system, making it the current working database for subsequent SQL commands. However, the USE statement is not part of the SQL standard defined by the American National Standards Institute (ANSI) or the International Organization for Standardization (ISO). Instead, it is a vendor-specific extension provided by some database management systems, such as MySQL and Microsoft SQL Server, to facilitate multi-database support within a single server instance.
> 
> So that means, we won't add it to our simple SQL grammar. Someone needs to step up and contribute a 3rd party MySQL grammar... we already have MS SQL and PostgreSQL grammars.
> 
> Closing this as wont-fix.

I'll gladly contribute to make a mysql extension soon.
Thanks for the feedback.

PR #3999, opened 2024-02-26
Added list of false positiitives

### LINKED ISSUE #3980: (cpp) Keywords followed by parens are treated as `built_in`
**Describe the issue**
Keywords, such as `static_assert`, `sizeof`, etc. are typically followed by parentheses. `static_assert;` will be highlighted as if it's a keyword, but `static_assert(true);` is highlighted as a built_in.

image

```
<code class="hljs cpp">
<code class="hljs c++">
```

`highlight`, I think?

**Sample Code to Reproduce**

In reveal.js, with a freshly-downloaded hljs, add the following:

```
<pre><code class="hljs c++" data-trim data-noescape data-line-numbers style="overflow: visible;">
  static_assert; // highlighted as keyword
  static_assert(true); // highlighted as built_in
</code></pre>
```

**Expected behavior**

Keywords continue to use their keyword colour when followed by parens.

akhtarmdsaad:
**Issue: (cpp) Keywords followed by parens are treated as built_in**
image

Resolves #3980
### Changes

Added list of false positives to `FUNCTION_DISPATCH`

The List: 
```
      /(?!decltype)/,
      /(?!if)/,
      /(?!for)/,
      /(?!switch)/,
      /(?!while)/,
      /(?!alignas)/,
      /(?!alignof)/,
      /(?!asm)/,
      /(?!catch)/,
      /(?!const_cast)/,
      /(?!dynamic_cast)/,
      /(?!noexcept)/,
      /(?!reinterpret_cast)/,
      /(?!sizeof)/,
      /(?!static_assert)/,
      /(?!static_cast)/,
      /(?!typeid)/,
      /(?!requires)/,
      /(?!explicit)/,
      /(?!case)/,
      /(?!delete)/,
      
      // compound operators
      /(?!bitand_eq)/,
      /(?!bitor_eq)/,
      /(?!xor_eq)/,
      /(?!not_eq)/,
      /(?!or_eq)/,
      /(?!and_eq)/,

      // operators
      /(?!and)/,
      /(?!or)/,
      /(?!bitand)/,
      /(?!bitor)/,
      /(?!xor)/,

      // unary operators
      /(?!not)/,
      /(?!compl)/,
      /(?!co_await)/,
      /(?!co_return)/,
      /(?!co_yield)/,
      
      // Reserved types
      /(?!int)/, 
      /(?!char)/, 
      /(?!double)/, 
      /(?!float)/, 
      /(?!bool)/,
      /(?!auto)/,
```

#### Before: 
image
#### After:
image

### Checklist


tomwt-awa (approved):

src/languages/cpp.js
```diff
@@ -428,10 +428,56 @@ export default function(hljs) {
...
+      /(?!co_await)/,
+      /(?!co_return)/,
+      /(?!co_yield)/,
+      
+      // Reserved types
+      /(?!int)/, 
+      /(?!char)/, 
+      /(?!double)/, 
+      /(?!float)/, 
+      /(?!bool)/,
```
tomwt-awa, line 475: Are these types really working properly?

tomwt-awa: 图片

tomwt-awa (changes requested):

src/languages/cpp.js
akhtarmdsaad: oops i missed that. 
i will fix it properly. 
Thanks for the review.

akhtarmdsaad: image

i think this is already properly. No need to change anything.

joshgoebel:
>  Keywords followed by parens are treated as built_in

Could we build the exclude list dynamically from our keywords list then instead of repeating ourselves?

akhtarmdsaad:
> > Keywords followed by parens are treated as built_in
> 
> Could we build the exclude list dynamically from our keywords list then instead of repeating ourselves?

Sure

akhtarmdsaad:
Its working fine now.

Screenshot 2024-07-09 at 12 01 01 PM

akhtarmdsaad:
hi @joshgoebel, can you please tell me if our engine has a priority based highlighting or it is just limited to regex. if no such feature exists, should we add this feature to our engine for better control over it?

src/languages/cpp.js
TOMWT-qwq: What I've mentioned above only happens without `;`

Just forget it.

TOMWT-qwq (approved):

CHANGES.md
```diff
@@ -61,6 +61,7 @@ CAVEATS / POTENTIALLY BREAKING CHANGES
 
 Core Grammars:
 
+- fix(cpp) Keywords followed by parens are treated as built_in [Md Saad Akhtar][]
```
joshgoebel, line 64: Please add your author link below as well.

joshgoebel (changes requested):
All we need is that link then we can get this merged! Thanks for your patience.

src/languages/cpp.js
```diff
@@ -423,15 +431,11 @@ export default function(hljs) {
       _hint: FUNCTION_HINTS },
     begin: regex.concat(
       /\b/,
-      /(?!decltype)/,
-      /(?!if)/,
-      /(?!for)/,
-      /(?!switch)/,
-      /(?!while)/,
+      // adds all keywords dynamically
+      ...allKeywords.map(keyword => new RegExp(`(?!${keyword})`)),
```
joshgoebel, line 435: Actually don't we only care about keywords here that can potentially have a `(` following them?  That's not ALL keywords, is it?

joshgoebel: We can ignore this for now since I already said this was ready but need your author link...

CHANGES.md
joshgoebel: Ping.

src/languages/cpp.js
```diff
@@ -415,6 +415,14 @@ export default function(hljs) {
     _type_hints: TYPE_HINTS
   };
 
+  const allKeywords = [
+    ...CPP_KEYWORDS.type,
+    ...CPP_KEYWORDS.keyword,
+    ...CPP_KEYWORDS.literal,
+    ...CPP_KEYWORDS.built_in,
```
joshgoebel, line 422: Let's drop `built_in` (and probably `_type_hints`) from this list for now and circle back in a follow-up if needed. Excluding the whole `built_in` table means real callables (`sin`, `printf`, …) never get `function.dispatch`.

For this PR, stick to the false-positive set: `keyword` / `type` / `literal` (or a curated list of control/cast/operator keywords).

src/languages/cpp.js
joshgoebel, line 435: Instead of spreading a huge list of `(?!kw)` regexes (and `new RegExp` per keyword), prefer one negative lookahead with `regex.either`, and a trailing `\b` so prefixes don’t false-negative (`for` vs `foreach`, `int` vs `interrupt`, same class of bug as #4200):

```js
begin: regex.concat(
  /\b/,
  '(?!',
  regex.either(...excludeFromDispatch),
  '\\b)',
  hljs.IDENT_RE,
  regex.lookahead(/(<[^<>]+>|)\s*\(/)
)
```

joshgoebel (changes requested):

PR #4002, opened 2024-02-29
Support namespaced tagged template strings in JavaScript

### LINKED ISSUE #4001: [Request] Support namespaced tagged template strings in JavaScript

[Kitten](https://codeberg.org/kitten/app) exposes `html` and `css` tagged template strings not from the global namespace but from `globalThis.kitten` (so as not to pollute the global namespace).

e.g.,

```js
export default () => kitten.html`
  <h1>I’m not highlighted *sadface*.</h1>
`
```

Currently, highlight.js does not recognise and highlight these as HTML.

I’d love to see the HTML and CSS tagged template string match criteria expanded so they match `.html` and `.css`, not just `html` and `css` so that, for example

```
someNamespace.html`…` 
```

would also be recognised as HTML.

Leaving a space after the dot works since JavaScript is whitespace agnostic but it looks and feels wrong to have to do that just to make the syntax highlighter happy :)

aral:
Resolves #4001

### Changes

Instead of just matching `html` and `css` (and, for completeness and consistency, `gql`), `\.?html`, `\.?css`, and `\.?gql` are matched, which allows for namespaced tagged template strings (e.g., [Kitten](https://codeberg.org/kitten/app)’s `kitten.html`) to be recognised and highlighted.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ CAVEATS / POTENTIALLY BREAKING CHANGES
 
 Core Grammars:
 
+- enh(js/ts) support namespaced tagged template strings [Aral Balkan][]
 - enh(perl) fix false-positive variable match at end of string [Josh Goebel][]
 - fix(cpp) not all kinds of number literals are highlighted correctly [Lê Duy Quang][]
 - fix(css) fix overly greedy pseudo class matching [Bradley Mackey][]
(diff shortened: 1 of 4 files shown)


joshgoebel (approved):

aral:
@joshgoebel Thank you :)

PR #4003, opened 2024-02-29
Add Expect support

OscarCreator:
### Changes
Added Expect language support

### Checklist
- [x] Added markup tests

Expect is an extension of tcl so it detects tcl files as expect files which I don't know how to solve.


src/languages/expect.js
```diff
@@ -0,0 +1,229 @@
...
+          { begin: regex.concat(
+            /\$/,
+            regex.optional(/::/),
+            EXPECT_IDENT,
+            '(::',
+            EXPECT_IDENT,
+            ')*'
+          ) },
+          {
+            begin: '\\$\\{(::)?[a-zA-Z_]((::)?[a-zA-Z0-9_])*',
```
OscarCreator, line 215: I copied this from the tcl file and it seems there is an error which I don't real know how to write in javascript, but I can explain it here:

Variable names can include special characters but only if wrapped in {} 
- https://wiki.tcl-lang.org/page/What+kinds+of+variable+names+can+be+used+in+Tcl

joshgoebel:
Please review our https://highlightjs.readthedocs.io/en/latest/language-contribution.html docs and then open a new issue using the language contributor template to walk you thru the process.

Issue #4004, opened 2024-03-01
(request org repo) highlightjs-unison


runarorama:
I would like to request the creation of a `highlightjs-unison` repository for my 3rd party grammar.

- [x] I have an existing personal repository ready to go

### Metadata

- Grammar Name: _Unison_
- Link to repo: https://github.com/runarorama/highlightjs-unison

### Requirements

- [x] I've read the [language Contributor Checklist](https://highlightjs.readthedocs.io/en/latest/language-contribution.html)
- [x] I've read the [Language Contribution Guide](https://github.com/highlightjs/highlight.js/blob/main/extra/3RD_PARTY_QUICK_START.md)
- [x] I'm currently willing to support and maintain this grammar over time
- [x] That repo includes a `dist` folder with CDN-ready `CJS` and `ESM` builds
- [x] That `README.md` includes examples of simple web usage
- [x] That `README.md` includes examples of Node.js usage

### Afterwards

Submit a PR that:

- [x] adds mention of your new language to `CHANGES.md`
- [x] updates `SUPPORTED_LANGUAGES.md` in the main library

joshgoebel:
Are you wanting us to link to your repo or to host the repo here?

joshgoebel:
Done.  Push your repo on over and I'll merge the other PR.

https://github.com/highlightjs/highlightjs-unison

runarorama:
It seems I don't have permission to push there. Should I make a PR against that repo?

joshgoebel:
You have a pending invite you have not accepted.

runarorama:
Ah yes. Pushed the repo

PR #4005, opened 2024-03-01
Add Unison 3rd party repo to CHANGES.md

runarorama:
### Checklist


joshgoebel:
https://github.com/highlightjs/highlightjs-unison

404

runarorama:
Right, so it was my understanding that once the project is hosted (see #4004) this would be good to go. Please let me know if I need to do something differently.

joshgoebel (approved):

PR #4007, opened 2024-03-03
Add New CSS Properties

carlrafting:
This pull request adds some new CSS properties that are currently missing, this means these properties doesn't get highlighted correctly. An example of this issue can be viewed at [this highlight.js demo page](https://highlightjs.org/demo#lang=css&v=1&code=OnJvb3QgewoJY29sb3Itc2NoZW1lOiBsaWdodCBkYXJrOwoJYWNjZW50LcUiOiByZWTEFHBwZWFyYW5jZTogbm9uZcQTc3BlY3QtcmF0aW86IGF1dG87CglzY2FsZTogMS41OwoJdHJhbnNsYXTKNHJvdMUPOTBkZWc7Cn0K).

I'm not sure what the relevant tests are for a PR like this, so any guidance are very much appreciated. Would `test/markup/css/sample.txt` be the right test file for this?

### References

- [accent-color](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color)
- [appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance)
- [color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme)
- [rotate](https://developer.mozilla.org/en-US/docs/Web/CSS/rotate)
- [scale](https://developer.mozilla.org/en-US/docs/Web/CSS/scale)
- [translate](https://developer.mozilla.org/en-US/docs/Web/CSS/translate)

### Changes

Added some new CSS properties to `ATTRIBUTES` in `src/languages/lib/css-shared.js`.

### Checklist


joshgoebel:
We don't need markup tests for keywords adds... but we still do need changes.md.

joshgoebel (approved):

PR #4023, opened 2024-03-21
Update gml.js to match 2024.2 version of GML language

gnysek:
Updates definitions of GameMakerLanguage to it's latest version, from March 2024.

Based on `fnames` file which is a part of GameMaker runtime (version `2024.2.0.163`) - copy is available here: https://gist.github.com/gnysek/feb6751e2ae0744f2355a8efbd0fff3c .


joshgoebel:
Please update changelog also.

gnysek:
It is already there ( https://github.com/highlightjs/highlight.js/blob/main/CHANGES.md?plain=1#L9 ) except it should be version `11.10.1` or something...

joshgoebel (approved):

PR #4025, opened 2024-03-22
Python self

### LINKED ISSUE #4024: Highlighting the self variable in Python
**Describe the issue**
Many syntax highlighters identify `self` in python as a distinct identifier, similar to `$this` in PHP. Highlight.js does not currently do this.

python

Both

**Sample Code to Reproduce**
```
class SelfTest:
    def __init__(self):
        self.text = True

    def method(self):
        pass
```

**Expected behavior**
(From Sublime Text)
image

**Actual behavior**
image

**Additional context**
While the python grammar definition [recognizes self](https://github.com/highlightjs/highlight.js/blob/8a3c8d5861f31ec0eb92cf092fc3a85cb9315fd6/src/languages/python.js#L376), it does not assign it a scope. In contrast, the PHP definition [assigns](https://github.com/highlightjs/highlight.js/blob/8a3c8d5861f31ec0eb92cf092fc3a85cb9315fd6/src/languages/php.js#L514) the `variable.language` scope to `$this`.

lfalin:
This change assigns a scope to the `self` variable in Python, similar to the PHP definition. This variable was already identified in the grammar parser (presumably for recognition), but no scope was assigned. 

Resolves #4024 

### Changes
- Assigns a scope to the `self` variable in the python language definition.
- Adds a markup test for the above change. 

### Checklist

### DIFF
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -373,6 +373,7 @@ export default function(hljs) {
       NUMBER,
       {
         // very common convention
+        scope: 'variable.language',
         begin: /\bself\b/
       },
       {
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -374,7 +374,7 @@ export default function(hljs) {
       {
         // very common convention
         scope: 'variable.language',
-        begin: /\bself\b/
+        match: /\bself\b/
       },
       {
         // eat "if" prior to string so that it won't accidentally be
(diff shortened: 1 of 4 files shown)


src/languages/python.js
```diff
@@ -373,6 +373,7 @@ export default function(hljs) {
       NUMBER,
       {
         // very common convention
+        scope: 'variable.language',
         begin: /\bself\b/
```
joshgoebel, line 377:
```suggestion
        match: /\bself\b/
```

joshgoebel (approved):

PR #4034, opened 2024-04-06
fix(gcode): stricter gcode line number matching

barthy-koeln:
### Changes

The current G-CODE syntax is too loose on the "line number" instruction matching.

Example:

```gcode
N1 M6
N2 SAVE_CONFIG
```

Rendered as:

```html
<span class="hljs-symbol">N1</span> <span class="hljs-name">M6</span>
<span class="hljs-symbol">N1</span> SAVE_CO<span class="hljs-symbol">NFIG</span>
```

`NFIG` is not a line number instruction or symbol. My proposed fix is to strictly match the letter N, followed by one or more digits.

___

This does not resolve all issues, such as matching `N\d+` that is NOT at the start of the line:

```
SKEW_PROFILE SAVE=GuteN8
```

Rendered as

```html
SKEW_PROFILE SAVE=Gute<span class="hljs-symbol">N8</span>
```

I haven't found a straightforward and reliable specification or source listing exactly which optional characters are allowed *before* line numbers. Unfortunately, I currently don't have the time to dive deeper into ISO specs.

But I think that this a good minimal change that resolves some issues and hopefully doesn't cause others.

### Checklist
- [x] Added markup tests


src/languages/gcode.js
```diff
@@ -51,13 +51,7 @@ export default function(hljs) {
...
     {
       className: 'symbol',
-      variants: [
-        {
-          begin: 'N',
-          end: '\\d+',
-          illegal: '\\W'
-        }
-      ]
+      begin: 'N\\d+'
```
joshgoebel, line 54:
```suggestion
      match: /N\d+/
```

barthy-koeln: Should I change the other 10 regexes with single quotes in that file, too?

barthy-koeln: Update: I found heaps more issues with samples of gcode I could find online. I will submit a PR with substantial changes. Until this, this suggestion was implemented and this PR can stay as small as it is.

barthy-koeln:
I suggest closing this in favor of #4040

joshgoebel:
Closing per suggestion.

PR #4040, opened 2024-04-14
enh(gcode): rewrote language for moden gcode support

barthy-koeln:
Complete rework of the `gcode` language to allow for extended uses cases.
Not all scope rules will apply to all implementations of g-code, but many applications of gcode have added a lot on top of the original spec.

This language implementation aims to be more extensive but still flexible.

My research has used the following documentations:
* [LinuxCNC](http://linuxcnc.org/docs/stable/html/)
* [RepRap](https://reprap.org/wiki/G-code)
* [ISO 6983](https://www.iso.org/standard/34608.html)

And countless code examples extracted from GitHub's search:
* [GitHub Seach For GCode](https://github.com/search?q=path%3A**%2F*.gcode&type=code)
* [GitHub Seach For NGC](https://github.com/search?q=path%3A**%2F*.ngc&type=code)

### Changes

* More keywords
* Stricter numbers matching (c-style numbers are too broad)
* Differentiation between functions (G-Codes, M-Codes, …), Axes (ABC, UVW, XYZ), Parameters (P, Q, R, S, …)
* Semicolon comments
* Quoted strings
* Word commands

#### Question about code

I have used this pattern to re-use complex regexes:

```js
const NUMBER = /[+-]?((\.\d+)|(\d+)(\.\d*)?)/;
const match = new RegExp(`(?<![A-Z])[FHIJKPQRS]\\s*${NUMBER.source}`)
```
I would like to use this for certain duplicated parts (e.g. `(?<![A-Z])`) as well. Is this acceptable?
The performance impact is surely negligible since this only happens during initialisation.
It greatly increases readability and IDE support in my opinion.

#### Screenshots

  |Name|Before|After|
  |---------|---------|--------|
  |Default|default_before|default_after|
  |Extended|extended_before|extended_after|

### Checklist
- [x] Added markup tests


src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
   const GCODE_KEYWORDS = {
-    $pattern: GCODE_IDENT_RE,
-    keyword: 'IF DO WHILE ENDWHILE CALL ENDIF SUB ENDSUB GOTO REPEAT ENDREPEAT '
-      + 'EQ LT GT NE GE LE OR XOR'
-  };
-  const GCODE_START = {
-    className: 'meta',
-    begin: '([O])([0-9]+)'
+    $pattern: /[A-Z]+|%/,
+    meta: [
```
joshgoebel, line 12:
```suggestion
    keyword: [
```

The default keywords for any grammar should be scoped `keyword`.

src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
-    hljs.C_LINE_COMMENT_MODE,
-    hljs.C_BLOCK_COMMENT_MODE,
+    // comments
     hljs.COMMENT(/\(/, /\)/),
-    NUMBER,
-    hljs.inherit(hljs.APOS_STRING_MODE, { illegal: null }),
-    hljs.inherit(hljs.QUOTE_STRING_MODE, { illegal: null }),
+    hljs.COMMENT(/;/, /$/),
+    hljs.APOS_STRING_MODE,
+    hljs.QUOTE_STRING_MODE,
```
joshgoebel, line 71: So we're sure we want the illegal matching?

src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
+    // gcodes
     {
-      className: 'name',
-      begin: '([G])([0-9]+\\.?[0-9]?)'
+      scope: 'title.function',
+      relevance: 10,
+      variants: [
+        // G General functions: G0, G5.1, G5.2, …
+        // M Misc functions: M0, M55.6, M199, …
+        { match: /(?<![A-Z])[GM]\s*\d+(\.\d+)?/ },
```
joshgoebel, line 81: We can't use look-behinds until version 12 (breaking change)... so we'll need another way, or this PR will have to wait till then... no ETA on version 12 yet.

src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
+      match: /-?#\s*\d+/,
     },
+
     {
-      className: 'built_in',
-      begin: '(ATAN|ABS|ACOS|ASIN|SIN|COS|EXP|FIX|FUP|ROUND|LN|TAN)(\\[)',
-      contains: [ NUMBER ],
-      end: '\\]'
+      scope: 'property', // Physical axes
+      match: new RegExp(`(?<![A-Z])[ABCUVWXYZ]\\s*${NUMBER.source}`),
```
joshgoebel, line 113: We try to avoid regex inside strings, you can use our own `regex.join` here instead to do the append and keep both regexes regexes in the source.

src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
-      className: 'attr',
-      begin: '(VC|VS|#)',
-      end: '(\\d+)'
+      scope: 'operator', // N Line number: N1, N2, N1020, …
+      match: /^N\s*\d+/
     },
+
     {
-      className: 'attr',
-      begin: '(VZOFX|VZOFY|VZOFZ)'
```
joshgoebel, line 44: Where did these go are these things GCODE doesn't care about anymore?

src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
+    hljs.APOS_STRING_MODE,
+    hljs.QUOTE_STRING_MODE,
+    hljs.C_NUMBER_MODE,
+
+    // gcodes
     {
-      className: 'name',
-      begin: '([G])([0-9]+\\.?[0-9]?)'
+      scope: 'title.function',
+      relevance: 10,
```
joshgoebel, line 77: All these `relevance` keys can be removed, as we're not really focusing on custom relevance anymore.

src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
+      'SIN',
+      'SQRT',
+      'TAN',
+      'EXISTS'
+    ]
   };
-  const NUMBER = hljs.inherit(hljs.C_NUMBER_MODE, { begin: '([-+]?((\\.\\d+)|(\\d+)(\\.\\d*)?))|' + hljs.C_NUMBER_RE });
+
+
+  const NUMBER = /[+-]?((\.\d+)|(\d+)(\.\d*)?)/;
```
joshgoebel, line 64:
```suggestion
  const NUMBER_RE = /[+-]?((\.\d+)|(\d+)(\.\d*)?)/;
```

src/languages/gcode.js
```diff
@@ -7,58 +7,116 @@
...
-      className: 'name',
-      begin: '([G])([0-9]+\\.?[0-9]?)'
+      scope: 'title.function',
+      relevance: 10,
+      variants: [
+        // G General functions: G0, G5.1, G5.2, …
+        // M Misc functions: M0, M55.6, M199, …
+        { match: /(?<![A-Z])[GM]\s*\d+(\.\d+)?/ },
+        // T Tools
+        { match: /(?<![A-Z])T\s*\d+/ },
```
joshgoebel, line 83: Is it really zero or an unlimited amount of spaces that can follow the first letter?

barthy-koeln: > Spaces and tabs are allowed anywhere on a line of code and do not change the meaning of the
line, except inside comments. This makes some strange-looking input legal. The line `g0x +0. 12
34y 7` is equivalent to `g0 x+0.1234 y7`, for example

In [The NIST RS274NGC Interpreter - Version 3, Section 3.3: Format of a line](https://www.nist.gov/publications/nist-rs274ngc-interpreter-version-3)

AFAIU, any amount of whitespace is ignored.

src/languages/gcode.js
barthy-koeln: Those seem to be implementation-specific for Okuma CNC machines and not language-specific.

src/languages/gcode.js
barthy-koeln: Quoted strings do not appear in the spec at all, but to my knowledge they are common enough to be included here. All implementations I've seen during my research were strict about 1 line equalling 1 command, which IS in the spec.

So I think newlines in quoted strings are a good indicator that the parsed text is not gcode.

src/languages/gcode.js
barthy-koeln: I used `regex.concat` which is probably what you meant to write?

src/languages/gcode.js
```diff
@@ -7,58 +7,115 @@
...
   const GCODE_KEYWORDS = {
-    $pattern: GCODE_IDENT_RE,
-    keyword: 'IF DO WHILE ENDWHILE CALL ENDIF SUB ENDSUB GOTO REPEAT ENDREPEAT '
-      + 'EQ LT GT NE GE LE OR XOR'
-  };
-  const GCODE_START = {
-    className: 'meta',
-    begin: '([O])([0-9]+)'
+    $pattern: /[A-Z]+|%/,
+    keywords: [
```
joshgoebel, line 13:
```suggestion
    keyword: [
```

Singular here. The scope name is `keyword`.

src/languages/gcode.js
```diff
@@ -7,58 +7,115 @@
...
+      'SIN',
+      'SQRT',
+      'TAN',
+      'EXISTS'
+    ]
   };
-  const NUMBER = hljs.inherit(hljs.C_NUMBER_MODE, { begin: '([-+]?((\\.\\d+)|(\\d+)(\\.\\d*)?))|' + hljs.C_NUMBER_RE });
+
+
+  const LETTER_BOUNDARY_RE = /(?<![A-Z])/;
```
joshgoebel, line 65:
```suggestion
// TODO: post v12 lets use look-behind for more accuracy, until then \b should suffice
//  const LETTER_BOUNDARY_RE = /(?<![A-Z])/;
  const LETTER_BOUNDARY_RE = /\b/;
```

barthy-koeln: I was still thinking about that — I unfortunately don't think that `\b` is good enough. It completely breaks the spaceless gcode snippet I included in the markup test, because `\b` includes `[0-9]`, meaning a sequence like `G1A2` does not match `G1` and `A2` separately like it should.

It's pretty common for gcode to be generated by software and not written by humans, and to reduce file size and transmission time to the machines, they strip all whitespace.

I wanted to look into `\b` and an `on:begin` filter  using `response.ignoreMatch()` if a digit is found before the first letter.

Would that be fine or discouraged?

joshgoebel: You'd have to do that per mode or course... we don't use that much because it's SO expensive... but I think if we were willing to remove GCODE from the list of auto-detectable languages with `disableAutodetect` that `on:begin` could then be used... since we'd then be certain we were at least doing all that work in the service of highlighting actual GCODE.

And lets still leave a comment somewhere that the better solution is look-behind - for the future when we can use that.

barthy-koeln:
@joshgoebel thank you for your patience.
 
 My latest commit added `disableAutodetect: true` to the language, as well as the aforementioned `on:begin` filter.
 I added the filter as a variant to full matches with `\b`.
 
 This means that readable gcode with sane spacing will rarely, if ever, run into the callback filter.

___

 I've tested both implementations with the existing markup test, as well as some 100lines of spaceless gcode.
 The results are always within 5% of each other, always in favor of the [v12 language](https://github.com/barthy-koeln/highlight.js/compare/feat/gcode_rework...barthy-koeln:highlight.js:feat/gcode-v12?expand=1) with lookbehinds.

<details>
<summary>Running Benchmark & Results</summary>

#### Shell commands to download v12 and install [benny](https://github.com/caderek/benny)

```shell
wget https://raw.githubusercontent.com/barthy-koeln/highlight.js/2c55db96e8a0523fdbdd6d4069c7007f75d5288b/src/languages/gcode.js -O src/languages/gcode-v12.js
npm run build gcode gcode-v12
cd build
npm install benny

# create benchmark script here e.g. bench.mjs

node bench.mjs
```

#### Benchmark script

```js
import b from 'benny'
import fs from 'fs'
import hljs from 'highlight.js'

const code = fs.readFileSync(import.meta.dirname + '/../test/markup/gcode/extended.txt').toString('utf-8')

b.suite(
  'gcode bench',

  b.add('v11', () => {
    hljs.highlight(code, { language: 'gcode' })
  }),

  b.add('v12', () => {
    hljs.highlight(code, { language: 'gcode-v12' })
  }),

  b.cycle(),
  b.complete(),
  b.save({ file: 'gcode', version: '1.0.0' }),
  b.save({ file: 'gcode', format: 'chart.html' })
)
```

#### Results

gcode-bench
</details>

src/languages/gcode.js
```diff
@@ -7,58 +7,166 @@
...
+    if (matchdata.index === 0) {
+      return;
+    }
+
+    const charBeforeMatch = matchdata.input[matchdata.index - 1];
+    if (charBeforeMatch >= '0' && charBeforeMatch <= '9') {
+      return;
+    }
+
+    response.ignoreMatch();
```
joshgoebel, line 79: If the regex is `![A-Z]` why wouldn't the conditional match?  ie:

```suggestion
    const charBeforeMatch = matchdata.input[matchdata.index - 1];
    if (charBeforeMatch >= 'A' && charBeforeMatch <= 'Z') {
      response.ignoreMatch();
    }
```

joshgoebel: Do we need to deal with lowercase a-z also?

barthy-koeln: If the `begin' with `on:begin` variant was the only matching logic, we'd have to check for `![a-zA-Z]`.

But since we have a variant with `\b` we can be sure that any instance that follows a word boundary `![a-zA-Z0-9_]` has been found already. So we only need to additionally find those that follow `0-9` and ignore everything else.

Doing that means we don't have to account for the case.

I just noticed the underscore `_` isn't handled by that callback, but it should be. I will push another commit (but will wait for feedback on the thing above).

barthy-koeln: Update: I've committed an additional check for underscores, to align the filter with what `\b` and `![A-Z]` mean.
I've added two lines to the markup test and verified in my v1 branch `feat/gcode-v12` that the result is the same.

barthy-koeln:
Hey @joshgoebel, this is just a careful ping to see if I can or should do anything else here?
I need this fix for a freelance project, but can use my fork for the initial release.

joshgoebel (approved):
Sorry this took so very long!

PR #4042, opened 2024-04-23
small changes in DocType of languages and formatting

BhJaipal:
### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,7 @@ Core Grammars:
 - enh(markdown) add entity support [David Schach][] [TaraLei][]
 - enh(css) add `justify-items` and `justify-self` attributes [Vasily Polovnyov][]
 - enh(css) add `accent-color`, `appearance`, `color-scheme`, `rotate`, `scale` and `translate` attributes [Carl Räfting][]
+- fix(fortran) fixes parsing of keywords delimited by dots [Julien Bloino][]
 
 New Grammars:
 
@@ -90,6 +91,7 @@ Themes:
 [Rúnar Bjarnason]: https://github.com/runarorama
 [Carl Räfting]: https://github.com/carlrafting
 [BackupMiles]: https://github.com/BackupMiles
+[Julien Bloino]: https://github.com/jbloino
 
 
 
--- a/src/languages/clojure.js
+++ b/src/languages/clojure.js
@@ -6,7 +6,7 @@ Website: https://clojure.org
 Category: lisp
 */
 
-/** @type {import("highlight.js").LanguageFn} */
+/** @type {LanguageFn} */
 export default function (hljs) {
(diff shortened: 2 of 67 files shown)


src/highlight.js
```diff
@@ -25,7 +25,7 @@ import HTMLInjectionError from "./lib/html_injection_error.js";
 @typedef {import('highlight.js').HLJSPlugin} HLJSPlugin
 @typedef {import('highlight.js').PluginEvent} PluginEvent
 @typedef {import('highlight.js').HLJSOptions} HLJSOptions
-@typedef {import('highlight.js').LanguageFn} LanguageFn
+@typedef {import('highlight.js').LanguageFn} {import("highlight.js").LanguageFn}
```
joshgoebel, line 28: What makes LanguageFn different from all the other types listed here?

BhJaipal: LanguageFn is type of this function that takes HLJSApi type parameter and returns Language type output 

But in DocCommets it wasn't used correctly 
```javascript 
/** @type LanguageFn */
```
It doesn't assign the LanguageFn type to it

So, it should be changed to 
```javascript 
/** @type {import("highlight.js").LanguageFn}
```
It will import all types from highlight.js and then we select LanguageFn type from it and assign it to function to assign type in DocComments, type must be put between {} after @type

BhJaipal: > What makes LanguageFn different from all the other types listed here?

It was his type
I didn't do anything more, I just correctly type assignment

joshgoebel: The whole point of `@typedef {import('highlight.js').LanguageFn} LanguageFn` should be to allow us to use the short names later on in the file - otherwise what's the point in typedef?

BhJaipal: Understood

BhJaipal: > The whole point of `@typedef {import('highlight.js').LanguageFn} LanguageFn` should be to allow us to use the short names later on in the file - otherwise what's the point in typedef?

I have updated everything as you asked

src/languages/abnf.js
```diff
@@ -6,8 +6,8 @@ Category: syntax
 Audit: 2020
 */
 
-/** @type LanguageFn */
-export default function(hljs) {
+/** @type {LanguageFn} */
```
joshgoebel, line 9: I'll need to glance at the docs, but this looks much more acceptable.  But please remove all your other linting changes.

joshgoebel:
Ping.  Please remove all linting changes so I can actually see what this PR includes.

joshgoebel:
Closing for inactivity but if you return and strip out all the linting I'm happy to take a second look.

BhJaipal:
> Closing for inactivity but if you return and strip out all the linting I'm happy to take a second look.  

These are done by vscode, so I can't just undo it all
So, no

joshgoebel:
Understood.  In the future perhaps turn of linting on save to avoid this type of issue.

Issue #4044, opened 2024-04-25
(YAML) Expand allowed characters in keys


heiskr:
**Describe the issue**

👋 Hello hello from GitHub Docs! We use highlight.js and are very grateful for your maintenance.

https://docs.github.com/en/copilot/managing-github-copilot-in-your-organization/configuring-content-exclusions-for-github-copilot#example-of-repositories-and-paths-in-organization-settings

We have a little syntax highlighting problem with a few YAML keys.

![Image](https://github.com/github/docs-engineering/assets/54933897/d9367e4e-3f66-4e7e-9b8f-47f9d6ef6d0d)

You can see GitHub.com (Ruby) handling it a little better:

```yaml
# Ignore all `.env` files in any location.
# This could also have been written on a single line as:
#
# "*": ["**/.env"]
"*":
  - "**/.env"

# In the `octo-repo` repository in this organization:
octo-repo:
  # Ignore the `/src/kernel.rs` file.
  - "/src/kernel.rs"

# In the `primer/react` repository on {% data variables.product.prodname_dotcom %}:
https://github.com/primer/react.git:
  # Ignore any files in or below the `/e2e` directory.
  - "/e2e/**"

# In the `copilot` repository of any {% data variables.product.prodname_dotcom %} organization:
git@github.com:*/copilot:
  # Ignore any files in or below the `/__tests__` directory.
  - "/__tests__/**"
  # Ignore any files in the `/script` directory.
  - "/script/*"

# In the `gitlab-org/gitlab-runner` repository on GitLab:
git@gitlab.com:gitlab-org/gitlab-runner.git:
  # Ignore the `/main_test.go` file.
  - "/main_test.go"
  # Ignore any files in any `packages` directories, anywhere in this repository.
  - "**/packages/*"
  # Ignore any files with names beginning `server` or `session`, anywhere in this repository.
  - "**/{server,session}*"
  # Ignore any files with names ending `.md` or `.mk`, anywhere in this repository.
  - "**/*.m[dk]"
```

I believe the relevant code would be here:

https://github.com/highlightjs/highlight.js/blob/583cb33627474ff577305dc787f9a2a09eaf3b88/src/languages/yaml.js#L16-L30

For our purposes we would need `@` and `*` supported, but I believe there's a wider range of supported characters for YAML keys.

**Which language seems to have the issue?**

YAML

**Are you using `highlight` or `highlightAuto`?**

highlight

**Sample Code to Reproduce**

See description above

[...]

PR #4045, opened 2024-04-29
fix: Expand allowed characters in keys #4044

uniqueeest:
### Changes
This modification extends the definition of the key to include spaces, colon, slash, hyphen, and parentheses. It also changes the key to include more character types while maintaining the pattern that ends in colon.

### Checklist


src/languages/yaml.js
```diff
@@ -20,12 +20,12 @@ export default function(hljs) {
...
     className: 'attr',
     variants: [
-      // added brackets support 
-      { begin: /\w[\w :()\./-]*:(?=[ \t]|$)/ },
-      { // double quoted keys - with brackets
-        begin: /"\w[\w :()\./-]*":(?=[ \t]|$)/ },
-      { // single quoted keys - with brackets
-        begin: /'\w[\w :()\./-]*':(?=[ \t]|$)/ },
+      // added brackets and other special characters support
+      { begin: /\w[\w\s:/.\-()]*:(?=[ \t]|$)/ },
```
joshgoebel, line 24: You just changed the order - other than `' '` and `\s`... why?

And is \s REALLY what we want?  That can include tabs and carriage returns... is there a reason a literal space we not sufficient?

uniqueeest: @joshgoebel I think I got it wrong. I changed it to just add * and @, what do you think?

src/languages/yaml.js
```diff
@@ -20,12 +20,12 @@ export default function(hljs) {
...
-      // added brackets support 
-      { begin: /\w[\w :()\./-]*:(?=[ \t]|$)/ },
+      // added brackets support
+      { begin: /\w[\w :()\./\-*@]*:(?=[ \t]|$)/ },
       { // double quoted keys - with brackets
-        begin: /"\w[\w :()\./-]*":(?=[ \t]|$)/ },
+        begin: /"\w[\w :()\./\-*@]*":(?=[ \t]|$)/ },
       { // single quoted keys - with brackets
-        begin: /'\w[\w :()\./-]*':(?=[ \t]|$)/ },
+        begin: /'\w[\w :()\./\-*@]*':(?=[ \t]|$)/ },
```
joshgoebel, line 28: Can we please keep the `-` at the very end of the character group so we don't need the extra escape?

Also, could we get a test case added to the markup tests to show that this works as intended?

uniqueeest: hello! @joshgoebel. Can you please review this part?

joshgoebel:
How are you generating the expect file?  All literal quotes need to be HTML escaped.

uniqueeest:
@joshgoebel I wrote it.

expect-test
```
<span class="hljs-attr">"*":</span>
```
test
```
"*":
```
But it comes out like the picture I posted.

joshgoebel:
Every `"` should be `&quot;` as the failing test is telling you.  The output from Highlight.js must be HTML encoded.

uniqueeest:
@joshgoebel I only changed the regular expression part from the original code. Should I add encoding in the test code part?

uniqueeest:
@joshgoebel Test file modification completed. It also passed without any problems.

test/markup/yaml/complex-keys.txt
```diff
@@ -1,18 +1,13 @@
-"*":
-  "**/.env"
+"**/.env"
 
-"octo-repo":
```
joshgoebel: Why did we lose all the keys?

uniqueeest: @joshgoebel It wasn't a key value that needed to be tested, but a title for that test😂. I was mistaken at first and added that part to the test.

uniqueeest: @joshgoebel It wasn't a key value that needed to be tested, but a title for that test😂. I was mistaken at first and added that part to the test.

src/languages/yaml.js
```diff
@@ -20,12 +20,12 @@ export default function(hljs) {
   const KEY = {
```
joshgoebel, line 20: Should we add a look-ahead for `:` at the end perhaps to make sure these are keys?

uniqueeest: @joshgoebel Should I add it if I think I need it? How should I do it

joshgoebel:
#4105 has solved this slightly more flexibly.  Sorry.

PR #4047, opened 2024-05-01
(chore) Remove unused submodule

jnooree:
### Changes

Removed unused submodule at the project root. This caused errors on projects which use hljs as a submodule.

### Checklist
- [x] Added markup tests, or they don't apply here because this does not change implementations


jnooree:
Any reason this not being merged in?

joshgoebel (approved):

PR #4049, opened 2024-05-14
Add Gleam to SUPPORTED_LANGUAGES.md

maxdeviant:
This PR adds [Gleam](https://gleam.run/) to the list of supported languages.

The grammar repo is available at [`gleam-lang/gleam-highlight.js`](https://github.com/gleam-lang/gleam-highlight.js).

### Changes

Added an entry for Gleam in `SUPPORTED_LANGUAGES.md`.

### Checklist
- [x] Added markup tests, or they don't apply here because... this is a docs-only change


joshgoebel:
Please also include the `dist` folder (see 3rd party contributor docs) so users have a standard CJS/auto-register version of the grammar they can drop in if they are still using CJS.  This need will disappear with v12, but while the core library is available in both CJS and ES6 module this makes sense.

joshgoebel:
Closing for inactivity.  When `dist` is available on the public repo please re-open and we can get this merged.

PR #4063, opened 2024-06-07
Erlang/OTP 27 enhancements

nixxquality:
### Changes
Erlang/OTP version 27 carried a bunch of syntax enhancements.
They are summarized over on their blog: https://www.erlang.org/blog/highlights-otp-27/
I've split up each feature into its own commit so far.

These are:
- Triple-quoted strings
- doc attribute
- Sigil type
- maybe statement

### Checklist
- [x] Added markup tests

Please let me know how it looks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,7 @@ CONTRIBUTORS
 [aneesh98]: https://github.com/aneesh98
 [BaliBalo]: https://github.com/BaliBalo
 [William Wilkinson]: https://github.com/wilkinson4
+[nixxquality]: https://github.com/nixxquality
 
 
 ## Version 11.10.0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,10 @@ Core Grammars:
 
 - fix(makefile) - allow strings inside `$()` expressions [aneesh98][]
 - enh(css) add all properties listed on MDN (96 additions including `anchor-name`, `aspect-ratio`, `backdrop-filter`, `container`, `margin-trim`, `place-content`, `scroll-timeline`, ...) [BaliBalo][]
+- enh(erlang) OTP 27 triple-quoted strings [nixxquality][]
+- enh(erlang) OTP 27 doc attribute [nixxquality][]
+- enh(erlang) OTP 27 Sigil type [nixxquality][]
+- enh(erlang) OTP25/27 maybe statement [nixxquality][]
 
 New Grammars:
 
--- a/src/languages/erlang.js
+++ b/src/languages/erlang.js
@@ -80,6 +80,11 @@ export default function(hljs) {
     scope: 'string',
     match: /\$(\\([^0-9]|[0-9]{1,3}|)|.)/,
   };
+  const TRIPLE_QUOTE = hljs.END_SAME_AS_BEGIN({
+    scope: 'string',
+    begin: /("{3,})/,
+    end: /("{3,})/,
+  });
 
   const BLOCK_STATEMENTS = {
     beginKeywords: 'fun receive if try case',
@@ -92,6 +97,7 @@ export default function(hljs) {
     hljs.inherit(hljs.APOS_STRING_MODE, { className: '' }),
     BLOCK_STATEMENTS,
     FUNCTION_CALL,
+    TRIPLE_QUOTE,
     hljs.QUOTE_STRING_MODE,
     NUMBER,
     TUPLE,
@@ -106,6 +112,7 @@ export default function(hljs) {
     NAMED_FUN,
     BLOCK_STATEMENTS,
     FUNCTION_CALL,
+    TRIPLE_QUOTE,
     hljs.QUOTE_STRING_MODE,
     NUMBER,
     TUPLE,
@@ -185,6 +192,7 @@ export default function(hljs) {
         contains: [ PARAMS ]
       },
       NUMBER,
+      TRIPLE_QUOTE,
       hljs.QUOTE_STRING_MODE,
       RECORD_ACCESS,
       VAR1,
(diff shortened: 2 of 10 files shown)


src/languages/erlang.js
```diff
@@ -80,9 +80,39 @@ export default function(hljs) {
     scope: 'string',
     match: /\$(\\([^0-9]|[0-9]{1,3}|)|.)/,
   };
+  const TRIPLE_QUOTE = hljs.END_SAME_AS_BEGIN({
```
joshgoebel, line 83: Since it looks like this has no sub-rules, can this be done with pure regex instead?  JS matches are very slow comparatively speaking.

src/languages/erlang.js
```diff
@@ -80,9 +80,39 @@ export default function(hljs) {
...
+  const SIGIL_START_END_PAIRS = {
+    '(': ')',
+    '[': ']',
+    '{': '}',
+    '<': '>',
+  };
+  const SIGIL = {
+    scope: 'string',
+    begin: /~\w?("{3,}|[\(\[\{<\/|'"`#])/,
+    end: /("{3,}|[\)\]\}>\/|'"`#])/,
```
joshgoebel, line 99: This looks like it could have expensive false positives in other languages (auto-detect), so we might need to disable auto-detect for Erlang if we go this direction...

src/languages/erlang.js
nixxquality: Not too familiar with that, I tried `/(?<open>"{3,}).*(\k<open>)/` but it didn't seem to work as I hoped.
Do you know another language syntax I can look at for reference?

joshgoebel: I _think) just the usual `\1` to reference a prior match group should work, you don't need any fancy named groups, etc...  Pretty sure if you dig around in the other grammars for strings you'll see this used in a few places.

src/languages/erlang.js
joshgoebel: That actually might help for languages that use `~` to define variables, etc... but is Erlang the only lisp that uses `~[bracket]` style patterns or would that apply to most LISPs?

nixxquality: There is some more technical information in the feature proposal [EEP 66](https://www.erlang.org/eeps/eep-0066).
It's based on previous work in Elixir (and that's a more popular Erlang-like so you may find information on how other tools deal with this online).

Note that some ideas in that proposal did not make it into the release, namely the `r` prefix for regular expressions.

src/languages/erlang.js
nixxquality: You're right, I found some previous work in csharp.js and used that.

src/languages/erlang.js
joshgoebel: What if we just used individual variants here instead of this matcher?

nixxquality: I thought about that as well. How would I do that in a way that fits the highlight.js code style?
Something like this?
```js
const SIGILS = [
  { ... },
  { ... },
];
// and then instead of SIGIL use
   ...SIGILS
```

joshgoebel: We also have the literal `variants` syntax, but I think `...SIGILS` is fine (as variants just gets expanded to the same thing anyways really).

nixxquality: `variants` is exactly what I was looking for :)

src/languages/erlang.js
```diff
@@ -82,32 +82,25 @@ export default function(hljs) {
   };
   const TRIPLE_QUOTE = {
     scope: 'string',
-    match: /"""("*)(?!")(.|\n)*?"""\1/,
+    match: /"""("*)(?!")[\s\S]*?"""\1/,
```
joshgoebel, line 85: What's the nuance here?

nixxquality: I was just looking at the code for the BACKSLASH_ESCAPE and I thought it made a bit more sense than `(.|\n)`.

nixxquality:
Taking a look at the failing tests now. I was sure I ran a check before I pushed...

nixxquality:
Silly mistake, after I made sure everything worked I went ahead and sorted the variants like the documentation
> Immediately after follows the sigil content; a character sequence between content delimiters. The allowed delimiters are these start-end delimiter pairs: () [] {} <>, or these characters that are both start and end delimiters: / | ' " ` #. [Triple-quote](https://www.erlang.org/doc/system/data_types#tqstring) string delimiters may also be used.
But placing the triple-quote matcher at the end made it prioritise the regular "" matcher above it.

joshgoebel (approved):

joshgoebel:
Looking good.  Rebase and add changelog?

nixxquality:
I hope that looks good.

CHANGES.md
```diff
@@ -9,6 +9,10 @@ Core Grammars:
 
 - fix(makefile) - allow strings inside `$()` expressions [aneesh98][]
 - enh(css) add all properties listed on MDN (96 additions including `anchor-name`, `aspect-ratio`, `backdrop-filter`, `container`, `margin-trim`, `place-content`, `scroll-timeline`, ...) [BaliBalo][]
+- enh(erlang) OTP 27 triple-quoted strings [nixxquality][]
+- enh(erlang) OTP 27 doc attribute [nixxquality][]
+- enh(erlang) OTP 27 Sigil type [nixxquality][]
+- enh(erlang) OTP25/27 maybe statement [nixxquality][]
```
joshgoebel, line 15: Also line ~32 ish for the link...

nixxquality: My bad.

joshgoebel (approved):

PR #4070, opened 2024-07-01
fix: C# raw string literals

jairbubbles:
Fix highlighting with nested double quotes. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string

I tweaked a bit the Regex and it seems to better handle the previous examples: https://regex101.com/r/16Vn7X/1

### DIFF
--- a/src/languages/csharp.js
+++ b/src/languages/csharp.js
@@ -166,7 +166,7 @@ export default function(hljs) {
   };
   const RAW_STRING = {
     className: 'string',
-    begin: /"""("*)(?!")(.|\n)*?"""\1/,
+    begin: /("{3,})(?!")(.|\n)*?\1/,
     relevance: 1
   };
   const VERBATIM_STRING = {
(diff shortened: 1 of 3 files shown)


joshgoebel:
Can you provide a failing test case?  If I try both the old and new regex with the text you provided (at regex101) the results look identical to me - ie the existing regex works as-is.  Am I missing something?

jairbubbles:
Hi @joshgoebel, yeah you're right there's something fishy with my fix.

On the demo, we can see there's an issue, maybe it's not using the latest code?

https://highlightjs.org/demo#lang=csharp&v=1&theme=arta&code=dmFyIE11bHRpTGluZVF1b3RlcyA9ICIiIiIKIMsBIiIiUmF3IHN0cmluZyBsaXRlcmFscyIiIiBjYW4gc3RhcnQgYW5kIGVuZCB3aXRoIG1vcmUgdGhhbiB0aHJlZSBkb3VibGUtccZid2hlbiBuZWVkZWQu0GgiOwo%3D

joshgoebel:
Yeah, we haven't released in a while.  I'll see if I can find some time this weekend.  If you build from source I imagine you'll find the issue fixed.

joshgoebel:
@allejo How do we bump the website to the latest release?

allejo:
Do an npm update against this file with the latest version

https://github.com/highlightjs/highlightjs.org/blob/main/package.json

PR #4073, opened 2024-07-05
(css) add all properties listed on MDN

BaliBalo:
### Changes

This adds all the missing properties listed on MDN. A number of those properties are already widely supported, like `aspect-ratio`.

Full list of added properties:
```
anchor-name
animation-composition
animation-range
animation-range-end
animation-range-start
animation-timeline
aspect-ratio
backdrop-filter
background-position-x
background-position-y
box-align
box-direction
box-flex
box-flex-group
box-lines
box-ordinal-group
box-orient
box-pack
contain-intrinsic-block-size
contain-intrinsic-height
contain-intrinsic-inline-size
contain-intrinsic-size
contain-intrinsic-width
container
container-name
container-type
counter-set
field-sizing
font-optical-sizing
font-palette
font-smooth
font-synthesis-position
font-synthesis-small-caps
font-synthesis-style
font-synthesis-weight
font-variant-alternates
font-variant-emoji
forced-color-adjust
hyphenate-character
hyphenate-limit-chars
initial-letter
initial-letter-align
inset-area
line-height-step
margin-trim
masonry-auto-flow
math-depth
math-shift
math-style
offset
offset-anchor
offset-distance
offset-path
offset-position
offset-rotate
overflow-anchor
overflow-block
overflow-clip-margin
overflow-inline
overlay
overscroll-behavior
overscroll-behavior-block
overscroll-behavior-inline
overscroll-behavior-x
overscroll-behavior-y
page
paint-order
place-content
place-items
place-self
position-anchor
position-visibility
print-color-adjust
ruby-align
ruby-position
scroll-behavior
scroll-timeline
scroll-timeline-axis
scroll-timeline-name
text-decoration-skip
text-size-adjust
text-wrap
text-wrap-mode
text-wrap-style
timeline-scope
touch-action
transition-behavior
user-modify
user-select
view-timeline
view-timeline-axis
view-timeline-inset
view-timeline-name
view-transition-name
white-space-collapse
zoom
```

------

For future reference, an easy way to get the list directly from MDN is using the following snippet in devtools on any CSS page (e.g. the [reference](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference) page):
```
$$('a', $x('//*[@id="sidebar-quicklinks"]//summary[text() = "Properties"]/..')[0]).map(link => link.textContent).filter(prop => !prop.startsWith('-'))
```
This extracts the links from the "Properties" section of the sidebar, while excluding the `-moz-` and `-webkit-` prefixed properties. These do not however list SVG presentation attributes, so it is important to not replace the existing list, but rather to get their union (for example with `[...new Set(currentArray.concat(valuesFromMDN))]`).

### Checklist
- [x] Added markup tests, or they don't apply here because... (keywords update)


CHANGES.md
```diff
@@ -45,6 +45,7 @@ Core Grammars:
 - enh(css) add `select`, `option`, `optgroup`, `picture` and `source` to list of known tags [Vasily Polovnyov][]
 - enh(css) add `inset`, `inset-*`, `border-start-*-radius` and `border-end-*-radius` attributes [Vasily Polovnyov][]
 - enh(css) add `text-decoration-skip-ink`, `text-decoration-thickness` and `text-underline-offset` attributes [Vasily Polovnyov][]
+- enh(css) add all properties listed on MDN (96 additions including `anchor-name`, `aspect-ratio`, `backdrop-filter`, `container`, `margin-trim`, `place-content`, `scroll-timeline`, ...) [BaliBalo][]
```
joshgoebel, line 48: You need to add your link to yourself on lines 72-105, etc...

joshgoebel:
Looks good after we fix attribution.

CHANGES.md
joshgoebel: Well now close to the top since I just did a release - will probably need to merge or rebase your changes.

BaliBalo: Of course ! Sorry I totally missed this

joshgoebel (approved):

Issue #4080, opened 2024-07-23
[Request] Support new digit-separators feature in Dart language


srawlins:
**Is your request related to a specific problem you're having?**

Yes. Highlight.js does not yet support the upcoming [digit-separators feature](https://github.com/dart-lang/language/blob/main/working/digit-separators/feature-specification.md) in Dart.

**The solution you'd prefer / feature you'd like to see added...**

Separate from `C_NUMBER_MODE` to allow for digit separators (underscores).

**Any alternative solutions you considered...**

**Additional context...**

PR #4081, opened 2024-07-23
Update Dart for digit-separators feature

srawlins:
Fixes https://github.com/highlightjs/highlight.js/issues/4080

### Changes

Step away from `C_NUMBER_MODE`, and implement a new NUMBER mode, similar to Python, Ruby, and others.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@ CONTRIBUTORS
 [BaliBalo]: https://github.com/BaliBalo
 [William Wilkinson]: https://github.com/wilkinson4
 [nixxquality]: https://github.com/nixxquality
+[srawlins]: https://github.com/srawlins
 
 
 ## Version 11.10.0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Core Grammars:
 - enh(erlang) OTP 27 doc attribute [nixxquality][]
 - enh(erlang) OTP 27 Sigil type [nixxquality][]
 - enh(erlang) OTP25/27 maybe statement [nixxquality][]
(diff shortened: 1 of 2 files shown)


src/languages/dart.js
```diff
@@ -25,6 +25,15 @@ export default function(hljs) {
     keywords: 'true false null this is new super'
   };
 
+  const NUMBER = {
+    className: 'number',
+    relevance: 0,
+    variants: [
+      { begin: "\\b[0-9][0-9_]*(\\.[0-9][0-9_]*)?([eE][+-]?[0-9][0-9_]*)?\\b" },
+      { begin: "\\b0[xX][0-9A-Fa-f][0-9A-Fa-f_]*\\b" }
```
joshgoebel, line 33: Nice, but please use a regex literal, not a string with tons of extra escapes.  And reach for the new `match` vs the older `begin` (which is more for begin/end pairs).

CHANGES.md
```diff
@@ -13,6 +13,7 @@ Core Grammars:
 - enh(erlang) OTP 27 doc attribute [nixxquality][]
 - enh(erlang) OTP 27 Sigil type [nixxquality][]
 - enh(erlang) OTP25/27 maybe statement [nixxquality][]
+- enh(dart) Support digit-separators in number literals [Sam Rawlins][]
```
joshgoebel, line 16: Need a footnote on lines 32-36 too...

src/languages/dart.js
srawlins: Thanks for the tip! Done

CHANGES.md
srawlins: Done

joshgoebel (approved):

PR #4087, opened 2024-08-01
enh(csharp) add Contextual keywords `file`, `args`, `dynamic`, `nint`, `record`, `required` and `scoped`

### LINKED ISSUE #4050: (C#) `file` keyword/accessibility modifier isn't highlighted
**Describe the issue**
The `file` accessibility modifier is left unhighlighted when applied to top-level types

C#

`highlight`

...

**Sample Code to Reproduce**

```cs
file class K {}
public class K {}
```
[Highlight.js demo](https://highlightjs.org/demo#lang=csharp&v=1&theme=an-old-hope&code=ZmlsZSBjbGFzcyBLIHt9CnB1YmxpY8sS)

**Expected behavior**
The file class keyword shares the same color as other accessibility modifiers such as public and private

**Additional context**
None

alvinsjoy:
Resolves #4050 

### Changes

Add new contextual keywords [file](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/file), [args](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/top-level-statements#args), [dynamic](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/reference-types#the-dynamic-type), [nint](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types#native-sized-integers), [record](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/types/records), [required](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required) and [scoped](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/declarations#scoped-ref).

### Checklist
- [x] Added markup tests, or they don't apply here because no major changes


src/languages/csharp.js
```diff
@@ -131,13 +134,17 @@ export default function(hljs) {
     'join',
     'let',
     'nameof',
+    'nint',
```
joshgoebel, line 137: Are some of these types?  We have a separate list of types and if so we should add types to that list.

joshgoebel:
Ok removed. But wouldn't we also want to add it to the list of data types?

alvinsjoy:
> I've removed `nint` from the contextual keywords. When you mentioned a "list of types," were you referring to the built-in keywords? If so `nint` is already present there. If you were referring to a different list, could you specify which one?

Hi, as I mentioned in the pull request review reply, could you please clarify which list you are referring to as "data types"?

joshgoebel:
Could be named better, but I'm referring to `BUILT_IN_KEYWORDS`.  If you wanted to rename that to TYPES and scope them all as `type` that'd be fine also.

alvinsjoy:
`nint` was already present in `BUILT_IN_KEYWORDS`, so should it be kept as it is for now? Are there any other keywords that has to be moved?

CHANGES.md
```diff
@@ -14,6 +14,7 @@ Core Grammars:
 - enh(erlang) OTP 27 Sigil type [nixxquality][]
 - enh(erlang) OTP25/27 maybe statement [nixxquality][]
 - enh(dart) Support digit-separators in number literals [Sam Rawlins][]
+- enh(csharp) add Contextual keywords `file`, `args`, `dynamic`, `nint`, `record`, `required` and `scoped` [Alvin Joy][]
```
joshgoebel, line 17: Please add link a well on line ~33.

alvinsjoy: Done

joshgoebel (approved):

joshgoebel:
THanks so much!

PR #4088, opened 2024-08-02
(themes) update felipec style

felipec:
### Changes

Update the felipec style to the latest version.

### DIFF
--- a/src/styles/felipec.css
+++ b/src/styles/felipec.css
@@ -7,26 +7,26 @@
 */
 
 .hljs {
-  color: #dddde1;
-  background: #1e1e22;
+  color: #dedde4;
+  background-color: #1d1c21;
 }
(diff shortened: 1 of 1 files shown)


src/styles/felipec.css
```diff
@@ -7,26 +7,26 @@
 */
 
 .hljs {
-  color: #dddde1;
-  background: #1e1e22;
+  color: #dedde4;
+  background-color: #1d1c21;
```
joshgoebel, line 11: Is there a reason for the attribute change?

felipec: No reason. I used `background` initially because `check-theme` failed otherwise, but I fixed that on 0e3735ac.

joshgoebel (approved):

PR #4093, opened 2024-08-18
Fix block parameter handling ruby

tachyons:
Fix block parameter handling

### Changes

in ruby block parameter can be enclosed with |. But ruby also has builtin operator |=. This PR ensure that |= is not marked as be the beginning of the block parameter.
Before: 

Screenshot 2024-08-19 at 5 44 09 AM

After:

Screenshot 2024-08-19 at 5 44 43 AM

### Checklist


joshgoebel:
Awesome, just please add an entry to CHANGES.md.

tachyons:
@joshgoebel Done

joshgoebel (approved):

PR #4094, opened 2024-08-19
fixed hex decimals on C

### LINKED ISSUE #4065: (C/C++) Hex float literal hilit
**Describe the issue**

The rendering of a hexadecimal floating-point constant may stop in the middle of the literal

```cpp
double a = -0x1.638e38e38e38ep5;
```
Live example: https://stackoverflow.com/q/78603151/1556746

Such literals are supported since C++17 / C99.

C/C++ (in stackoverflow what's renderd for the `cpp` "language").

Dunno. Using stackoverflow, and they said at Meta that I should report it here at highlight.js
...

**Sample Code to Reproduce**

```cpp
double a = -0x1.638e38e38e38ep5;
```
though the artifact looks different here than on stackoverflow.

**Expected behavior**

A literal should either be highlightet completey, or not at all.

**Additional context**

Popped up in the context of stackoverflow

Dxuian:
Resolves #4065

### Changes

Changed the regex for hex numbers to accomdate decimal numbers 
(first pr pls be nice 😁😁)
[(https://github.com/highlightjs/highlight.js/issues/4065)](url)
### Checklist

before
after
Checklist
- [x]  Updated the changelog at CHANGES.md

### DIFF
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -57,11 +57,11 @@ export default function(hljs) {
   const NUMBERS = {
     className: 'number',
     variants: [
-      { begin: '\\b(0b[01\']+)' },  
-      { begin: '(-?)\\b([\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)' },  
-      { begin: '(-?)(0[xX][a-fA-F0-9]+(?:\'[a-fA-F0-9]+)*(?:\\.[a-fA-F0-9]*(?:\'[a-fA-F0-9]*)*)?(?:[pP][-+]?[0-9]+)?)' },  
-      { begin: '(-?)\\b\\d+(?:\'\\d+)*(?:\\.\\d*(?:\'\\d*)*)?(?:[eE][-+]?\\d+)?' }  
-    ],
+      { match: /\b(0b[01']+)/ },  
+      { match: /(-?)\b([\d']+(\.[\d']*)?|\.[\d']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)/ },  
+      { match: /(-?)(0[xX][a-fA-F0-9]+(?:'[a-fA-F0-9]+)*(?:\.[a-fA-F0-9]*(?:'[a-fA-F0-9]*)*)?(?:[pP][-+]?[0-9]+)?(l|L)?(u|U)?)/ },  
+      { match: /(-?)\b\d+(?:'\d+)*(?:\.\d*(?:'\d*)*)?(?:[eE][-+]?\d+)?/ }  
+  ],
     relevance: 0
   };
   
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -58,12 +58,12 @@ export default function(hljs) {
     className: 'number',
     variants: [
       { match: /\b(0b[01']+)/ },  
-      { match: /(-?)\b([\d']+(\.[\d']*)?|\.[\d']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)/ },  
+      { match: /(-?)([\d']+(\.[\d']*)?|\.[\d']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)/ },  
       { match: /(-?)(0[xX][a-fA-F0-9]+(?:'[a-fA-F0-9]+)*(?:\.[a-fA-F0-9]*(?:'[a-fA-F0-9]*)*)?(?:[pP][-+]?[0-9]+)?(l|L)?(u|U)?)/ },  
       { match: /(-?)\b\d+(?:'\d+)*(?:\.\d*(?:'\d*)*)?(?:[eE][-+]?\d+)?/ }  
-  ],
+    ],
     relevance: 0
-  };
+  };  
   
   const PREPROCESSOR = {
     className: 'meta',
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -57,14 +57,14 @@ export default function(hljs) {
   const NUMBERS = {
     className: 'number',
     variants: [
-      { begin: '\\b(0b[01\']+)' },
-      { begin: '(-?)\\b([\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)' },
-      { begin: '(-?)(0[xX][a-fA-F0-9]+(?:\'[a-fA-F0-9]+)*(?:\\.[a-fA-F0-9]*(?:\'[a-fA-F0-9]*)*)?(?:[pP][-+]?[0-9]+)?|\\b\\d+(?:\'\\d+)*(?:\\.\\d*(?:\'\\d*)*)?(?:[eE][-+]?\\d+)?)' }
+      { begin: '\\b(0b[01\']+)' },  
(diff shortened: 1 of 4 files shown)


src/languages/c.js
```diff
@@ -54,15 +54,16 @@ export default function(hljs) {
...
-  const NUMBERS = {
-    className: 'number',
-    variants: [
-      { begin: '\\b(0b[01\']+)' },
-      { begin: '(-?)\\b([\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)' },
-      { begin: '(-?)(\\b0[xX][a-fA-F0-9\']+|(\\b[\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)([eE][-+]?[\\d\']+)?)' }
-    ],
-    relevance: 0
-  };
+const NUMBERS = {
```
joshgoebel, line 57: Tabbing/spacing here should note have changed...

src/languages/c.js
```diff
@@ -54,15 +54,16 @@ export default function(hljs) {
     ]
   };
 
-  const NUMBERS = {
-    className: 'number',
-    variants: [
-      { begin: '\\b(0b[01\']+)' },
-      { begin: '(-?)\\b([\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)' },
-      { begin: '(-?)(\\b0[xX][a-fA-F0-9\']+|(\\b[\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)([eE][-+]?[\\d\']+)?)' }
```
joshgoebel, line 62: Where did `[eE]` go in the new code?

joshgoebel:
Also we really should have some tests here to validate both the old behavior and confirm the new behavior.  

Could we perhaps pull over the `number-literals` tests from `cpp` and modify them to be C specific?

Dxuian:
> Could we perhaps pull over the `number-literals` tests from `cpp` and modify them to be C specific?

i have also come up with a new test and expect file for number literals in C since all C++ number literals cannot be directly ported over due to c(gcc) not having digit seprators 
the testing mainly modified the C++ files and its mostly the same 
but i dont know why C doesnt have a separate testing file as of yet despite some compatibility issues in C++ and C code

src/languages/c.js
joshgoebel: Are `p` and `e` exclusive? If so could we split this into two variants?

Dxuian: we could but the new commit i made checks for both using a regex or

Dxuian:
i have also noticed some keywords in the C keywords file that dont belong there for e.g -'auto' these work in c++ and not c.
i would really want to work on correcting the C/C++ thing

src/languages/c.js
joshgoebel: Sure, but is then becomes longer and harder to read. Let's have two different ones here, please.

joshgoebel:
>  C keywords file that dont belong there for e.g -'auto' these work in c++ and not c.

Please make sure your knowledge is 100% up-to-date before offering to start removing keywords.  `auto` is totally valid since C23 and does what you'd expect.

src/languages/c.js
Dxuian: i've split the  regex into two  @joshgoebel please review changes

src/languages/c.js
Dxuian: i have fixed this

joshgoebel:
> I believe we should have a seperate testing file for all others features in C too ... and i urge the maintainers to open a issue for the same

More tests would be welcome, but I don't think we need a separate issue for "add more tests to C"... but perhaps a meta-issue could be good - it would just be a placeholder for all grammars tagged with "good for beginners".  Many grammars could use additional tests.

src/languages/c.js
```diff
@@ -57,14 +57,14 @@ export default function(hljs) {
   const NUMBERS = {
     className: 'number',
     variants: [
-      { begin: '\\b(0b[01\']+)' },
-      { begin: '(-?)\\b([\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)' },
-      { begin: '(-?)(0[xX][a-fA-F0-9]+(?:\'[a-fA-F0-9]+)*(?:\\.[a-fA-F0-9]*(?:\'[a-fA-F0-9]*)*)?(?:[pP][-+]?[0-9]+)?|\\b\\d+(?:\'\\d+)*(?:\\.\\d*(?:\'\\d*)*)?(?:[eE][-+]?\\d+)?)' }
+      { begin: '\\b(0b[01\']+)' },  
+      { begin: '(-?)\\b([\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)' },  
+      { begin: '(-?)(0[xX][a-fA-F0-9]+(?:\'[a-fA-F0-9]+)*(?:\\.[a-fA-F0-9]*(?:\'[a-fA-F0-9]*)*)?(?:[pP][-+]?[0-9]+)?)' },  
+      { begin: '(-?)\\b\\d+(?:\'\\d+)*(?:\\.\\d*(?:\'\\d*)*)?(?:[eE][-+]?\\d+)?' }
```
joshgoebel, line 63: Now if we could please just switch to `match` vs `begin` and literal regex: `/.../` (vs strings)... then this is good to go I think. I'm also happy to do this cleanup myself, just might be a little while.

Dxuian: @joshgoebel i've changed the match to begin , i also managed to correct for a issue in detectign hex longs that went previously undetected please review if any changes are needed
before
after

src/languages/c.js
```diff
@@ -57,13 +57,14 @@ export default function(hljs) {
   const NUMBERS = {
     className: 'number',
     variants: [
-      { begin: '\\b(0b[01\']+)' },
-      { begin: '(-?)\\b([\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)' },
-      { begin: '(-?)(\\b0[xX][a-fA-F0-9\']+|(\\b[\\d\']+(\\.[\\d\']*)?|\\.[\\d\']+)([eE][-+]?[\\d\']+)?)' }
-    ],
+      { match: /\b(0b[01']+)/ },  
+      { match: /(-?)\b([\d']+(\.[\d']*)?|\.[\d']+)((ll|LL|l|L)(u|U)?|(u|U)(ll|LL|l|L)?|f|F|b|B)/ },  
+      { match: /(-?)(0[xX][a-fA-F0-9]+(?:'[a-fA-F0-9]+)*(?:\.[a-fA-F0-9]*(?:'[a-fA-F0-9]*)*)?(?:[pP][-+]?[0-9]+)?(l|L)?(u|U)?)/ },
```
joshgoebel, line 62: Do we want need `\b` here also?

Dxuian: i did see a redundant \b and ive removed it

joshgoebel: I thought these were needed to prevent false positives... like say `g2324` where the `234` would be identified as a number without the `\b`, no?

Dxuian: test
the current behavior seems to work fine but ig we can do with a \b just to be sure but wouldnt that take a performance hit. should i add \b just to be sure or is it fine ?

joshgoebel: Screen Shot 2024-08-24 at 8 39 40 PM

An example of false positives, wouldn't these be valid variable names?  The `\b` is needed to prevent this.

> wouldnt that take a performance hit.

Not really, no - I don't think so - but we don't really talk about regex performance here outside of security concerns (run-away exponential regex, etc).

joshgoebel: > should i add \b

Yep. I think we need it.

Dxuian: image
@joshgoebel  ive fixed the false positive issue please review changes

joshgoebel (approved):

PR #4095, opened 2024-08-19
Resolve the memory leak problem when creating multiple HLJS instances (#4086)

### LINKED ISSUE #4086: Memory leaks when using `newInstance` to create multiple hljs instances

Calling `newInstance` multiple times poses a risk of memory leaks.

see https://github.com/remarkjs/react-markdown/issues/791#issuecomment-1878795432

**Expected behavior**

`newInstance` should not call `window.addEventListener('DOMContentLoaded', boot, false)`.

**Additional context**

In function HLJS:

https://github.com/highlightjs/highlight.js/blob/e1fa2eacf673ef6f8aab2f5c287fa9198edf4fa5/src/highlight.js#L838-L840

Related:

remarkjs/react-markdown#791 rehypejs/rehype-highlight#31

immccn123:
Fixes the memory leak problem mentioned in the above issue, using https://github.com/highlightjs/highlight.js/issues/4086#issuecomment-2266730648 as the solution.

Resolves #4086.

### Changes

The event listener is added to the window only when `highlightAll()` is called, and the listener will be removed by itself when DOM content is loaded.

### Checklist

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -819,24 +819,24 @@ const HLJS = function(hljs) {
    * auto-highlights all pre>code elements on the page
    */
   function highlightAll() {
+    function boot() {
+      // if a highlight was requested before DOM was loaded, do now
+      highlightAll();
+      window.removeEventListener('DOMContentLoaded', boot, false);
+    }
+
     // if we are called too early in the loading process
     if (document.readyState === "loading") {
+      // make sure the event listener is only added once
+      if (!wantsHighlight) {
+        window.addEventListener('DOMContentLoaded', boot, false);
+      }
       wantsHighlight = true;
       return;
     }
 
     const blocks = document.querySelectorAll(options.cssSelector);
     blocks.forEach(highlightElement);
-  }
-
-  function boot() {
-    // if a highlight was requested before DOM was loaded, do now
-    if (wantsHighlight) highlightAll();
-  }
-
-  // make sure we are in the browser environment
-  if (typeof window !== 'undefined' && window.addEventListener) {
-    window.addEventListener('DOMContentLoaded', boot, false);
   }
 
   /**
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -822,7 +822,6 @@ const HLJS = function(hljs) {
     function boot() {
       // if a highlight was requested before DOM was loaded, do now
       highlightAll();
-      window.removeEventListener('DOMContentLoaded', boot, false);
     }
 
     // if we are called too early in the loading process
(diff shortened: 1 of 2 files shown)


src/highlight.js
```diff
@@ -819,8 +819,18 @@ const HLJS = function(hljs) {
    * auto-highlights all pre>code elements on the page
    */
   function highlightAll() {
+    function boot() {
+      // if a highlight was requested before DOM was loaded, do now
+      highlightAll();
+      window.removeEventListener('DOMContentLoaded', boot, false);
```
joshgoebel, line 825: What is the thinking here?  Is this type of cleanup truly necessary?

immccn123: This cleanup might only matter in rare cases, like when an `hljs` instance is used just once and `highlightAll` is called before `DOMContentLoaded`. In such a scenario, the instance would stay in memory, unable to be released. If this corner case isn't a concern, the cleanup is likely unnecessary.

If this contradicts the core philosophy, the cleanup can be removed.

joshgoebel: I'm just going to call it unnecessary -- unless you have some docs (from MDN or such) that says this is somehow best practice... I've seen tons of `onload` hooks like this over the years but never code to unwind a single-use event listner afterwards - seems a browser concern... if the browser knows that event that just fired can never be fired again on the same page it could do the cleanup itself.

immccn123: I understand your point. While the event could technically be triggered using `window.dispatchEvent(new Event("DOMContentLoaded"))`, which isn't common practice, this means the browser wouldn’t automatically handle the cleanup. However, another approach to ensure the listener is only triggered once and automatically cleaned up is to use `window.addEventListener(..., { once: true })`. Is this one better, or just remove the cleanup?

Would you prefer switching to `{ once: true }`, or should we remove the cleanup entirely?

joshgoebel: https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event

I find zero examples of `once` being suggested.  Lets just remove the cleanup.

immccn123: Thanks for the clarification and patience! I've removed the cleanup.

joshgoebel:
Awesome, now can you just add an entry to CHANGES and then I think this is good to merge.

immccn123:
Changelog entry has been added as requested. Thanks for your guidance throughout this process!

joshgoebel (approved):

erha19:
@immccn123 hi, is there any new version for this fix?

joshgoebel:
Fixed in 11.11.0.

PR #4098, opened 2024-08-27
fixed 'foreign key' and 'primary key' do not highlight 'key'.

### LINKED ISSUE #4057: (sql) 'foreign key' and 'primary key'  do not highlight 'key'.
**Describe the issue**
When I write `foreign key` or `primary key` is sql language, `primary` or `foreign` is highlighted, however `key` is not.

sql.

`highlightAll()`.

**Sample Code to Reproduce**

```
<!DOCTYPE html>
<html lang="en">
<head>
<title>test</title>
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/styles/agate.min.css">
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/highlight.min.js"></script>
<script>
hljs.highlightAll();
</script>
</head>
<body>
<pre>
<code class="sql">
create table Book (
    bookId      int,
    publisherId int,
    primary key(bookId),
    foreign key(PublisherId) references Publisher(id)
);
</code>
</pre>
</body>
</html>
```

**Expected behavior**
'key' is also highlighted.

**Additional context**
Even though I add the following to my script to force the word 'key' as keyword,

```
const custom_keywords = ['key'];
hljs.getLanguage('sql').keywords.keyword.push(...custom_keywords);
```
`key` won't be highlighted.  (However this addition will now highlights `foreign  key` or `foreign&nbsp;key`  `primary&#8288; key` etc when the separator is not a single space.)

I think the regexp using in COMBO does not recognize space as a part of keyword, that is the problem.
Supposing the intention of this code is to treat `foreign key` as one keyword.
In my script, changing the regexp so as to include space character

```
hljs.getLanguage('sql').contains[0].keywords.$pattern = /\b[\w\. ]+/;
```

solve the problem.  

However I have not understand yet why `create table` is working as intended, so I am not sure this is correct fix.

Also it does not highlights when I write `foreign  key` using 2 spaces etc, unless I added the `key` in keyword.

Though I have not understand the whole code yet, my suggestion would be the regexp at

https://github.com/highlightjs/highlight.js/blob/15d3b627fa7c99cb98d7b6760a6fbdbfd519d1a0/src/languages/sql.js#L660

should be changed to recognize space as a part of keyword. (Also it need to add `key` in reserved word list 

https://github.com/highlightjs/highlight.js/blob/15d3b627fa7c99cb98d7b6760a6fbdbfd519d1a0/src/languages/sql.js#L99

in case there are 2 or more whitespaces `between` foreign and `key`.)

Dxuian:
Resolves #4057
before 
image

after 
image

### Checklist

### DIFF
--- a/src/languages/sql.js
+++ b/src/languages/sql.js
@@ -622,6 +622,35 @@ export default function(hljs) {
     keywords: { built_in: FUNCTIONS }
   };
 
+  //COMBOS generator
+  // const regexPatterns = COMBOS.map(phrase => {
+  //   const escapedPhrase = phrase.replace(/ /g, "\\s+"); // Replace spaces with \s+ to match any whitespace
+  //   return new RegExp(escapedPhrase, "gi"); // Create case-insensitive regex
+  // });
+
+  const COMBOSLIST = {
+    className: 'type',
+    variants: [
+      { match: /\bcreate\s+table\b/ },
+      { match: /\binsert\s+into\b/ },
+      { match: /\bprimary\s+key\b/ },
+      { match: /\bforeign\s+key\b/ },
+      { match: /\bnot\s+null\b/ },
+      { match: /\balter\s+table\b/ },
+      { match: /\badd\s+constraint\b/ },
+      { match: /\bgrouping\s+sets\b/ },
+      { match: /\bon\s+overflow\b/ },
+      { match: /\bcharacter\s+set\b/ },
+      { match: /\brespect\s+nulls\b/ },
+      { match: /\bignore\s+nulls\b/ },
+      { match: /\bnulls\s+first\b/ },
+      { match: /\bnulls\s+last\b/ },
+      { match: /\bdepth\s+first\b/ },
+      { match: /\bbreadth\s+first\b/ }
+    ],
+    relevance: 0
+  };
+  
   // keywords with less than 3 letters are reduced in relevancy
   function reduceRelevancy(list, {
     exceptions, when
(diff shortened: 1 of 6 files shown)


joshgoebel:
_Except `key` isn't a keyword on it's own._ 

To properly fix this we need a small helper function to run thru COMBOS and generate a list of separate keywords that's then would be used only for the COMBO matcher.

Dxuian:
> To properly fix this we need a small helper function to run thru COMBOS and generate a list of separate keywords that's then would be used only for the COMBO matcher.
@joshgoebel  i managed to make a helper funcction but i want you to look at it to see if it seems acceptable. because it does seem to work
before :
image

result/after :
image

src/languages/sql.js
```diff
@@ -655,11 +661,11 @@ export default function(hljs) {
     },
     contains: [
       {
-        begin: regex.either(...COMBOS),
+        begin: regex.either(...generateComboKeywords(COMBOS)),
```
joshgoebel, line 664: This can't use the split version or it'll give the same false positives on `key`.... only the "internal" `keywords` needs the full expanded list (for matching individual words inside larger keyword combos)... but I have a feeling then we're going to find that compound keywords spread across multiple lines don't work, ugh.

Dxuian: i believe ive managed to do it 
ive even added support for newline but i had to make a new comboslist @joshgoebel  please review 
now 
image

PR #4099, opened 2024-08-27
Improve documentation: include usage of ES6 Modules in browser

sudo-barun:
### Changes

1. change "Download prebuilt CDN assets" from h4 to h3
2. include usage of ES Modules in browser

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,13 +292,14 @@ import 'highlight.js/styles/github.css';
 
 ### Browser ES6 Modules
 
-*Note: You need to install `@highlightjs/cdn-assets` package instead of `highlight.js`. See [Download prebuilt CDN assets](#download-prebuilt-cdn-assets)*
+*Note: For now you'll want to install `@highlightjs/cdn-assets` package instead of `highlight.js`.
+See [Download prebuilt CDN assets](#download-prebuilt-cdn-assets)*
 
 To import the library and register only those languages that you need:
 
 ```js
-import hljs from './node_modules/@highlightjs/cdn-assets/es/core.js';
-import javascript from './node_modules/@highlightjs/cdn-assets/es/languages/javascript.min.js';
+import hljs from './assets/js/@highlightjs/cdn-assets/es/core.js';
+import javascript from './assets/js/@highlightjs/cdn-assets/es/languages/javascript.min.js';
 
 hljs.registerLanguage('javascript', javascript);
 ```
@@ -306,10 +307,12 @@ hljs.registerLanguage('javascript', javascript);
 To import the library and register all languages:
 
 ```js
-import hljs from './node_modules/@highlightjs/cdn-assets/es/highlight.js';
+import hljs from './assets/js/@highlightjs/cdn-assets/es/highlight.js';
 ```
 
-*Note: The above code works only when the file containing the JavaScript code and the `node_modules` folder are inside same folder. If that is not the case, you have to modify the path in import statement accordingly.*
+*Note: The path to these files will depend on how your individiaul project is configured,
+the path `./assets/js` above is only an example.  You'll likely need to modify the path
+for your own unmique setup.*
 
 You can also use [`importmap`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) to import in similar way as Node:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -292,6 +290,50 @@ import hljs from 'highlight.js';
...
 
+### Browser ES6 Modules
+
+*Note: You need to install `@highlightjs/cdn-assets` package instead of `highlight.js`. See [Download prebuilt CDN assets](#download-prebuilt-cdn-assets)*
+
+To import the library and register only those languages that you need:
+
+```js
+import hljs from './node_modules/@highlightjs/cdn-assets/es/core.js';
+import javascript from './node_modules/@highlightjs/cdn-assets/es/languages/javascript.min.js';
```
joshgoebel, line 301: Isn't this assuming a lot about someone's file structure?  I feel like publishing `node_modules` onto a public site is a bit of a smell, is that normal these days?

joshgoebel: Ah you addressed that below... I'll mull on this.

sudo-barun: Thanks.

First of all, use of web technologies (HTML, CSS, JS) is no longer limited to website/webapp. With the help of Electron, web technologies can used in desktop app as well. This means importing from `node_modules` is no longer a concern in desktop app.

Now, about the public site, you are right that it is not normal (or lets say, not common in practice) to publish `node_modules`. I think it is because of lack of awareness about different ways npm packages can be installed in a project. Most of people think that `package.json` stays at the root directory and all packages should be installed at the root directory. However, we can create separate `package.json` at public directory for browser specific packages.

```
/
/package.json          (may contain server-specific packages like Express, or may only contain devDependencies like Webpack in case the app is not using Node for server (for e.g. PHP app))
/node_modules/
/public/package.json          (contains browser-specific packages like highlight.js, bootstrap)
/public/node_modules/
/public/index.html
```

This way, packages at top level and any other server specific code does not get exposed to the public.

There is a change of security issue if a npm package contain file that can be directly executed by server (for e.g. PHP file). In that case, a CVE should be created.

Lastly, we can generate different code for the `prod` environment. Tools like [Vite](https://vitejs.dev/guide/build) may be helpful regarding that.  This way, we can use original code for development and generated code for production.

I can also update the doc to include warning:
```
Warning: publishing `node_modules` can lead to security problem. Use with caution.
```

If you think it is not a good idea at all to have `node_modules` being publicly exposed, I can update the doc to manually copy code from `@highlightjs/cdn-assets` package and paste in public directory (or create symlink).

Let me know what you think.

README.md
```diff
@@ -292,6 +290,50 @@ import hljs from 'highlight.js';
...
 ```
 
+### Browser ES6 Modules
+
+*Note: You need to install `@highlightjs/cdn-assets` package instead of `highlight.js`. See [Download prebuilt CDN assets](#download-prebuilt-cdn-assets)*
+
+To import the library and register only those languages that you need:
+
+```js
+import hljs from './node_modules/@highlightjs/cdn-assets/es/core.js';
```
joshgoebel, line 300: What is the `./`?  Is this assuming a build system or nobuild?  Don't real in-browser ES6 imports need to be full URLs?

sudo-barun: You need to check this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#importing_features_into_your_script
Also check below for using bare specifier (`import hljs from '@highlightjs/cdn-assets/es/core.js';`) using importmap.
Code in `@highlightjs/cdn-assets` package is browser compatible, so nobuild.

tagliala: Hello, is this change somehow related to https://github.com/rails/importmap-rails/issues/270 ?

joshgoebel: I'm familiar with this actually, I was thinking of someone pulling from from a CDN I suppose...   I'll look at this and tweak the examples.  I'm not going to recommend that people use `node_modules` though...

I was also thinking of the FIRST import, which is (surely?) going to be relative to the HTML file... relative references for assets are such a pain (IMHO) on a large website.  I imagine named imports vs importmap solve a lot of that though.

Taking a look at this more now.

joshgoebel:
Thoughts?

README.md
```diff
@@ -292,42 +292,46 @@ import 'highlight.js/styles/github.css';
...
 
 To import the library and register all languages:
 
 ```js
-import hljs from './node_modules/@highlightjs/cdn-assets/es/highlight.js';
+import hljs from './assets/js/@highlightjs/cdn-assets/es/highlight.js';
 ```
 
-*Note: The above code works only when the file containing the JavaScript code and the `node_modules` folder are inside same folder. If that is not the case, you have to modify the path in import statement accordingly.*
+*Note: The path to these files will depend on how your individiaul project is configured,
```
sudo-barun, line 313: In addition to this, I think it is better to be more explicit by including instruction to copy `@highlightjs/cdn-assets` from node_modules (or CDN) into `assets/js/@highlightjs/cdn-assets`.

sudo-barun, line 313: Seems like there is a typo here as well.

joshgoebel (approved):

PR #4104, opened 2024-08-30
fixed css escape chars

### LINKED ISSUE #3965: (css) Escaped special character in selector class name breaks highlighting
**Describe the issue**

CSS class and ID selectors contain arbitrary strings (e.g. `.myclassname`), which can contain escaped special characters (e.g. `.foo\:myclassname` to match `class="foo:myclassname"`). Currently, the backslash that escapes a special character is not recognised by highlight.js as a valid part of the class name or ID.

css

`highlight`

**Sample Code to Reproduce**

```css
.dark\:hover\:bg-sky-500:hover {
  --tw-bg-opacity: 1;
  background-color: rgb(14 165 233 / var(--tw-bg-opacity));
}
```

Current highlighting:

CleanShot 2024-01-16 at 11 45 54

**Expected behavior**

The escaped special characters (`\:`) should not interfere with the highlighting of the class name, but actual pseudoclasses (`:hover`) should still be highlighted.

GitHub:

CleanShot 2024-01-16 at 11 43 34

VS Code:

CleanShot 2024-01-16 at 11 45 17

**Additional context**

TailwindCSS makes extensive use of escaped characters in CSS selectors for its [modifiers](https://tailwindcss.com/docs/hover-focus-and-other-states).

Other special characters are escaped for features like [color opacity](https://tailwindcss.com/docs/text-color#changing-the-opacity):

```css
.dark\:hover\:bg-sky-500\/25:hover {
  background-color: rgb(14 165 233 / 0.25);
}
```

…and [arbitrary values](https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values):

```css
.dark\:hover\:bg-\[\#ff0000\]\/25:hover {
  background-color: rgb(255 0 0 / 0.25);
}
```

Dxuian:
Resolves #3965
(fixed the branch issue)
before
image

after
image

### Checklist

### DIFF
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -14,7 +14,7 @@ export default function(hljs) {
   const VENDOR_PREFIX = { begin: /-(webkit|moz|ms|o)-(?=[a-z])/ };
   const AT_MODIFIERS = "and or not only";
   const AT_PROPERTY_RE = /@-?\w[\w]*(-\w+)*/; // @-webkit-keyframes
-  const IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*';
+  const IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*(\\\\:[a-zA-Z0-9_-]+)*';
   const STRINGS = [
     hljs.APOS_STRING_MODE,
     hljs.QUOTE_STRING_MODE
@@ -37,7 +37,7 @@ export default function(hljs) {
       modes.CSS_NUMBER_MODE,
       {
         className: 'selector-id',
-        begin: /#[A-Za-z0-9_-]+/,
+        begin: '#' + IDENT_RE,
         relevance: 0
       },
       {
@@ -49,8 +49,8 @@ export default function(hljs) {
       {
         className: 'selector-pseudo',
         variants: [
-          { begin: ':(' + css.PSEUDO_CLASSES.join('|') + ')' },
-          { begin: ':(:)?(' + css.PSEUDO_ELEMENTS.join('|') + ')' }
+          { begin: '(?<!\\\\):(' + css.PSEUDO_CLASSES.join('|') + ')'},
+          { begin: '(?<!\\\\):(:)?(' + css.PSEUDO_ELEMENTS.join('|') + ')'}
         ]
       },
       // we may actually need this (12/2020)
--- a/src/languages/css.js
+++ b/src/languages/css.js
@@ -13,8 +13,8 @@ export default function(hljs) {
   const modes = css.MODES(hljs);
   const VENDOR_PREFIX = { begin: /-(webkit|moz|ms|o)-(?=[a-z])/ };
   const AT_MODIFIERS = "and or not only";
-  const AT_PROPERTY_RE = /@-?\w[\w]*(-\w+)*/; // @-webkit-keyframes
-  const IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*(\\\\:[a-zA-Z0-9_-]+)*';
+  const AT_PROPERTY_RE = /@-?\b\w[\w-]*\b/; // @-webkit-keyframes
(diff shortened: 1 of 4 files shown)


Dxuian:
@joshgoebel  any input ?

src/languages/css.js
```diff
@@ -14,7 +14,7 @@ export default function(hljs) {
   const VENDOR_PREFIX = { begin: /-(webkit|moz|ms|o)-(?=[a-z])/ };
   const AT_MODIFIERS = "and or not only";
   const AT_PROPERTY_RE = /@-?\w[\w]*(-\w+)*/; // @-webkit-keyframes
-  const IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*';
+  const IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*(\\\\:[a-zA-Z0-9_-]+)*';
```
joshgoebel, line 17: Is `::` escaping supported? Or escaping any other chars?  Feels like this is just handling this one edge case vs the concept of escaping in general - are there other uses?

src/languages/css.js
```diff
@@ -49,8 +49,8 @@ export default function(hljs) {
       {
         className: 'selector-pseudo',
         variants: [
-          { begin: ':(' + css.PSEUDO_CLASSES.join('|') + ')' },
-          { begin: ':(:)?(' + css.PSEUDO_ELEMENTS.join('|') + ')' }
+          { begin: '(?<!\\\\):(' + css.PSEUDO_CLASSES.join('|') + ')'},
+          { begin: '(?<!\\\\):(:)?(' + css.PSEUDO_ELEMENTS.join('|') + ')'}
```
joshgoebel, line 53: These (look-behind) are problematic until we hit v12 - as this would be a breaking change to introduce to v11 since it would change which browsers v11 supports (Safari was very late to this party).

test/markup/css/pseudo-selector.txt
```diff
@@ -22,3 +22,47 @@ p::first-letter {
   font-weight: bold;
   color: brown;
 }
+.dark\:hover\:bg-sky-500:hover {
```
joshgoebel, line 25: Two or three (tops) of these longer examples should be sufficient unless there is some big difference here I'm missing.

Dxuian: okay ill reduce the newer tc

src/languages/css.js
Dxuian: ive managed to generalize to include other special chars in the new commit pls review

src/languages/css.js
Dxuian: fixed it in the new pr

src/languages/css.js
```diff
@@ -13,8 +13,9 @@ export default function(hljs) {
   const modes = css.MODES(hljs);
   const VENDOR_PREFIX = { begin: /-(webkit|moz|ms|o)-(?=[a-z])/ };
   const AT_MODIFIERS = "and or not only";
-  const AT_PROPERTY_RE = /@-?\w[\w]*(-\w+)*/; // @-webkit-keyframes
-  const IDENT_RE = '[a-zA-Z-][a-zA-Z0-9_-]*';
+  const AT_PROPERTY_RE = /@-?\b\w[\w-]*\b/; // @-webkit-keyframes
```
joshgoebel, line 16: What are the changes here trying to do?

Dxuian:
@joshgoebel ?

joshgoebel:
Sorry for the delay.  While coming back to this I just remembered that we support many CSS-like grammars - and many of those allow plain old CSS as well as their other syntactic improvements.

SO... to accept a `css` improvement you would also need to support:

- less
- stylus
- scss

There is a script to help with this. 

- You'd first need to add a small representative sample set of escapes to `markup/css/css_consistency.txt`... 
  - with a paired result file... 
- then use the `tools/css` script - 
  - which will copy the consistency test (as-is) into the `less`, `stylus`, and `scss` folders.

After that you need to work with the less, stylus, and scss grammars to get the markup tests passing - ie their on CSS output must exactly match the output of the CSS grammar.

joshgoebel:
Also perhaps look at the `css_shared.js` file, which might be useful for shared regex, etc.

joshgoebel:
Closing for inactivity.

PR #4106, opened 2024-08-30
fixed the spacing issue in primary key and similar combos

Dxuian:
Resolves #4057

before 
image

after 
image

### Checklist
- [x] Added markup tests, or they don't apply here because...
- [x] Updated the changelog at `CHANGES.md`

PR #4105, opened 2024-08-30
fix for special chars in yaml

Dxuian:
Resolves #4044
### Changes

added special char support to YAML keys via updating regex
before
image
after 
image
### Checklist
- [x] Added markup tests, or they don't apply here because...
- [ ] Updated the changelog at `CHANGES.md`

PR #4107, opened 2024-08-30
Fixed highlight C++ std types in C 

### LINKED ISSUE #4103: (C) Don't highlight C++ std types
**Describe the issue**
C++ std types are highlighted in C sources.

`C`.

highlight.

**Expected behavior**
C++ std types should't be highlighted in C sources.

**Additional context**
From `src/languages/c.js`:
```js
  const KEYWORDS = {
    keyword: C_KEYWORDS,
    type: C_TYPES,
    literal: 'true false NULL',
    // TODO: apply hinting work similar to what was done in cpp.js
    built_in: 'std string wstring cin cout cerr clog stdin stdout stderr stringstream istringstream ostringstream '
      + 'auto_ptr deque list queue stack vector map set pair bitset multiset multimap unordered_set '
      + 'unordered_map unordered_multiset unordered_multimap priority_queue make_pair array shared_ptr abort terminate abs acos '
      + 'asin atan2 atan calloc ceil cosh cos exit exp fabs floor fmod fprintf fputs free frexp '
      + 'fscanf future isalnum isalpha iscntrl isdigit isgraph islower isprint ispunct isspace isupper '
      + 'isxdigit tolower toupper labs ldexp log10 log malloc realloc memchr memcmp memcpy memset modf pow '
      + 'printf putchar puts scanf sinh sin snprintf sprintf sqrt sscanf strcat strchr strcmp '
      + 'strcpy strcspn strlen strncat strncmp strncpy strpbrk strrchr strspn strstr tanh tan '
      + 'vfprintf vprintf vsprintf endl initializer_list unique_ptr',
  };
```

Why?

Dxuian:
Resolves #4103

 added C lingo and rm C++ lingo
### Checklist


joshgoebel:
Could we break this into two commits please, the first one just rewriting the keywords to be one per line (see some of the newer grammars), then the second actually making the changes - this way it's possible to review a diff and actually see what was changed - which is very difficult now.

I used to have a VS Code plugin for this but I think I've lost it.

Dxuian:
> Could we break this into two commits please, the first one just rewriting the keywords to be one per line (see some of the newer grammars), then the second actually making the changes - this way it's possible to review a diff and actually see what was changed - which is very difficult now.
> 
> I used to have a VS Code plugin for this but I think I've lost it.

done but the order might still give you should i sort it add two commits for easier diffing
also please review my other prs 🙏🙏
[](https://github.com/highlightjs/highlight.js/pull/4106)
[](https://github.com/highlightjs/highlight.js/pull/4105)
[](https://github.com/highlightjs/highlight.js/pull/4104)

Dxuian:
@joshgoebel  should i sort this alphabetically for easier diffing ?

Dxuian:
> Could we break this into two commits please, the first one just rewriting the keywords to be one per line (see some of the newer grammars), then the second actually making the changes - this way it's possible to review a diff and actually see what was changed - which is very difficult now.
> 
> I used to have a VS Code plugin for this but I think I've lost it.

or should i break up the whole thing into different lists for easier diffing ?

joshgoebel:
To even review this this would need to only include the C changes, not all the SQL stuff.

src/languages/c.js
```diff
@@ -180,19 +180,142 @@ export default function(hljs) {
...
-      + 'auto_ptr deque list queue stack vector map set pair bitset multiset multimap unordered_set '
-      + 'unordered_map unordered_multiset unordered_multimap priority_queue make_pair array shared_ptr abort terminate abs acos '
-      + 'asin atan2 atan calloc ceil cosh cos exit exp fabs floor fmod fprintf fputs free frexp '
-      + 'fscanf future isalnum isalpha iscntrl isdigit isgraph islower isprint ispunct isspace isupper '
-      + 'isxdigit tolower toupper labs ldexp log10 log malloc realloc memchr memcmp memcpy memset modf pow '
-      + 'printf putchar puts scanf sinh sin snprintf sprintf sqrt sscanf strcat strchr strcmp '
-      + 'strcpy strcspn strlen strncat strncmp strncpy strpbrk strrchr strspn strstr tanh tan '
-      + 'vfprintf vprintf vsprintf endl initializer_list unique_ptr',
+    literal: 
+        'true '
```
joshgoebel, line 184: This should now be an array (one item per line), not a large string.

joshgoebel:
Closing for inactivity and multiple grammar changes in a single PR.  This needs to be a PR from a branch with ONLY the C grammar changes.  This can be reopened at any time you have the time to submit a new PR with isolated changes and address the questions asked.

PR #4110, opened 2024-09-01
enh(lua) add 'pluto' as an alias

Sainan:
Rationale: My understanding is that the aliases are used for file extension matching. [Pluto](https://pluto-lang.org/) is a fork of Lua that uses the .pluto extension. There would be benefit for its users if .pluto files were recognised and highlighted as Lua by this package.

There are some differences in syntax, but I don't want to impose the burden of you having to maintain highlighting for it.


joshgoebel (approved):

joshgoebel:
Sorry for pushing to your main, suggest you learn to use branches for PRs in the future.

Thanks for the contrib!

PR #4111, opened 2024-09-01
 added fix for newline and updated tests

### LINKED ISSUE #4090: (YAML) Multiline strings don't support empty lines
**Describe the issue**
YAML multiline strings break on first empty line even when the rest of the block is properly indented.

YAML

`highlight`

**Sample Code to Reproduce**
```yaml
foo:
  bar: |
    still: a string

    not: anymore
```

**Additional context**
I was able to workaround the issue by replacing [this regex](https://github.com/highlightjs/highlight.js/blob/263980c6285f77f8637f7f11884060d492ac339f/src/languages/yaml.js#L123) with `[\\|>]([1-9]?[+-])?[ ]*\\n+( +)[^ ][^\\n]*\\n+(\\2[^\\n]+\\n*)*`.
It's not perfect and still has some false positives/negatives but should cover more valid cases.

Dxuian:
Resolves #4090
before
beforebig
beforeexample
beforesmall
after:
afterbig
afterexaple
aftersmall

### Checklist

### DIFF
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -21,11 +21,11 @@ export default function(hljs) {
     className: 'attr',
     variants: [
       // added brackets support 
-      { begin: /\w[\w :()\./-]*:(?=[ \t]|$)/ },
-      { // double quoted keys - with brackets
-        begin: /"\w[\w :()\./-]*":(?=[ \t]|$)/ },
-      { // single quoted keys - with brackets
-        begin: /'\w[\w :()\./-]*':(?=[ \t]|$)/ },
+      { begin: /\w[\w :()\./-]*:(?=\s|$)/ },
+      // double quoted keys - with brackets
+      { begin: /"\w[\w :()\./-]*":(?=\s|$)/ },
+      // single quoted keys - with brackets
+      { begin: /'\w[\w :()\./-]*':(?=\s|$)/ },
     ]
   };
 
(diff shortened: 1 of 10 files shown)


src/languages/yaml.js
```diff
@@ -107,20 +107,34 @@ export default function(hljs) {
     relevance: 0
   };
 
+  // Optimized block scalar handling to avoid polynomial backtracking
+  const BLOCK_SCALAR = {
+    className: 'string',
+    begin: /[|>][-+\d]*\s*$/, // Block scalar indicator with modifiers
```
joshgoebel, line 113: Is there a reason the regex here around +-number is much fuzzier than the tighter one below?

src/languages/yaml.js
```diff
@@ -107,20 +107,34 @@ export default function(hljs) {
...
+    BLOCK_SCALAR, // handle block scalars (| or >)
     { // multi line string
       // Blocks start with a | or > followed by a newline
       //
       // Indentation of subsequent lines must be the same to
       // be considered part of the block
       className: 'string',
-      begin: '[\\|>]([1-9]?[+-])?[ ]*\\n( +)[^ ][^\\n]*\\n(\\2[^\\n]+\\n?)*'
+      begin: '[\\|>]([-+]?[0-9]*)?[ ]*\\n', // Support for block string indicators
+      relevance: 0
```
joshgoebel, line 137: Now that we have BLOCK_SCALAR which cases is this covering again?

src/languages/yaml.js
Dxuian: are you referring to the regex on line 113 and 116 ?

joshgoebel: Referring to this looking much more correct:

`'[\\|>]([-+]?[0-9]*)?[ ]*\\n'`

ie, +- followed by digits, etc...

This regest would allow all sorts of `-43+23423---13+++`...

joshgoebel:
ping

joshgoebel:
Closing for inactivity - this can be reopened at any time when you have time to work on it.

PR #4113, opened 2024-09-04
Add `time` and `coproc` reserved keywords to Bash

mondeja:
### Changes

Adds the keywords `time` and `coproc` to Bash. See https://www.gnu.org/software/bash/manual/bash.html#Reserved-Words

- `coproc` [was added in bash-4.0-alpha](https://github.com/bminor/bash/blob/142bbdd89e4d5bb62aea4469d1d2c24cf470afd8/CHANGES#L5270)
- `time` seems that has been added previous to v2 but was not being documented as a reserved word.

In fact, many of the reserved words should only be taken as such when followed by certain others, but it seems excessive in this context.

### Checklist
- [x] Tests don't apply here because is a trivial change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ Core Grammars:
 - enh(erlang) OTP25/27 maybe statement [nixxquality][]
 - enh(dart) Support digit-separators in number literals [Sam Rawlins][]
 - enh(csharp) add Contextual keywords `file`, `args`, `dynamic`, `record`, `required` and `scoped` [Alvin Joy][]
+- enh(bash) add reserved keywords `time` and `coproc` [Álvaro Mondéjar][]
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -15,6 +15,7 @@ Core Grammars:
 - enh(erlang) OTP25/27 maybe statement [nixxquality][]
 - enh(dart) Support digit-separators in number literals [Sam Rawlins][]
 - enh(csharp) add Contextual keywords `file`, `args`, `dynamic`, `record`, `required` and `scoped` [Alvin Joy][]
+- enh(bash) add reserved keywords `time` and `coproc` [Álvaro Mondéjar][]
```
joshgoebel, line 18: You need to also add a link to yourself around line ~40 or so.

joshgoebel (approved):

PR #4119, opened 2024-09-22
fix(php) improves attribute highlighting

### LINKED ISSUE #4115: (PHP) Attributes with \ highlighted as comment instead of meta
**Describe the issue**
Using Attributes with `\` are hilighted as comments instead of meta. If it's only a leading `\` it is working as expected. But as soons as we add a `\` after that, the highlighting is not correct anymore. This is a common situation when using attributes with PHP for example using Doctrine or Swagger etc. 

`php`

`highlight`

**Sample Code to Reproduce**
```php
use Doctrine\ORM\Mapping as ORM;

#[\ORM\Entity]
final class Account
{
    #[ORM\Id]
    #[ORM\Column('uuid')]
    private AccountId $id;

    #[Column('string')]
    #[\Column('string')]
    private string $name;
}
```

Bildschirmfoto 2024-09-10 um 11 28 09

**Expected behavior**
Same meta highlighting in these cases.

**Additional context**
Using highlight js through https://github.com/rehypejs/rehype-highlight which is using https://github.com/wooorm/lowlight. But i dont think the problem comes from there, since it's also wrong highlighted in the demo.

joshgoebel:
Fixes #4115.

### Checklist

### DIFF
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -22,12 +22,15 @@ export default function(hljs) {
   const PASCAL_CASE_CLASS_NAME_RE = regex.concat(
     /(\\?[A-Z][a-z0-9_\x7f-\xff]+|\\?[A-Z]+(?=[A-Z][a-z0-9_\x7f-\xff])){1,}/,
     NOT_PERL_ETC);
+  const UPCASE_NAME_RE = regex.concat(
+    /[A-Z]+/,
+    NOT_PERL_ETC);
   const VARIABLE = {
     scope: 'variable',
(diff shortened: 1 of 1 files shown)


DanielBadura (approved):

joshgoebel:
Just needs a changes entry and rebase.

wkania:
And maybe change one of the attributes to uppercase inside of [test](https://github.com/highlightjs/highlight.js/blob/main/test/markup/php/attributes.txt#L6). `#[ASSERT(value: 1234)]`

PR #4124, opened 2024-09-30
add Cybertopia themes

noraj:
New theme with 4 variants, source here (with screenshots): https://github.com/noraj/cybertopia-highlightjs

PR #4125, opened 2024-10-01
remove implicit style

noraj:
Remove styles that are added automatically by the build system. cf. https://github.com/highlightjs/highlight.js/pull/4124#discussion_r1781871020

### DIFF
--- a/src/styles/base16/3024.css
+++ b/src/styles/base16/3024.css
@@ -34,16 +34,6 @@ base0E  #a16a94  Keywords, Storage, Selector, Markup Italic, Diff Changed
 base0F  #cdab53  Deprecated, Opening/Closing Embedded Language Tags, e.g. <?php ?>
 */
 
-pre code.hljs {
-  display: block;
-  overflow-x: auto;
-  padding: 1em;
-}
-
-code.hljs {
-  padding: 3px 5px;
-}
-
 .hljs {
   color: #a5a2a2;
   background: #090300;
--- a/VERSION_11_UPGRADE.md
+++ b/VERSION_11_UPGRADE.md
@@ -16,7 +16,7 @@
 
 ## Overview of Breaking Changes
 
-Welcome to version 11.0. This a major release and therefore contains breaking changes. Below is a complete list of those such changes.
+Welcome to version 11.0.  This a major release and therefore contains breaking changes.  Below is a complete list of those such changes.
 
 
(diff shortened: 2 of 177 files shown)


src/styles/base16/3024.css
```diff
@@ -34,16 +34,6 @@ base0E  #a16a94  Keywords, Storage, Selector, Markup Italic, Diff Changed
 base0F  #cdab53  Deprecated, Opening/Closing Embedded Language Tags, e.g. <?php ?>
 */
 
-pre code.hljs {
```
joshgoebel, line 37: I'm not sure if the base16 files go thru our internal build process, they have their own external build process, but I'll double check.

joshgoebel:
I can't tell what if anything you changed in the README, please revert all the unnecessary linting/spacing changes.

The other changes for the base16 styles seem to be correct. I'm thinking they ship "complete" from the Base16 repo they are copied directly from to allow people to use them "as-is" from that repo... where-as the themes here are not meant to be used until they are "cooked" and added to a distribution.

So I think this can be merged just as soon as we remove the unnecessary changes to the README.

noraj:
> I can't tell what if anything you changed in the README, please revert all the unnecessary linting/spacing changes.

You mean in VERSION_11_UPGRADE.md (aebf8d061ecf5a96b82dc3d5e99101cedb6e87a5)? I removed all the unnecessary extra/double spaces.

> So I think this can be merged just as soon as we remove the unnecessary changes to the README.

You want me to add a new revert commit? Or do you prefer to drop this commit while squashing the PR?

joshgoebel:
> You want me to add a new revert commit? 

That'd be great.

joshgoebel:
Actually we may need more work here, but I'm not sure what to suggest.  It just clicked that these are actually assets copied directly from [another build system](https://github.com/highlightjs/base16-highlightjs).  So modifying them in our repo makes no sense - we'd need a script to make this repeatable.  

Otherwise all this work will be lost when the Base16 themes are rebuilt and copied over whenever we decide to refresh these in the future.  I'm thinking it maybe best to leave these "as is" and clean up the duplication in our own build step.

Thoughts?

joshgoebel:
See last comment.  Closing this.  I feel like a tiny improvement to clean this up inside our CSS processing pipeline would be a different PR entirely - not a reuse of this one.

PR #4130, opened 2024-10-06
enh(php) add support for Anonymous functions (closures)

### LINKED ISSUE #4006: (php) Semicolon inside string parsed as end of statement
**Describe the issue**
It looks like a semicolon inside a string is parsed as terminating the statement, at least sometimes.

`php`

`highlight`

**Sample Code to Reproduce**
Fiddle here:  https://jsfiddle.net/c3sLm2ga/.  See how the highlighting gets messed up halfway through the string in the `throw` statement.  Oddly, double-quoted strings are also affected but in a slightly different way.

**Expected behavior**
```php
public static function mutate() {
  $result = null;
  DB::transaction(function () use (): void {
    if (static::$currentMutator !== null) {
      throw new LoggingMutatorException('LoggingMutator::mutate called while already in a mutation; nested mutations are not supported.');
    }
    $operationLogger = new OperationLogger($user);
  })
}
```

wkania:
Resolves #4006 

### Changes
Anonymous function can be declared inside function invoke.

### Checklist
- [☑️] Added markup tests,
- [☑️] Updated the changelog at `CHANGES.md`

### DIFF
--- a/src/languages/php.js
+++ b/src/languages/php.js
@@ -401,6 +401,40 @@ export default function(hljs) {
     scope: 'attr',
     match: regex.concat(IDENT_RE, regex.lookahead(':'), regex.lookahead(/(?!::)/)),
   };
+  
+  const FUNCTION_DECLARATION = {
+    scope: 'function',
+    relevance: 0,
+    beginKeywords: 'fn function',
+    end: /[;{]/,
+    excludeEnd: true,
+    illegal: '[$%\\[]',
+    contains: [
+      { beginKeywords: 'use', },
+      hljs.UNDERSCORE_TITLE_MODE,
+      {
+        begin: '=>', // No markup, just a relevance booster
+        endsParent: true
+      },
+      {
+        scope: 'params',
+        begin: '\\(',
+        end: '\\)',
+        excludeBegin: true,
+        excludeEnd: true,
+        keywords: KEYWORDS,
+        contains: [
+          'self',
+          VARIABLE,
+          LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+          hljs.C_BLOCK_COMMENT_MODE,
+          STRING,
+          NUMBER
+        ]
+      },
+    ]
+  };
+  
   const PARAMS_MODE = {
     relevance: 0,
     begin: /\(/,
@@ -414,8 +448,10 @@ export default function(hljs) {
       STRING,
       NUMBER,
       CONSTRUCTOR_CALL,
+      FUNCTION_DECLARATION,
     ],
   };
+  
   const FUNCTION_INVOKE = {
     relevance: 0,
     match: [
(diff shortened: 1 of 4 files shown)


joshgoebel (commented):
Thanks for this — nesting `FUNCTION_DECLARATION` in `PARAMS_MODE` is the right fix for #4006. A few notes below to get it merge-ready.

Also: please enable **Allow edits from maintainers** on this PR so we can help push small fixes if needed.

src/languages/php.js
```diff
@@ -401,6 +401,40 @@ export default function(hljs) {
...
+        excludeBegin: true,
+        excludeEnd: true,
+        keywords: KEYWORDS,
+        contains: [
+          'self',
+          VARIABLE,
+          LEFT_AND_RIGHT_SIDE_OF_DOUBLE_COLON,
+          hljs.C_BLOCK_COMMENT_MODE,
+          STRING,
+          NUMBER
```
joshgoebel, line 432: **Rebase / params contents:** On current `main`, function-params `contains` also has `ATTRIBUTES`, `hljs.C_LINE_COMMENT_MODE`, and `hljs.HASH_COMMENT_MODE`. This extract still has the older list only — please rebase and keep those so we don’t regress attribute/`//`/`#` highlighting in signatures.

test/markup/php/functions.txt
```diff
@@ -7,6 +7,15 @@ $fn2 = function ($x) use ($y) {
...
 };
 
+/**
+ * Anonymous function
+ */
+function hello(string $name) {
+    return trim(function() use ($name) {
+        echo "Hello; $name";
+    });
+}
```
joshgoebel, line 17: **Test coverage for #4006:** This is helpful, but it doesn’t cover the issue sample. Please add a case closer to the report, e.g. a method call with a closure argument, `use (...): void`, and a `;` inside a string in the body — that’s what was breaking via the top-level `use`…`;` rule.

test/markup/php/functions.txt
```diff
@@ -7,6 +7,15 @@ $fn2 = function ($x) use ($y) {
     return $x + $y;
 };
 
+/**
+ * Anonymous function
+ */
+function hello(string $name) {
```
joshgoebel, line 13: `function hello` is a normal named function; the change under test is the closure inside `trim(...)`. Consider moving this comment onto the inner `function() use` (or wording it that way) so the intent is obvious.

wkania: Ok, will check it during weekend.

joshgoebel:
@wkania Could you turn on **Allow edits from maintainers** on this PR? (`maintainerCanModify` is currently false.) That lets us rebase, drop `CHANGES.md` if it conflicts, and land the remaining review nits (params `contains` vs `main`, #4006 fixture, comment wording) if you do not get to them.

Thanks.

PR #4131, opened 2024-10-07
fix(css) line break disappearing

### LINKED ISSUE #4101: (css) Line breaks disappear
**Describe the issue**
Line breaks disappear

css

`hljs.highlight`

**Sample Code to Reproduce**
```
hljs.highlight(`[\n1`,
          {
            language:"css",
            ignoreIllegals: true
          }
        ).value
```

QQ_1724903357711

**Expected behavior**

Line breaks do not disappear

somya-05:
Fixed the line break disappearing in case of CSS.

Resolves #4101

### Changes

Before:
image

After:
image

### Checklist

### DIFF
--- a/src/languages/lib/css-shared.js
+++ b/src/languages/lib/css-shared.js
@@ -17,7 +17,7 @@ export const MODES = (hljs) => {
       scope: 'selector-attr',
       begin: /\[/,
       end: /\]/,
-      illegal: /[$]/,
+      illegal: '$',
       contains: [
         hljs.APOS_STRING_MODE,
         hljs.QUOTE_STRING_MODE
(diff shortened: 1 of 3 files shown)


src/languages/lib/css-shared.js
```diff
@@ -17,7 +17,7 @@ export const MODES = (hljs) => {
       scope: 'selector-attr',
       begin: /\[/,
       end: /\]/,
-      illegal: '$',
+      illegal: /[$]/,
```
joshgoebel, line 20: This is not equivalent and therefore not a _correct_ fix.  One matches end of line (which is what we want)... your new version matches the literal `$` character...

This is actually a bug that must be fixed in the parser itself, not any particular grammar.

PR #4133, opened 2024-10-09
fix(app): #4101 Fixes parser to parse new lines directly to output

### LINKED ISSUE #4101: (css) Line breaks disappear

mohit-nagaraj:
Fixes parser to parse new lines directly to output. 

Closes #4101 

### Changes

This modification to the addText method ensures that newline characters are properly handled and preserved in the token tree

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Core Grammars:
 - fix(ruby) - fix `|=` operator false positives (as block arguments) [Aboobacker MK]
 - fix(sql) - Fixed sql primary key and foreign key spacing issue   [Dxuian]
 - fix(cpp) added flat_set and flat_map as a part of cpp 23 version [Lavan]
+
   
 New Grammars:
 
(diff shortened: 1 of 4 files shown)


joshgoebel:
I think the needed fix here is much simpler.  The issue (in my testing) was only when we have an illegal match on `$` (end of line) that we then insert no content since technically $ matches an empty string not the `\n` itself...  If you'd like to help you could add a markup test with a problematic sample of this - then I'll throw my simper fix and we'll see if things are still green.

joshgoebel:
Could you please test https://github.com/highlightjs/highlight.js/pull/4142

joshgoebel:
If you believe you have test cases that my PR doesn't solve, just point me to them and I'll take another look.

joshgoebel:
@mohit-nagaraj ping.  Could you test my fix against your test cases?

joshgoebel:
I believe this is closed by https://github.com/highlightjs/highlight.js/pull/4142

PR #4135, opened 2024-10-10
Add the `using` keyword per the "Explicit Resource Management" proposal

sufianrhazi:
### Changes

This ECMAScript proposal adds a new keyword (`using`). It's reached stage 3, which is "recommended for implementation":
* Proposal: https://github.com/tc39/proposal-explicit-resource-management
* ECMAScript proposal process: https://tc39.es/process-document/

The change itself is very small, I've just added the `using` keyword to the list of keywords in `src/languages/lib/ecmascript.js`. I've also updated an existing keyword test adding a case to ensure `using` is rendered as `<span class="hljs-keyword">using</span>` in a declaration.

I didn't update the `CHANGES.md`, since I'm not sure if I should cut a new version for this? But let me know what to do and I can do it.

### Checklist

### DIFF
--- a/src/languages/lib/ecmascript.js
+++ b/src/languages/lib/ecmascript.js
@@ -40,7 +40,8 @@ const KEYWORDS = [
   "import",
   "from",
   "export",
-  "extends"
+  "extends",
+  "using"
 ];
 const LITERALS = [
   "true",
--- a/src/languages/lib/ecmascript.js
+++ b/src/languages/lib/ecmascript.js
@@ -41,6 +41,7 @@ const KEYWORDS = [
   "from",
   "export",
   "extends",
+  // It's reached stage 3, which is "recommended for implementation":
   "using"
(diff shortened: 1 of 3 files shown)


src/languages/lib/ecmascript.js
```diff
@@ -40,7 +40,8 @@ const KEYWORDS = [
   "import",
   "from",
   "export",
-  "extends"
+  "extends",
+  "using"
```
joshgoebel, line 44:
```suggestion
  "extends",
  // It's reached stage 3, which is "recommended for implementation":
  "using"
```

joshgoebel (changes requested):
Lets add the comment then this is good to merge!

sufianrhazi:
Thank you! Just added the suggested changes.

joshgoebel (approved):

PR #4136, opened 2024-10-11
potential fix for issue#4134

### LINKED ISSUE #4134: dos/bat/cmd Disk switch "D:" destroys highlight
**Describe the issue**
If the code snippet contains a disk switch line, e.g. "D:", the subsequent highlight becomes all red. It seems to be connected with the colon.
bild

"dos"

Seems to be the same with tags "bat" and "cmd".

Seems to work the same with and without space between ``` and the language tag.

Here in GitHub it seems to work correctly with tag "bat", see below under "Expected behavior".

Not sure. I was using this in an Azure DevOps Wiki page.

**Sample Code to Reproduce**
```
triplebacktick dos
rem This highlight works.
copy file.txt other.txt 
rem After next line it becomes all red.
D:
cd "D:\Folder"
md Tmp
del Tmp\*.* /q
triplebacktick
```

**Expected behavior**
The highlight should work normally even after a disk switch command, as here in GitHub:
``` bat
rem This highlight works.
copy file.txt other.txt 
rem After next line it becomes all red.
D:
cd "D:\Folder"
md Tmp
del Tmp\*.* /q
```

**Additional context**
Found and tested in an Azure DevOps Wiki page, which seems to point to this repo in the docs: https://learn.microsoft.com/en-us/azure/devops/project/wiki/markdown-guidance?view=azure-devops#code-highlighting

the-dg04:
### Potential fix for the problem
dos/bat/cmd Disk switch "D:" destroys highlight

Resolves #4134 

### Changes
in `src/languages/dos.js line:151` replaced `end:'goto:eof'` to `end:'\n'`

### Checklist

### DIFF
--- a/src/languages/dos.js
+++ b/src/languages/dos.js
@@ -15,14 +15,27 @@ export default function(hljs) {
 
   // for matching comments starting with ::
   const COMMENT_2 = hljs.COMMENT(
-    /^::/, /$/,
+    /^::.*$/, /$/,
     { relevance: 10 }
   );
   const LABEL = {
     className: 'symbol',
-    begin: '^:',
+    begin: '^:[A-Za-z._?][A-Za-z0-9_$#@~.?]*',
     relevance: 0
   };
+
+  const DISK_CHANGE = {
+    className: 'symbol',
+    begin: '^[A-Za-z]:\\?$',
+    relevance: 0
+  };
+
+  const OUTPUT_REDIRECT = {
+    className: 'symbol',
+    begin: '[1-2]?[>]>{1}\s*[^&\s]+',
+    relevance: 0
+  };
+
   const KEYWORDS = [
     "if",
     "else",
@@ -157,11 +170,23 @@ export default function(hljs) {
         className: 'function',
         begin: LABEL.begin,
         end: /\n/,
-        illegal: '[::].+', // to ignore lines starting with "::" [COMMENTS]
         contains: [
           hljs.inherit(hljs.TITLE_MODE, { begin: '([_a-zA-Z]\\w*\\.)*([_a-zA-Z]\\w*:)?[_a-zA-Z]\\w*' }),
-          COMMENT
+          COMMENT,
+          COMMENT_2
         ]
+      },
+      {
+        className: 'function',
+        begin: DISK_CHANGE.begin,
+        end: /\n/,
+        contains: [ hljs.inherit(hljs.TITLE_MODE, { begin: '([_a-zA-Z]\\w*\\.)*([_a-zA-Z]\\w*:)?[_a-zA-Z]\\w*' }) ]
+      },
+      {
+        className: 'function',
+        begin: OUTPUT_REDIRECT.begin,
+        end: /\n/,
+        contains: [ hljs.inherit(hljs.TITLE_MODE, { begin: '([_a-zA-Z]\\w*\\.)*([_a-zA-Z]\\w*:)?[_a-zA-Z]\\w*' }) ]
       },
       {
         className: 'number',
--- a/src/languages/dos.js
+++ b/src/languages/dos.js
@@ -12,9 +12,15 @@ export default function(hljs) {
     /^\s*@?rem\b/, /$/,
     { relevance: 10 }
   );
+
+  // for matching comments starting with ::
+  const COMMENT_2 = hljs.COMMENT(
+    /^::/, /$/,
+    { relevance: 10 }
+  );
   const LABEL = {
     className: 'symbol',
-    begin: '^\\s*[A-Za-z._?][A-Za-z0-9_$#@~.?]*(:|\\s+label)',
+    begin: '^:',
     relevance: 0
   };
   const KEYWORDS = [
@@ -141,6 +147,8 @@ export default function(hljs) {
       built_in: BUILT_INS
     },
     contains: [
+      COMMENT,
+      COMMENT_2,
       {
         className: 'variable',
         begin: /%%[^ ]|%[^ ]+?%|![^ ]+?!/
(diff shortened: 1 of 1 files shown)


src/languages/dos.js
```diff
@@ -148,7 +148,7 @@ export default function(hljs) {
       {
         className: 'function',
         begin: LABEL.begin,
-        end: 'goto:eof',
+        end: '\n',
```
joshgoebel, line 151: Ok, that was a weird end condition, but now `D:` is incorrectly detected as a label, is it not?  From what I'm reading labels always start with `:` not end with it... so could we just update our label matcher to something like?

```
[start of line]:[label_name]
```

I think that would wrap this up nicely.

the-dg04: That makes sense, ive added expressions like `[start of line]:[end of line]` to illegal so they wont be matched anymore.

src/languages/dos.js
```diff
@@ -12,9 +12,15 @@ export default function(hljs) {
...
+
+  // for matching comments starting with ::
+  const COMMENT_2 = hljs.COMMENT(
+    /^::/, /$/,
+    { relevance: 10 }
+  );
   const LABEL = {
     className: 'symbol',
-    begin: '^\\s*[A-Za-z._?][A-Za-z0-9_$#@~.?]*(:|\\s+label)',
+    begin: '^:',
```
joshgoebel, line 23: Wouldn't this be something more like:

```
/^:[A-Za-z._?][A-Za-z0-9_$#@~.?]*/
```

Or whatever value characters for a label are?

joshgoebel: I'll have to take a closer look to see how this is currently working.  This grammar pattern is pretty old... I think we could have a much simpler rule here, or perhaps a multi-match if we truly want the `:` to be different than the name of the label.

the-dg04:
I've added multiple possible cases for the use of `:` in a dos-script. Although I'm a bit muddled on what the relevance values should be for the new symbols. Mind helping me with that?

joshgoebel:
We can just forget about relevance, we'll be dropping it from v12 anyways.  

It would be nice to add some markup examples here so I can see which cases you are trying to cover explicitly.   Once I have some solid test cases I'll probably clean this up a bit further before merging.

src/languages/dos.js
```diff
@@ -12,11 +12,30 @@ export default function(hljs) {
...
   const LABEL = {
     className: 'symbol',
-    begin: '^\\s*[A-Za-z._?][A-Za-z0-9_$#@~.?]*(:|\\s+label)',
+    begin: '^:[A-Za-z._?][A-Za-z0-9_$#@~.?]*',
+    relevance: 0
+  };
+
+  const DISK_CHANGE = {
+    className: 'symbol',
+    begin: '^[A-Za-z]:\\?$',
```
joshgoebel, line 29: This should be a literal regex, not a string.

src/languages/dos.js
```diff
@@ -12,11 +12,30 @@ export default function(hljs) {
     /^\s*@?rem\b/, /$/,
     { relevance: 10 }
   );
+
+  // for matching comments starting with ::
+  const COMMENT_2 = hljs.COMMENT(
+    /^::.*$/, /$/,
```
joshgoebel, line 18: The begin should start and the end should end, so I think you want:

start: `/^::/`
end : `/$/`

The end shouldn't be included in the begin.

src/languages/dos.js
```diff
@@ -12,11 +12,30 @@ export default function(hljs) {
...
+
+  const DISK_CHANGE = {
+    className: 'symbol',
+    begin: '^[A-Za-z]:\\?$',
+    relevance: 0
+  };
+
+  const OUTPUT_REDIRECT = {
+    className: 'symbol',
+    begin: '[1-2]?[>]>{1}\s*[^&\s]+',
```
joshgoebel, line 35: Please add a markup test or two for this so I can understand what exactly you're trying to match.  This looks a bit overly complex to me, but maybe I don't understand the goal.

src/languages/dos.js
```diff
@@ -141,25 +160,39 @@ export default function(hljs) {
...
-          COMMENT
+          COMMENT,
+          COMMENT_2
         ]
       },
+      {
+        className: 'function',
+        begin: DISK_CHANGE.begin,
+        end: /\n/,
+        contains: [ hljs.inherit(hljs.TITLE_MODE, { begin: '([_a-zA-Z]\\w*\\.)*([_a-zA-Z]\\w*:)?[_a-zA-Z]\\w*' }) ]
```
joshgoebel, line 183: Please explain what this is attempting to do.  Are you trying to highlight all path names?

src/languages/dos.js
```diff
@@ -141,25 +160,39 @@ export default function(hljs) {
...
+      },
+      {
+        className: 'function',
+        begin: OUTPUT_REDIRECT.begin,
+        end: /\n/,
+        contains: [ hljs.inherit(hljs.TITLE_MODE, { begin: '([_a-zA-Z]\\w*\\.)*([_a-zA-Z]\\w*:)?[_a-zA-Z]\\w*' }) ]
```
joshgoebel, line 189: Same question, is this another attempt to highlight pathname?

joshgoebel:
The DOS grammar is simple enough we could try adding path highlighting, but we'd do it by recognizing the FULL path... ie something like: (**pseudo code**)

```
disk = [A-Z]:\\
path_component = [A-Za-z.....]
[disk]([path component]\\)*[path component]?
```

So first the disk, followed by any number of \ terminated path segments possibly following be a single non-slash terminated segment... and maybe one sep rule to match non-slash terminated disks...

Does that make sense?

joshgoebel:
Ping.

joshgoebel:
Closing in favor of #4479

PR #4137, opened 2024-10-12
fix(basic) - closing quotation marks not required for a PRINT statement 

### LINKED ISSUE #3996: (basic) Closing quotation marks are not required for a PRINT statement (gwbasic)
**Describe the issue**
I may be wrong and this probably depends on _basic_ dialect, but it seems (I found some old code that runs) that for _gwbasic_ closing quotation marks are not required for a PRINT statement. The interpreter seems to assume that the end of the line closes the string (like there was a quotation mark).

In this case, when an ending quotation mark is omitted, the following line(s) are part of the string until a quotation mark is found.

afbeelding

basic. And in my case specifically for gwbasic code.

highlight: language-basic

**Sample Code to Reproduce**

```
10	CLS
20	INPUT "Hoeveel paarden ";H:CLS
30	LOCATE 1,1:PRINT "╔════════╦════════════════════╗
40	PRINT "║ nummer ║ naam               ║
50	PRINT "╠════════╬════════════════════╣
60	FOR F=1 TO H
70	LOCATE 23,2:PRINT "Paard";F:LOCATE 23,10:INPUT P(F)
80	GET #1,P(F)
```

**Expected behavior**
At line 30 the symbols after **"** should be highlighted as string for the rest of the line. Line 40 however should start with the normal highlighting as it is not part of a continued string.

It mentioned the PRINT statement but this is also the case for e.g. CHAIN and LOAD.

somya-05:
Fixed the closing quotations marks considered for string highlighting in basic.

Resolves #3996 

### Changes
Before
image

After
image

### Checklist

### DIFF
--- a/src/languages/basic.js
+++ b/src/languages/basic.js
@@ -203,7 +203,6 @@ export default function(hljs) {
         scope: 'string',
         begin: /"/,
         end: /"|$/,
-        illegal: /\n/,
         contains: [ hljs.BACKSLASH_ESCAPE ]
       },
       hljs.COMMENT('REM', '$', { relevance: 10 }),
--- a/src/languages/basic.js
+++ b/src/languages/basic.js
@@ -203,7 +203,7 @@ export default function(hljs) {
         scope: 'string',
         begin: /"/,
         end: /"|$/,
-        illegal: '\\n',
+        illegal: /\n/,
         contains: [ hljs.BACKSLASH_ESCAPE ]
       },
       hljs.COMMENT('REM', '$', { relevance: 10 }),
--- a/src/languages/basic.js
+++ b/src/languages/basic.js
@@ -198,7 +198,14 @@ export default function(hljs) {
       keyword: KEYWORDS
     },
     contains: [
-      hljs.QUOTE_STRING_MODE,
+      {
+        // Match strings that start with " and end with " or a line break
+        className: 'string',
+        begin: '"',
+        end: '"|$',
+        illegal: '\\n',
+        contains: [ hljs.BACKSLASH_ESCAPE ]
+      },
       hljs.COMMENT('REM', '$', { relevance: 10 }),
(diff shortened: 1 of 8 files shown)


src/languages/basic.js
```diff
@@ -198,7 +198,14 @@ export default function(hljs) {
       keyword: KEYWORDS
     },
     contains: [
-      hljs.QUOTE_STRING_MODE,
+      {
+        // Match strings that start with " and end with " or a line break
+        className: 'string',
```
joshgoebel, line 203:
```suggestion
        scope: 'string',
```

src/languages/basic.js
```diff
@@ -198,7 +198,14 @@ export default function(hljs) {
...
+        begin: '"',
```
joshgoebel, line 204: Please use regex literals for all the regex, rather than strings.

somya-05:
@joshgoebel the changes for regex literals are made, please review them once.

src/languages/basic.js
```diff
@@ -198,7 +198,14 @@ export default function(hljs) {
       keyword: KEYWORDS
     },
     contains: [
-      hljs.QUOTE_STRING_MODE,
+      {
+        // Match strings that start with " and end with " or a line break
+        scope: 'string',
+        begin: /"/,
+        end: /"|$/,
+        illegal: /\n/,
```
joshgoebel, line 206: I'm not sure this has any effect at all since the `$` should match first (or does it match after)... are you seeing different behavior with/without `illegal`?

somya-05: I added the illegal keyword to ensure the string doesn't span multiple lines. But I can't see a different behavior without it. This can be removed to optimize the code.

joshgoebel: It can't span multiple lines because the `$` will match the end of the first line.

joshgoebel (approved):

PR #4140, opened 2024-10-13
nix: major cleanup

h7x4:
### Changes

This PR contains an overall cleanup of the nix language. It fixes some bugs with string handling, adds a bunch of new keywords that are now part of the language, and adds a bunch of extra support for different language constructs.

- enh(nix) update keywords
- enh(nix) support paths
- enh(nix) support lookup paths
- enh(nix) support operators
- enh(nix) support URLs
- enh(nix) support REPL keywords
- enh(nix) support markdown comments
- enh(nix) support basic function params
- enh(nix) better parsing of attrsets
- fix(nix) remove `add` builtin
- fix(nix) mark `or` as builtin instead of literal
- fix(nix) handle `'''` string escapes
- fix(nix) handle backslash string escapes
- fix(nix) don't mix escapes for `"` and `''` strings

### Checklist


src/languages/nix.js
```diff
@@ -9,85 +9,291 @@ Category: system
...
+      "removeAttrs",
+      "scopedImport",
+      "throw",
+      "toString",
+    ],
+  };
+
+  const BUILTINS = {
+    className: 'built_in',
+    match: new RegExp('(' + [
```
joshgoebel, line 54: Please use our regex built-in tools such as `concat`... so here you'd just wrap your map in `concat`.  And rememver it takes arguments, not an array, so we explode the array with `...`.

```
// something like
hljs.regex.concat(...LIST.map( /* add namespaces*/ ))
```

src/languages/nix.js
```diff
@@ -9,85 +9,291 @@ Category: system
...
         begin: /\S+/,
         relevance: 0.2
-      }
-    ]
+      },
+    ],
+  };
+
+  const NORMAL_ESCAPED_DOLLAR = {
+    className: 'char.escape',
```
joshgoebel, line 208: Please use the new `scope` everywhere in place of `className` since this is such a big rewrite.

src/languages/nix.js
```diff
@@ -9,85 +9,291 @@ Category: system
...
+          ANTIQUOTE,
+          ESCAPED_LITERAL,
+        ],
+      },
+    ],
   };
+
+  const URL = {
+    className: 'link',
+    match: RegExp("https?://[A-Za-z0-9\\-._~:/?#[\\]@!$&'()*+,=]+"),
```
joshgoebel, line 256: Please use regex `//` literal rather than `RegExp()`

src/languages/nix.js
```diff
@@ -9,85 +9,291 @@ Category: system
...
   ANTIQUOTE.contains = EXPRESSIONS;
+
+  const REPL = [
+    {
+      className: 'meta.prompt',
+      match: /nix-repl> /
+    },
+    {
+      className: 'meta',
+      match: /:[a-z]+/,
```
joshgoebel, line 289: Can we pin to the beginning of line here perhaps with `^`... not sure, just a suggestion.

h7x4: I have pinned `nix-repl>` above with a `^`. `:[a-z]+` is meant to highlight commands used in the REPL after the prompt. There isn't anything else that can come after a colon in the nix language for now, so I'm not sure how much better we could make this one.

h7x4:
I have rewritten the test to test more language features, catching a few bugs in the process. Also added support for basic function parameters.

There are some more advanced function parameters as well (you can find an example at the top of the test), but I've not been able to properly parse those without breaking parser backtracking.

src/languages/nix.js
```diff
@@ -9,85 +9,319 @@ Category: system
...
+      '=',
+      '\\+\\+',
+      '\\+',
+      '<=',
+      '<\\|',
+      '<',
+      '>=',
+      '>',
+      '->',
+      /(?<=(\s|[0-9\.]+))-/,
```
joshgoebel, line 203: Sadly, we can't use look-behind until version 12. :-/

src/languages/nix.js
```diff
@@ -9,85 +9,319 @@ Category: system
...
+    ATTRS,
+    OPERATOR,
   ];
+
   ANTIQUOTE.contains = EXPRESSIONS;
+
+  const REPL = [
+    {
+      scope: 'meta.prompt',
+      match: /(?<=^\s*)nix-repl> /
```
joshgoebel, line 313: You can use `^` to match the beginning of a line.

h7x4: The idea was to leave some leeway for quickly copypasting stuff, and forgetting to remove whitespace at the start of the line. Is it okay if I just let any initial whitespace be part of the scope as a temporary workaround for the lookbehind?

src/languages/nix.js
h7x4: Hmm, I might need some time to think about a workaround for this one

src/languages/nix.js
joshgoebel: Does the repl prompt itself not start at the beginning of the ilne?

h7x4: Usually yes, but I'd expect the prompt to be highlighted even if you happen to have a space or two before it. It's not a big deal, I could maybe readdress it once lookbehinds are available.

h7x4: Fixed

h7x4:
@joshgoebel I have gotten rid of all the lookbehinds by using `beforeMatch` and handling `-` on its own with various cases. There are still some edgecases not covered (in particular with identifiers like `abc-123-def`), but I think it's better than nothing :)

src/languages/nix.js
```diff
@@ -6,88 +6,372 @@ Website: http://nixos.org/nix
...
+          ANTIQUOTE,
+          ESCAPED_LITERAL,
+        ],
+      },
+    ],
+  };
+
+  const URL = {
+    scope: 'link',
+    match: /https?:\/\/[A-Za-z0-9\-._~:/?#[\]@!$&'()*+,=]+/,
```
joshgoebel, line 324: Does nix support ftp, ssh, or other protocols?

h7x4: Nix actually supports the entirety of rfc2396 (see last paragraph of https://nix.dev/manual/nix/2.24/language/syntax#string-literal), but I wasn't able to make direct use of the regex provided in the rfc because it kept picking up all sorts of false positives.

The feature is quite discouraged these days, mostly replaced with strings, and with no instances left that I'm aware of in nixpkgs (the largest and most important repository for the ecosystem). So I used a slightly suboptimal regex, hoping to catch some basic instances.

I could add a few more common protocols, but I'm not entirely sure where to stop.

joshgoebel: If it's deprecated/discouraged perhaps lets drop it altogether so people don't complain it's not working in their own special use cases... What are they doing now instead, quoted strings?

h7x4: I have removed the commits related to it. It will pick instances up like a literal path, but maybe it's better to actively not support it (at least for time being, until someone comes up with a better regex). Quoted strings is the recommended replacement, yes.

``` nix
Nix 2.24.9
Type :? for help.
nix-repl> https://nixos.org == "https://nixos.org"
true
```

a-kenji: I think this is more than OK, as there are linters that will automatically fix it and both active lsp implementation warn and provide code actions for fixes:

image

PR #4143, opened 2024-10-20
Add scope to the python `self` keyword

lfalin:
Added a scope to the python `self` keyword.

### Changes
Added a scope to the python `self` keyword.

### Checklist

### DIFF
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -373,6 +373,7 @@ export default function(hljs) {
       NUMBER,
       {
         // very common convention
+        scope: 'variable.language',
         begin: /\bself\b/
       },
       {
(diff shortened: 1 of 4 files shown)


src/languages/python.js
```diff
@@ -373,6 +373,7 @@ export default function(hljs) {
       NUMBER,
       {
         // very common convention
+        scope: 'variable.language',
         begin: /\bself\b/
       },
```
joshgoebel, line 378: You'll need to pull this out into a `SELF_MODE` and make sure it's include both here and above around line 346 (so it's highlighting inside params too) - or was that something you omitted on purpose?

And ignore the literal `"self"` that's just telling it params can include more params inside it's self recursively to allow for nested `()` expressions...

joshgoebel: Ping.

lfalin: Hi @joshgoebel,

Sorry for the late response. Good thoughts. I'll close this pull request and open another once I have those changes ready.

PR #4150, opened 2024-10-23
Update hljs function

TusharNaugain:
This pull request introduces the HLJS function, which sets up the essential components of the Highlight.js library. It includes storage for registered languages and aliases, the ability to handle plugins for extensibility, and a safe mode flag to control unsafe operations

### Checklist

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -47,36 +47,28 @@ const MAX_KEYWORD_HITS = 7;
  * @param {any} hljs - object that is extended (legacy)
  * @returns {HLJSApi}
  */
+// Main function to initialize the Highlight.js library
 const HLJS = function(hljs) {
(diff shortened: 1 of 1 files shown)


joshgoebel:
Ok, thanks truly for finally a small, readable PR. But...

These types of changes are not desired or helpful.  

- We have very different opinions about how comments should be used.
- You continue to remove type information.
- You changed an error message [removing half of it] for no stated reason.

---

As I suggested before, perhaps find an existing issue that needs to be fixed and start there.  Here are the issue tagged "good first issue", you could have a look thru those.

https://github.com/highlightjs/highlight.js/issues?q=is%3Aissue+is%3Aclosed+label%3A%22good+first+issue%22

PR #4154, opened 2024-10-29
fix(swift) - add proper syntax highlighting for class func/var declarations

### LINKED ISSUE #4121: (Swift) `class func foo()` not highlighted properly
**Describe the issue**

In Swift 
```swift
class func foo()
```
works like
```swift
static func foo()
```

But highlight.js interpret `func` as the class name in this case. Both `class` and `func` should be interpreted as keywords.

For example,
```swift
class func foo() async throws -> [Bar]
```
becomes
```html
<span class="hljs-keyword">class</span>
<span class="hljs-title class_">func</span>
 foo() 
<span class="hljs-keyword">async</span>
<span class="hljs-keyword">throws</span>
 -&gt; [
<span class="hljs-type">Bar</span>
```
Whereas,
```swift
static func foo() async throws -> [Bar]
```
becomes
```html
<span class="hljs-keyword">static</span>
<span class="hljs-keyword">func</span>
<span class="hljs-title function_">foo</span>
() 
<span class="hljs-keyword">async</span>
<span class="hljs-keyword">throws</span>
 -&gt; [
<span class="hljs-type">Bar</span>
```

Swift, manually selected and tested on [highlight.js demo page](https://highlightjs.org/demo)

I believe when the language is set to `Swift` on the demo page it uses `highlight`.

**Sample Code to Reproduce**

```swift
class func foo() async throws -> [Bar]
```

**Expected behavior**

```html
<span class="hljs-keyword">class</span>
<span class="hljs-keyword">func</span>
<span class="hljs-title function_">foo</span>
() 
<span class="hljs-keyword">async</span>
<span class="hljs-keyword">throws</span>
 -&gt; [
<span class="hljs-type">Bar</span>
```

**Additional context**

[swift-docc-render](https://github.com/swiftlang/swift-docc-render/tree/main) uses a customized function to resolve this issue. See [here](https://github.com/swiftlang/swift-docc-render/blob/main/src/utils/custom-highlight-lang/swift.js)

guuido:
Fixes the syntax highlighting for class functions and variables

Resolves #4121

### Changes
Before the declaration of `class func` and `class var` was not handled correctly, for example
```swift
class func foo() async throws -> [Bar]
class var someProperty: Int { get }
```
was interpreted as
```html
<span class="hljs-keyword">class</span> <span class="hljs-title class_">func</span> foo() <span class="hljs-keyword">async</span> <span class="hljs-keyword">throws</span> -&gt; [<span class="hljs-type">Bar</span>]
<span class="hljs-keyword">class</span> <span class="hljs-title class_">var</span> someProperty: <span class="hljs-type">Int</span> { <span class="hljs-keyword">get</span> }
```
After the fix the result is the following:
```html
<span class="hljs-keyword">class </span><span class="hljs-keyword">func</span> <span class="hljs-title function_">foo</span>() <span class="hljs-keyword">async</span> <span class="hljs-keyword">throws</span> -&gt; [<span class="hljs-type">Bar</span>]
<span class="hljs-keyword">class </span><span class="hljs-keyword">var</span> someProperty: <span class="hljs-type">Int</span> { <span class="hljs-keyword">get</span> }
```

### Checklist

### DIFF
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -460,6 +460,31 @@ export default function(hljs) {
     end: /}/
   };
 
+  const CLASS_FUNC_DECLARATION = {
+    begin: [
+      /(class)\s+/,          
+      /(func)\b/,
+      /\s+/,
+      /\b[A-Za-z_][A-Za-z0-9_]*\b/ 
+    ],
+    beginScope: {
+      1: "keyword",
+      2: "keyword",
+      4: "title.function"
+    }
+  };
+
+  const CLASS_VAR_DECLARATION = {
+    begin: [
+      /(class)\s+/,          
+      /(var)\b/, 
+    ],
+    beginScope: {
+      1: "keyword",
+      2: "keyword"
+    }
+  };
+
   const TYPE_DECLARATION = {
     begin: [
       /(struct|protocol|class|extension|enum|actor)/,
@@ -524,6 +549,8 @@ export default function(hljs) {
       ...COMMENTS,
       FUNCTION_OR_MACRO,
       INIT_SUBSCRIPT,
+      CLASS_FUNC_DECLARATION,
+      CLASS_VAR_DECLARATION,
       TYPE_DECLARATION,
       OPERATOR_DECLARATION,
       PRECEDENCEGROUP,
--- a/src/languages/swift.js
+++ b/src/languages/swift.js
@@ -461,27 +461,29 @@ export default function(hljs) {
   };
 
   const CLASS_FUNC_DECLARATION = {
-    begin: [
-      /(class)\s+/,          
-      /(func)\b/,
+    match: [
+      /class\b/,          
+      /\s+/,
+      /func\b/,
       /\s+/,
       /\b[A-Za-z_][A-Za-z0-9_]*\b/ 
     ],
-    beginScope: {
+    scope: {
       1: "keyword",
-      2: "keyword",
-      4: "title.function"
+      3: "keyword",
+      5: "title.function"
     }
   };
 
   const CLASS_VAR_DECLARATION = {
-    begin: [
-      /(class)\s+/,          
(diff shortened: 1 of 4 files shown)


src/languages/swift.js
```diff
@@ -460,6 +460,31 @@ export default function(hljs) {
     end: /}/
   };
 
+  const CLASS_FUNC_DECLARATION = {
+    begin: [
+      /(class)\s+/,          
+      /(func)\b/,
```
joshgoebel, line 466: You shouldn't need the `()`.

src/languages/swift.js
```diff
@@ -460,6 +460,31 @@ export default function(hljs) {
...
+    beginScope: {
+      1: "keyword",
+      2: "keyword",
+      4: "title.function"
+    }
+  };
+
+  const CLASS_VAR_DECLARATION = {
+    begin: [
+      /(class)\s+/,
```
joshgoebel, line 479: Please split out the `\s+` as you did above.

src/languages/swift.js
```diff
@@ -460,6 +460,31 @@ export default function(hljs) {
...
   };
 
+  const CLASS_FUNC_DECLARATION = {
+    begin: [
+      /(class)\s+/,          
+      /(func)\b/,
+      /\s+/,
+      /\b[A-Za-z_][A-Za-z0-9_]*\b/ 
+    ],
+    beginScope: {
```
joshgoebel, line 470: Pretty sure if there is no `end` you can use the simpler:

- `match`
- `scope`

joshgoebel (changes requested):
Some small thoughts, but looks like we're headed the right direction.

guuido:
@joshgoebel thanks for your patience (it's my first contribution to the project), now it should be ok

joshgoebel (approved):
LGTM!

PR #4156, opened 2024-10-31
fix(rust) emoji supported in single quote strings

### LINKED ISSUE #3933: (rust) quotes around a char emoji are not colored properly
**Describe the issue**
In rust, single quotes around an emoji are not colored properly.
However, autodetect (on csharp) successfully colors it.

Rust

`highlight`

**Sample Code to Reproduce**
https://jsfiddle.net/cawyx173/

**rust, coloring doesn't work**
image

**autodetect (csharp), works**
image

**Expected behavior**
When using rust highlighting, single quotes around an emoji should have the same color as single quotes around any other character.

**Additional context**
Syntax highlighting works properly using double quotes around emojis.
Problem seen in [rust book](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html).

joshgoebel:
### Changes

Resolves #3933.  Uses begin/end to wrap strings instead of a regex that tries to match the whole string - which we do not recommend - and this is why.  Now emoji work "for free" because they are just more text inside an existing string.

### Checklist


tomsarry (approved):

steffahn:
I’m not quite familiar with grammar format here, but it turns out this change seems to be **very** breaking for a lot of Rust code:

Screenshot_20241220_070444

Screenshot_20241220_070504

steffahn:
11.11.0 is completely unusable like this, I guess this needs a revert and point release?

joshgoebel:
Thanks for the link to the example test.  Let me add that as a test case.  I think I see what went wrong here.

PR #4159, opened 2024-11-04
fix(yaml) - correct escaping behavior in single-quoted YAML strings

### LINKED ISSUE #4068: (YAML) Handling of '\'
Highlight.js seems to treat `'\'` in YAML as escaping the `'` but with `'` quoting the backslash does not escape the apostrophe in YAML.

image

guuido:
Fixes wrong escaping behavior in single-quoted strings

Resolves #4068 

### Changes
Before, backslashes in single-quoted strings were incorrectly treated as escape characters:

image

Now single-quoted strings follow YAML specification where only single quotes can be escaped (by doubling them):

image

### Checklist

### DIFF
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -42,14 +42,28 @@ export default function(hljs) {
       }
     ]
   };
-  const STRING = {
+
+  const SINGLE_QUOTE_STRING = {
     className: 'string',
     relevance: 0,
     variants: [
       {
         begin: /'/,
         end: /'/
-      },
+      }
+    ],
+    contains: [
+      {
+        begin: /''/,
+        relevance: 0
+      }
+    ]
+  };
+
+  const STRING = {
+    className: 'string',
+    relevance: 0,
+    variants: [
       {
         begin: /"/,
         end: /"/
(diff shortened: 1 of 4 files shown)


src/languages/yaml.js
```diff
@@ -42,14 +42,28 @@ export default function(hljs) {
       }
     ]
   };
-  const STRING = {
+
+  const SINGLE_QUOTE_STRING = {
     className: 'string',
     relevance: 0,
     variants: [
```
joshgoebel, line 49: We no longer need variants if there is only one.

src/languages/yaml.js
```diff
@@ -42,14 +42,28 @@ export default function(hljs) {
...
       {
         begin: /'/,
         end: /'/
-      },
+      }
+    ],
+    contains: [
+      {
+        begin: /''/,
```
joshgoebel, line 57: I wonder if this should be `char.escape` scoped?

joshgoebel:
Lets also make sure we add a single quote escaping to our markup tests - esp if we go with `char.escape`.

src/languages/yaml.js
guuido: if we use `char.escape` scope both quotes are included
```
<span class="hljs-char escape_">&#x27;&#x27;</span>
```
should I change the pattern to target only the first quote in the pair?

joshgoebel: > both quotes are included

Makes sense to me, I'd say both quotes are the entire escape sequence - which is what we're wanting to make stand out...

joshgoebel (approved):

PR #4167, opened 2024-11-11
Add Vue as 3rd party language

noraj:


joshgoebel:
Please see our 3rd party guide. You can use our own build process to create the `dist` folder which you can then check into your project - that will provide a packaged and ready to go versions (CJS and ES6) that can just be dropped into a browser - like how all our CDN grammars are packaged.

noraj:
> Please see our 3rd party guide. You can use our own build process to create the `dist` folder which you can then check into your project - that will provide a packaged and ready to go versions (CJS and ES6) that can just be dropped into a browser - like how all our CDN grammars are packaged.

Hi 👋🏻, I'm not sure to understand that comment. I saw SUPPORTED_LANGUAGES.md was listing the natively supported languages as well as the languages supported via 3rd party repositories. I saw a 3rd party repository existed for Vue.js but was missing from the list. I thought it was a mistake, so I PRed so it is listed.

As you mentioned it, I quickly checked [docs/language-contribution.rst](https://github.com/highlightjs/highlight.js/blob/main/docs/language-contribution.rst) and [extra/3RD_PARTY_QUICK_START.md](https://github.com/highlightjs/highlight.js/blob/main/extra/3RD_PARTY_QUICK_START.md). Maybe you mean the state the vue repository is in is so that it's not ready? But by no mean I meant to offer to maintain it.

joshgoebel:
Yeah this isn't ready/good to use as is, we'd like it to be "drop in" line every other grammar we publish.

PR #4169, opened 2024-11-12
fix(dart) - add highlighting for class and function names

### LINKED ISSUE #4091: (Dart) Highlight colors for Class and Constructor Calls
**Describe the issue**

Dart  (the issue is impacting Flutter)

highlight

...

**Sample Code to Reproduce**

```dart

import 'package:flutter/material.dart';

void main() {
  runApp(MaterialApp(
    home: Scaffold (
      appBar: AppBar(
        title: const Text('Scaffold Example'),
      ),
      body: const Center(
        child: Text('Hello, World!'),
      ),
      drawer: Drawer(
        child: ListView(
          children: const <Widget>[
            ListTile(title: Text('Home')),
            ListTile(title: Text('Profile')),
          ],
        ),
      ),
      bottomNavigationBar: BottomNavigationBar(
        items: const [
          BottomNavigationBarItem(icon: Icon(Icons.home), label: 'Home'),
          BottomNavigationBarItem(icon: Icon(Icons.search), label: 'Search'),
        ],
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () {},
        child: Icon(Icons.add),
      ),
    ),
  ));
}
```

**Expected behavior**
Highlight colors for Class and Constructor Calls

**Additional context**

Here are two images: 
(1) - displaying the current output from the Highlight library , where all elements appear in the same color, making it difficult to read.
(2) - a different library, which shows custom colors for classes and constructors. Both images feature code written in Dart for Flutter.

highlight

other

guuido:
Resolves #4091

### Changes
Before class and function names were not decorated and this could impact readability

image

Now these elements are correctly highlighted

image

### Checklist


src/languages/dart.js
```diff
@@ -221,6 +224,42 @@ export default function(hljs) {
...
+    /\b([A-Z]+[a-z0-9]+)+[A-Z]+/,
+  );
+
+  const CLASS_REFERENCE = {
+    relevance: 0,
+    variants: [
+      {
+        // prevent built-in types with capital letter from being selected as title.class
+        match: BUILT_IN_TYPES_REGEX,
+        skip: true
```
joshgoebel, line 243: Why skip?

src/languages/dart.js
```diff
@@ -221,6 +224,42 @@ export default function(hljs) {
...
+        scope: "title.class"
+      } 
+    ]
+  };
+
+  const FUNCTION_REFERENCE = {
+    relevance: 0,
+    match: [
+      /[a-z][A-Za-z0-9]*/,
+       /\(/
```
joshgoebel, line 256: You don't need a multi-matcher for this, just use a look-ahead to match the trailing `(`.

src/languages/dart.js
joshgoebel: Why wouldn't a built-in class also get `title.class`?  That or perhaps `built_in` I'd think...

guuido: Using this snippet as an example:
```dart
class Person {
  String name;
  int age;

  Person(this.name, this.age);
}
```
Currently the library already marks String as `built_in`, so I added this regex to prevent String from being matched as `title.class`. I may have misunderstood what skip is used for (still learning how the library works) - I set `skip: true` because otherwise String wasn't decorated at all. I thought we wanted to preserve the current behavior, but if it's ok to use `title.class` for built-in classes I can remove this part

guuido: No, without the skip String is not decorated at all. I've already removed this part because I understood it was ok to mark String with title.class, but if you want I can add it again

kevmoo:
@joshgoebel ?

Yang062 (approved):

joshgoebel (approved):

PR #4178, opened 2024-11-28
fix(ruby) don't highlight interpolation in single quoted strings

### LINKED ISSUE #3676: (Ruby) #{} substitution shouldn't be highlighted in single quotes
**Describe the issue**
In Ruby, single quote and double quote strings work differently:

> In addition to disabling interpolation [(`#{}` syntax)], single-quoted strings also disable all escape sequences except for the single-quote (`\'`) and backslash (`\\`).

https://docs.ruby-lang.org/en/3.1/syntax/literals_rdoc.html#label-Strings

But in ruby.js, they are implemented identically

https://github.com/highlightjs/highlight.js/blob/9c49a4298a981e960b6d112c746d998d51e1d1e8/src/languages/ruby.js#L132-L139

which leads it to incorrectly highlight `#{}` substitutions in single quoted strings which actually have no effect, they are just literal `#{}` characters (in the "`github`" theme):

Screenshot 2022-12-11 at 19 00 50

instead of highlighting it as a string, like it does if we remove the `#`:

Screenshot 2022-12-11 at 19 01 01

Ruby

N/A

**Expected behavior**
Single and double quoted strings should have separate highlighting rules for Ruby. 

**Additional context**
https://docs.ruby-lang.org/en/3.1/syntax/literals_rdoc.html#label-Strings

Unrelated, but for double quoted strings, that page also lists a couple interpolation syntax shorthands that are not implemented in the highlight.js grammar:

> You can also use `#@foo`, `#@@foo` and `#$foo` as a shorthand for, respectively, `#{ @foo }`, `#{ @@foo }` and `#{ $foo }`.

verhovsky:
Fixes #3676

### Changes

Not all Ruby strings can contain (`contains:`) string substitutions (`SUBST`).

### DIFF
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -15,8 +15,7 @@ export default function(hljs) {
     /\b([A-Z]+[a-z0-9]+)+/,
     // ends in caps
     /\b([A-Z]+[a-z0-9]+)+[A-Z]+/,
-  )
-  ;
+  );
   const CLASS_NAME_WITH_NAMESPACE_RE = regex.concat(CLASS_NAME_RE, /(::\w+)*/)
   // very popular ruby built-ins that one might even assume
   // are actual keywords (despite that not being the case)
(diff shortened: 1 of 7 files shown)


src/languages/ruby.js
```diff
@@ -122,56 +121,128 @@ export default function(hljs) {
...
+  }
+
+  const STRING_DELIMITERS = [
+    "()",
+    "[]",
+    "{}",
+    "<>",
+    "\\/\\/",
+    "%%",
+    "--"
```
joshgoebel, line 141: Would we want to flush this out a fit further while we're here?

joshgoebel:
I jumped in here on top of your work to organize things a bit nicer and try to DRY it up just a little... any additional thoughts?  I left some work for the future, but I think it's broken out a lot clearer now.

src/languages/ruby.js
```diff
@@ -122,56 +121,128 @@ export default function(hljs) {
...
+      }
+    })
+  }
+
+  const STRING_DELIMITERS = [
+    "()",
+    "[]",
+    "{}",
+    "<>",
+    "\\/\\/",
```
verhovsky, line 139: Did you mean

```suggestion
    "//",
```

`d.charAt(1)` won't work because it's 4 characters long.

```js
> d = "\\/\\/"
'\\/\\/'
> d.charAt(0)
'\\'
> d.charAt(1)
'/'
```

https://docs.ruby-lang.org/en/3.1/syntax/literals_rdoc.html#label-Percent+Literals says

> Each of the literals in described in this section may use these paired delimiters
> - [...] 
> - Any other character, as both beginning and ending delimiters.

Which seems like something regex can't handle.

joshgoebel: Yes if you're just using it for the function then it should be two characters just like all the rest... you're already calling `escape` in your function...

src/languages/ruby.js
joshgoebel: Clarifying what I meant: while we’re here, finish splitting/DRYing the rest instead of stopping mid-refactor.

We already carved out single-quoted / `%q` vs double-quoted / `%Q` (and heredoc). The leftover `OLD_STRINGS_TOO_MANY_VARIANTS` + TODO is still a grab bag (`%w`/`%W`/`%x`, backticks, `?…`, etc.). Ideally those get the same treatment—small discrete modes (and the right `contains` for interpolation)—rather than leaving the half-finished bucket.

joshgoebel (changes requested):

PR #4184, opened 2024-12-14
Overhaul GML highlighting

thennothinghappened:
### Changes
I've created a proper grammar for GameMaker Language, which supports many of the language's newer features, and goes a lot further than the existing one - which currently just matches a master list of keywords and function names that must be kept up to date.

The work I've done here is a port of my previous PRs to the GameMaker Manual (https://github.com/YoYoGames/GameMaker-Manual/pull/160, https://github.com/YoYoGames/GameMaker-Manual/pull/220), though since they use a bit of a customised fork, required some extra work to then bring upstream. As this PR encompasses the changes I listed in those PRs above I've avoided re-stating the same changes, though if it would be preferable I can bring them over into this PR's description.

Additional to the PRs however, I've also made some further minor fixes and improvements:
- JSDoc is now highlighted, in all forms [stated in the GameMaker Manual](https://manual.gamemaker.io/beta/en/The_Asset_Editors/Code_Editor_Properties/JSDoc_Script_Comments.htm).
- Multi-line `#macro`s are now parsed properly, though this isn't currently a noticeable change.
- Many obsolete or incorrect keyword entries have been removed, and some new ones have been added.
- The `gml.css` theme has been updated accordingly with the changes with scopes to more correctly follow the GM IDE's colour scheme.

#### Comparison
As a visual overview of the changes, here's a before and after!

##### Before
image

##### After
image

### Checklist
- [x] Updated markup tests.


joshgoebel:
This is a huge change and will need a bit of going thru.  I'll drop comments.

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+    "rollback_api_server",
+    "player_avatar_sprite",
+    "player_avatar_url",
+    "player_user_id",
+    "player_type"
+  ];
+  /**
+   * Regex for some sort of identifier - i.e, a valid name of something in code.
+   */
+  const VALID_IDENTIFIER_REG = /[a-zA-Z_][a-zA-Z0-9_]*/;
```
joshgoebel, line 1078: :naming_conventions:

Use `_RE` as the suffix, not `_REG`.

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+  /**
+   * Expressions, which form part of a valid statement.
+   * @type {Object[]}
+   */
+  const EXPRESSION = [];
+
+  /**
+   * A single-line comment.
+   */
+  const COMMENT_LINE = hljs.COMMENT('//', /\$|\n/);
```
joshgoebel, line 1094: Did you mean to escape `$` here?

What is the need for both `$` and `\n` rather than just `$` which matches end of line already?

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+  const JSDOC_TYPE_PARAM_REG = /{ *?[a-zA-Z_][a-zA-Z0-9_\.<>\|]*? *?} *?/;
+
+  const JSDOC_ANNOTATIONS = [
+    {
+      match: /@((desc(ription)?)|ignore|pure|deprecated|(func(tion)?))\b/,
+      className: "keyword"
+    },
+    {
+      match: [
+        /@(self|context) *?/,
```
joshgoebel, line 1105: What is the goal of ` *?`?

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+        JSDOC_TYPE_PARAM_REG
+      ],
+      className: {
```
joshgoebel, line 1108: :convention:

Use `scope` now, it's preferred vs the older `className`.

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+  const STRING_NUMERICAL_SUBSTITUTION = {
+    match: /{[0-9]+}/,
+    scope: "literal"
+  };
+
+  /**
+   * An escape sequence in a string.
+   */
+  const STRING_ESCAPE = {
+    scope: "literal",
```
joshgoebel, line 1198:
```suggestion
    scope: "char.escape",
```

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+        ]
+      }
+    ]
+  };
+
+  /**
+   * Various representations of numbers
+   */
+  const NUMBER = {
+    className: "literal",
```
joshgoebel, line 1253:
```suggestion
    className: "number",
```

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+   * A struct variable declaration, of `<ident>:`
+   */
+  const STRUCT_LITERAL_MEMBER = {
+    match: [
+      /\b/,
+      VALID_IDENTIFIER_REG,
+      /\s*:/
+    ],
+    className: {
+      2: "variable-instance"
```
joshgoebel, line 1399: This is not one of the scopes we support.

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+      },
+    }
+  );
+
+  doLateConcat();
+
   return {
     name: 'GML',
-    case_insensitive: false, // language is case-insensitive
+    case_insensitive: false, // language is case-sensitive
```
joshgoebel, line 1483: Case sensitive is the default and this line can be removed.
```suggestion
```

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+      COMMENT_LINE,
+      hljs.C_BLOCK_COMMENT_MODE,
+    ]
+  };
+
+  /**
+   * A template string substitution. `contains` is filled in after `EXPRESSION` is defined due to
+   * nesting.
+   */
+  const STRING_SUBSTITUTION = {
```
joshgoebel, line 1177: The scope for something like this would usually be `subst`.

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+   * for keywords such as `if (<condition>)`, so has lower priority in the mode `contains` list.
+   */
+  const FUNCTION_CALL = {
+    begin: [
+      VALID_IDENTIFIER_REG,
+      /\s*?/,
+      /\(/
+    ],
+    className: {
+      1: "function"
```
joshgoebel, line 1351: `function` is deprecated, I think you want `title.function` or `title.function.invoke`.

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+    }
+  };
+
+  /**
+   * The manual likes using `obj_` and such to define assets. Sneaky trick to make it look nicer :P
+   */
+  const USER_ASSET_CONSTANT = {
+    className: "literal",
```
joshgoebel, line 1359: Do these refer to some sort of variable/object?  If so I think `variable` would be correct.  If they can't be redefined then perhaps `variable.constant`.

We typically use `literal` only to refer to literal values that "are themselves", such as:

- null
- true
- false

Ie, most grammars have very few things scoped as `literal` - typically a very short list of keywords.  Most places this grammar uses literal another scope should likely be preferred.  Did you see our official scope list?

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+
+  /**
+   * A function declaration matching for:
+   * ```gml
+   * function <ident>(
+   * ```
+   */
+  const FUNCTION_DECLARATION = {
+    match: [
+      "function",
```
joshgoebel, line 1411:
```suggestion
      "title.function",
```

src/languages/gml.js
```diff
@@ -1,18 +1,55 @@
...
+    lateConcatTargets.push(arrays);
+    return arrays;
+  }
+
+  /**
+   * Perform the late-concatenation operation on all target arrays, converting them into effectively
+   * an array of the members that were to be concatenated.
+   */
+  function doLateConcat() {
+
```
joshgoebel, line 36: Could you explain why these helpers are necessary?

src/languages/gml.js
thennothinghappened: Ah, think that was a leftover from back when I was battling with the GM Manual's setup where it was possible for `<br>` to make its way to the parser, so there were traces here and there of various things I was trying to resolve that before I realised what the root issue was and then sanitised the inputs in their script.

I've been using `\n` over `$` mostly as it felt less "magic-ey", but that was moreso due to my limited experience with Regex as a whole (of which taking on this project has helped immensely with!)

I've now changed out `\n` to `$` wherever it is appropriate.

src/languages/gml.js
```diff
@@ -2994,135 +937,572 @@ export default function(hljs) {
...
+  /**
+   * Various types of strings supported in the engine.
+   */
+  const STRING = {
+    variants: [
+      {
+        begin: /\$"/,
+        end: "\"",
+        beginScope: "string",
+        endScope: "string",
```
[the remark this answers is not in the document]
thennothinghappened: I was having a lot of issues getting template literals to behave as I wanted. The issue was that if I simply set `scope: "string"`, then the entire thing would be wrapped with that class, which would then make anything within templates, e.g.,`$"{someVariable}"` appear string-coloured if not claimed by another expression type (e.g., a number would still look like a number, since it would claim it.)

I might have a look at how some other languages approached this issue as I wasn't happy with this solution regardless, but I recall going down a rabbit hole of greater and greater complexity language-knowledge by the parser to fix this - there's probably an easy solution I missed though :P

src/languages/gml.js
thennothinghappened: Leftover from some fine-grained extra scopes I had in the manual.

src/languages/gml.js
thennothinghappened: I've swapped that out for `variable.constant`. I was not aware of the proper full scopes list earlier as I was just working off of what was present in `gml.css` in the GM Manual initially, I've gone ahead and had a proper look over the scopes list and made various fixes here and there.

src/languages/gml.js
thennothinghappened: Effectively, they're a workaround for `EXPRESSION` being a big array of "everything that is an expression", of which some expression types (e.g., template strings) can, themselves, contain expressions. I saw that some other languages handled this by juggling `thing.contains.push(...)` but this felt pretty unwieldy, so that's what I came up with to work around it.

Though, I actually only ended up using it in one place in the current setup, and that's somewhere that it can be removed from, so I might go ahead and take it out, perhaps revisit if need be in the future? Dunno. I'll remove it. :)

src/languages/gml.js
thennothinghappened: I'm matching 0 or more spaces there, so that any of these are valid:
```gml
/// @self {Struct.Thing}
/// @self                                  {Struct.Thing}
/// @self{Struct.Thing}
```
...which matches the behaviour of GameMaker's IDE. I tried sticking `\b` there instead, but the exact behaviour of `\b` continues to elude me, so as it didn't work, I just replaced it with that instead.

thennothinghappened:
Thanks for the thorough review! I've addressed everything I could, though there's two left unresolved with comments as I was unsure how to proceed with them so leaving for further feedback.

src/languages/gml.js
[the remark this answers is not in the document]
thennothinghappened: I've come back to this again and came up with a middleground that I think is a lot more reasonable - this has been kept, but I've fixed the issue where string characters were being matched one-by-one which was making a ridiculous number of individual `<span>`s for each character. This has been fixed, but the `string` scope is still not applied to the entire span, so that the inner templates are not interfered with by `string` in themes.

src/languages/gml.js
joshgoebel: Right, but why ` *?` vs just ` *`... Isn't the default of greedy just fine here without the `?`?

thennothinghappened: To be honest I dunno why I wrote that. I think I got in my head at some point or other that `?` was "generally preferable" for no particular reason. In this case it does absolutely nothing, so I'll take that out!

src/languages/gml.js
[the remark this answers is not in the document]
joshgoebel: Maybe we'd allow some liberty with the GML theme... not sure... but typically nested items like:

```
string -> subst -> number
```

The subst color would override string, and the number color would override subst.  If a theme doesn't provide a style for `subst` that theme is broken and should be fixed.  Background can be a real problem, but not a lot of themes do crazy things with backgrounds - is GML one of them?

Still not entire I'm completely following your issue.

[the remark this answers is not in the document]
thennothinghappened: The root of the issue is that identifiers for things like local variables do not have any matching parse rule, and are unhighlighted. I didn't add any "catch all" rule at the bottom for this, though I guess I could throw in such a rule for anything that's a valid identifier with a relevance of 0. As this rule does not currently exist, here's what happens if I let the scopes be nested:
image

...the result being that these variables become string-coloured.
However, using the setup that's currently in, the issue is fine:
image

Another thing going for this to me was that it avoids the issue where a nested scope would inherit things you don't want it to - e.g., if you had a theme which made strings bold, you don't really want the contents of their templates to also become bold with them.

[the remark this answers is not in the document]
joshgoebel: Ok, I looked closely at your mode for string templates now - yeah this is not the preferred way - matching string like chunks inside strings. :).  Neat trick though.

Isn't this same effect achievable thru pure CSS with just a ~ `.string .subst` rule that sets the styling (color, bg, bold/italic) back to whatever the theme considers "standard"?

This is the kind of thing I"d prefer to solve at the engine/CSS level rather than encouraging every individual grammar to start adding "hacks" like this to accomplish stylistic goals.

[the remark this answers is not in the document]
thennothinghappened: Totally understand about avoiding these sorts of hacks.

As for the alternative, if I had `.subst` reset the colour scheme in CSS, it'd also cause the surrounding `{}` to become an un-highlighted colour. I've taken that idea though and modified it such that `.subst` resets colours, but the `{}` is given `char.escape` instead - would this be an acceptable solution?

thennothinghappened:
Hey, just checking in on the status of this PR - is there anything else I need to do to get it merged? I've not modified `CHANGES.md` as I wasn't sure when it'd be merged and that seemed like a source of issues to edit that file too soon.

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+  };
+
+  /**
+   * Various types of strings supported in the engine.
+   */
+  const STRING = {
+    variants: [
+      {
+        begin: /\$"/,
+        end: /\"/,
```
joshgoebel, line 1178: Is this \ an escape?  Do we need to escape `"`?

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+        illegal: /\n/,
+        contains: [
+          STRING_ESCAPE,
+          STRING_SUBSTITUTION
+        ]
+      },
+      {
+        scope: "string",
+        begin: "@'",
+        end: "'",
```
joshgoebel, line 1190: regex literals please

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+        relevance: 10,
+        contains: [STRING_NUMERICAL_SUBSTITUTION]
+      },
+      {
+        scope: "string",
+        begin: "@\"",
+        end: "\"",
```
joshgoebel, line 1197: regex literals please

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+      },
+      {
+        scope: "string",
+        begin: "@'",
+        end: "'",
+        relevance: 10,
+        contains: [STRING_NUMERICAL_SUBSTITUTION]
+      },
+      {
+        scope: "string",
```
joshgoebel, line 1195: If you're using variants, move everything that doesn't vary to the top-most object.

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+      }
+    },
   ];
+
+  /**
+   * Function call sites, just looking for `<ident>(`. This creates false positives
+   * for keywords such as `if (<condition>)`, so has lower priority in the mode `contains` list.
+   */
+  const FUNCTION_CALL = {
+    begin: [
```
joshgoebel, line 1263:
```suggestion
    match: [
```

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+
+  /**
+   * Pre-processor modes for macro definitions and regions.
+   */
+  const PREPROCESSOR = {
+    end: /$/,
+    variants: [
+      {
+        begin: [
+          /#macro\s+/,
```
joshgoebel, line 1415: Can you move the `\s+` out of the `meta` scope?

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+          VALID_IDENTIFIER_RE
+        ],
+        end: /$/,
```
joshgoebel, line 1418: Isn't this already covered by the top-level `end`?

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+        begin: /#(end)?region\b/,
+        beginScope: "meta",
+        contains: [COMMENT_LINE_INNER]
+      },
+    ]
+  };
+
   return {
     name: 'GML',
-    case_insensitive: false, // language is case-insensitive
```
joshgoebel, line 3113: Was this wrong?

src/languages/gml.js
```diff
@@ -37,2043 +35,21 @@ export default function(hljs) {
...
-    "ds_grid_add",
-    "ds_grid_add_disk",
-    "ds_grid_add_grid_region",
-    "ds_grid_add_region",
-    "ds_grid_clear",
-    "ds_grid_copy",
-    "ds_grid_create",
-    "ds_grid_destroy",
-    "ds_grid_get",
-    "ds_grid_get_disk_max",
```
joshgoebel, line 583: what's the thinking with all the removals?

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+
+  /**
+   * A template string substitution. `contains` is filled in after `EXPRESSION` is defined due to
+   * nesting.
+   */
+  const STRING_SUBSTITUTION = {
+    begin: /{/,
+    end: /}/,
+    beginScope: "char.escape",
+    endScope: "char.escape",
```
joshgoebel, line 1146: I don't know that we have a scope for this other than `punctuation`, but pretty sure `char.escape` doesn't fit.  That's for things like `\n`, etc...

src/languages/gml.js
```diff
@@ -2994,135 +898,569 @@ export default function(hljs) {
...
+  };
+
+  /**
+   * An escape sequence in a string.
+   */
+  const STRING_ESCAPE = {
+    scope: "char.escape",
+    variants: [
+      { match: /\\u[a-fA-F0-9]{1,6}/ },
+      { match: /\\[^\n]/ }
```
joshgoebel, line 1167: Hmmm... this is really broad, does something simpler like `\w` work maybe?

src/languages/gml.js
thennothinghappened: The way the GML highlighting for function calls has worked historically has been effectively a gigantic hard-coded, and maintained by-hand list of the standard library's functions. In 2020, GameMaker 2.3 introduced a bunch of syntax changes that brought the language's syntax a lot closer to that of say, JavaScript, for instance. One of these changes was _proper_ support of user-defined functions. This was previously done by creating assets in the IDE, and functions (or "scripts" as they were called prior to 2.3), looked a bit like this:

```gml
/// @function some_function(blah)
/// @desc Does stuff!
/// @param blah Some parameter.
/// @returns blah plus three

var blah = argument0;
return blah + 3;
```

Post-2.3, functions are now defined in a manner very reminiscent to JS:
```gml
/// Does stuff!
/// @param {Real} blah Some parameter.
/// @returns {Real} blah plus three
function some_function(blah) {
    return blah + 3;
}
```

This effectively made functions a first-class feature. As such, your average GM project nowadays uses plenty of 'em, and thus one of the very first changes I made was to highlight functions by looking for invocations: `<ident>(` will highlight the identifier `<ident>` as a function, which to my knowledge is pretty much how highlighting is done in most of the languages here. This also by extension covers `function <ident>(`, since there's no special handling for `function` other than it being a keyword.

I was, a while ago, looking into highlighting function _arguments_ too, but that spiraled into more and more complex language parsing, i.e., due to [IIFEs](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) existing in GML just as they do in JS, and that GML supports default argument values, someone could technically have an IIFE as a default argument, thus meaning that to parse arguments would require the parser to understand a much more in-depth version of the language grammar than I am currently willing to pursue.

TL;DR: functions are first-class in modern GML, and user-functions are now commonplace, so it makes sense that this hard-coded list of function names be removed (so it doesn't need to be kept up-to-date anymore), and in its place anything looking like a function call just be highlighted as one, since there isn't any kind of semantic highlighting to decide whether that identifier is known, or anything like that here.

src/languages/gml.js
thennothinghappened: I removed this as the default value is already false. If it's preferable to define it explicitly, I'll put it back :)

src/languages/gml.js
thennothinghappened: The escape is to escape the functionality of `$`, since this is matching the literal text `$"`, as GML's template strings/string interpolation is in the form `$"some text with {some_variable}, etc"`. They are effectively identical to C#'s string interpolation.

src/languages/gml.js
thennothinghappened: Good call, though I've switched that for `\S` instead of `\w`, since one of the main uses is escaping quotes that would otherwise end the string - i.e., non-alphabetical characters are also escape-able.

thennothinghappened:
Thanks for the persistence in helping me clean this up, should hopefully be up to snuff at this point!

thennothinghappened:
I've just gone ahead and finally updated the changelog, though I noted that the next version is a patch release, so let me know if I should revert that commit until later.

src/languages/gml.js
tinkerer-red: Though this does mean less maintaining, removing already existing features of the highlighter probably isn't preferred, coming from gml knowing we have arbitrary function callers like `script_execute` and `script_execute_ext`, does mean that we could still keep the list of function names and they would remain highlighted accurately.
Logic being doing nothing produces better results, so if leaving it in isnt jeopardizing anything, and it allows for catching a few extra highlights accurately then I'd say its worth keeping. Additionally getting the most up to date list is easier then ever with the GML.spec file openly available.
Though to be fair this really only means `var _func = ds_grid_create;` and other similar use cases wouldn't be accurately highlighted.

Given 2.3 is essentially feature complete with the coming GMRT, there shouldn't be any new functions or constants added with in some time. If I opted in to offering all of the modern function names from Monthly would we still be able to keep this if I update the list?

thennothinghappened: It definitely does have some value in auto-detect relevance through the GML-specific stdlib function names, but the only other cases I can think of that help would be what you pointed at above. If you're happy to take on maintaining the list I can go ahead and put it back again :)

thennothinghappened: @tinkerer-red I've gone ahead and re-added that built-in function list, and I've re-generated a fresh list from GmlSpec, so that's got all the new flex panel functions and whatnot.

Yang062 (approved):

thennothinghappened:
@joshgoebel is there anything else needing to be resolved to merge this?

Yang062 (approved):

thennothinghappened:
@joshgoebel apologies for the double-ping now, but I'm still unsure whether you need me to do anything else to merge this. If you've not had the time to look at it, no stress.

src/languages/gml.js
joshgoebel: Referring to the quotation, which is not special in regex.
```suggestion
        end: /"/,
```

src/languages/gml.js
```diff
@@ -2062,18 +2319,38 @@ export default function(hljs) {
...
     "zip_unzip",
-    "zip_unzip_async"
+    "zip_unzip_async",
   ];
-  const SYMBOLS = [
-    "AudioEffect",
-    "AudioEffectType",
-    "AudioLFOType",
+  
+  const LITERALS = [
```
joshgoebel, line 2287: **Dropped constants vs old `SYMBOLS`**

Compared to the previous grammar’s `SYMBOLS` list, about **54 names** are gone from the new `LITERALS` table (and only a handful were added, e.g. `audio_3d` renaming `audio_3D`). Can you confirm these removals are intentional (deprecated / never real constants / renamed elsewhere) rather than an accidental GmlSpec diff?

Dropped (old → missing in new):

- **Types / audio:** `AudioEffect`, `AudioEffectType`, `AudioLFOType`, `audio_3D` (see `audio_3d`)
- **Events:** `ev_audio_playback`, `ev_audio_playback_ended`, `ev_audio_recording`, `ev_dialog_async`, `ev_draw_normal`, `ev_pre_create`, `ev_push_notification`, `ev_social`, `ev_system_event`, `ev_web_async`, `ev_web_cloud`, `ev_web_iap`, `ev_web_image_load`, `ev_web_networking`, `ev_web_sound_load`, `ev_web_steam`
- **Draw / font / FX-ish:** `fa_none`, `alignmentH`, `alignmentV`, `characterSpacing`, `lineSpacing`, `paragraphSpacing`, `frameSizeX`, `frameSizeY`, `thickness`, `wrap`, `coreColor(u)r`, `glow*`, `outline*`, `shadow*`, `dropShadowEnabled`, `effectsEnabled`
- **Other:** `of_challenge_{win,lose,tie}`, `rollback_connect_error`, `wallpaper_config`, `wallpaper_subscription_data`

Some look like they may have moved under new names (`ev_*` → `ev_async_*` etc.) — a short note on the rename map (or restoring still-valid ones) would help avoid silent regressions for existing GML snippets.

---

**Side note on scope:** these ~800 entries are almost entirely **built-in named constants** from the runtime (event ids `ev_*`, virtual keys `vk_*`, colours `c_*`, blend modes `bm_*`, gamepad `gp_*`, …) — i.e. predefined constant *identifiers*, not language literals in the hljs sense (`true` / `false` / `undefined` / `NaN` / …). Per earlier review, `variable.constant` (or similar) fits that better than `literal` for the bulk of this table; happy to bikeshed naming separately from the drop list.

src/languages/gml.js
```diff
@@ -2913,13 +3141,11 @@ export default function(hljs) {
     "vk_subtract",
     "vk_tab",
     "vk_up",
-    "wallpaper_config",
-    "wallpaper_subscription_data",
-    "wrap"
   ];
+  // many of these look like enumerables to me (see comments below)
   const LANGUAGE_VARIABLES = [
```
joshgoebel: **Language variables vs old grammar**

Compared to the previous `LANGUAGE_VARIABLES` list, roughly these names disappeared:

`background_color(u)r*`, `debug_mode`, `event_action` / `event_number` / `event_object` / `event_type`, `iap_data`, `longMessage`, `message`, `room_speed`, `script`, `sequence_instance`, `stacktrace` (plus the obvious junk `colour?ColourTrack`).

And some were added (`view_*`, `room_caption`, `wallpaper_*`, …).

Can you confirm the removals are intentional (deprecated / removed from the runtime / GmlSpec) rather than an accidental list diff? Happy to restore any that are still valid in current GM.

src/languages/gml.js
```diff
@@ -2994,135 +3199,589 @@ export default function(hljs) {
...
+    STRING,
+    DS_MAP_ACCESS,
+    TERNARY,
+    PROP_ACCESS,
+    NUMBER,
+    FUNCTION_CALL,
+    USER_ASSET_CONSTANT,
+    {
+      match: VALID_IDENTIFIER_RE,
+      keywords: {
```
joshgoebel, line 3642: **Nested `EXPRESSION` keywords**

Top-level `keywords` include `"title.function": BUILT_IN_FUNCTIONS`, but this nested identifier mode only has `keyword` / `literal` / `variable.language`.

So inside template subst, ternary arms, enum values, etc., a bare built-in like `show_debug_message` may not get `title.function` unless `FUNCTION_CALL` matches (`ident(`).

Is that an intentional tradeoff, or a deficiency we should fix (e.g. mirror the full keyword set here)?

src/languages/gml.js
```diff
@@ -2994,135 +3199,589 @@ export default function(hljs) {
...
+    variants: [
+      { match: /\\u[a-fA-F0-9]{1,6}/ },
+      { match: /\\\S/ }
+    ]
+  };
+
+  /**
+   * Various types of strings supported in the engine.
+   */
+  const STRING = {
```
joshgoebel, line 3418: **Single-quoted strings**

The old grammar used `hljs.APOS_STRING_MODE` (plain `'…'`). The new one only has `@'…'` (verbatim) plus `"…"` / `$"…"` / `@"…"`.

Can you confirm GML has no plain single-quoted string literals (so dropping `APOS_STRING_MODE` is correct)?

src/languages/gml.js
```diff
@@ -2994,135 +3199,589 @@ export default function(hljs) {
...
   ];
+  /**
+   * Regex for some sort of identifier - i.e, a valid name of something in code.
+   */
+  const VALID_IDENTIFIER_RE = /[a-zA-Z_][a-zA-Z0-9_]*/;
+  /**
+   * Regex for a dot separating some LHS and RHS expression with optional whitespace (as this is
+   * supported in the engine.)
+   */
+  const DOT_ACCESSOR_RE = /\b\.\b/;
```
joshgoebel: **Dot accessor whitespace**

The comment says optional whitespace around `.` is supported; the regex is `/\b\.\b/` (no spaces), so `foo . bar` will not match.

What does GML actually allow — `foo.bar` only, or spaces around `.`? We should either allow whitespace in the pattern or fix the comment to match reality.

src/languages/gml.js
```diff
@@ -2994,135 +3199,589 @@ export default function(hljs) {
...
+   * Various representations of numbers
+   */
+  const NUMBER = {
+    scope: "number",
+    variants: [
+      { match: /(\B|^)\$[0-9a-fA-F]+/ },
+      { match: /(\B|^)#[0-9a-fA-F]+/ },
+      { match: /\b0x[0-9a-fA-F][0-9a-fA-F_]*/, relevance: 0 },
+      { match: /\b0b[01][01_]*/, relevance: 0 },
+      { match: /\b[0-9][0-9_.]*/, relevance: 0 }
```
joshgoebel: **Tighten decimal numbers**

`/\b[0-9][0-9_.]*/` is quite loose (e.g. `1.`, `1.2.3`, odd `_` placements).

Please tighten to proper int/float variants (same spirit as other grammars — explicit optional fractional part, etc.), keeping the GML-specific `$hex` / `#hex` / `0x` / `0b` forms.

src/languages/gml.js
```diff
@@ -2994,135 +3199,589 @@ export default function(hljs) {
...
+    relevance: 0
+  };
+
+  /**
+   * The manual likes using `obj_` and such to define assets. Sneaky trick to make it look nicer :P
+   */
+  const USER_ASSET_CONSTANT = {
+    scope: "variable.constant",
+    end: VALID_IDENTIFIER_RE,
+    relevance: 2,
```
joshgoebel: **Remove hand-coded `relevance`**

Please drop **all** explicit `relevance` fields in this grammar (not only this `2` — also the `10`s / `0`s on strings, numbers, `FUNCTION_CALL`, etc.).

Mode-level `relevance` is deprecated and on the way out; we do not want new/updated grammars depending on it.

src/languages/gml.js
```diff
@@ -2994,135 +3199,589 @@ export default function(hljs) {
...
+   */
+  const DS_MAP_ACCESS = {
+    match: /\[\?/
+  };
+
+  /**
+   * A ternary expression, matching partial ternary as `? <EXPRESSION> :`.
+   * Effectively exists to prevent {@link STRUCT_LITERAL_MEMBER} from stealing `<EXPRESSION> :`.
+   */
+  const TERNARY = {
```
joshgoebel, line 3541: **Ternary / struct / `[?` disambiguation — need markup**

This coupling is subtle. Please add markup tests for the edge cases so we can see where things land, then we can circle back on fixes:

- `a ? b : c`
- nested `a ? b : c ? d : e`
- `{ x: 1 }`
- `cond ? { x: 1 } : y` (struct inside ternary)
- `map[? "key"]` / `map[? key]`
- `case foo:` vs struct-looking `ident:`

If some cases are still wrong, **leave them in the markup suite** with a short comment that they are known-broken and can be fixed later — failing/annotated expectations are fine and better than no coverage.

joshgoebel:
### Markup tests for new functionality (blocker)

Beyond the ternary/struct/`[?` edges called out inline: this overhaul adds a lot of real grammar that the old keyword-dump did not have. Before merge we should have **markup tests covering the new functionality**, not only a single `default` sample. For example:

- string forms: `"…"`, `$"…{ }…"`, `@"…"`, `@'…'`, escapes, `"{0}"` subst
- `#macro` / config-scoped macros / `#region`
- JSDoc `///` and `/** … */`
- `function` / `enum` / `constructor`
- GML numbers: `$hex`, `#hex`, `0x`, `0b`, decimals
- invokes / property access
- (plus the disambiguation cases on the ternary/struct/`[?` thread)

Small focused files are better than one giant dump. Known limitations can stay as commented known-broken cases.

src/languages/gml.js
tinkerer-red: Can confirm that those are deprecated functions from Gamemaker.
Other category is all un implemented functions, things which were worked on but never finished.
Audio types are all outdated old language, while they are still technically reserved keywords they are discoraged and not suppported, the change from `3D` to `3d` was an error which was reported by me actually found here: https://github.com/YoYoGames/GameMaker-Bugs/issues/5966
Events, all of those use a dated system which has been deprocated, which is why they are all using the same syntax.
image
while they are highlighted in this Long Term Support build of gamemaker they are no longer supported, and do not work, the fact they are highlighted is more an issue with the GM IDE, and following the spec guideline supports this better.
All of the draw functions are similar all deprocated, the `fa_none` is a Font Alignment, so we have fa_left, fa_center, fa_right, fa_top, fa_middle, and fa_bottom, obviously there would be no such thing as an `fa_none` because you cant set a font anchoring to nothing. all, deprecated in full.
image


As for our side note. GML doesnt really distinguish from one or the other a constant and a literal have both semantic equivalent and highlighting equivalent. apart from longstanding dated bugs where you couldnt build a struct with very specific keywords/constant, they are exactly the same thing in our language. As an example we technically dont have `bool` as a type, we simply have signed 32 bit and signed 64 bit. So apart from the compiled code with `undefined` and `null` meaning differing things per platform, ultimately they are internally just 64 bit floats, where the first 32 bits are the "type" this is also true for all instance, audio, and any other handles, everything is essentially just a number which gets handled by the engine. It could be changed if you prefer, but ultimately would not effect anything except highlight.js renderers. it wouldnt actually reflect any better or worse for real gml syntax parsing.

src/languages/gml.js
tinkerer-red: Should be Re-added:
- debug_mode
- event_number
- event_object
- event_type
- sequence_instance

background_color - was replaced with room layers, so a layer can be your background, this no longer does anything.
```
<Function Name="room_set_background_color" Deprecated="true" ReturnType="Undefined" Locale="US" Pure="false">
	<Parameter Name="ind" Type="Asset.GMRoom" Optional="false" />
	<Parameter Name="col" Type="Constant.Color" Optional="false" />
	<Parameter Name="show" Type="Bool" Optional="false" />
</Function>
```
background_colour - same as above, the `color` vs `colour` is support for both USA and EU English support, no difference otherwise
```
<Function Name="room_set_background_colour" Deprecated="true" ReturnType="Undefined" Locale="GB" Pure="false">
	<Parameter Name="ind" Type="Asset.GMRoom" Optional="false" />
	<Parameter Name="col" Type="Constant.Color" Optional="false" />
	<Parameter Name="show" Type="Bool" Optional="false" />
</Function>
```
debug_mode - should be re-added
```
<Variable Name="debug_mode" Type="Bool" Deprecated="false" Get="true" Set="false" Instance="false">This read-only variable returns true when the game is being played in debug mode and false when being played as normal.</Variable>
```
event_action - this is so old i dont even remember what this does...
```
<Variable Name="event_action" Type="Real" Deprecated="true" Get="true" Set="false" Instance="false" />
```
event_number - should be re-added
```
<Variable Name="event_number" Type="Constant.EventNumber" Deprecated="false" Get="true" Set="false" Instance="true">This read-only variable returns the number of the event currently being called, where the number is actually referring to the "sub event" of the event, i.e.: for the step event the event number could be any one of the constants ev_step_normal, ev_step_begin, or ev_step_end. For a full list of constants that are available for the specific sub-events see event_perform().</Variable>
```
event_object - should be re-added
```
<Variable Name="event_object" Type="Asset.GMObject" Deprecated="false" Get="true" Set="false" Instance="true">This read-only variable returns the object index of the instance which is running the event being checked.</Variable>
```
event_type - should be re-added
```
<Variable Name="event_type" Type="Constant.EventType" Deprecated="false" Get="true" Set="false" Instance="true">This read-only variable returns the type of event currently being executed.</Variable>
```
iap_data - extensions now handle "In App Purchases" per store front, these old functions are all deprocated and have not seen updates in years.
```
<Variable Name="iap_data" Type="Undefined" Deprecated="true" Get="true" Set="false" Instance="false" />
```
room_speed - deprocated because each room no longer controls update rate, and that is a product of the game settings, this does nothing anymore
```
<Variable Name="room_speed" Type="Real" Deprecated="true" Get="true" Set="true" Instance="false" />
```

sequence_instance - should be readded
https://manual.gamemaker.io/lts/en/#t=GameMaker_Language%2FGML_Reference%2FAsset_Management%2FSequences%2Fsequence_instance.htm
```
<Variable Name="sequence_instance" Type="Struct.SequenceInstance" Deprecated="false" Get="true" Set="false" Instance="true">This is a built-in variable that is part of the instance variables created for every object instance in your game. If the instance is being controlled by a sequence, this variable will hold the sequence instance struct for the Sequence controlling the instance, otherwise it will be undefined. This is a read-only variable and cannot be changed.</Variable>
```

---

the following are all struct 'keynames' which is not required, and not a builtin keyword or anything special, it only helps young devs with auto complete for their error handling, could be included, could be removed, really either or are fine. here is a link to that documentation: https://manual.gamemaker.io/lts/en/#t=GameMaker_Language%2FGML_Reference%2FDebugging%2Fexception_unhandled_handler.htm&rhsearch=longMessage&rhhlterm=longMessage

longMessage
message
script
stacktrace
```
<Structure Name="Exception">
	<Field Name="message" Type="String" Get="true" Set="true">Short message for this exception.</Field>
	<Field Name="longMessage" Type="String" Get="true" Set="true">Long message for this exception.</Field>
	<Field Name="script" Type="String" Get="true" Set="true">Describes the script where this exception came from.</Field>
	<Field Name="stacktrace" Type="Array.String" Get="true" Set="true">The stack frame that the exception was generated from.</Field>
</Structure>
```

src/languages/gml.js
tinkerer-red: correct, gamemaker does not support `'abc'` this was at one point supported, about 10+ years ago, but no longer supported, and not planned for support in the future.

image


triple quotes also do not exist in our language, incase that needs clarifying too.

src/languages/gml.js
tinkerer-red: image

all of which are supported. though rarely used.

src/languages/gml.js
tinkerer-red: the liberal use/acceptance of `_` is intended, mostly for spacing out your code in a more human readable way.
image

however the `1.2.3` is technically supported, it will throw an error on compile, so not practically supported, i would opt for only allowing a single `.` unless someone else could confirm why this would be allowed in any numeric representation, maybe binary uses it for some arcane reason.

however i will say that `1.` is perfectly valid syntax. it would compile to 1.0
image
Reason for this is mostly to do with binary operations on a unsigned int, being allowed to convert it into a signed int autonomously, though extremely rarely used as often one would simply use `real()`

src/languages/gml.js
tinkerer-red: As a note for nested ternary, while they are syntax highlighted supported, they are not compile supported, so they will appear as though they work, but ultimately they would not. and would suggest we favor not what the dated syntax highlighter of the IDE says "works", but instead what the actual language supported. one day gamemaker will receive this update, however in LTS 2026 this is still not the case, and would prefer we catch that ahead of time.
image

src/languages/gml.js
joshgoebel: > if you prefer, but ultimately would not effect anything except highlight.js renderers.

Yeah, how it looks is often the point (when it's unclear which to pick), but we can leave it as is for now to get this shipped and see if anyone objects to the classification later.

src/languages/gml.js
joshgoebel: > could be included, could be removed, 

I'll defer to you then. :-)

src/languages/gml.js
joshgoebel: >  but instead what the actual language supported.

Sure, agreed.

src/languages/gml.js
tinkerer-red: Feel like we should stick to the standards already introduced by the built in instance variables, examples include `x` `y` and `id` which technically an instance is just a typed struct, so since those are included and syntax highlighted we should probably favor these to also be highlighted as they require nothing special for syntax highlighting (like requiring a prefixed dot accessor, like an enum would).

image

So would argue they should probably be re-added as well.

src/languages/gml.js
joshgoebel: I suppose we could come back and support later then...

src/languages/gml.js
joshgoebel: Ok, we can leave for now.

src/languages/gml.js
thennothinghappened: `fa_none` should be in there as GML re-uses the prefix `fa_` for both "font alignment" and "file attributes". From memory I generated this from Feather's GmlSpec, so no idea why that wouldn't be in there.

src/languages/gml.js
thennothinghappened: I think I'll leave out `longMessage` etc as Code Editor 2 in the IDE doesn't highlight these anymore, at least from what I've seen.

src/languages/gml.js
thennothinghappened: I vaguely recall having some kind of issue with built-ins back when I wrote this, but copying that line there doesn't seem to break anything I can see today, so I'll put it back.

src/languages/gml.js
thennothinghappened: Writing e.g. `self.x` now highlights `x` as a built-in.

thennothinghappened:
I've added tests for everything you've named above - is there anything else I should test for, or anything else I should do for this to be ready for merging? Also, would you like the default test case gone, or to leave it be?

joshgoebel (approved):

thennothinghappened:
Hooray!! Thanks for your patience with this :)

PR #4188, opened 2024-12-18
fix(ruby): stricter block params, allowing bitwise OR

### LINKED ISSUE #4187: (ruby) Bitwise OR pipe confused with Block Parenthesis pipes
**Describe the issue**
Bitwise OR operators which use pipe "|", i.e `1 | 2` get confused with block parenthesis syntax wich are two pipes, i.e.
```ruby
[1,2,3].each do |num|
   # ...
end
```
Currently, anything after a pipe is forever highlighted as a parameter until another pipe appears.

ruby

Either one.

...

**Sample Code to Reproduce**
```ruby
1 | 2 # Bitwise OR
puts "between '2' and '5.times' are all hljs-params"
5.times do |num|
 puts num
end
```
https://jsfiddle.net/oh57e1s2/

**Expected behavior**
Bitwise OR operator "|" should work just like any other &^~+-*/ etc.

dysonreturns:
Block parenthesis, which are two pipe symbols are applicable after a `do |param|` or inside of an inline block `{ |param| ... }`.

Resolves #4187

### Changes

This change ensures the pipes `|param|` are preceded by an open bracket `{` or a `do`.
By being specific when these parenthesis occur, bitwise OR operations `|` don't read what follows as params indefinitely.

### Checklist

### DIFF
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -243,6 +243,16 @@ export default function(hljs) {
     ]
   };
 
+
+  const BLOCK_PARAMS = {
+      className: 'params',
+      begin: /(do|\{)\s\|(?!=)/,
+      end: /\|/,
+      excludeBegin: true,
+      excludeEnd: true,
+      keywords: RUBY_KEYWORDS,
+  };
+
   const INCLUDE_EXTEND = {
     match: [
       /(include|extend)\s+/,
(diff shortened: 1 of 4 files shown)


jimtng:
Thanks! I've just noticed this bug too

src/languages/ruby.js
```diff
@@ -243,6 +243,16 @@ export default function(hljs) {
     ]
   };
 
+
+  const BLOCK_PARAMS = {
+      className: 'params',
+      begin: /(do|\{)\s\|(?!=)/,
```
joshgoebel, line 249: Do we need a look-ahead for the do?  How do we know it's going to have params at all? I feel like we need some new markup test cases here to make sure we get the expected behavior here.

Wouldn't this break?

```
setup do
  # code
end
```

joshgoebel: Ping.

PR #4198, opened 2025-01-05
Provide highlighting of Oberon source code

miasapkarl:
This supports highlighting for the final revision of the original Oberon language from 2016 (also called Oberon-07) as defined by the Oberon report https://miasap.se/obnc/oberon-report.html

### Checklist


joshgoebel:
Please see our 3rd party contrib docs:

https://highlightjs.readthedocs.io/en/latest/language-contribution.html#

We don't accept new grammars into core.  But host it in your own repo and we'll be happy to add it to SUPPORTED_LANGUAGES.md with a link to your repo.

PR #4199, opened 2025-01-22
(build) allow third-party packages to override default languages

Lysxia:
### Changes

For example, I can define my own `rust.js` as a third-party package in `extra/`, then `node tools/build.js rust` will pick up my version instead of the default version.

### Checklist
- [x] ~~Added markup tests, or~~ they don't apply here because this just modifies the build script

### DIFF
--- a/tools/lib/dependencies.js
+++ b/tools/lib/dependencies.js
@@ -41,6 +41,8 @@ const isGroup = (id) => id[0] === ":";
  *
  * This also resolves any requires and dependency ordering issues.
  * The returned list can be registered sequentially and should "just work".
+ * If multiple languages have the same name, the last one wins.
+ * This gives priority to third-party packages so they can override built-in languages.
  *
  * @param {array<Lanuage>} allLanguages - full list of languages
  * @param {array<name|group_name>} includes - which languages or groups to include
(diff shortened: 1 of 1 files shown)


joshgoebel:
I think this is ok, since one would only ever do this on purpose - but I'm going to give it just a little more thought.

joshgoebel:
Which build or builds are you using?  The current implementation really wasn't designed to deal with overlapping names at all.  I worry any behavior here may be accidental so I need to take a closer look.

Think I'm fine with the idea of this change though (and it likely was my original intent - that 3rd party get added last, though I'm not sure I considered of dups).

joshgoebel:
Ah, `filter` makes this work because it's scanning the list top-down to find matches (and it will find the first `rust` when there are many)... I think a better solution here would be to first call `reverse` on all languages in `dependencies.js#filter` so that the filter works from bottom to top - so that languages added later to the list win when there is a name match.

Can you give that a shot and update this PR?

Technically this would be a breaking change, but I may let it slide since this strikes me as very edge case.

Lysxia:
Changing `filter` makes sense to me. How this commit?

At the moment I'm using the browser build.

joshgoebel:
> Changing filter makes sense to me. How this commit?

Looks good.  Safe to assume you tested and it also gets the job done?

Lysxia:
Indeed this works for me.

joshgoebel (approved):
Tested by PR submitter, looks good to me.  CI green.

Yang062 (approved):

PR #4213, opened 2025-02-25
fix(ruby) symbols, string interpolation, class names with underscores

### LINKED ISSUE #3676: (Ruby) #{} substitution shouldn't be highlighted in single quotes

jimtng:
Also fix #3676

Change from:

image
image
image

To:

image
image
image

### Checklist
* [x]  Added markup tests
* [x]  Updated the changelog at `CHANGES.md`

### DIFF
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -115,7 +115,7 @@ export default function(hljs) {
     begin: /#\{/,
     end: /\}/,
     keywords: RUBY_KEYWORDS,
-    relevance: 10
+    relevance: 2
   };
   const STRING_INTERPOLABLE = {
     className: 'string',
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -11,7 +11,7 @@ export default function(hljs) {
   const regex = hljs.regex;
   const RUBY_METHOD_RE = '([a-zA-Z_]\\w*[!?=]?|[-+~]@|<<|>>|=~|===?|<=>|[<>]=?|\\*\\*|[-/+%^&*~`|]|\\[\\]=?)';
   // TODO: move concepts like CAMEL_CASE into `modes.js`
-  const CLASS_NAME_RE = /\b([A-Z]+[a-z0-9_]+)+[A-Z]*/;
+  const CLASS_NAME_RE = /\b([A-Z]+[a-z0-9_]*)+[A-Z]*/;
   // very popular ruby built-ins that one might even assume
   // are actual keywords (despite that not being the case)
   const PSEUDO_KWS = [
@@ -362,7 +362,7 @@ export default function(hljs) {
     },
     {
       className: 'symbol',
-      begin: hljs.UNDERSCORE_IDENT_RE + '(!|\\?)?:',
+      begin: hljs.UNDERSCORE_IDENT_RE + '(!|\\?)?:(?!:)',
       relevance: 0
     },
     {
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -115,7 +115,7 @@ export default function(hljs) {
     className: 'subst',
     begin: /#\{/,
     end: /\}/,
-    keywords: RUBY_KEYWORDS,
+    keywords: RUBY_KEYWORDS
   };
   const STRING_INTERPOLABLE = {
     className: 'string',
@@ -134,35 +134,35 @@ export default function(hljs) {
       },
       {
         begin: /%[QWx]?\(/,
-        end: /\)/,
+        end: /\)/
       },
       {
         begin: /%[QWx]?\[/,
-        end: /\]/,
+        end: /\]/
       },
       {
         begin: /%[QWx]?\{/,
-        end: /\}/,
+        end: /\}/
       },
       {
         begin: /%[QWx]?</,
-        end: />/,
+        end: />/
       },
       {
         begin: /%[QWx]?\//,
-        end: /\//,
+        end: /\//
       },
       {
         begin: /%[QWx]?%/,
-        end: /%/,
+        end: /%/
       },
       {
         begin: /%[QWx]?-/,
-        end: /-/,
+        end: /-/
       },
       {
         begin: /%[QWx]?\|/,
-        end: /\|/,
+        end: /\|/
       },
       // heredocs
       {
@@ -194,35 +194,35 @@ export default function(hljs) {
       },
       {
         begin: /%[qw]?\(/,
-        end: /\)/,
+        end: /\)/
       },
       {
         begin: /%[qw]?\[/,
-        end: /\]/,
+        end: /\]/
       },
       {
         begin: /%[qw]?\{/,
-        end: /\}/,
+        end: /\}/
       },
       {
         begin: /%[qw]?</,
-        end: />/,
+        end: />/
       },
       {
         begin: /%[qw]?\//,
-        end: /\//,
+        end: /\//
       },
       {
         begin: /%[qw]?%/,
-        end: /%/,
+        end: /%/
       },
       {
         begin: /%[qw]?-/,
-        end: /-/,
+        end: /-/
       },
       {
         begin: /%[qw]?\|/,
-        end: /\|/,
+        end: /\|/
       },
       // in the following expressions, \B in the beginning suppresses recognition of ?-sequences
       // where ? is the last character of a preceding identifier, as in: `func?4`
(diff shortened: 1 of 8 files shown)


src/languages/ruby.js
```diff
@@ -120,19 +114,16 @@ export default function(hljs) {
     className: 'subst',
     begin: /#\{/,
     end: /\}/,
-    keywords: RUBY_KEYWORDS
+    keywords: RUBY_KEYWORDS,
+    relevance: 10
```
joshgoebel, line 118: Please do not relevance spike.  Relevance is going away entirely in v12.

src/languages/ruby.js
```diff
@@ -142,45 +133,45 @@ export default function(hljs) {
         end: /`/
       },
       {
-        begin: /%[qQwWx]?\(/,
```
joshgoebel, line 145: Why removal of the lowercase sigils?

joshgoebel: Ah I see now.

src/languages/ruby.js
```diff
@@ -11,13 +11,7 @@ export default function(hljs) {
...
   const RUBY_METHOD_RE = '([a-zA-Z_]\\w*[!?=]?|[-+~]@|<<|>>|=~|===?|<=>|[<>]=?|\\*\\*|[-/+%^&*~`|]|\\[\\]=?)';
   // TODO: move concepts like CAMEL_CASE into `modes.js`
-  const CLASS_NAME_RE = regex.either(
-    /\b([A-Z]+[a-z0-9]+)+/,
-    // ends in caps
-    /\b([A-Z]+[a-z0-9]+)+[A-Z]+/,
-  )
-  ;
-  const CLASS_NAME_WITH_NAMESPACE_RE = regex.concat(CLASS_NAME_RE, /(::\w+)*/)
+  const CLASS_NAME_RE = /\b([A-Z]+[a-z0-9_]+)+[A-Z]*/;
```
joshgoebel, line 14: Doesn't this break (at least) the inheritance highlighting?

jimtng: Can you give an example syntax of what you mean?

```ruby
Class::With::Namespace
```
will work fine and better, because it doesn't highlight the `::` much like it does here in github, or vscode.

src/languages/ruby.js
```diff
@@ -326,20 +375,28 @@ export default function(hljs) {
...
+    },
     {
       className: 'symbol',
       begin: hljs.UNDERSCORE_IDENT_RE + '(!|\\?)?:',
       relevance: 0
     },
     {
       className: 'symbol',
-      begin: ':(?!\\s)',
+      begin: '(?<!:):(?!\\s|:)',
```
joshgoebel, line 387: We can't use negative lookbehind until v12 as that would be a breaking change.

joshgoebel: The rule on line 378 is designed to deal with `::` and prevent false symbol positives - until v12 where it would be simpler to use negative lookbehind.

jimtng: Removing this negative lookbehind will cause this:
image

The rule on line 378 doesn't seem to do its job.

So I had to adjust line 378 so I can remove the negative lookbehind.

src/languages/ruby.js
jimtng: I've set this to 2

joshgoebel: Please remove all relevance changes you've introduced.  You can leave what is already there... this should reduce merge conflicts with my other relevance changes.

src/languages/ruby.js
joshgoebel:
```
class C::E < A::B
end
```

We have special rules to match syntax where we KNOW an identifier is a class (or class with namespaces).

>  it doesn't here in github, or vscode

Striving for exact matches against other highlighting tools isn't a thing we care about here.

---

In this case I'm sympathetic to not highlighting the `::`, but only if it can be done without adding a lot of complexity or advanced parsing to the grammar.

joshgoebel: You really want something like `[class](::[class])*` but our multi-match engine can't really handle that - with discrete coloring of the different pieces.  Maybe if you tried a long sequence with some items that matched 0 length, but I'm not sure I've tested multi-matching in that type of scenario before.

src/languages/ruby.js
jimtng: removed

src/languages/ruby.js
jimtng: This case would've worked fine, except it didn't recognise single-letter classes. I've adjusted it to do so.
```
class C::E < A::B
end
```

image

src/languages/ruby.js
```diff
@@ -11,13 +11,7 @@ export default function(hljs) {
...
   const RUBY_METHOD_RE = '([a-zA-Z_]\\w*[!?=]?|[-+~]@|<<|>>|=~|===?|<=>|[<>]=?|\\*\\*|[-/+%^&*~`|]|\\[\\]=?)';
   // TODO: move concepts like CAMEL_CASE into `modes.js`
-  const CLASS_NAME_RE = regex.either(
-    /\b([A-Z]+[a-z0-9]+)+/,
-    // ends in caps
-    /\b([A-Z]+[a-z0-9]+)+[A-Z]+/,
-  )
-  ;
-  const CLASS_NAME_WITH_NAMESPACE_RE = regex.concat(CLASS_NAME_RE, /(::\w+)*/)
+  const CLASS_NAME_RE = /\b([A-Z]+[a-z0-9_]*)+[A-Z]*/;
```
joshgoebel, line 14: I think we need the `+`... On it's own, without other context, `A` is not a class, it's a constant.  Classes are CamelCase. Constants are UPCASE.

```suggestion
  const CLASS_NAME_RE = /\b([A-Z]+[a-z0-9_]+)+[A-Z]*/;
```

src/languages/ruby.js
joshgoebel: Please use "show the structure". as you develop.  If you're matching all those "one off" with just your class rule then you're missing out on the fact that the portion after the `<` is not just a `title.class`, it's a `title.class.inherited`.

This is why the simple multi-match rules exist - to flesh out more detail.

jimtng:
@joshgoebel can we use a positive lookbehind?

src/languages/ruby.js
jimtng: I wasn't aware of the different handling for `title.class.inherited`. 

Whilst special casing `class A` works using a multi-match rule, it wouldn't help for `A.method(foo)` which will still be matched as a constant. So I'd suggest we simplify and not handle the `class A` case at all.

In the current version, it doesn't work either: https://highlightjs.org/demo#lang=ruby&v=1&theme=atom-one-dark&code=Y2xhc3MgQQplbmQKCscNOjpCywsgPCBDOjpExxJGb286OkZvbyA8yQvKGsYVyEjEC8VXQSA9ICdYJw%3D%3D

jimtng: image

src/languages/ruby.js
jimtng: Reverted to `+`

joshgoebel:
> an we use a positive lookbehind? 

No.  Until v12 we have to support older versions of Safari that don't have this built-in.  What issue are we still trying to solve with lookbehind?

jimtng:
> What issue are we still trying to solve with lookbehind?

I used a positive lookbehind for `title.class.inherited`. I'll try again using a different approach.

jimtng:
@joshgoebel I've managed to do the inheritance without the use of a lookbehind

jimtng:
@joshgoebel are the any other problems with this PR that I should address?

Yang062 (approved):

jimtng:
Ping

Issue #4214, opened 2025-02-26
(C, C++) Implement C2y and C++26 changes


Eisenwave:
**Describe the issue**
The grammars for C and C++ do not yet include the latest C2y and C++26 features.

**Additional context**
This list is quite likely incomplete, but here are some of the changes:
- C++26 contracts (P2900) have been accepted (contextual keywords `pre`, `post`, and true keyword `contract_assert`)
- `#embed` is in C++, and on track to be added to C2y (add keyword in preprocessor mode)
- C++26 `delete("reason")` is possible, but `delete` is currently highlighted as a function name instead of a keyword in that situation
- C2y is on track to add `defer` as a feature (add keyword)
- C2y already supports `break label` and there are ongoing efforts to add this to C++; do we need dedicated label highlighting?

PR #4217, opened 2025-02-28
fix(cpp) exclude keywords from function calls

Eisenwave:
Related to https://github.com/highlightjs/highlight.js/issues/4214.

The highlighter for C++ behaves treats keywords followed by parentheses as function calls (with some exceptions), and this is harmful:

```cpp
// highlighted as a function call, but is grammatically a declaration
static_assert("");

// providing a reason for deletion should not be a function call!
void awoo() = delete("reason");

// highlighted as a function call, but that is slightly forgivable
// because it's called "function-style cast"
int(0);
// ... but then why is a function-style cast with braces not a function call?!
int { 0 };
// ... and why is this other cast not a function call?!
(int) x;
```
In general, it is wrong to highlight keywords followed by parentheses as function calls. This is always wrong, not just in specific cases like `for()` or `if()`.

### Changes
Exclude all keywords followed by parentheses from being classified as `function.dispatch = built_in`.

### Checklist
- [x] Added markup tests

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -425,11 +425,7 @@ export default function(hljs) {
       _hint: FUNCTION_HINTS },
     begin: regex.concat(
       /\b/,
-      /(?!decltype)/,
-      /(?!if)/,
-      /(?!for)/,
-      /(?!switch)/,
-      /(?!while)/,
+      `(?!${RESERVED_KEYWORDS.join('|')})`,
       hljs.IDENT_RE,
       regex.lookahead(/(<[^<>]+>|)\s*\(/))
   };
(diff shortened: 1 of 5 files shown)


src/languages/cpp.js
```diff
@@ -425,11 +425,7 @@ export default function(hljs) {
       _hint: FUNCTION_HINTS },
     begin: regex.concat(
       /\b/,
-      /(?!decltype)/,
-      /(?!if)/,
-      /(?!for)/,
-      /(?!switch)/,
-      /(?!while)/,
+      RegExp(`(?!${RESERVED_KEYWORDS.join('|')})`),
```
joshgoebel, line 428: Pretty sure we don't need the RegExp since `concat` handles all that behind the scenes?  Otherwise looks good.

joshgoebel (approved):

PR #4218, opened 2025-02-28
enh(cpp) add C++26 keywords, #embed, and compiler-builtin types

eisenwave:
Related to https://github.com/highlightjs/highlight.js/issues/4214.

### Changes
- (C++26) treat `#embed` as a known preprocessor directive
- (C++26) add `contract_assert`, `pre`, and `post` keywords
- (Clang) add `_Atomic` and `_BitInt` keywords

As for the last change, Clang supports `_Atomic` and `_BitInt` keywords in C++ mode as well, with the same semantics as in C. I am also working on a proposal that introduces bit-precise integers to C++, and either through standardization or common practice, at least `_BitInt` is expected to become a common non-standard keyword.

### Checklist
- [x] Added markup tests

The test expectations depend on decisions made in https://github.com/highlightjs/highlight.js/pull/4217, so I haven't written tests yet.

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -108,7 +108,11 @@ export default function(hljs) {
     scope: 'meta',
     begin: /#\s*include\b/,
     end: /$/,
-    keywords: { keyword: 'include' },
+    keywords: {
+      keyword: [
+        'include'
+      ]
+    },
     contains: [
       {
         // the `\` at the end of a line signaling continuation
@@ -128,9 +132,25 @@ export default function(hljs) {
     className: 'meta',
     begin: /#\s*[a-z]+\b/,
     end: /$/,
-    keywords: { keyword:
-        'if else elif endif define undef warning error line '
-        + 'pragma _Pragma ifdef ifndef include' },
+    keywords: {
+      keyword: [
+        'if',
+        'else',
+        'elif',
+        'embed',
+        'endif',
+        'define',
+        'undef',
+        'warning',
+        'error',
+        'line',
+        'pragma',
+        '_Pragma',
+        'ifdef',
+        'ifndef',
+        'include'
+      ]
+    },
     contains: [
       {
         begin: /\\\n/,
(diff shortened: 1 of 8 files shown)


Yang062 (approved):

joshgoebel:
Looks good.

Need rebase and and entry added to changelog. Thanks.

Eisenwave:
@joshgoebel Could you help me out with this one? I'm kinda clueless as to how I can make the following test cases work, which are currently failing:
```cpp
void throwing(int x) noexcept(false);
void deleted() = delete("reason");
void contracts(int x) pre(x > 10) post(r : r != 0);
```
highlight.js treats `noexcept` `pre`, and `post` as function `title`s and the content in parentheses as `parameters`.

I think the way to fix it is to simply exclude keywords from being function titles, but doing so breaks highlighting for the function completely.

CHANGES.md
```diff
@@ -60,6 +60,7 @@ Core Grammars:
 - fix(yaml) - Fixed wrong escaping behavior in single quoted strings [guuido]
 - enh(nim) - Add `concept` and `defer` to list of Nim keywords [Jake Leahy]
 - fix(cpp) - Exclude keywords from highlighting as function calls [Eisenwave]
+- enh(cpp) add C++26 keywords, `#embed`, and compiler-builtin types [Eisenwave]
```
joshgoebel, line 63: You also need to add your footnote on line ~90...

joshgoebel (approved):

CHANGES.md
Eisenwave: It's already there.

joshgoebel: Ah, I see.  I thiknk you need to move it up to the `Version 11.11.2` section though now that time has moved on.

joshgoebel:
> Could you help me out with this one? 

Maybe it's because you're trying to hard to think like a full parser.  Use of `function` is deprecated now - we no longer try and figure out what a function is (from start to finish, or even the function definition)... we only care about highlighting types, `title.function`, etc... if you simplify along those lines, does that help?  And also `title.function.invoke` if we can detect the difference between definition and invokation.

src/languages/cpp.js
```diff
@@ -99,7 +99,7 @@ export default function(hljs) {
     begin: /#\s*[a-z]+\b/,
     end: /$/,
     keywords: { keyword:
-        'if else elif endif define undef warning error line '
+        'if else elif embed endif define undef warning error line '
```
joshgoebel: Please go ahead and refactor this to one keyword per line (as seen elsewhere).

joshgoebel (changes requested):
Almost there! Thanks.

joshgoebel (approved):

PR #4228, opened 2025-03-19
Fix unified diff hunk header highlight to allow for single-number case

### LINKED ISSUE #4227: (diff) Won't highlight hunk headers with a single number instead of a number pair
**Describe the issue**
Diff is looking for hunk header that looks like this:

```
@@ -<num>,<num> +<num>,<num> @@
```

However, it's OK to omit the second number of each pair in some cases, see https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html

> **If a hunk contains just one line, only its start line number appears.** Otherwise its line numbers look like ‘start,count’. An empty hunk is considered to start at the line that follows the hunk.
> 
> If a hunk and its context contain two or more lines, its line numbers look like ‘start,count’. **Otherwise only its end line number appears.** An empty hunk is considered to end at the line that precedes the hunk.

So these should also be valid, for example:

```
@@ -<num> +<num> @@
```

```
@@ -<num>,<num> +<num> @@
```

This is using diff.

highlight

**Sample Code to Reproduce**

See above

**Expected behavior**

The above examples should also be highlighted as `hljs-meta`. 

**Additional context**

I can see the exact regexp causing this at https://github.com/highlightjs/highlight.js/blob/85b20421a2620b765508f16289e0592e0563cfcb/src/languages/diff.js#L20

It can probably be simply replaced with this:

```
/^@@ +-\d+(?:,\d+)? +\+\d+(?:,\d+)? +@@/
```

sushicw:
Resolves #4227

### Changes
Update the regex for matching a unified hunk header to allow for single numbers as well as comma-paired numbers.

From https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html, emphasis mine:

> **If a hunk contains just one line, only its start line number appears.** Otherwise its line numbers look like ‘start,count’. An empty hunk is considered to start at the line that follows the hunk.
> 
> If a hunk and its context contain two or more lines, its line numbers look like ‘start,count’. **Otherwise only its end line number appears.** An empty hunk is considered to end at the line that precedes the hunk.

### Checklist


src/languages/diff.js
```diff
@@ -17,7 +17,7 @@ export default function(hljs) {
         className: 'meta',
         relevance: 10,
         match: regex.either(
-          /^@@ +-\d+,\d+ +\+\d+,\d+ +@@/,
+          /^@@ +-\d+(?:,\d+)? +\+\d+(?:,\d+)? +@@/,
```
joshgoebel, line 20: Lets keep this simple by keeping the original regex line and then adding a new one for the single number case.

sushicw: We would actually need three new lines for a total of four if we did it that way, since `-1 +1,2` and `-1,2 +1` are also both valid:

```
/^@@ +-\d+,\d+ +\+\d+,\d+ +@@/
/^@@ +-\d+ +\+\d+,\d+ +@@/
/^@@ +-\d+,\d+ +\+\d+ +@@/
/^@@ +-\d+ +\+\d+ +@@/
```

IMHO this does not end up simpler to read, but I don't mind changing it if you prefer, just let me know!

I'll also add test coverage for those other cases, which is probably a good idea in any case.

joshgoebel: > I'll also add test coverage for those other cases, which is probably a good idea in any case.

Yes, awesome.

> IMHO this does not end up simpler to read,

Yep, matter of opinion.  What might help further though is put an example in a comment after each of the regex.  That should make it 100% clear I think for all readers.

sushicw: Sounds good, done.

Yang062 (approved):

Yang062 (approved):

joshgoebel (approved):

Issue #3890, opened 2023-10-15
Time to finally allow negative lookbehind?


joshgoebel:
I just checked and the latest versions of Safari (that now support negative lookbehind) [16.4, 16.5, etc] go all the way back to Big Sur, supporting:

- Big Sur (November 12, 2020)
- Montery (October 25, 2021)
- Ventura (October 24, 2022)
- Sonoma (September 26, 2023)

Big Sur support [extends all the way back](https://support.apple.com/en-us/HT211238) to Mac's from 2013.  I have some sympathy for those who hold off a while on OS releases, but little for those who would avoid installing the newest version of Safari (since often there are security fixes, etc)... so I think if we allow support for negative lookbehind we'd only be dropping support for Mac's 10+ years old or those refusing to upgrade their Safari - in which case I'd suggest they use another browser instead, such as Chrome.

**So I say it's time to allow look-behind...**

Thoughts, @allejo ?

allejo:
As long as the introduction of negative lookbehinds in grammars is introduced in 12.0 and not in our 11.x series, you have my support :+1:

Solely on principle for semantic versioning, I'd be hesitant to introduce a breaking change in browser support in a minor release

kuenzign:
I have quite a few personal lexers that really require the use of lookbehinds as my current workarounds do not work for all cases.  Is there a timeline on when lookbehinds will be allowed?

joshgoebel:
Well if they are truly personal, just use them... it's only the core library that we aren't allowing them in v11 because of semantic versioning.  I haven't had much time lately to do a major release but it's a long weekend so I'll take a look at where we stand with v12...

@allejo The old website handled us being able to publish two major versions at once - and it just did the right thing (pay attention to the new one)... does the new website share this characteristic - what if say we pushed a few 12.0.0-beta or alpha relases?

kuenzign:
Ah it looks like all I needed to do was change the `ecmaVersion` to `2018` in the config to allow it to work.  Thanks!

joshgoebel:
Closing as the decision has been made, this will be allowed when v12 ships.

PR #4239, opened 2025-04-11
fix(bash): correctly highlight doctags in comments again (#4234)

### LINKED ISSUE #4234: (bash) regression: “TODO” in comments no longer highlighted
**Describe the issue**
```bash
# This is a comment with a TODO item in it
```
should have the `TODO` highlighted as scope `doctag`, and *used* to, but no longer does as of d78749a59410387f7232beb296e4606c9f4b7d08 (#3918).

bash

highlight

**Sample Code to Reproduce**
```bash
# This is a comment with a TODO item in it
```

**Expected behavior**
**TODO** is highlighted as `doctag` scope.

**Additional context**
I upgraded highlight.js and this stopped working. I bisected the problem to d78749a59410387f7232beb296e4606c9f4b7d08 (#3918); in particular the relevant problem seems to be somewhere in this code:

https://github.com/highlightjs/highlight.js/blob/85b20421a2620b765508f16289e0592e0563cfcb/src/languages/bash.js#L41-L52

As far as I can tell, the `scope: {2: 'comment'}` is somehow screwing up the `mode.contains.push({scope: 'doctag', ...})` detection in `modes.js`, but I don’t have a good enough grasp on highlight.js internals to really understand what's going on or how to fix it.

Test case which I expect to pass:
```diff
diff --git a/test/markup/bash/not-comments.txt b/test/markup/bash/not-comments.txt
--- a/test/markup/bash/not-comments.txt
+++ b/test/markup/bash/not-comments.txt
@@ -1,3 +1,5 @@
 echo asdf#qwert yuiop
 
 echo asdf #qwert yuiop
+
+# TODO: this *is* a comment
diff --git a/test/markup/bash/not-comments.expect.txt b/test/markup/bash/not-comments.expect.txt
index ab128642..0f57ee8b 100644
--- a/test/markup/bash/not-comments.expect.txt
+++ b/test/markup/bash/not-comments.expect.txt
@@ -1,3 +1,5 @@
 <span class="hljs-built_in">echo</span> asdf#qwert yuiop
 
 <span class="hljs-built_in">echo</span> asdf <span class="hljs-comment">#qwert yuiop</span>
+
+<span class="hljs-comment"># <span class="hljs-docstring">TODO:</span> this *is* a comment</span>
```

wolfgang42:
### Changes
Previously, because whitespace needed to be *detected* as part of the comment, but not *highlighted* that way, the match was split and only the second part of the match was assigned a scope. However, for reasons I couldn’t figure out, it seems that the presence of multiple `scope` values breaks the `contains` highlighting logic for a mode. (Or something to that effect.)

Rather than resolving whatever’s causing that problem, this PR changes the regex to use a [lookbehind assertion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion), thus requiring the whitespace to be there without capturing it as part of the match.

I *think* this is a **breaking change:** #3890 says that negative lookbehinds aren’t allowed in Highlight.js until v12 because of Safari support; it seems like the same is true of *positive* lookbehind, in which case merging this requires a major version bump.

Closes #4234.

### Checklist

(Full disclosure: this patch was prepared with the assistance of an LLM. However I have prepared this PR description myself, and am confident that the changes are correct.)

### DIFF
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -38,18 +38,7 @@ export default function(hljs) {
     end: /\)/,
     contains: [ hljs.BACKSLASH_ESCAPE ]
   };
-  const COMMENT = hljs.inherit(
-    hljs.COMMENT(),
-    {
-      match: [
-        /(^|\s)/,
-        /#.*$/
-      ],
-      scope: {
-        2: 'comment'
-      }
-    }
-  );
+  const COMMENT = hljs.COMMENT(/(?<=^|\s)#/, /$/);
   const HERE_DOC = {
     begin: /<<-?\s*(?=\w+)/,
     starts: { contains: [
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ New Grammars:
 
 Core Grammars:
 
+- fix(bash) resolve regression in comment doctag highlighting [wolfgang42][]
 - enh(csp) add missing directives / keywords from MDN (7 more) [Max Liashuk][]
 - enh(ada) add new `parallel` keyword, allow `[]` for Ada 2022 [Max Reznik][]
 
--- a/test/markup/bash/not-comments.expect.txt
+++ b/test/markup/bash/not-comments.expect.txt
@@ -1,3 +1,5 @@
 <span class="hljs-built_in">echo</span> asdf#qwert yuiop
 
 <span class="hljs-built_in">echo</span> asdf <span class="hljs-comment">#qwert yuiop</span>
+
+<span class="hljs-comment"># <span class="hljs-doctag">TODO:</span> this *is* a comment</span>
--- a/test/markup/bash/not-comments.txt
+++ b/test/markup/bash/not-comments.txt
@@ -1,3 +1,5 @@
 echo asdf#qwert yuiop
 
 echo asdf #qwert yuiop
+
+# TODO: this *is* a comment


Yang062 (approved):

src/languages/bash.js
```diff
@@ -38,18 +38,7 @@ export default function(hljs) {
...
-      match: [
-        /(^|\s)/,
-        /#.*$/
-      ],
-      scope: {
-        2: 'comment'
-      }
-    }
-  );
+  const COMMENT = hljs.COMMENT(/(?<=^|\s)#/, /$/);
```
joshgoebel, line 41: We can't use look-behind until v12 because it's a breaking change in older Safari.

How did we do this previously?  I also think we do not want any spacing in front of the comment to be scoped as part of the comment.

wolfgang42: > We can't use look-behind until v12 because it's a breaking change in older Safari.

Yeah, I thought as much. This is not a particularly pressing bug for me so I’d be fine waiting for whenever that happens, unless we can find another way to implement this first.

> How did we do this previously?

Prior to d78749a59410387f7232beb296e4606c9f4b7d08 it was just using `hljs.HASH_COMMENT_MODE` directly, but that doesn’t handle the whitespace requirement.

> I also think we do not want any spacing in front of the comment to be scoped as part of the comment.

Agreed; if it was OK to scope it as part of the comment we wouldn’t need the lookbehind, the regex could just be `/(^|\s)#/`. There are also some existing test cases which cover this.

joshgoebel: What was breaking if we just matched it as `/#/`?

wolfgang42: See https://github.com/highlightjs/highlight.js/pull/3918 and associated tests:

https://github.com/highlightjs/highlight.js/blob/d21e285f0df83699d460798f28c489e00a7c7ae9/test/markup/bash/not-comments.expect.txt#L1-L3

For example, `echo foo#bar` will print `foo#bar`; the `#bar` part is not a comment.

Presumably the author of that change had a specific scenario in mind, but I went looking for some real-world examples in the [Oil Shell test corpus](https://github.com/oilshell/wild-corpus), and instead found [another example](https://github.com/oilshell/wild-corpus/blob/a6fe3886095ae03afd528aefe296a169738984bc/git/git-rebase.sh#L535) of something that it has *broken* the highlighting for:

```bash
		branch_name=HEAD ;# detached
```
...which ought to be highlighted as a comment (since it's not inside a command), even though it doesn't have whitespace before it.

So, while that change was well-meaning, on closer inspection I think *very* careful reading of [POSIX §2.3](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_03) would be needed to figure out what the *actual* requirements are here; at a glance I see at least `|;<>&`; sometimes `()` can also be immediately followed by a comment, but (I think) only when indicating a standalone subshell; I haven’t looked at it closely but I think in the general case you may need a full-on stateful parser to determine exactly where in the grammar you are and what should be done with a `#` character in that state.

Given that, this is maybe a question of determining which kinds of brokenness are most acceptable? *I* certainly don't figure like wrangling the grammar to make sure that every edge case is covered correctly...

wolfgang42: (Also cc'ing @iFreilicht as author of that PR in case they have any more insight into their use case)

iFreilicht: @wolfgang42 Thank you! The important part for me is that a command like `nix run nixpkgs#btop` should be correctly highlighted, i.e. `#btop` is not a comment in this case, and `nixpkgs#btop` is a single token. POSIX-compatible shells require a comment to be at the beginning of a line or be preceeded by whitespace.

PR #4240, opened 2025-04-11
fix(javascript) correctly highlight 'for await' again

### LINKED ISSUE #4235: (javascript) regression: “for await“ highlighted as function instead of keyword
**Describe the issue**
In the statement `for await (foo) {}`, the keyword `await` is being incorrectly highlighted as a function since 063876f86c53713a665de0b17b8d795ae5fce0a1 (#3940).

JavaScript

highlight

**Sample Code to Reproduce**
```js
for await (foo) {}
```

**Expected behavior**
`for await` are both highlighted as keywords.

**Additional context**

I upgraded highlight.js and this stopped working. I bisected the problem to 063876f86c53713a665de0b17b8d795ae5fce0a1 (#3940), but I don’t really understand the internals of highlight.js well enough to figure out why that code is being triggered in this position or what to do about it.

wolfgang42:
Fixes highlighting for `for await (foo) {}`, which had regressed.

Closes #4235.

### Changes
The `FUNCTION_CALL` highlighting function has a list of excluded keywords; adding `await` there fixed the highlighting.

However, I’m not sure this is the correct fix. I notice there is another special case as well:

https://github.com/highlightjs/highlight.js/blob/85b20421a2620b765508f16289e0592e0563cfcb/src/languages/javascript.js#L557-L561

This implies to me that perhaps something in this grammar is in the wrong order, or needs its relevance adjusted, but that seems like a much bigger change and I’m having a hard time figuring out what it would need to look like. 

Also, it is technically possible (if very confusing) to have a function called `await`, and this would be highlighted incorrectly:
```js
async function await() {}
await await()
```

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ New Grammars:
 
 Core Grammars:
 
+- fix(javascript) correctly highlight 'for await' again [wolfgang42][]
 - enh(csp) add missing directives / keywords from MDN (7 more) [Max Liashuk][]
 - enh(ada) add new `parallel` keyword, allow `[]` for Ada 2022 [Max Reznik][]
 
--- a/src/languages/javascript.js
+++ b/src/languages/javascript.js
@@ -387,7 +387,8 @@ export default function(hljs) {
       noneOf([
         ...ECMAScript.BUILT_IN_GLOBALS,
         "super",
-        "import"
+        "import",
+        "await",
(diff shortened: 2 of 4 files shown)


Yang062 (approved):

joshgoebel:
>  to have a function called await

That's a terrible idea so I don't mind if that edge case isn't covered.

joshgoebel (approved):

PR #4241, opened 2025-04-18
fix(ex) #4238 and add test

### LINKED ISSUE #4238: (Elixir) Apostrophe code point character literal syntax (`?'`) comments out the rest of the code
**Describe the issue**
If you use the legitimate sequence `?'` to represent the code point of the apostrophe character in Elixir code, highlight.js treats everything after the `'` as a comment:

Image

`elixir`

`highlight`

**Sample Code to Reproduce**

```elixir
defmodule Acronym do
  defguardp apostrophe?(c) when c == ?'
  defguardp upper?(c) when (c >= ?A and c <= ?Z)
end
```

...as seen at https://jsfiddle.net/5gkn4hyw/

**Expected behavior**
The code after the `?'` sequence should continue to highlight normally.  (Above, GitHub gets it right.)

codingthat:
Adds support for the `?'` char literal and the missing `defguardp` keyword.

Resolves #4238 — see more detail there.

### Changes

- Adds a simple regex just for one char literal, `?'`, that otherwise breaks formatting significantly by turning on comment mode.
- Adds support for the `defguardp` keyword, which happened to be missing.
- Adds a line that covers both of these to the detector sample for Elixir.

### Checklist
- [x] Added markup tests


test/markup/elixir/char-literal.txt
```diff
@@ -0,0 +1,2 @@
+defguardp apostrophe?(c) when c == ?'
+defguardp upper?(c) when (c >= ?A and c <= ?Z)
```
joshgoebel, line 2: What are `?A` and `?Z` then?  Are they not also char literals?

codingthat: They are, but char literals can be [any character](https://hexdocs.pm/elixir/1.18.3/binaries-strings-and-charlists.html#unicode-and-code-points).  To keep the change small and the matching efficient, I just targeted the most problematic case (at least for now).  If you want to leave them absolutely consistent with other char literals, disabling highlighting `?'` (I guess with `className: null`?) could also work.

Issue #4244, opened 2025-04-24
(shell) Period not match as part of promt


ianw:
A period in the prompt name is not matched by the shell highlighter, e.g.

```console
aprompt $ ls -l  # matches
a-prompt $ ls -l # matches
a.prompt $ ls -l # no match
```

It matches other special characters such as `[]-` so I think `.` is a reasonable addition.  I noticed this when writing some documentation that had a branch-name in the prompt, so `foo-1` was working but `foo-1.0` wasn't.

joshgoebel:
Merged in #4245

PR #4245, opened 2025-04-24
enh(shell) allow period in prompt match (#4244)

ianw:
Allow a "." to match in the prompt section of shell/console examples.

### Checklist
- [x] Modified existing markup test to also cover this


test/markup/shell/prompt-with-slash.txt
```diff
@@ -1,6 +1,6 @@
-pao-sw-a04-40(config)# interface range ethernet 1/1/1-1/1/9
-pao-sw-a04-42(conf-range-eth1/1/1-1/1/9)# switchport trunk allowed vlan 1102
-pao-sw-a04-42(conf-range-eth1/1/1-1/1/9)# show configuration
+pao.sw.a04-40(config)# interface range ethernet 1/1/1-1/1/9
+pao.sw.a04-42(conf-range-eth1/1/1-1/1/9)# switchport trunk allowed vlan 1102
+pao.sw.a04-42(conf-range-eth1/1/1-1/1/9)# show configuration
```
joshgoebel, line 3: Why would we remove the old tests?  The goal here is to support both, yes?  If so lets put the old tests back so that we're testing both formats.  

Otherwise seems reasonable. Thanks!

ianw: Since this replaces a “-“ with a “.” but retains other “-“ it feels to me this covers both cases OK.  If you feel strongly I can make another separate test.

Yang062 (approved):

test/markup/shell/prompt-with-slash.txt
joshgoebel: Yes, lets keep the old tests and just add the 3 new lines of tests right after.  That'd be perfect!

ianw: Going back and re-opening the file I realised it's called `prompt-with-slash.txt` and there's other tests named like that -- so I agree that overloading it to test multiple things isn't keeping with what's there.  So switched to a individual test as first suggested.

Yang062 (approved):

Yang062 (approved):

joshgoebel (approved):

PR #4248, opened 2025-04-28
Fix bug in Leaf keyword highlighting

fpseverino:
### Changes
Fix a bug in the [Leaf](https://docs.vapor.codes/leaf/overview/) syntax where keywords are highlighted when at the start of a variable name (see the third code block [here](https://docs.vapor.codes/leaf/overview/#template-syntax))

### Checklist


joshgoebel:
Thanks!  Please also add a changelog entry as well as markup test case.

fpseverino:
> Please also add a changelog entry as well as markup test case.

@joshgoebel done! I hope I did it correctly

joshgoebel:
Except we need a test that would have failed before - so that we prevent this regression in the future... so we need an example of what would have failed before.

joshgoebel:
Ah, I see. :)

joshgoebel (approved):

Yang062 (approved):

Yang062 (approved):

fpseverino:
Hi @joshgoebel, I just wanted to kindly follow up on this PR, just making sure it hasn’t been lost in the shuffle.
Let me know if there’s anything you'd like me to adjust

Yang062 (approved):

Yang062 (approved):

fpseverino:
Hi, can we please get this merged?

fpseverino:
Hi, just another friendly ping on this PR

test/markup/leaf/default.txt
```diff
@@ -7,7 +7,7 @@
 
 #if(count(visit) == 1):
     Hello new user!
-#elseif(title == "Welcome back!"):
```
joshgoebel, line 10: What is this change showing me?

src/languages/leaf.js
```diff
@@ -26,7 +26,7 @@ export default function(hljs) {
       },
       {
         scope: 'keyword',
-        match: LITERALS.join("|"),
+        match: LITERALS.map(lit => `\\b${lit}\\b`).join("|"),
```
joshgoebel: Could we do all this with a single set of `\b`?  Something like:

```suggestion
        match: `\\b${regex.either(LITERALS...)}\\b`)
```

test/markup/leaf/default.txt
fpseverino: A word like `index` inside a Leaf tag is highlighted incorrectly right now without the fix I'm proposing (see the third code block [here](https://docs.vapor.codes/leaf/overview/#template-syntax)), I changed it so we can keep it tested here

src/languages/leaf.js
fpseverino: I'm committing this suggestion. I can't test it right now on my machine, so I'll let CI do it

fpseverino:
Hi @joshgoebel, thanks for the new review! I've addressed your comments, let's see if CI passes and let me know if you need anything else. Pretty excited to finally fix this bug!

joshgoebel (approved):

joshgoebel:
I wrote that off the top of my head, looks like it needs work... does the `...` come first?

fpseverino:
@joshgoebel I think we got it now

joshgoebel:
Tests are broken now.

fpseverino:
> Tests are broken now.

@joshgoebel Sorry about that, I've tested it locally, https://github.com/highlightjs/highlight.js/pull/4248/commits/b13f83bdb6d7ced015882a8d989987ad48e92691 fixes it

PR #4250, opened 2025-04-30
fix(docs): fixed typo and incorrect git command

witch-factory:
### Changes

- Fix incorrect quote syntax in git commit example
- Correct "simply" to "simplify" in the Future section

### Checklist
- [x] Added markup tests, or they don't apply here because it's only a document fix


Yang062 (approved):

joshgoebel:
How is the quoting incorrect?

joshgoebel:
Oh, it was so minute I missed it.  Thanks!

Issue #4253, opened 2025-05-05
(CSS) `unicode-range` keyword not supported


kshetline:
**Describe the issue**

CSS `unicode-range` is not recognized, and subsequent range values are badly parsed.

**Which language seems to have the issue?**

CSS

**Are you using `highlight` or `highlightAuto`?**

highlight

**Sample Code to Reproduce**

```css
@font-face {
  font-family: DroidSans;
  src: url(DroidSans.ttf);
  unicode-range: U+000-5FF, U+1e00-1fff, U+2000-2300;
}
```

Gets highlighted this way:

![Image](https://github.com/user-attachments/assets/a6989eaa-8139-473f-bb0f-5575d338b59f)

**Expected behavior**

More like the way GitHub shows the code above.

divyam-r25:
Hii @kshetline, I would like to work on this issue,can you please assign me this issue

joshgoebel:
@divyam-r25  You're going to get banned if I continue to see the **exact** same text offering to work on like **every** single issue.  You look like a bot or AI or something.   

**No one needs to ask here to work on approved issues.**. If you want to help out, just do so and submit a PR.

joshgoebel:
Closed by #4454

PR #4254, opened 2025-05-06
Fix for CSS unicode-range support, issue #4253.

kshetline:
Added support for CSS unicode-range keyword and associate range arguments.
Resolves issue #4253

### Checklist

- [x] Added markup tests

### DIFF
--- a/src/languages/lib/css-shared.js
+++ b/src/languages/lib/css-shared.js
@@ -7,7 +7,11 @@ export const MODES = (hljs) => {
     BLOCK_COMMENT: hljs.C_BLOCK_COMMENT_MODE,
     HEXCOLOR: {
       scope: 'number',
-      begin: /#(([0-9a-fA-F]{3,4})|(([0-9a-fA-F]{2}){3,4}))\b/
+      begin: /#(([0-9A-F]{3,4})|(([0-9A-F]{2}){3,4}))\b/i
+    },
+    UNICODE_RANGE: {
+      scope: 'number',
+      begin: /\bU\+[0-9A-F][0-9A-F?]{0,4}(-[0-9A-F][0-9A-F]{0,4})?/i
     },
     FUNCTION_DISPATCH: {
       className: "built_in",
@@ -777,6 +781,7 @@ export const ATTRIBUTES = [
   'transition-timing-function',
   'translate',
   'unicode-bidi',
+  'unicode-range',
   'user-modify',
   'user-select',
   'vector-effect',
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Core Grammars:
 - fix(ex) adds support for `?'` char literal and missing `defguardp` keyword [Kevin Bloch][]
 - fix(diff) fix unified diff hunk header regex to allow unpaired numbers [Chris Wilson][]
 - enh(php) support single line and hash comments in attributes, constructor and functions [Antoine Musso][]
+- fix(css) `unicode-range` parsing, issue #4253 [Kerry Shetline][]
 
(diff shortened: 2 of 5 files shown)


kshetline:
Resolves issue #4253 

(Adding "Resolves" again because the link didn't function completely the first time.)

test/markup/scss/css_consistency.txt
```diff
@@ -54,7 +54,7 @@ a[href*="example"] {}
   font-variant: no-common-ligatures proportional-nums;
   font-feature-settings: "liga" 0;
   font-variation-settings: "xhgt" 0.7;
-  /* unicode-range: U+0025-00FF, U+4??; */
```
Yang062, line 57: []()haha wag

Yang062 (approved):

src/languages/lib/css-shared.js
```diff
@@ -7,7 +7,11 @@ export const MODES = (hljs) => {
     BLOCK_COMMENT: hljs.C_BLOCK_COMMENT_MODE,
     HEXCOLOR: {
       scope: 'number',
-      begin: /#(([0-9a-fA-F]{3,4})|(([0-9a-fA-F]{2}){3,4}))\b/
+      begin: /#(([0-9A-F]{3,4})|(([0-9A-F]{2}){3,4}))\b/i
+    },
+    UNICODE_RANGE: {
+      scope: 'number',
+      begin: /\bU\+[0-9A-F][0-9A-F?]{0,4}(-[0-9A-F][0-9A-F]{0,4})?/i
```
joshgoebel, line 14: We can't use `/i`... the entire grammar is either case insensitive or not (the engine handles this) - the tags on any individual regex are irrelevent... right now 3 or our CSS grammars are marked case in-sensitive while the 4th is not.  So we can either:

- Revert the first, and add case variants to the second
- You could make the argument that Stylus should also be case-insensitive, then we'd get this for free... but AI says that Stylus isn't 100% case insensative... :-/

kshetline: Aren't all of these languages case-insensitive on the particular issue of parsing hexadecimal numbers? I'm not familiar with Stylus, but since the existing regex that I modified contained `0-9a-fA-F` that made it just as case-insensitive as adding the `i` flag does.

joshgoebel: > but the functional end result is the same

It's not.

*The `/i` flag you've added will have no effect*.  In Stylus right now your `/[A-F]/i` would be broken.  It would NOT match lowercase hex digits.  If a grammar is marked insensitive EVERY regex will have the `i` flag (no need to specify it per regex), but if not NONE of the regex will have the `i` flag (even if specified, it'll be removed).  This is because of how the parsing engine combines rules into one giant regex.

> make my own regex... 

Yes, because Stylus is still case-sensitive you'll need to add `a-f` to your regex to make it insensitive.

kshetline: > ...because of how the parsing engine combines rules into one giant regex.

Whoa. That's unexpected. As a suggestion, perhaps a warning in the contributor documentation about that?

I've now fixed this PR to take the case handling/regex combining issue into account.

Yang062 (approved):

src/languages/lib/css-shared.js
kshetline: I can do that. I certainly don't understand this code base well enough yet to tell anyone else much about how to work with it in general, but this one detail I can handle. 🙂

As for the rest of this PR, is there anything left I should do, or is it good to go?

Yang062 (approved):

joshgoebel (approved):

PR #4256, opened 2025-05-06
Add support for JSON5 as an alias of JavaScript.

### LINKED ISSUE #4255: [Request] Add JSON5 as an alias for JavaScript
I'm submitting this issue as a small enhancement request, hoping it doesn't qualify as a request for a whole new language.

Since valid JSON5 is also valid JavaScript at the same time, the current JavaScript parser does a good job with highlighting it – better than JSON/JSONC.

I'm using highlight.js to enhance Markdown syntax highlighting within code fences, where users would be expected to annotate their code fences like this:

Image

...and not tag them as `javascript` instead, just to get the content parsed correctly.

A simple change to the alias list in `javascript.js`, which I'm prepared to submit myself.

n/a

kshetline:
Simple addition of JSON5 as an alias for JavaScript.

Resolves #4255

### Changes

In addition to simply adding 'json5' to the javascript.js alias list, I noticed a very suspicious looking regex just a few lines below the alias list, so I've included a fix for that issue as well. I can make this fix and all tests still pass, so I think including this change is a good idea. The suspicious regex, at line 460, is this:

```javascript
    illegal: /#(?![$_A-z])/,
```

It seems highly doubtful this was meant to be the range from uppercase A to lowercase z, which isn't just letters, but the punctuation " [ \ ] _ ` " as well.

### Checklist

- [x] Added markup tests
- [x] Updated aliases listed for JavaScript at `SUPPORTED_LANGUAGES.md`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Core Grammars:
 - fix(ex) adds support for `?'` char literal and missing `defguardp` keyword [Kevin Bloch][]
 - fix(diff) fix unified diff hunk header regex to allow unpaired numbers [Chris Wilson][]
 - enh(php) support single line and hash comments in attributes, constructor and functions [Antoine Musso][]
+- enh(javascript) add json5 as alias for javascript [Kerry Shetline][]
 
 CONTRIBUTORS
 
@@ -23,6 +24,7 @@ CONTRIBUTORS
 [Chris Wilson]: https://github.com/sushicw
 [Antoine Musso]: https://github.com/hashar
 [Chester Moses]: https://github.com/Chester-Moses-HCL
+[Kerry Shetline]: https://github.com/kshetline
 
 
 ## Version 11.11.1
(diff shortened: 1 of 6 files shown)


Yang062 (approved):

joshgoebel:
Why would json5 not be an alias to json?

kshetline:
Because the JSON parser, as implemented, doesn't handle unquoted identifiers or single quoted strings. It only handles the comments extension of JSON (JSONC).

joshgoebel:
I'm not sure what to do with this.  Here are current stats:

```
```

kshetline:
So you prefer expanding the current json.js to handle JSON5 features over a separate json5.js?

By the way, do you have any objections to making ES2018 the compile target instead of ES2015 so that more regex features (like lookahead and lookbehind) are available?

joshgoebel:
> So you prefer expanding the current json.js to handle JSON5 features over a separate json5.js?

Yep.  Should be pretty simple I think.  It's a superset IIRC, yes?

> By the way, do you have any objections to making ES2018 the compile target instead of ES2015 so that more regex features (like lookahead and lookbehind) are available?

We can't do this until version 12, as this is technically a breaking change.

PR #4259, opened 2025-05-12
Add support for JSON5 as extension of json.js.

### LINKED ISSUE #4260: [Request] Add support for JSON5 as extension of current JSON support
I'm submitting this issue as a small enhancement request, hoping it doesn't qualify as a request for a whole new language.

I'm using highlight.js to enhance Markdown syntax highlighting within code fences, where users would be expected to annotate their code fences like this:

Image

...and not tag them as `javascript` instead, just to get the content parsed correctly.

Updates to json.js to cover JSON5 syntax, for which I have already submitted a PR.

n/a

kshetline:
Addition of JSON5 handling to json.js.

Resolves #4260 

### Changes

In addition to adding 'json5' to json.js, I noticed a suspicious looking regex in javascript.js. I can make this fix and all tests still pass, so I think including this change is a good idea. The suspicious regex, at line 460, is this:

```javascript
    illegal: /#(?![$_A-z])/,
```

It seems highly doubtful this was meant to be the range from uppercase A to lowercase z, which isn't just letters, but the punctuation " [ \ ] _ ` " as well.

### Checklist

- [x] Added markup tests
- [x] Updated aliases listed for JSON at `SUPPORTED_LANGUAGES.md`

### DIFF
--- a/src/languages/json.js
+++ b/src/languages/json.js
@@ -2,14 +2,15 @@
 Language: JSON
 Description: JSON (JavaScript Object Notation) is a lightweight data-interchange format.
 Author: Ivan Sagalaev <maniac@softwaremaniacs.org>
-Website: http://www.json.org
+Author: Kerry Shetline <kerry@shetline.com> (added JSON5 support)
+Websites: http://www.json.org, https://www.json5.org
 Category: common, protocols, web
 */
 
 export default function(hljs) {
   const ATTRIBUTE = {
     className: 'attr',
-    begin: /"(\\.|[^\\"\r\n])*"(?=\s*:)/,
+    begin: /(("(\\.|[^\\"\r\n])*")|('(\\.|[^\\'\r\n])*'))(?=\s*:)/,
     relevance: 1.01
   };
   const PUNCTUATION = {
@@ -34,16 +35,17 @@ export default function(hljs) {
 
   return {
     name: 'JSON',
-    aliases: ['jsonc'],
+    aliases: ['jsonc', 'json5'],
     keywords:{
       literal: LITERALS,
     },
     contains: [
       ATTRIBUTE,
       PUNCTUATION,
+      hljs.APOS_STRING_MODE,
       hljs.QUOTE_STRING_MODE,
       LITERALS_MODE,
-      hljs.C_NUMBER_MODE,
+      hljs.EXTENDED_NUMBER_MODE,
       hljs.C_LINE_COMMENT_MODE,
       hljs.C_BLOCK_COMMENT_MODE
     ],
--- a/src/languages/json.js
+++ b/src/languages/json.js
@@ -1,11 +1,11 @@
 /*
 Language: JSON
 Description: JSON (JavaScript Object Notation) is a lightweight data-interchange format.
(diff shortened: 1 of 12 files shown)


src/lib/modes.js
```diff
@@ -10,6 +10,7 @@ export const IDENT_RE = '[a-zA-Z]\\w*';
 export const UNDERSCORE_IDENT_RE = '[a-zA-Z_]\\w*';
 export const NUMBER_RE = '\\b\\d+(\\.\\d+)?';
 export const C_NUMBER_RE = '(-?)(\\b0[xX][a-fA-F0-9]+|(\\b\\d+(\\.\\d*)?|\\.\\d+)([eE][-+]?\\d+)?)'; // 0x..., 0..., decimal, float
+export const EXTENDED_NUMBER_RE = '([-+]?)(\\b0[xX][a-fA-F0-9]+|(\\b\\d+(\\.\\d*)?|\\.\\d+)([eE][-+]?\\d+)?)|NaN|[-+]?Infinity'; // 0x..., 0..., decimal, float
```
joshgoebel, line 13: Is this all valid ECMAScript?  If so perhaps this belongs in `lib/ecmascript`, but I don't think we want to add another global here since this is a very specific use case.

src/languages/json.js
```diff
@@ -2,14 +2,15 @@
 Language: JSON
 Description: JSON (JavaScript Object Notation) is a lightweight data-interchange format.
 Author: Ivan Sagalaev <maniac@softwaremaniacs.org>
-Website: http://www.json.org
+Author: Kerry Shetline <kerry@shetline.com> (added JSON5 support)
```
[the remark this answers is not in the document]
kshetline: Done.

types/index.d.ts
```diff
@@ -74,6 +74,7 @@ declare module 'highlight.js' {
         HASH_COMMENT_MODE: Mode
         NUMBER_MODE: Mode
         C_NUMBER_MODE: Mode
+        EXTENDED_NUMBER_MODE: Mode
```
joshgoebel, line 77: Dead code?

joshgoebel (changes requested):
Looks good, almost there!

types/index.d.ts
kshetline: Well, the code passes tests without that line, so I guess so. I was just following what seemed to be an established pattern, thinking everything like that should be declared there.

src/languages/lib/ecmascript.js
```diff
@@ -1,4 +1,12 @@
 export const IDENT_RE = '[A-Za-z$_][0-9A-Za-z$_]*';
+export const EXTENDED_NUMBER_RE = '([-+]?)(\\b0[xX][a-fA-F0-9]+|(\\b\\d+(\\.\\d*)?|\\.\\d+)([eE][-+]?\\d+)?)|NaN|[-+]?Infinity'; // 0x..., 0..., decimal, float
+
+export const EXTENDED_NUMBER_MODE = {
+  scope: 'number',
+  begin: EXTENDED_NUMBER_RE,
```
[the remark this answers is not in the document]
kshetline: This change is done as well.

joshgoebel (approved):

PR #4267, opened 2025-05-26
Fix dead reference to non-existing template repository

### LINKED ISSUE #3038: Task: 3rd party grammar template repository
Copying in content from @tajmone at https://github.com/highlightjs/highlightjs-alan/issues/5#issuecomment-789643133

---

I think it would make sense if you created a repository template for custom HLJS syntaxes, which developers can use to create the codebase for their new languages. 

- [GitHub Docs » Creating a template repository]

The template would provide a boilerplate, with all the required files and directories structure (a sample language to use as a reference), repository settings (`.giattributes`, `.gitignore`, `.editroconfig`, etc.) and a sample `README.md` with instructions, which end users will only need to adapt to their new language (instructions for the developer, but also those for end users, as for this Issue).

This would simplify creating new languages for both newbies and experts alike, for with a single click they'd have a robust starting point, and newbies will also find the needed instructions.

At least, this is what I do when I'm dealing with commonly shared repository settings and structures, to avoid useless and repetitive tasks.

Furthermore, having an officially sanctioned template would dispel any confusion regarding the proper repository structure to adopt, since not all independent syntax repos are abiding to the same structure right now.

[GitHub Docs » Creating a template repository]: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-template-repository

sebastiaanspeck:
Resolves #3038

### Changes

Added a link to an up-to-date and working template repository. Feel free to migrate my template repository to highlight.js organisation.

### Checklist
- [x] Added markup tests, or they don't apply here because it doesn’t fix/update any grammar
- [x] Updated the changelog at `CHANGES.md`, doesn't apply here because nothing really changed

### DIFF
--- a/extra/3RD_PARTY_QUICK_START.md
+++ b/extra/3RD_PARTY_QUICK_START.md
@@ -10,7 +10,7 @@ So you'd like to create and share your own language definition for Highlight.js.
 - [ ] Clone the main [highlight-js](https://github.com/highlightjs/highlight.js) repository from GitHub
 - [ ] Read our [Language Contributor Checklist](https://highlightjs.readthedocs.io/en/latest/language-contribution.html)
 - [ ] Review the [Language Definition Guide](https://highlightjs.readthedocs.io/en/latest/language-guide.html)
-- [ ] Start with our [repository template](https://github.com/highlightjs/highlightjs-language-template) to more easily follow the suggested layout. (this isn't ready yet!)
+- [ ] Start with our [repository template](https://github.com/sebastiaanspeck/highlightjs-language-template) to more easily follow the suggested layout. (this isn't ready yet!)
 
 ## Create your repository
 


extra/3RD_PARTY_QUICK_START.md
```diff
@@ -10,7 +10,7 @@ So you'd like to create and share your own language definition for Highlight.js.
 - [ ] Clone the main [highlight-js](https://github.com/highlightjs/highlight.js) repository from GitHub
 - [ ] Read our [Language Contributor Checklist](https://highlightjs.readthedocs.io/en/latest/language-contribution.html)
 - [ ] Review the [Language Definition Guide](https://highlightjs.readthedocs.io/en/latest/language-guide.html)
-- [ ] Start with our [repository template](https://github.com/highlightjs/highlightjs-language-template) to more easily follow the suggested layout. (this isn't ready yet!)
+- [ ] Start with our [repository template](https://github.com/sebastiaanspeck/highlightjs-language-template) to more easily follow the suggested layout. (this isn't ready yet!)
```
joshgoebel, line 13: Do you consider yours "ready"?  If so perhaps we change this to something more like:

```
- [ ] You might find Sebastian Speck's [repository template] helpful as a starting point.
```

Or some such. Thoughts?

Yang062 (approved):

Issue #4284, opened 2025-06-12
(Python) Methods don't get the proper highlight color


fibbo:
**Describe the issue**

I have the following Python code.

```python
sweeper = daq.sweep()
sweeper.set('device', 'dev3225')
sweeper.subscribe('/dev3225/demods/0/sample')
sweeper.execute()
```

Highlighting of strings and literals is okay. However, it highlights `set` as built-in and doesn't highlight the methods `subscribe` and `execute` at all.

**Which language seems to have the issue?**

As mentioned, the problem above is with Python, but there are similar issues with e.g., .NET, where method names are also consistently not highlighted.

**Are you using `highlight` or `highlightAuto`?**

I'm using highlightElement.
...

**Sample Code to Reproduce**
https://jsfiddle.net/Ldpfg657/

**Expected behavior**
I would expect the methods (like `subscribe`, `execute`) to have a distinct color, and especially a `set` that is not the built-in one, not using the built-in highlight.

**Additional context**

joshgoebel:
Some grammars have better highlighting rules than others.  This is another common case of that.  We generally don't leave tickets open for "make grammar xyz better" style issues or we'd end up with 200 open tickets - one for every grammar.  Instead, all of this is covered by #2500.  

We're totally open to highlighting these as either `property` or `title.function` (given the trailing `(`)... If you'd like to contribute, PRs are definitely welcome for Python or the other languages you mentioned in this department.

Closing this as a duplicate of #2500.

PR #4285, opened 2025-06-16
Add better support for Python functions and props

fibbo:
This is an attempt to improve the syntax highlighting for Python (#4284)

It should now correctly highlight function calls, chained function calls, and properties.

### Changes
Added three more regex rules to the Python rules.

### Checklist
- [x] Added markup tests ~~TBD~~
- [x] Updated the changelog at `CHANGES.md` ~~TBD~~

### DIFF
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -383,6 +383,39 @@ export default function(hljs) {
         relevance: 0
       },
       { match: /\bor\b/, scope: "keyword" },
+      // Method calls with parentheses
+      {
+        match: [
+          /\./,
+          IDENT_RE,
+          /(?=\s*\()/
+        ],
+        scope: {
+          2: "title.function.method"
+        }
+      },
+      // Chained method calls
+      {
+        match: [
+          /\./,
+          IDENT_RE,
+          /(?=\s*\.\s*\w)/
+        ],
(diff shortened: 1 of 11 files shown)


Yang062 (approved):

src/languages/python.js
```diff
@@ -383,6 +383,39 @@ export default function(hljs) {
...
+          IDENT_RE,
+          /(?=\s*\.\s*\w)/
+        ],
+        scope: {
+          2: "title.function.method"
+        }
+      },
+      {
+        match: [
+          /(?<!(?:import|from)\s+[\w.]*)/,  // Negative lookbehind for import context
```
joshgoebel, line 410: Do we have a test case for this?  I'd like to see an example or two of what exactly we're avoiding here.

fibbo: Imagine the following line:

```python
from base.derive import test
```

I wouldn't want `derive` to be highlighted as a property.

Yang062 (approved):

Yang062 (approved):

fibbo:
@joshgoebel, would it be possible to get this merged at some point?

joshgoebel (approved):

PR #4289, opened 2025-07-02
Update string color in Monokai

SmartManoj:


Yang062 (approved):

joshgoebel:
Any docs or pointers to the theme elsewhere to show this has always been wrong?

SmartManoj:
d/f between monokai and monokai-sublime?

joshgoebel:
Well the file is named monokai, not monokai sublime (which would perhaps need to be a diff theme)... so if someone can just show me the official monokai prefers this color I think we can change it.

Yang062 (approved):

joshgoebel:
> so if someone can just show me the official monokai prefers this color I think we can change it.

This never happened, closing.  If someone wants to show this is indeed a correct change, happy to re-review.

PR #4292, opened 2025-07-03
rust: add `safe` keyword

fw-immunant:
This is a "weak" keyword, only applicable in Rust 2024+ `unsafe extern` blocks: https://doc.rust-lang.org/reference/keywords.html#r-lex.keywords.weak.safe

### Changes
Adds the `safe` keyword within a mode for `unsafe extern` blocks.

### Checklist

### DIFF
--- a/src/languages/rust.js
+++ b/src/languages/rust.js
@@ -17,7 +17,7 @@ export default function(hljs) {
   const IDENT_RE = regex.concat(RAW_IDENTIFIER, hljs.IDENT_RE);
   // ============================================
   const FUNCTION_INVOKE = {
-    className: "title.function.invoke",
+    scope: "title.function.invoke",
     relevance: 0,
     begin: regex.concat(
       /\b/,
@@ -202,7 +202,7 @@ export default function(hljs) {
         illegal: null
       }),
       {
-        className: 'symbol',
+        scope: 'symbol',
         // negative lookahead to avoid matching `'`
         begin: /'[a-zA-Z_][a-zA-Z0-9_]*(?!')/
       },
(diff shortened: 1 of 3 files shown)


src/languages/rust.js
```diff
@@ -318,6 +318,12 @@ export default function(hljs) {
         className: "punctuation",
         begin: '->'
       },
+      {
+        className: 'unsafe_extern',
```
joshgoebel, line 322:
```suggestion
        scope: 'unsafe_extern',
```

joshgoebel: Wait, this isn't one of our named scopes... this value must be one of the scopes from our documentation.  We'll also need a markup test to show this works as intended.

Yang062 (approved):

Yang062 (approved):

src/languages/rust.js
```diff
@@ -230,24 +230,35 @@ export default function(hljs) {
         ],
         relevance: 0
       },
+      {
+        begin: [
+          /\bsafe/,
```
joshgoebel, line 238: Is there a fear of false positives on *safe extern?  usually we don't `\b` defensively.

src/languages/rust.js
```diff
@@ -230,24 +230,35 @@ export default function(hljs) {
...
         ],
-        className: {
+        scope: {
           1: "keyword",
           3: "title.function"
         }
       },
       {
-        className: 'meta',
+        scope: 'meta',
```
joshgoebel, line 259: Thanks for this cleanup.

src/languages/rust.js
joshgoebel: Ah this is to ward of grabbing `unsafe`?

fw-immunant: Yes, precisely.

joshgoebel (approved):

PR #4293, opened 2025-07-07
fix(bash): arguments highlight

### LINKED ISSUE #4288: (Bash) Highlighted colors doesn't look correct
**Describe the issue**
Issues:
- Only highlight part of command paramters
- Inconsistent color between lines

Bash

`highlight`

...

**Sample Code to Reproduce**

1st example:

```bash
epi deployment start \
  --project my-project \
  --deployment-package my-project.cms.app.1.0.0.nupkg \
  --env Integration \
  --direct-deploy true \
  --use-maintainance-page true \
  --wait true \
  --show-progress true
```

2nd example:

```bash
epi deployment start \
  --project my-project \
  --deployment-package my-project.cms.app.1.0.0.nupkg \
  --env Integration \
  --direct-deploy true \
  --use-maintainance-page true \
  --wait true \
  --show-progress true
```

For partial highlight, I see one part has `hljs-built_in` class, while the other doesn't has any class.

**Expected behavior**
- Should highlight the whole command paramter name
- Command parameter names should have same colors across lines

**Additional context**
Live example:
<https://cli.eshn.dev/epi-deployment-start-command.html#samples>

Image

Reproduce on Highlight.js demo:

Image

tuyen-at-work:
Highlight arguments in Bash/Shell

Resolve #4288

### Changes

- fix(bash) arguments highlight for Bash/Shell, issue

### Checklist


Yang062 (approved):

src/languages/bash.js
```diff
@@ -117,6 +117,16 @@ export default function(hljs) {
     relevance: 0
   };
 
+  const BASH_ARGUMENTS = {
+    variants: [
+      {
+        className: "attr",
+        begin: /[\t\u0020](-){1,2}[\w\d-]+/,
```
joshgoebel: **Blocker: leading whitespace is included in the `attr` span**

`begin: /[\t\u0020](-){1,2}[\w\d-]+/` puts the space/tab inside `hljs-attr`, so markup becomes e.g. `tr<span class="hljs-attr"> -d</span>` / `tail<span class="hljs-attr"> -10</span>`. Themes then color the separator, which looks wrong.

v11 still can’t use lookbehind, but multi-match can keep the separator unscored and only scope the flag:

```js
const BASH_ARGUMENTS = {
  match: [/[\t ]+/, /-{1,2}[\w-]+/],
  scope: { 2: "attr" },
  relevance: 0
};
```

Please update the mode and regenerate the affected markup expects so the space is outside the span.

joshgoebel: **Blocker: flags at the start of a continued line are missed**

Only `\t` / space are allowed before `-`/`--`. Common layout after a line continuation fails:

```bash
epi deployment start \
--project my-project \
--env Integration
```

Here the flag sits after a newline, not a space on the same line. Prefer a whitespace class on the separator group (e.g. `\s+` in a multi-match group 1 with no scope) so BOL options after `\` still match, while only `-{1,2}[\w-]+` gets `attr`.

(See also the suggestion on #4288 about using `[\s]`.)

src/languages/bash.js
```diff
@@ -117,6 +117,16 @@ export default function(hljs) {
     relevance: 0
   };
 
+  const BASH_ARGUMENTS = {
+    variants: [
+      {
+        className: "attr",
```
joshgoebel: **Blocker: prefer current grammar API / simplify the mode**

- Prefer `scope` over deprecated `className` for new grammar code.
- A single-entry `variants: [...]` wrapper isn’t needed — use one mode.
- `(-){1,2}` can be `-{1,2}`; `\d` is already covered by `\w`, so `[\w\d-]+` → `[\w-]+`.

Suggested shape (combined with the other two comments):

```js
const BASH_ARGUMENTS = {
  match: [/\s+/, /-{1,2}[\w-]+/],
  scope: { 2: "attr" },
  relevance: 0
};
```

joshgoebel (changes requested):
Thanks for tackling #4288 — the approach (match flags with a leading separator so the keyword engine never sees `--wait` / `--env`) is right, and the subcommand regression test is useful.

Requesting a few changes before merge: leading whitespace should not be inside the `attr` span, line-start flags after `\` continuations should match, and the mode should follow current grammar style (`scope`, no single-variant wrapper).

joshgoebel (changes requested):
See comments.

joshgoebel (approved):

PR #4299, opened 2025-07-28
add support for FreeDesktop config file

gg582:
### Changes
- Add XDG Desktop Shortcut's syntax highlighting


gg582:
You can see the example at the end of this post.
[Highlight JS Custom Highlighter](https://chi-blog-seven.vercel.app/posts/highlightjs-custom-highlighter-develop-a-highlighter-for-xdg-desktop-shortcut)

gg582:
https://chi-blog-seven.vercel.app/posts/systemd-making-a-blog-server-as-a-daemon


Here's an example for *.service support.

gg582:
This should support similar files as a whole.
So, I will add it as a Systemd syntax.

gg582:
You can see some examples from here.
https://chi-blog-seven.vercel.app/posts/systemdxdg-config-files-are-so-similar-highlight-them-as-a-systemdxdg-config-at-custom-highlightjs

gg582:
Latest patch: Freedesktop formats do not use `;` as a delimiter.

gg582:
So the basic principal of the highlighting is to support FreeDesktop related config files

gg582:
Currently Supported Formats
- NetworkManager Configuration
- Systemd Units
  - Service Units
  - Mount Units
  - Socket Units
  - Timer Units
- XDG Desktop Shortcuts

Yang062 (approved):

Yang062 (approved):

gg582:
Hi @Yang062, since this PR was approved a few months ago, I was wondering if there are any remaining steps before it can be merged? I'd love to see XDG support in the next release.

gg582:
Hi, I wonder if there are any plans to merge this.

Yang062 (approved):

joshgoebel:
@allejo  You think freedesktop config is important/popular enough to warrant inclusion in core (against the general policy)?  It crossed my mind that it might be.

allejo:
I'd be in favor of adding freedesktop/XDG to core.

After some brief research ([Wikipedia](https://en.wikipedia.org/wiki/Freedesktop.org), [Medium.com article](https://thamizhelango.medium.com/freedesktop-org-building-universal-standards-for-linux-desktop-interoperability-44d1d9f1244e)), it looks like a lot of the Linux desktops are getting behind this effort of having a shared syntax for their configuration files. I feel like any Linux hacking community (blog, forums, etc.) would make use of this.

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
+/*
+Language: FreeDesktop Configs
+Description: FreeDesktop Config Specification file format
+Contributors: Lee Yunjin <gzblues61@daum.net>
```
joshgoebel: Please remove this, we now give credit via github.com/project/contributors or whatever the URL is rather than having people comment in individual fiels.

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
...
+Category: common, config
```
joshgoebel: `config` I think, not common.

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
...
+Website: https://www.freedesktop.org/
+*/
+export default function(hljs) {
+  const FIELD_CODES = {
+    className: 'variable',
```
joshgoebel: Please use `scope` everything, className is being deprecated.

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
...
+    match: /%[a-zA-Z]/,
+    relevance: 1
```
joshgoebel: You can drop all `relevance` keys. This is being deprecated in v12.

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
...
+    className: 'section',
+    begin: /^\[(Desktop Entry|Unit|Service|Install|Socket|Mount|Automount|Swap|Path|Timer|Slice|Scope|Manager|connection|ipv4|wifi|wifi-security|ipv6|802-3-ethernet|802-11-wireless|802-11-wireless-security|vpn|Journal|Bridge|Desktop Action\s+[A-Za-z0-9_-]+)\]$/,
+    relevance: 10
+  };
+
+  const KEY_NAME = {
+    className: 'attr',
+    begin: /^[A-Za-z0-9_-]+(\[[A-Za-z0-9_@.]+\])?/,
+    end: /\s*=/,
+    excludeEnd: true,
```
joshgoebel: Prefer an end that uses `regex.lookahead` now vs `excludeEnd`.  That's also being deprecated.

joshgoebel: Or better yet should the look ahead be patch of a singular `match`?

Right now this can match attributes with no `=` at all, is that intentional?

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
...
+  const OPERATOR = {
+    className: 'operator',
+    match: /=/,
+    relevance: 0
+  };
+
+  const VALUE = {
+    begin: /=\s*/,
+    end: /$/,
+    excludeBegin: true,
```
joshgoebel: Is this necessary if we have no `scope` here?  A match without a scope won't get tagged...

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
...
+      STRING,
+      {
+        className: 'literal',
+        match: /\b(Application|Link|Directory|forking|oneshot|OneShot|true|false|True|False)\b/,
+        relevance: 10
+      }
+    ]
+  };
+
+  const KEY_VALUE = {
```
joshgoebel, line 86: What is the intention of KEY_VALUE vs `attr` above?

joshgoebel: Oh I see... hmm...

src/languages/freedesktop.js
```diff
@@ -0,0 +1,86 @@
...
+      FIELD_CODES,
+      KEY_NAME,
+      OPERATOR,
+      VALUE
+    ]
+  };
+
+  return {
+    name: 'FreeDesktop configuration format',
+    aliases: ['desktop', 'service', 'mount', 'socket', 'timer', 'nmconnection', 'systemd', 'freedesktop'],
```
joshgoebel: This might be a bit much... can we whittle this down a lot?

src/languages/freedesktop.js
joshgoebel: Is there an issue for this?  If so could you post a few sample freedesktop.org samples there so I can see what we're trying to match here and provide some better guidance.

Perhaps just add those examples as markup tests, which we'll need anyways to merge this as a grammar into core.

gg582:
Hi, thanks for remembering this pull request. Currently I started working on this pull request. I appreciate your feedback, and I will update this shortly(and submit an example of highlight samples)

gg582:
ssr
Hi, unfortunately my previous attached blog link was broken, so I removed them from this conversation. I am now attaching this rendered example. Currently this highlights some important service type and format specifier like `%F`, `%U`.

src/languages/freedesktop.js
gg582: Hi, I've removed this comment.

src/languages/freedesktop.js
gg582: Yup, I agree with your suggestion. This was misplaced. Now this is classified as `config` only.

src/languages/freedesktop.js
gg582: I rewrote this part into `scope` now:

## AS-IS

```javascript

const FIELD_CODES = {
    className: 'variable',
    match: /%[a-zA-Z]/,
    relevance: 1
};

// others remain same...
```

## TO-BE(current patch)

```javascript
const FIELD_CODES = {
    scope: 'variable',
    match: /%[a-zA-Z]/
};

// other constants are patched too..
```

src/languages/freedesktop.js
gg582: Now this part gets `freedesktop`, `desktop`, and `systemd`.

We may discuss if we shall narrow down more.

src/languages/freedesktop.js
gg582: I rewrote this part into `scope` now:

## AS-IS

```javascript

const FIELD_CODES = {
    className: 'variable',
    match: /%[a-zA-Z]/,
    relevance: 1
};

// others remain same...
```

## TO-BE(current patch)

```javascript
const FIELD_CODES = {
    scope: 'variable',
    match: /%[a-zA-Z]/
// no relevalance flag
};

// other constants are patched too..
```

src/languages/freedesktop.js
gg582: Now it seems to be removed well :+1:

joshgoebel:
Now I'm debating... this really looks like an ini/toml file and perhaps we should just parse it as such.  What is the advantage of a separate parser JUST for FreeDesktop - if freedesktop is NOT really a custom format at all.

gg582:
One concern I have with extending the generic INI grammar is that some of the required changes are actually incompatible with ordinary INI files.

For example, many INI implementations legitimately support `;` as a comment delimiter:

    key=value ; comment

However, in the FreeDesktop specifications, `;` is part of the syntax itself and is commonly used as a list separator:

    MimeType=image/jpeg;image/png;
    Categories=Network;WebBrowser;

Treating `;` as a comment would incorrectly terminate these values.

Likewise, the parser needs to recognize locale-qualified keys such as:

    Name[ko]=...
    Comment[fr]=...

as a single key rather than arbitrary bracket syntax.

Another difference is the handling of Linux-specific format specifiers:

    Exec=firefox %U
    Exec=myapp %F

These are not ordinary text substitutions. They are defined by the Desktop Entry Specification and are important syntax elements that users often expect to be highlighted.

Conversely, extending an INI grammar to recognize these constructs globally would introduce Linux-specific semantics into files that are not FreeDesktop configuration files. That increases the chance of false positives in ordinary INI documents.

Similarly, using a TOML grammar would introduce parsing rules that simply do not exist in the FreeDesktop specifications (quoted strings, arrays, multiline strings, etc.), leading to different classes of false positives.

So while the visual layout resembles INI, I think the syntax has diverged enough that a dedicated grammar remains simpler, more accurate, and less likely to affect existing INI highlighting behavior.

gg582:
A real example from Debian 13:

```ini
[Desktop Entry]
Name=Web
GenericName=Web Browser
Comment=Browse the web
# Translators: Search terms to find this application. Do NOT translate or localize the semicolons! The list MUST also end with a semicolon!
Keywords=web;browser;internet;epiphany;
Exec=epiphany %U
StartupNotify=true
Terminal=false
Type=Application
# Translators: Do NOT translate or transliterate this text (this is an icon file name)!
Icon=org.gnome.Epiphany
Categories=Network;GNOME;GTK;WebBrowser;
MimeType=text/html;application/xhtml+xml;x-scheme-handler/http;x-scheme-handler/https;multipart/related;application/x-mimearchive;message/rfc822;application/x-xpinstall;
Actions=new-window;Incognito;
# Translators: Do NOT translate or transliterate this text (these are enum types)!
X-Purism-FormFactor=Workstation;Mobile;
X-Ubuntu-Gettext-Domain=epiphany

[Desktop Action new-window]
Name=New Window
Exec=epiphany --new-window %U

[Desktop Action Incognito]
Name=New Incognito Window
Exec=epiphany --incognito-mode %U
```

gg582:
## Alternatives

Another possible approach would be to implement FreeDesktop support as an extension of the existing INI grammar instead of introducing a dedicated language.

For example, the INI grammar could expose optional behaviors such as:

- Disabling `;` as a comment delimiter
- Recognizing locale-qualified keys (`Name[ko]`, `Comment[fr]`, ...)
- Highlighting FreeDesktop field codes (`%U`, `%F`, `%i`, `%c`, ...)
- Recognizing FreeDesktop-specific keywords and values

### Advantages

- Reuses the existing INI parser.
- Reduces duplicated parsing logic for sections and key-value pairs.
- Makes the visual relationship between INI-like formats explicit.

### Trade-offs

As more FreeDesktop-specific rules are added, the generic INI grammar gradually becomes aware of Linux-specific semantics.

For example:

- Traditional INI commonly treats `;` as a comment delimiter, while FreeDesktop uses it as a value separator.
- FreeDesktop introduces Desktop Entry field codes (`%U`, `%F`, `%i`, ...), which have no meaning in generic INI files.
- Systemd, Desktop Entry, and NetworkManager define many domain-specific keywords that are irrelevant outside the FreeDesktop ecosystem.

Supporting all of these through optional branches inside a generic INI grammar may eventually become more complex than maintaining a dedicated grammar derived from the same key-value structure.

I am also open to this design if it is considered preferable by the maintainers. My primary goal is accurate FreeDesktop syntax highlighting while keeping the implementation maintainable.

gg582:
Personally, I think that both implementation plans are well-structured.

If we are planning to rewrite this into an extended `ini` variant, the result should render this same result.

IMO the way to implement is not the core problem, the real problem weighs on `FreeDesktop` support itself :+1: 

epiphany

joshgoebel:
Fair points.

src/languages/freedesktop.js
```diff
@@ -0,0 +1,63 @@
...
+    contains: [
+      OPERATOR,
+      STRING,
+      LITERALS,
+      FIELD_CODES
+    ]
+  };
+
+  return {
+    name: 'FreeDesktop configuration format',
```
joshgoebel:
```suggestion
    name: 'FreeDesktop config',
```

src/languages/freedesktop.js
```diff
@@ -0,0 +1,63 @@
...
+    aliases: ['freedesktop', 'desktop', 'systemd'],
```
joshgoebel:
```suggestion
    aliases: ['desktop', 'systemd'],
```

src/languages/freedesktop.js
```diff
@@ -0,0 +1,63 @@
...
+    match: /=/
+  };
+
+  const LITERALS = {
+    scope: 'literal',
+    match: /\b(Application|Link|Directory|forking|oneshot|OneShot|true|false|True|False)\b/
+  };
+
+  const KEY_VALUE = {
+    begin: /^[A-Za-z0-9_-]+(\[[A-Za-z0-9_@.]+\])?/m,
```
joshgoebel: What is `/m`... I don't think you can do this anyways since this regex is not used, they are all together compile into one huge regex - with a single set of flags... you can't control whether it has /m or not.

joshgoebel:
Dropped a few notes, looking good so far!

taufik-nurrohman:
Maybe just extend it to `*.ini` language definition then add a rule for comments started with `;` that it needs at least 1 white-space before the comment, and rules for others such as that `%U` placeholder.

gg582:
@taufik-nurrohman That's a good point. I'll try to append Josh Goebel's feedback and reconsider.

If the code is still dirty enough, rewriting the whole form into an extension is a good manner.

src/languages/freedesktop.js
gg582: yup, I removed /m. That didn't make any difference when rendering colors.

src/languages/freedesktop.js
gg582: good. now I applied that :) freedesktop and desktop are actually saying the same

src/languages/freedesktop.js
gg582: Yup, that comment was too long and verbose.

joshgoebel:
Ok, to clarify what I said earlier so it's not misunderstand.

1. Perhaps `ini/toml` is "good-enough" **as-it-is** to parse FreeDesktop files nicely/adequately - if so, that might be the suggestion.  Double so if FreeDesktop considers itself an INI or TOML file.
2. I was not suggesting we add things that are NOT common to ALL ini/toml files to the `ini/toml` grammar... there should NOT be FreeDesktop specific things in the INI grammar.

It's maybe bad that we have a combined INI/TOML... but that's legacy... we don't want to just shove every INI-like file into `ini`... for example if FreeDesktop has unique keywords/keys (as it seems to)... those should never be added to the more generic INI grammar.  

So I've come around to thinking it's ok to add a new grammar for this case - and this grammar is also largely more of a "one and done", not something that we'll spend lots of time maintaining for years to come... (as it's a simpler ruleset/syntax)

gg582:
Yes, I think so. The biggest problem to use `ini` as a freedesktop config parser, a comment section is the biggest problem.

`;` is considered as a delimiter for Freedesktop files, just like adding commas into a sentence.

However, classic `ini` uses semicolons as a comment section.

As I mentioned earlier, this causes a problem(which is commonly found when opening common shortcuts:

```ini
## A Type
Type=Application
## A Key and values
Key1=value1;value2;value3
## A Command
Exec=command %F
```

In this case, a parser should not consider `value2` and `value3`.
`Application` should be highlighted.
`%F` should be highlighted.

As a result, just adding `#` to enhance FreeDesktop config files can be considered as `not-enough` solution.
I appreciate that there are few programmers who uses Linux, or Unix.
I don't know much about American worksites, but at least in South Korea, many `FreeDesktop` files are commonly found at worksites(usually NetworkManager configs, `*.timer`, or `*.service` type for on-premise servers).

In my opinion, this is not limited for *another shortcut types for graphical target*, but a common file format which can be found at reproducible on-premise server setup.

Recently, many web-based code management systems are aggressively adopted even for closed-network infrastructures. I guess this may be useful for industrial sites. :+1: 

## P.S

In standard `ini` style config files, this is an equivalent expression:


```ini
; A Key and values
Key1=value1,value2,value3
```

gg582:
Hi, should I refactor, or freeze current status before any decisions?

src/languages/freedesktop.js
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  const OPERATOR = {
+    scope: 'operator',
+    match: /=/
+  };
+
+  const LITERALS = {
+    scope: 'literal',
+    match: /\b(Application|Link|Directory|forking|oneshot|OneShot|true|false|True|False)\b/
```
joshgoebel: Lets write this list first as a constant array of strings then join then to build this regex. (lots of examples in other grammars).  Makes for easier maintenance in the future.

src/languages/freedesktop.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const COMMENT = {
+    scope: 'comment',
+    begin: /#/,
+    end: /$/
+  };
+
+  const SECTION = {
+    scope: 'section',
+    begin: /^\[(Desktop Entry|Unit|Service|Install|Socket|Mount|Automount|Swap|Path|Timer|Slice|Scope|Manager|connection|ipv4|wifi|wifi-security|ipv6|802-3-ethernet|802-11-wireless|802-11-wireless-security|vpn|Journal|Bridge|Desktop Action\s+[A-Za-z0-9_-]+)\]$/
```
joshgoebel: Lets write this list first as a constant array of strings then join then to build this regex. (lots of examples in other grammars). Makes for easier maintenance in the future.

joshgoebel (changes requested):
Dropped two thoughts on small refactors.  But if you're happy I think this is in decent shape after those changes.

gg582:
In my opinion those two comments were valid. I'll try to refactor tomorrow!

gg582 (commented):
Currently I am pretty happy with current status. This implemnents freedesktop grammar as an individual syntax. There are few test markups(desktop shortcut, network manager, systemd unit) and code style is pretty much refined. Deprecated grammar is correctly updated, and regex result is okay. But, there are other methods that can support freedesktop format, which can be a long discussion.

src/languages/freedesktop.js
```diff
@@ -25,19 +25,60 @@ export default function(hljs) {
...
+    'wifi-security',
+    'ipv6',
+    '802-3-ethernet',
+    '802-11-wireless',
+    '802-11-wireless-security',
+    'vpn',
+    'Journal',
+    'Bridge',
+    'Desktop Action\\s+[A-Za-z0-9_-]+'
+  ];
```
joshgoebel, line 54: Nice, but longer matches must come first (or else just sort at the very end).

Otherwise `wifi` will prevent `wifi-security` from every matching.

joshgoebel (changes requested):

src/languages/freedesktop.js
gg582: ah yes! I get it, that must be right. I'll reorder those two.

gg582: I found a similar bug for `802-11-wireless` and `802-11-wireless-security`. Now I'm patching both :+1:

gg582: Yep, I patched both pairs. How about sorting all keywords by length? That is safe but I am worrying that may look too terribly to read what it is(and what keywords are in a same family)

joshgoebel: Sorting them in reverse alpha does the trick, we use that trick elsewhere in the code base.  If we never want to think about this again, just do that.

joshgoebel (approved):

PR #4307, opened 2025-08-25
fix(kotlin) number literals

### LINKED ISSUE #4305: (Kotlin) Invalid number literal highlighting
**Describe the issue**

highlight.js highlighting of Kotlin number literals doesn't match the actual number literals Kotlin accepts, including false-positives and false-negatives.

`kotlin`

No
...

**Sample Code to Reproduce**

https://jsfiddle.net/u3fn8t9m/

```kotlin
// Correct
val croissants = 34L
val muesli = 34f
val porridge = 34F

// False Negatives
// valid Kotlin - wrong highlighted
val schnitzel = 100u
val hagelslag = 23U
val churros = 99UL

// False Positives
// invalid Kotlin - still highlighted
val muesli = 34l
val melange = 34d
val bagel = 34D
```

**Expected behavior**

I expect the highlighting to match the official Kotlin grammar.

**Additional context**

From the [Kotlin Spec](https://kotlinlang.org/spec/pdf/kotlin-spec.pdf) section _1.2.3 Literals_ there are only 3 Rules that define suffixes for number literals:
```ebnf
FloatLiteral:
    DoubleLiteral ('f' |'F')
    | DecDigits ('f' |'F')

UnsignedLiteral
    (IntegerLiteral | HexLiteral | BinLiteral) ('u' |'U') ['L']

LongLiteral
    (IntegerLiteral | HexLiteral | BinLiteral)'L'
```

However, the current grammar for Kotlin [reuses Java numerical parsing](https://github.com/highlightjs/highlight.js/blob/5697ae5187746c24732e62cd625f3f83004a44ce/src/languages/kotlin.js#L9C1-L9C41) which does have significant differences.

flofriday:
Previously the parser for Kotlin reused the Java number parsing but there are significant differences in how the two languages express numbers causing false-negatives and false-positives.

Resolves #4305

### Changes
Implemented a custom parser for Kotlin number literals since they are different enough from Java literals to justify their own parser.

My hunch is that the orginal implementor misinterpreted the [release notes from Kotlin 1.1](https://kotlinlang.org/docs/reference/whatsnew11.html#underscores-in-numeric-literals) which they referenced since this only applies to how seperators in Kotlin follow the same rules as Java 8 and not how all the number parsing is identical to Java 8.
Moreover, in the 9 years since Kotlin 1.1 the language has changed a lot and for example back then unsigned number literals surely weren't a thing.

### Checklist


src/languages/kotlin.js
```diff
@@ -100,10 +98,31 @@ export default function(hljs) {
     ]
   };
 
-  // https://kotlinlang.org/docs/reference/whatsnew11.html#underscores-in-numeric-literals
-  // According to the doc above, the number mode of kotlin is the same as java 8,
-  // so the code below is copied from java.js
-  const KOTLIN_NUMBER_MODE = NUMERIC;
+  const decimalDigits = '[0-9](_*[0-9])*';
+  const frac = `\\.(${decimalDigits})`;
```
joshgoebel:
```suggestion
  const fractional = `\\.(${decimalDigits})`;
```

src/languages/kotlin.js
```diff
@@ -100,10 +98,31 @@ export default function(hljs) {
...
+  const KOTLIN_NUMBER_MODE = {
+    className: 'number',
```
joshgoebel:
```suggestion
    scope: 'number',
```

src/languages/kotlin.js
```diff
@@ -100,10 +98,31 @@ export default function(hljs) {
...
+    variants: [
+      // DecimalFloatingPointLiteral
+      // including ExponentPart
+      { begin: `(\\b(${decimalDigits})((${frac})|\\.)?|(${frac}))` +
```
joshgoebel:
```suggestion
      { match: `(\\b(${decimalDigits})((${frac})|\\.)?|(${frac}))` +
```

Prefer match for all these if there is no `end`.

flofriday: thanks :)

flofriday:
Thanks for the suggestions, I implemented them all :)

joshgoebel:
Did you? I'm not seeing the changes.

flofriday:
Uff, that's embarrassing, no I forgot to push 🙈

src/languages/kotlin.js
```diff
@@ -100,10 +98,31 @@ export default function(hljs) {
...
+      { match: `\\b(${decimalDigits})[fF]\\b` },
+  
+      // DecimalIntegerLiteral
+      { match: '\\b(0|[1-9](_*[0-9])*)[uU]?L?\\b' },
+  
+      // HexIntegerLiteral
+      { match: `\\b0[xX]([0-9a-fA-F](_*[0-9a-fA-F])*)[uU]?L?\\b` },
+  
+      // BinaryIntegerLiteral
+      { match: '\\b0[bB][01](_*[01])*[uU]?L?\\b' },
```
joshgoebel: Can we make all these actual regex to avoid the double \\ escaping?

src/languages/kotlin.js
```diff
@@ -100,10 +98,31 @@ export default function(hljs) {
...
+    ],
+    relevance: 0
```
joshgoebel: Don't need relevance anymore.

joshgoebel (changes requested):

joshgoebel:
Please lets go backwards (before rebase), the cedar doc change should NOT be a part of this PR.

flofriday:
> [..] the cedar doc change should NOT be a part of this PR.

Totally agree, I cannot find any cedar changes though, maybe it was in an older commit by accident which resolved itself on its own.

I now also included all the other changes.

src/languages/kotlin.js
```diff
@@ -100,10 +98,30 @@ export default function(hljs) {
...
-  // so the code below is copied from java.js
-  const KOTLIN_NUMBER_MODE = NUMERIC;
+  const decimalDigits = '[0-9](_*[0-9])*';
+  const fractional = `\\.(${decimalDigits})`;
+  const KOTLIN_NUMBER_MODE = {
+    scope: 'number',
+    variants: [
+      // DecimalFloatingPointLiteral
+      // including ExponentPart
+      { match: new RegExp(`(\\b(${decimalDigits})((${fractional})|\\.)?|(${fractional}))` +
```
joshgoebel: `new RegExp(...)` isn’t needed here — hljs accepts string patterns the same as regex literals. Same style as `src/languages/lib/java.js`:

```js
{ match: `(\\b(${decimalDigits})((${fractional})|\\.)?|(${fractional}))` +
  `[eE][+-]?(${decimalDigits})[fF]?\\b` },
{ match: `\\b(${decimalDigits})((${fractional})[fF]?\\b|\\.([fF]\\b)?)` },
{ match: `(${fractional})[fF]?\\b` },
{ match: `\\b(${decimalDigits})[fF]\\b` },
```

(Applies to the four float variants below too.)

joshgoebel (changes requested):

src/languages/kotlin.js
```diff
@@ -100,10 +98,30 @@ export default function(hljs) {
     ]
   };
 
-  // https://kotlinlang.org/docs/reference/whatsnew11.html#underscores-in-numeric-literals
-  // According to the doc above, the number mode of kotlin is the same as java 8,
-  // so the code below is copied from java.js
-  const KOTLIN_NUMBER_MODE = NUMERIC;
+  const decimalDigits = '[0-9](_*[0-9])*';
```
joshgoebel, line 101: Could we get a markup test for `_`... and truly is `5______9` a valid number?

flofriday: Yes `5______9` is really a valid number, here is a [Kotlin Playground example](https://pl.kotl.in/3WNqc6Hmw).

joshgoebel (approved):

PR #4308, opened 2025-08-29
fix(haxe): after colon, highlight non-identifiers like numbers and strings

joshtynjala:
### Changes

Issue: Numbers and strings after a colon were treated as type names, which could cause problems with highlighting for anonymous structures, like this one:

```haxe
var obj = { num: 123.4, str: "hello" };
```

With this change, non-identifiers are excluded from being treated as a type name.

### Checklist

### DIFF
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -74,7 +74,7 @@ export default function(hljs) {
       },
       {
         className: 'type', // function types
-        begin: /:[ \t]*/,
+        begin: ":[ \t]*(?=\\(|" + IDENT_RE + ")",
         end: /[^A-Za-z0-9_ \t\->]/,
         excludeBegin: true,
         excludeEnd: true,
@@ -82,7 +82,7 @@ export default function(hljs) {
       },
       {
         className: 'type', // types
-        begin: /:[ \t]*/,
+        begin: ":[ \t]*(?=" + IDENT_RE + ")",
         end: /\W/,
         excludeBegin: true,
         excludeEnd: true
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -73,19 +73,14 @@ export default function(hljs) {
         keywords: { keyword: 'if else elseif end error' }
       },
       {
-        className: 'type', // function types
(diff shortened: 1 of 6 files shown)


Yang062 (approved):

src/languages/haxe.js
```diff
@@ -74,15 +74,15 @@ export default function(hljs) {
       },
       {
         className: 'type', // function types
-        begin: /:[ \t]*/,
+        begin: ":[ \t]*(?=\\(|" + IDENT_RE + ")",
```
joshgoebel: These two “type after `:`” rules would be a better fit for multi-match + `scope` (prefer `scope` over deprecated `className`):

```js
// types
{
  begin: [/:/, /[ \t]*/, IDENT_RE],
  scope: { 3: "type" }
}

// function types — keep the wider `end` / contains behavior as needed
{
  begin: [/:/, /[ \t]*/, /(?=\(|)/],
  // ...
}
```

One match, per-group scopes, no `excludeBegin` / empty type spans.

src/languages/haxe.js
```diff
@@ -74,15 +74,15 @@ export default function(hljs) {
...
         end: /[^A-Za-z0-9_ \t\->]/,
         excludeBegin: true,
         excludeEnd: true,
         relevance: 0
       },
       {
         className: 'type', // types
-        begin: /:[ \t]*/,
+        begin: ":[ \t]*(?=" + IDENT_RE + ")",
```
joshgoebel: Worth a negative lookahead so `true` / `false` / `null` after `:` aren’t scoped as types (object fields still show them as `type` in the current expect):

```js
{
  begin: [
    /:/,
    /[ \t]*/,
    regex.concat(/\b(?!(?:true|false|null)\b)/, IDENT_RE)
  ],
  scope: { 3: "type" }
}
```

If you want them as `literal` instead of plain text, a separate literal rule that wins over this is fine too — lookahead is the lightest fix when this mode currently eats them.

joshgoebel (changes requested):

joshgoebel (approved):

joshtynjala:
Thank you!

PR #4309, opened 2025-08-29
fix(haxe) Restore highlighting of type after new keyword

joshtynjala:
### Changes

	
PR #3993 fixed incorrect highlighting of variable names starting with new (like `newValue`), but at the same time, it removed highlighting of type names that follow the `new` keyword.

This PR restores the lost highlighting for type names after the `new` keyword, while still ensuring that new at the start of a variable name is not highlighted as a keyword.

### Checklist

### DIFF
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -89,7 +89,7 @@ export default function(hljs) {
       },
       {
         className: 'type', // instantiation
-        beginKeywords: 'new',
+        begin: /new[ \t]+/,
         end: /\W/,
         excludeBegin: true,
         excludeEnd: true
--- a/src/languages/haxe.js
+++ b/src/languages/haxe.js
@@ -88,11 +88,15 @@ export default function(hljs) {
         excludeEnd: true
       },
       {
-        className: 'type', // instantiation
-        begin: /new[ \t]+/,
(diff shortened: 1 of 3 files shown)


Yang062 (approved):

src/languages/haxe.js
```diff
@@ -89,7 +89,7 @@ export default function(hljs) {
       },
       {
         className: 'type', // instantiation
-        beginKeywords: 'new',
+        begin: /new[ \t]+/,
```
joshgoebel: Today we'd usually express this as a multi-match rather than `begin` + `excludeBegin`/`end`:

```js
{
  begin: [/\bnew/, /\s+/, hljs.IDENT_RE],
  scope: {
    1: "keyword",
    3: "type"
  }
}
```

Same idea as java/scala/rust — one match, per-group scopes, no empty type spans, and `new` stays a keyword without `beginKeywords`. You'd still want a fuller type-path matcher if dotted names like `haxe.ds.IntMap` should all be scoped as `type`.

joshgoebel (changes requested):

joshgoebel (approved):

joshtynjala:
Thank you!

PR #4312, opened 2025-08-31
feat: add `npm` to Dependabot

sebastiaanspeck:
### Checklist


Yang062 (approved):

joshgoebel:
What is the actual real-world effect of this change?

sebastiaanspeck:
> What is the actual real-world effect of this change?

Right now Dependabot only watches GitHub Actions here, so npm dev/build dependencies (nothing runtime, since highlight.js has zero deps) just go stale until someone bumps them manually. This adds a weekly `npm` ecosystem entry so Dependabot opens PRs for outdated build tooling automatically, same as it already does for Actions. No effect on build output, just keeps dev dependencies from drifting.

joshgoebel:
I usually find dependabot more annoying than useful in other projects, but we can try it here and see if it'll proves useful.

joshgoebel (approved):

PR #4322, opened 2025-10-03
Fix #3928: support digit separators (') in C++ integer literals in #define

### LINKED ISSUE #3928: C++ Integer literals with quote delimiters in #defines break highlighting
**Describe the issue**
[Integer literals](https://en.cppreference.com/w/cpp/language/integer_literal) in C++ may be written as 1'000'000'000 instead of 1000000000 for greater readability. When this is used inside a `#define`, if breaks highlighting which treats the rest of the code as being inside a string.

cpp

highlight

**Sample Code to Reproduce**

```
#include <iostream>
#define INF 1'000'000'000

int main() {
	int a = INF;
    return 0;
}
```
image

**Expected behavior**
The token 1'000'000'000 should be highlighted as a number, and the following code should also be highlighted normally, instead of as if it is inside a string:

image

**Additional context**

The snippet below seems to be highlighted fine, so the issue is likely the interplay between quote-separated literals and defines.

```
#include <iostream>

int f(int x) { return f(2'000); }

int main() {
	int a = 1'000'000'000;
    int b = f(1'000);
    return 0;
}
```

aalokitbh:
#### Changes made:
 - Updated src/languages/cpp.js to include NUMBERS in PREPROCESSOR.contains so numbers in preprocessor lines are highlighted
 - Added a test snippet in test/markup/cpp/preprocessor.txt

### Closes#3928

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -112,7 +112,8 @@ export default function(hljs) {
         begin: /<.*?>/
       },
       C_LINE_COMMENT_MODE,
-      hljs.C_BLOCK_COMMENT_MODE
+      hljs.C_BLOCK_COMMENT_MODE,
+      NUMBERS
     ]
   };
 
--- a/test/markup/cpp/preprocessor.txt
+++ b/test/markup/cpp/preprocessor.txt
@@ -28,6 +28,10 @@ if (p) {
 #endif
 }
 
+// Test : digit separators in integer literals
+#define MILLION 1'000'000
+int x = 1'000'000;
+
 // this is a continued\
 comment.
 end


Yang062 (approved):

aalokitbh:
Thanks for the review, @Yang062! 😊
Please let me know if any further changes are needed. Also, is there any chance this PR could be merged or labeled for Hacktoberfest?

arturict (commented):
Can you add the expected markup for this case? preprocessor.txt gains the regression input, but preprocessor.expect.txt is unchanged, so the fixture does not assert that 1'000'000 is scoped as hljs-number without swallowing the following code.

src/languages/cpp.js
```diff
@@ -112,7 +112,8 @@ export default function(hljs) {
         begin: /<.*?>/
       },
       C_LINE_COMMENT_MODE,
-      hljs.C_BLOCK_COMMENT_MODE
+      hljs.C_BLOCK_COMMENT_MODE,
+      NUMBERS
```
joshgoebel, line 116: I think this would need to be more precise... we don't want this behavior for EVERY pre-processor, do we?  I'd guess mostly define?  Are there cases where we could hit false positives?

joshgoebel: Ping.

Issue #4331, opened 2025-10-21
(Python) Wrong coloring in python while wrongly using just one quote to include multiple lines string.


Locked-chess-official:
**Describe the issue**

**Which language seems to have the issue?**

**Are you using `highlight` or `highlightAuto`?**

...

**Sample Code to Reproduce**

**Expected behavior**

**Additional context**

<img width="340" height="292" alt="Image" src="https://github.com/user-attachments/assets/9699af7e-4487-487f-86c5-ec36a1801352" />

The invalid string's second line was wrongly drawn as string.

Here is the code to show this issue:
```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="UTF-8">
<title>Highlight.js Python test for quote</title>

<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.11.1/styles/default.min.css">
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.11.1/highlight.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.11.1/languages/python.min.js"></script>

<script>
hljs.highlightAll();
</script>
</head>

<body>
<h2>Highlight.js Python</h2>

<pre><code class="language-python">
# 1. valid string
a = 'abc'
print(a)

# 2. invalid string
b = 'abc
def'
print(b)
</code></pre>

</body>
</html>
```

JamesParrott:
I'm new to this library, but relevant definitions of single quoted Python strings + variants are here:

https://github.com/highlightjs/highlight.js/blob/5697ae5187746c24732e62cd625f3f83004a44ce/src/languages/python.js#L217

Is fixing this as simple as adding `illegal: '\\n',` or `illegal: /\n/`?  

(*) Kotlin and Scala also forbid new lines in single quoted strings.  In their syntax implementations elsewhere in highlight.js, we see the "illegal" field, that is absent in the Python ones:

```
      {
        begin: '"',
        end: '"',
        illegal: '\\n',
        contains: [ hljs.BACKSLASH_ESCAPE ]
      },
```
https://github.com/highlightjs/highlight.js/blob/5697ae5187746c24732e62cd625f3f83004a44ce/src/languages/scala.js#L35C1-L40C9

```
      {
        begin: '\'',
        end: '\'',
        illegal: /\n/,
        contains: [ hljs.BACKSLASH_ESCAPE ]
      },
```
[...]

PR #4333, opened 2025-10-22
Fix: #4331 Prevent wrongly coloring multilines in Python string patterns in non-thrice quotes

Locked-chess-official:
Added 'illegal' property to string patterns to prevent newlines coloring.

### Checklist


joshgoebel:
Please provide some exact examples of the problem and before/after states.

PR #4342, opened 2025-11-11
fix(markdown) Ignore mid_word underscores

### LINKED ISSUE #4279: (markdown) snake_case term causes italics in rest of file
**Describe the issue**

If you have an unescaped term in snake_case in a Markdown file, it puts highlight.js into "italic mode" when it should not. GitHub-Flavored Markdown (GFM) and others seem to not italicize underscores in the middle of a word.

- markdown

- `highlight`

**Sample Code to Reproduce**

https://jsfiddle.net/mn15jwg7/

```
Header_under

    code listing; italic

More text, still italic.
```

Here's how highlight.js renders it on the playground. Note how the `_` puts it into Italic mode and it never leaves, even in subsequent blocks:

Image

Here's how GitHub renders the same markdown:

Image

danvk:
Resolves #4279 

### Changes

This changes the markdown rule to ignore mid-word `_` characters. This brings highlight.js in line with GFM (GitHub-Flavored Markdown). There's no official standard for Markdown and different renderers vary in how they handle `mid_word` underscores. Not italicizing them is the more conservative behavior, and avoids the common situation where a large fraction of your Markdown file turns italic with highlight.js when it does not with GitHub.

### Checklist

- [x] Added markup tests

### DIFF
--- a/src/languages/markdown.js
+++ b/src/languages/markdown.js
@@ -156,8 +156,8 @@ export default function(hljs) {
         end: /\*/
       },
       {
-        begin: /_(?![_\s])/,
-        end: /_/,
+        begin: /(?<![a-zA-Z0-9])_(?![_\s])/,
+        end: /_(?![a-zA-Z0-9])/,
         relevance: 0
       }
     ]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,7 @@ Core Grammars:
 - enh(json) add json5 support [Kerry Shetline][]
 - fix(css) `unicode-range` parsing, issue #4253 [Kerry Shetline][]
 - fix(csharp) Support digit separators [te-ing][]
+- fix(markdown) Ignore mid_word underscores, issue #4279 [Dan Vanderkam]
 
 Documentation:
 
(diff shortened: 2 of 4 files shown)


src/languages/markdown.js
```diff
@@ -156,8 +156,8 @@ export default function(hljs) {
         end: /\*/
       },
       {
-        begin: /_(?![_\s])/,
-        end: /_/,
+        begin: /(?<![a-zA-Z0-9])_(?![_\s])/,
+        end: /_(?![a-zA-Z0-9])/,
```
joshgoebel, line 160: This will have to sit until v12 unless you know how to do this without negative look-behind... this would be a breaking change for older versions of Safari that v11 still supports.

danvk: Thanks for taking a look @joshgoebel. Is there any guess when there might be a v12? Is it months, years, maybe never?

Claude came up with a more complex version that doesn't have any negative lookbehind. If this seems reasonable, I can move it over to this PR:
https://github.com/highlightjs/highlight.js/compare/main...danvk:highlight.js:markdown-midword-underscore-no-lookbehind

joshgoebel: Hard to say, I would have thought a year ago but then I got busy with life.

I guess if we're never going to need to match on that last character for another mode then it could be gobbled up by the "guard" here, yes... When you pull that over please also leave your negative look-behind in place, just comment it out and add a short comment that it's for v12+.

PR #4346, opened 2025-12-05
(java) Improve detection of types, including generic and array types

### LINKED ISSUE #3678: (java) Highlighting fails on square brackets ([]) of array type
**Describe the issue**
When the return type of a method is an array type, the function name is not highlighted.

Java

`hightlightAll()` with `class="language-java"` on the `<code>`-block

**Sample Code to Reproduce**
```html
<pre><code class="language-java">
interface FooBar {
  String foo();
  String[] bar1();
  String&#91;&#93; bar2();
}
</code></pre>
```
http://jsfiddle.net/Elewyth/Lxmgwosk/

**Expected behavior**
`foo`, `bar1`, and `bar2` should all be highlighted as a function name.

**Additional context**
Tested with 11.7.0.

### LINKED ISSUE #4017: (Java) Generic type variable not highlighted
**Describe the issue**
In a Java statement that declares and assigns a variable with a generic type, the type and variable don't get highlighted. Without the generic type they are properly highlighted.

java

highlightElement
...

**Sample Code to Reproduce**
<pre><code class="language-java">Map&lt;String, String&gt; params = request.getParams();</code></pre>

**Expected behavior**
Without the generic type, the type (Map) and variable (params) get highlighted. They should be similarly highlighted also when the generic type is present (and the generic type itself should be highlighted as well).

**Additional context**
The example shows two generic types, but the issue also occurs with a single type such as List<String>. This is the proper way of defining such collections.

### LINKED ISSUE #4345: (java) Method return type is not highlighted
**Describe the issue**
The return type in method declarations is not highlighted. 

Java

highlightAuto

**Sample Code to Reproduce**
```
Object get() { ... }
```

**Expected behavior**
The return type `Object` should be highlighted with class `hljs-type`, as it is in variable declarations

**Additional context**
In methods with return type `void` the return type is highlighted as `hljs-keyword`. Although `void` is a keyword in Java, it represents a type and should also be highlighted as `hljs-type`.

hns:
This improves detection of types in Java variable and method declarations, including support for generic and array types. 

Resolves #3678
Resolves #4017
Resolves #4345

### Changes

 - The regexes for type name and type parameters for generic types are split so they can be highlighted independently.
 - The regex for type parameters in generic types is improved to support wildcards and bounded types.
 - A regex for array brackets is added to support array types.
 - Regexes for type parameters and array brackets are added to variable and method declarations so they support generic and array types.
 - Like types of variable declarations, return types in methods are highlighted as `hljs-type`, including the `void` type. 

### Checklist


src/languages/java.js
```diff
@@ -30,8 +30,9 @@ function recurRegex(re, substitution, depth) {
 export default function(hljs) {
   const regex = hljs.regex;
   const JAVA_IDENT_RE = '[\u00C0-\u02B8a-zA-Z_$][\u00C0-\u02B8a-zA-Z_$0-9]*';
-  const GENERIC_IDENT_RE = JAVA_IDENT_RE
-    + recurRegex('(?:<' + JAVA_IDENT_RE + '~~~(?:\\s*,\\s*' + JAVA_IDENT_RE + '~~~)*>)?', /~~~/g, 2);
+  const TYPE_ARG_RE = '(?:(?:' + JAVA_IDENT_RE + '~~~)|(?:\\?\\s+(?:extends|super)\\s+' + JAVA_IDENT_RE + '~~~)|(?:\\?))';
```
nizarbenalla, line 33: This accepts `List<String> xs = ...` and `List<String>[] xs = ...` but not `List<String[]> xs = ...` and `Map<? extends Foo[], Bar> m = ..` which are valid array type aguments

src/languages/java.js
```diff
@@ -188,7 +189,7 @@ export default function(hljs) {
       {
         begin: [
           regex.concat(/(?!else)/, JAVA_IDENT_RE),
-          /\s+/,
+          regex.concat(GENERIC_RE, ARRAY_RE, /\s+/),
```
nizarbenalla, line 192: This accepts `int[] values` and `byte[] getBytes()` but not `int values[]` and `int getValues()[]` (the brackets are outside the type match)

nizarbenalla (approved):
Change looks good, I left a couple of inline comments.
The cases I mentioned are pretty harmless, you may only notice them if you're using snippets.
I also see that the copyright notices are being applied correctly.

src/languages/java.js
joshgoebel: Ok can we break this down with comments and built it across multiple lines?  This is starting to get pretty gnarly and hard to read...

src/languages/java.js
```diff
@@ -223,11 +224,15 @@ export default function(hljs) {
       },
       {
         begin: [
-          '(?:' + GENERIC_IDENT_RE + '\\s+)',
```
joshgoebel, line 226: So we don't need generic here anymore?

joshgoebel: Maybe if ARRAY_RE is optional name it `ARRAY_OPTIONAL_RE`?

src/languages/java.js
```diff
@@ -223,11 +224,15 @@ export default function(hljs) {
...
-          hljs.UNDERSCORE_IDENT_RE,
+          JAVA_IDENT_RE,
+          regex.concat(GENERIC_RE, ARRAY_RE, /\s+/),
+          JAVA_IDENT_RE,
           /\s*(?=\()/
         ],
-        className: { 2: "title.function" },
+        className: {
+          1: "type",
+          3: "title.function"
```
joshgoebel, line 252: So we don't highlight          ` regex.concat(GENERIC_RE, ARRAY_RE, /\s+/),` ?

joshgoebel:
I like where you're headed with this, just need to make it all a bit more readable, and I dropped a few questions.

hns:
Thanks for looking into this, Josh. I'll update the PR to address your feedback in the next few days.

src/languages/java.js
hns: No, this PR is just to recognize method and variable declarations that use generic or array types. Highlighting those parts is left for a future enhancement.

src/languages/java.js
hns: Renamed to `ARRAY_BRACKETS_OPTIONAL_RE` in my latest commit. Let me know if you think this is too verbose.

src/languages/java.js
hns: Addressed in my latest commit.

hns (commented):
@joshgoebel @nizarbenalla I addressed your review feedback in a followup commit. 

I also fixed a bug I discovered where a keyword was highlighted as type if it is followed by an assignment, such as in:

```
return foo = 3;
```

Let me know if you think this should be fixed as a separate issue.

src/languages/java.js
hns: > So we don't need generic here anymore?

It's covered in line 246 in `regex.concat(TYPE_ARGS_OPTIONAL_RE, ...)`

src/languages/java.js
```diff
@@ -186,18 +203,25 @@ export default function(hljs) {
...
       },
+      {
+        // Expression keywords prevent keyword-led expressions from being
+        // recognized as variable or method declarations.
+        beginKeywords: 'new throw return else yield assert',
+        relevance: 0
+      },
       {
         begin: [
-          regex.concat(/(?!else)/, JAVA_IDENT_RE),
```
joshgoebel, line 190: So we do not need to guard against else any more because you moved the `beginKeywords` rule above?

src/languages/java.js
```diff
@@ -186,18 +203,25 @@ export default function(hljs) {
...
-          regex.concat(GENERIC_RE, ARRAY_RE, /\s+/),
```
joshgoebel: So what happened to our matching of generics?

joshgoebel: Is this named TYPE_ARGS_OPTIONAL_RE now?

src/languages/java.js
joshgoebel: I'm just trying to follow the renaming since these names don't have a ton of meaning for me.  i think I see now you've just renamed a few things.

src/languages/java.js
hns: Yes, we moved those keywords up so they cover this section, and they make this exclusion of else obsolete. There's already an existing test in bugs.txt for `else` that would otherwise fail.

src/languages/java.js
hns: Yes, the old `GENERIC_IDENT_RE` was meant to match java generic type arguments, but it only matched the most simple forms such as:

```
List<String>
```

The new more complex `TYPE_ARGS_OPTIONAL_RE` with `ARRAY_BRACKETS_OPTIONAL_RE` match more complex forms, such as 

```
Map<? extends Key, List<? super Value[]>>
```

src/languages/java.js
hns: Yes, "type args" is the more precise term for the part in angle brackets than "generic".

joshgoebel:
how you feeling? Ready to undraft? Anything further we need to do here?

hns:
I had kind of forgotten this was still in draft mode. I think it looks good so I marked it as ready.

joshgoebel (approved):

hns:
Thanks, Josh!

PR #4357, opened 2026-02-13
The language Coq has been renamed into Rocq by its community

Mbodin:
I didn't open an issue for this. Is this a problem?

### Changes
- The language Coq has been renamed to Rocq by its community. I'm thus renaming it (while keeping an alias to `coq`).
- I added some new keywords that were missing.
- I sorted the keywords in the files.

### Checklist
- [X] I didn't add markup tests, as the language didn't fundamentally changed.


src/languages/rocq.js
```diff
@@ -421,7 +464,11 @@ export default function(hljs) {
     "with"
   ];
   return {
-    name: 'Coq',
+    name: 'Rocq',
+		aliases: [
+			'rocq',
```
joshgoebel: Doesn't need to list itself.

src/languages/ocaml.js
```diff
@@ -11,7 +11,11 @@ export default function(hljs) {
   /* missing support for heredoc-like string (OCaml 4.0.2+) */
   return {
     name: 'OCaml',
-    aliases: [ 'ml' ],
+    aliases: [
+			'ocaml',
```
joshgoebel: Doesn't need to list itself.

joshgoebel:
> The language Coq has been renamed to Rocq by its community. I'm thus renaming it (while keeping an alias to coq).

Link to a news release or blog or something on this?

src/languages/ocaml.js
joshgoebel: How is this related to rocq?

Mbodin: Both communities are very close to each other (and they have the tendency to remove the initial O of Ocaml). I can remove this from this PR if it causes any issue.

Mbodin:
> Link to a news release or blog or something on this?

Would https://github.com/rocq-prover/rfcs/blob/coq-roadmap/text/069-coq-roadmap.md#change-of-name-coq---the-rocq-prover be enough?

The old webpage coq.inria.fr now redirects to https://rocq-prover.org/ The old account https://github.com/coq/ now features a single webpage ( https://github.com/coq/coq.github.io/blob/main/404.html ) with a redirection to https://rocq-prover.org/
The change is not fully done, though: some packages in https://rocq-prover.org/platform still have names of the form coq-*. I think that this justifies that both names should be aliases.

PR #4368, opened 2026-03-03
fix: skip re-highlighting to avoid false unescaped HTML warnings fixes #3761

### LINKED ISSUE #3761: Repeated highlighting of already highlighted code causes "*One of your code blocks includes unescaped HTML. This is a potentially serious security risk.*"
## **Describe the issue/behavior that seems buggy**
This issue exists at least in the current *highlight.js* version `11.7.0`.

When *highlight.js* has already highlighted some code, then when this code is highlighted again, *highlight.js* will warn with "*One of your code blocks includes unescaped HTML. This is a potentially serious security risk.*".

This e.g. happens with consecutive executions of `hljs.highlightAll()`.

I *believe* this is a *false* alarm, as my assumption for the root cause is that *highlight.js* detects its own previously added `<span class="hljs-syntaxfoo>...</span>` styling elements.

## **Sample Code or Instructions to Reproduce**
* Visit https://highlightjs.org/ . This website presents random highlighted code sample.
* Open DevTools and execute `hljs.highlightAll()`.
* *highlight.js* will give the warning, including "*The element with unescaped HTML: `<code class="hljs language-foo">...</code>`*".

Other highlighting options also provoke this warning, e.g. with `hljs.highlightElement(..)`:

```
let highlightedCodes = [...document.getElementsByClassName('hljs')];
highlightedCodes.forEach( highlightedCode => hljs.highlightElement(highlightedCode) );
```

## **Expected behavior**
On consecutive highlighting attempts, *highlight.js* shall notice already-highlighted code and handle the situation gracefully:
### Option 1
Use a marker to convey that the code is already highlighted, e.g. by adding a data attribute such as `data-hljs-highlighted="true"` to the wrapping `<code>...</code>` element (`<code class="hljs language-foo" data-hljs-highlighted="true">..</code>`), and have *highlight.js* skip highlighting alltogether when it detects such marker.

But this highlighting skipping may not be desirable for scenarios where to-be-highlighted code is dynamically updated, e.g. a poor man's code editor with `<pre contenteditable><code>...</code></pre>`.

### Option 2
Tolerate *highlight.js*' own styling HTML and suppress warnings for them.

I am not sure if this could unintentionally swallow correct warnings, e.g. when wrapping *malicious* unescaped HTML within highlight.js syntax highlighting styling elements (`<span class="hljs-comment"><script>doMaliciousThings()</script></span>`).

## **Additional context**
While consecutive highlighting causes DOM updates to the already-highlighted code, the end result stays the same (i.e., the highlighting is effectively idempotent), which I consider a good thing/expected. E.g., consecutive highlighting does not create undesireable nested `<span class="hljs-comment"><span class="hljs-comment">...</span></span>` styling tags. 👍

Also see related discussion in https://github.com/highlightjs/highlight.js/issues/2886 ("*Detect HTML/JS injection attacks and warn users pro-actively*").

KJyang-0114:
## Summary

When `highlightElement` is called on already-highlighted code (e.g., consecutive `hljs.highlightAll()` calls), the existing `<span class="hljs-*">` tags from the previous highlighting are incorrectly flagged as "unescaped HTML" - a false positive security warning.

## Fix

This fix checks for the `data-highlighted` attribute early and silently skips re-highlighting, matching the proposed solution in issue #3761.

## Test

The fix prevents the false warning when calling `hljs.highlightAll()` multiple times on the same code blocks.

Fixes #3761

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -751,7 +751,8 @@ const HLJS = function(hljs) {
       { el: element, language });
 
     if (element.dataset.highlighted) {
-      console.log("Element previously highlighted. To highlight again, first unset `dataset.highlighted`.", element);
+      // Already highlighted - skip to avoid false "unescaped HTML" warnings
+      // caused by hljs's own <span> tags from previous highlighting
       return;
     }
 


joshgoebel:
This behavior is warned on purpose.  If you don't want to highlight something twice, don't. This is a problem we implementation, not the library.

PR #4374, opened 2026-03-08
fix: allow unclosed single quotes in php-template (fixes #4152)

### LINKED ISSUE #4152: (PHP Template) A single apostrophe ' breaks php-template highlighting
**Describe the issue**

`php-template`

 Highlight, not auto.

**Sample Code to Reproduce**
Code: See expected behaviour

**Not highlighted:**

```
<?php
// single apostrophe later.
?>
```

image

**Correctly highlighted:**
```
<?php
// single apostrophe later.
?>
```
image

**Expected behavior**

```php
<?php
// single apostrophe later.
?>
```

**Additional observations**  

KJyang-0114:
## Description

Fixes issue #4152 where a single apostrophe (unclosed quote) in php-template breaks highlighting for everything after it.

## Problem
In php-template, an unclosed single quote like  causes the entire remainder of the document to not be highlighted correctly. The issue does not occur in regular PHP mode.

## Solution
Removed the `skip: true` property from the inherited APOS_STRING_MODE and QUOTE_STRING_MODE in php-template.js. This allows the parser to handle unclosed quotes more gracefully without breaking subsequent highlighting.

## Testing
Added a test case in `test/markup/php-template-apostrophe/default.txt` that reproduces the issue from the bug report.

## Reproduction
```
<?php
// This won\'t be highlighted unless there is another 
// single apostrophe later.
?>
```

### DIFF
--- a/src/languages/php-template.js
+++ b/src/languages/php-template.js
@@ -33,17 +33,16 @@ export default function(hljs) {
             end: '\'',
             skip: true
           },
+          // Allow unclosed single quotes (fixes issue #4152)
           hljs.inherit(hljs.APOS_STRING_MODE, {
             illegal: null,
             className: null,
-            contains: null,
-            skip: true
+            contains: null
           }),
           hljs.inherit(hljs.QUOTE_STRING_MODE, {
             illegal: null,
             className: null,
-            contains: null,
-            skip: true
+            contains: null
           })
         ]
       }


joshgoebel:
`skip` is important so that entire chunks of skipped code are correctly handed to `php` for it's own subprocessing.  Perhaps if there was another simple way to fix this without touching `skip`...

As a general policy though we don't care a lot about incomplete/invalid code chunks (which often carry all sorts of challenges to get right)... _we are optimized for highlighting syntactically valid code._.

PR #4380, opened 2026-03-16
Add batch script support and fix minor issues

megyerikenez:
### Changes
This pull request introduces syntax highlighting support for Windows batch files to the project. The main addition is a new language definition for batch scripts, along with relevant tests to ensure proper detection and markup of batch syntax elements.

Batch language support:

* Added a new language definition file `src/languages/batch.js` for Windows batch files, supporting `.bat`, `.batch`, and `.cmd` extensions, with highlighting for built-in commands, operators, comments, strings, variables, and numbers.

Testing and validation:

* Added detection test file `test/detect/batch/default.txt` to verify that batch files are correctly identified by the language detector.
* Added markup test input `test/markup/batch/default.txt` and expected output `test/markup/batch/default.expect.txt` to validate syntax highlighting for various batch commands, comments, variables, and control structures. [[1]](diffhunk://#diff-c2236e3cff6c92db54b4f5a2ed0406bee8aedfc0f6e535a05012c954e52ee095R1-R43) [[2]](diffhunk://#diff-1fbcb7b92716c0d5b57de9c3dfe88b7c30c37495202102b4f329ff4c615f2aa2R1-R43)

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,14 @@
+## Version 11.11.3
+
+New Grammars:
+
+- added Batch file language support [Kenez Megyeri][]
+
+CONTRIBUTORS
+
+[Kenez Megyeri]: https://github.com/kenez-megyeri
+
+
 ## Version 11.11.2
 
 Themes:
@@ -55,6 +66,7 @@ CONTRIBUTORS
 [te-ing]: https://github.com/te-ing
 [Anthony Martin]: https://github.com/anthony-c-martin
 [NriotHrreion]: https://github.com/NriotHrreion
+[Kenez Megyeri]: https://github.com/kenez-megyeri
 
 
 ## Version 11.11.1
--- a/src/languages/batch.js
+++ b/src/languages/batch.js
@@ -2,7 +2,7 @@
 Language: Batch
 Author: Kenez Megyeri <m.kenz99@gmail.com>
 Website: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/windows-commands
-Description: Syntax highlighting for windows batch files and scripts eg.(.bat, .batch, .cmd)
+Description: Syntax highlighting for Windows batch files and scripts eg.(.bat, .batch, .cmd)
 Category: common, scripting
 */
 
@@ -12,8 +12,8 @@ export default function(hljs) {
     name: "Batch file",
     aliases: [
       "bat",
-      "cmd",
-      "dos"
+      "batch",
+      "cmd"
     ],
     case_insensitive: true,
     keywords: {
--- a/src/languages/batch.js
+++ b/src/languages/batch.js
@@ -1,0 +1,61 @@
+/*
+Language: Batch
+Author: Kenez Megyeri <m.kenz99@gmail.com>
+Website: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/windows-commands
+Description: Syntax highlighting for windows batch files and scripts eg.(.bat, .batch, .cmd)
+Category: common, scripting
+*/
+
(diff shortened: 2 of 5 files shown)


joshgoebel:
Don't we already support batch files?  I just added an alias for `batch` to one of our grammars...

joshgoebel:
I believe it may be time to rename our dos grammer to dos_batch... is this a replacement/better grammar for DOS than dos already is?

megyerikenez:
Thanks for looking, @joshgoebel — answering both questions directly.

**"Don't we already support batch files?"** Yes, `dos.js` does, and I saw #4395 land the `batch` alias there. So to be clear about intent: I'm **not** proposing this as an additional language. I'm proposing it as a **replacement for the `dos` grammar** — which is exactly the question in your second comment. Framed that way it isn't a new grammar under the "no new core grammars" policy, it's a rewrite of an existing one.

**"Is this a better grammar for DOS than dos already is?"** On the things modern `.bat`/`.cmd` files actually contain, yes — and these aren't edge cases:

| input | `dos.js` today | this grammar |
|---|---|---|
| `:: colon comment` | not highlighted at all | `comment` |
| `set "MSG=hi there"` | quotes plain — **there is no string mode in `dos.js`** | `string` |
| `robocopy`, `icacls`, `schtasks`, `whoami`, `netsh`, `tasklist`, `reg`, `sc`, `wmic`, `timeout` | missing | `built_in` |
| `prompt $P$G` | missed — keyword list has the typo `promt` | `built_in` |
| `fc a.txt b.txt` | missed — keyword list has `fs`, likely a typo for `fc` | `built_in` |

`::` is arguably the dominant comment idiom in modern batch, and having no string mode whatsoever affects nearly every real-world script. The two typos suggest that keyword list hasn't been audited in a long while. Meanwhile a good chunk of what `dos.js` does carry is DOS-era device names — `lpt3`, `com4`, `aux`, `prn` — which are mostly noise for anything written this century. That mismatch is, I think, the same thing that prompted your `dos_batch` rename idea.

**Where `dos.js` is currently better, and what I'll do about it.** Being straight with you rather than having you find these in CI:

- `dos.js` highlights `%VAR%`, `%%i` loop vars and `!DELAYED!` outside strings; mine only handles `%VAR%` nested inside a string. That's the biggest gap and I'll port `dos.js`'s variable mode over wholesale.
- `dos.js` has a `function` mode for `:label` … `goto:eof`. I'll keep it.
- `dos.js` matches `@rem` via `/^\s*@?rem\b/`; mine requires a trailing space and misses the `@` form. Fixing.
- I'll restore the `keyword` vs `built_in` split (`if`/`else`/`for`/`goto` should be `keyword`, not `built_in`), drop `while` (not a batch keyword), and move the keyword lists to arrays to match current house style.
- I'll drop `Category: common` — that's your call to make, not mine, and `dos` isn't in the common bundle. Together with re-adding `illegal: /\/\*/`, `relevance: 0` on numbers, and tightening the bare `[|&><]` operator rule, that fixes the auto-detect scoring; right now this grammar wins detection on some non-batch samples and fails the XML/PHP sub-language test, which I'll have green before asking for another look.

So the end state I'm proposing: rename to `dos_batch` as you suggested, this grammar becomes its body with `dos.js`'s variable/label/`@rem` handling folded in, and `dos` + `bat` + `cmd` + `batch` all stay as aliases so nothing breaks for existing users. I'll also rebase — this was opened off my fork's `main` and has drifted.

Before I put in that work: would you rather have it as a single replacement PR along those lines, or as an incremental `enh(dos)` series against the existing file? Happy to go either way, just don't want to build the wrong one.

PR #4385, opened 2026-03-25
Add 'limit' keyword to SQL language support

rozhnev:
SQL language keywords list updated by 'limit' keyword.


src/languages/sql.js
```diff
@@ -408,6 +417,7 @@ export default function(hljs) {
     "stddev_samp",
     "submultiset",
     "subset",
```
Yang062, line 419: 96db/test

Yang062 (approved):

joshgoebel:
Please show documentation that this is part of the actual SQL standard, not just popular in modern DBs.  The SQL syntax is only for official SQL, it does not include many modern SQL tropes.  If you wanter better SQL support I'd recommending using a syntax for your particular database rather than `sql`.

joshgoebel:
Closing as dup of #3952  (and others) unless you can show proof this is now a part of ANSI SQL.

PR #4386, opened 2026-03-27
(docs) List languages with category as "common"

mbtools:
There's no list of languages which are included when using `require('highlight.js/lib/common')`.

### Changes

Adds a "Common" column to the list of supported languages.

Based on [code search](https://github.com/search?q=repo%3Ahighlightjs%2Fhighlight.js++%2F%5Ecategory%3A.*common%2F+path%3A%2F%5Esrc%5C%2Flanguages%5C%2F%2F&type=code) for "common" category.

### Checklist
- [x] Added markup tests: n/a


Yang062 (approved):

joshgoebel:
I see some value, but I think this will also increase maintenance burden.   It's also recommended people build their own distribution rather than use the default set.  If we're going to do this we need a script capable of updating this automatically during builds/CI.

I also wonder if this shouldn't just be a bullet point rather than mixing it with the table.

joshgoebel:
> I also wonder if this shouldn't just be a bullet point rather than mixing it with the table.

Yes, I think this is the way.

mbtools:
It's just a simple list now

joshgoebel (approved):

PR #4398, opened 2026-05-19
enh(python): Add support for Python 3.15 lazy import syntax

ZeroIntensity:
### Changes

Support for `lazy import`/`lazy from ... import ...` in Python 3.15+. See [PEP 810](https://peps.python.org/pep-0810/) for the specification.

Add `lazy` to the reserved names list to highlight it as a keyword. It is worth noting that this will also highlight cases where `lazy` is not an actual keyword (such as `lazy = 42`), but this seems to be the case for soft keywords already (`match = 42` is also highlighted, for example).

### Checklist


AA-Turner (approved):

ZeroIntensity:
cc @joshgoebel

ZeroIntensity:
@allejo, are you around for a review?

joshgoebel (approved):

joshgoebel:
Please rebase.

joshgoebel:
@copilot resolve the merge conflicts in this pull request

joshgoebel:
It needs to be rebased (not merged) and the conflcits with CHANGELOG reconciled.

PR #4399, opened 2026-05-21
Fix C++ function highlighting for keyword prefixes

### LINKED ISSUE #4394: (cpp) Inconsistent highlighting of top-level identifiers that start with a keyword
**Describe the issue**
Identifiers at the top level that start with certain keywords (such as `for`, `while`, `if`, etc.) are not highlighted, while others get the `hljs-built_in` class.  It's not clear to me which one is right but they should be treated the same.

Image

`cpp`

`highlight`

**Sample Code to Reproduce**
https://jsfiddle.net/2asbwfpq/

**Expected behavior**
```cpp
testing();
format();
for_this();
std::forceSomething();
whilexyz();
if (ifsyz()) {}
returnxyz();
thisxyz();
putforce();
meanwhile();
mean_while();
```

puneetdixit200:
Fixes #4394

### Changes
- Treat only exact C++ reserved keyword identifiers as excluded from function dispatch highlighting.
- Add C++ markup coverage for function names that start with keywords, including `format`, `for_this`, and `whilexyz`.

### Checklist

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -417,6 +417,10 @@ export default function(hljs) {
     _type_hints: TYPE_HINTS
   };
 
+  const RESERVED_KEYWORD_RE = regex.either(
+    ...RESERVED_KEYWORDS.map(keyword => keyword.replace(/\|\d+$/, ''))
+  );
+
   const FUNCTION_DISPATCH = {
     className: 'function.dispatch',
     relevance: 0,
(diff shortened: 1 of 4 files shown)


arturict:
This parser-keyword fix appears scoped; if no docs update is needed, confirming no false positives in function-like headers would be useful.

src/languages/cpp.js
```diff
@@ -425,7 +429,7 @@ export default function(hljs) {
       _hint: FUNCTION_HINTS },
     begin: regex.concat(
       /\b/,
-      `(?!${RESERVED_KEYWORDS.join('|')})`,
+      `(?!${RESERVED_KEYWORD_RE}\\b)`,
```
joshgoebel, line 493: \b is to allow reserved words to still be part of identifiers if they are mere prefixes?

joshgoebel (commented):
One question

joshgoebel:
Ping.

joshgoebel (approved):

PR #4414, opened 2026-05-31
Enhance NSIS rules and tests

idleberg:
### Changes
This is my second attempt at https://github.com/highlightjs/highlight.js/pull/3822. It updates the syntax with missing language features, improves the highlighting logic and adds more test cases.

### Checklist


src/languages/nsis.js
```diff
@@ -103,41 +108,53 @@ export default function(hljs) {
...
+    "elseif",
     "else",
     "endif",
     "error",
     "execute",
     "finalize",
     "getdllversion",
     "gettlbversion",
-    "if",
-    "ifdef",
```
joshgoebel, line 131: Please justify these order changes.  Alphabetical order is preferred.

src/languages/nsis.js
```diff
@@ -175,15 +192,19 @@ export default function(hljs) {
   const PARAMETERS = {
     // command parameters
     className: 'params',
-    begin: regex.either(...PARAM_NAMES)
+    begin: regex.concat(
+      regex.either(...PARAM_NAMES),
+      /\b/
```
joshgoebel, line 197: Give an example of the false positives you were seeing?

src/languages/nsis.js
```diff
@@ -542,15 +575,15 @@ export default function(hljs) {
...
+      { beginKeywords: 'Function PageEx Section SectionGroup FunctionEnd PageExEnd SectionEnd SectionGroupEnd', },
       STRING,
       COMPILER,
       DEFINES,
       VARIABLES,
       LANGUAGES,
       PARAMETERS,
       PLUGINS,
-      hljs.NUMBER_MODE
+      hljs.C_NUMBER_MODE
```
joshgoebel, line 586: Explain please.

test/markup/nsis/default.expect.txt
```diff
@@ -3,36 +3,59 @@
   for highlight.js
 */</span>
 
+<span class="hljs-comment">; Semicolon comment</span>
+<span class="hljs-comment"># Hash comment</span>
+
 <span class="hljs-keyword">!assert</span> <span class="hljs-variable">${NSIS_CHAR_SIZE}</span> = <span class="hljs-number">2</span> <span class="hljs-string">&quot;Unicode required&quot;</span>
 
-<span class="hljs-comment">; Includes</span>
```
joshgoebel, line 8: Why did you remove these? They also serve to document the tests themselves.

src/languages/nsis.js
idleberg: NSIS supports int, float and hex

src/languages/nsis.js
idleberg: - `HKLM` vs `HKLM32`/`HKLM64`
- `SW_SHOW` vs `SW_SHOWMAXIMIZED`/`SW_SHOWMINIMIZED`/`SW_SHOWNORMAL`

joshgoebel: Hmmmm... ok, makes sense... but... idea for future: what if we changed the behavior of `either` to always perform a reverse sort... making sure that the longest terms naturally come first... guaranteeing the order with nothing more than the name itself?

joshgoebel (approved):

PR #4415, opened 2026-06-04
fix(bash): keep heredoc continuations highlighted

### LINKED ISSUE #4377: (shell/bash) Wrong syntax highlight after a heredoc, until the end of the line.
**Describe the issue**

hljs assumes that what comes after a heredoc is a string.

However, the end of the line may be the continuation of the command line.

shell, bash

highlight

**Sample Code to Reproduce**

```bash
echo "ok" <<-EOF | cat
	Hello
EOF
```

**Expected behavior**

`cat` should be highlighted as a command instead of being highlighted as a string.

puneetdixit200:
Fixes #4377.

## Summary
- Keep Bash heredoc continuation commands on the command line instead of folding them into the heredoc string.
- Start the heredoc string span on the following line and end it at the matching delimiter.
- Add a Bash markup regression for `<<-EOF | cat`.
- Update the existing Bash heredoc expectation and changelog entry.

## Tests
- Red before fix: the reported sample highlighted `EOF | cat` inside the heredoc string span, so `cat` was not highlighted as a command.
- `ONLY_LANGUAGES=bash npm run test-markup`
- `npx eslint --no-eslintrc -c .eslintrc.lang.js src/languages/bash.js`
- `npm run build`
- `git diff --check`

Note: I also ran the full `npm run test-markup`; it reached 537 passing tests and then hit the existing 2s Mocha timeout on `http/default`, unrelated to the Bash fixture. The Bash markup suite passed in both focused runs.

### DIFF
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -50,16 +50,6 @@ export default function(hljs) {
       }
     }
   );
-  const HERE_DOC = {
-    begin: /<<-?\s*(?=\w+)/,
-    starts: { contains: [
-      hljs.END_SAME_AS_BEGIN({
-        begin: /(\w+)/,
-        end: /(\w+)/,
-        className: 'string'
-      })
-    ] }
-  };
   const QUOTE_STRING = {
     className: 'string',
     begin: /"/,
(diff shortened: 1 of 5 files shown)


src/languages/bash.js
```diff
@@ -368,27 +358,60 @@ export default function(hljs) {
...
+    keywords: BASH_KEYWORDS,
+    contains: [
+      ARITHMETIC,
+      COMMENT,
+      PATH_MODE,
+      QUOTE_STRING,
+      ESCAPED_QUOTE,
+      APOS_STRING,
+      ESCAPED_APOS,
+      VAR
```
joshgoebel, line 399: All of this is because of the pipe and piping the string into any sequence of bash, right?

src/languages/bash.js
```diff
@@ -368,27 +358,60 @@ export default function(hljs) {
...
+      "shopt",
+      ...ZSH_BUILT_INS,
+      ...GNU_CORE_UTILS
+    ]
+  };
+  let heredocDelimiter = "";
+  const HERE_DOC_BODY = {
+    className: 'string',
+    begin: /\n/,
+    end: /^\t*(\w+)$/m,
```
joshgoebel, line 379: What is the requirement for heredoc naming in Bash?  Could this be tightened up any further than `\w+`?

arturict: I think this needs broadening rather than tightening. Bash takes a `word` after `<<` and applies quote removal; it is not limited to `\w+`.

For example, this branch captures only `END` here and never finds the real terminator, so `echo done` stays inside the string:

```
cat <<END-OF | cat
hello
END-OF
echo done
```

`<<'EOF'` is valid too, but is not recognized as a heredoc by this matcher. Could we add both as markup cases before settling on the delimiter match?

src/languages/bash.js
arturict: Not only pipes. The same split is needed for any shell syntax after the delimiter word, for example `<<EOF >out` or `<<EOF && echo queued`. The body should start at the following newline; everything before that newline should remain normal Bash.

src/languages/bash.js
joshgoebel: Sure. Add all the cases we know of, then try to get the highlighted correctly - that's the way to do it.

arturict:
Nice rust escaping coverage. Could you include minimal unit case for escaped quotes/backslashes in char literals to aid review traceability?

Issue #4388, opened 2026-04-02
Localization of The Highlight.js Documentation


hwhsu1231:
## Announcement

Hello Highlight.js Community,

I am the author of the @localizethedocs organization. And I’m glad to announce that the 🎉 **highlightjs-docs-l10n** 🎉 project is published now:

- 🚀 **Preview:** [highlightjs-docs-l10n](https://projects.localizethedocs.org/highlightjs-docs-l10n)
- 🌐 **Crowdin:** [highlightjs-docs-l10n](https://localizethedocs.crowdin.com/highlightjs-docs-l10n)
- 🐙 **GitHub:** [highlightjs-docs-l10n](https://github.com/localizethedocs/highlightjs-docs-l10n)

The goal of this project is to translate The Highlight.js Documentation into multiple languages. Translations are contributed via the Crowdin platform, automatically synchronized with the GitHub repository, and can be previewed on GitHub Pages.

## How to Contribute Translations?

To contribute the translations to the project, just follow the following steps:

1. Create an account on [Crowdin Enterprise](https://localizethedocs.crowdin.com/u/signup) if you don't have one.
2. Log in and go the [highlightjs-docs-l10n](https://localizethedocs.crowdin.com/highlightjs-docs-l10n) project.
3. Choose the language you would like to contribute.

If translators want to translate a language that is not yet supported, they just need to open a new [issue](https://github.com/localizethedocs/highlightjs-docs-l10n/issues/new?template=language_request.yml) to request the new language. Once the requested language is added, they can begin translating.

## How to Keep Content Up-to-date?

An FAQ that is sure to come up in the Highlight.js community is: ***How to keep translations up-to-date?***

First and foremost, in my opinion, the focus should be on how to keep the **"Translatable Content (`msgid`)**" up-to-date, rather than the **"Translations (`msgstr`)**". I will focus on the most critical part of the whole infrastructure here.

In short, the [ci-sphinx-update-pot.yml](https://github.com/localizethedocs/highlightjs-docs-l10n/blob/main/.github/workflows/ci-sphinx-update-pot.yml) workflow will be executed weekly to check the upstream project for any required updates to the document content, which in turn triggers the update of the `.pot` files. If an update is needed, a Pull Request (PR) is automatically created to merge into the `l10n` branch. For example:

- https://github.com/localizethedocs/highlightjs-docs-l10n/pull/3
- https://github.com/localizethedocs/highlightjs-docs-l10n/pull/4

[...]

PR #4420, opened 2026-06-24
build(deps): bump python and sphinx requirements

hwhsu1231:
### Changes

- Bump Python version to 3.12 in `.readthedocs.yaml`.
- Bump Sphinx requirements in `docs/requirements.txt`.

### Checklist
- [x] Added markup tests, or they don't apply here because this is a documentation/infrastructure change only 

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
-myst-parser == 0.15.0
-sphinx-rtd-theme == 0.5.2
+sphinx == 9.1.0
+myst-parser == 5.1.0
+sphinx-rtd-theme == 3.1.0
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,6 +5,12 @@
 # Required
 version: 2
 
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-24.04
+  tools:
+    python: "3.12"
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
    configuration: docs/conf.py
@@ -15,6 +21,5 @@ sphinx:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-   version: 3.7
    install:
    - requirements: docs/requirements.txt
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,9 +4,14 @@ Core Grammars:
 
 - fix(groovy) support underscores in numeric literals [greymoth-jp][]
 
+Documentation:
+
+- build(deps): bump python and sphinx requirements [Haowei Hsu][]
+
 CONTRIBUTORS
 
 [greymoth-jp]: https://github.com/greymoth-jp
+[Haowei Hsu]: https://github.com/hwhsu1231
 
 
 ## Version 11.11.2


joshgoebel:
What was the why behind this change? Were there CVEs or issues or?

hwhsu1231:
Hello @joshgoebel 

> What was the why behind this change? Were there CVEs or issues or?

One of the main reasons, of course, is that the dependency versions specified in the previous [`requirements.txt`](https://github.com/highlightjs/highlight.js/blob/4cd65730547d595ee63ac5e38bc92b931f4f18b8/docs/requirements.txt) were simply too outdated! The previously pinned `myst-parser==0.15.0` and `sphinx-rtd-theme==0.5.2` restricted the installed Sphinx version to `<5`. Furthermore, newer versions of Sphinx offer faster build times and undoubtedly include more bug fixes.

In particular, I recently published the [highlightjs-docs-l10n](https://github.com/localizethedocs/highlightjs-docs-l10n) project, which is dedicated to maintaining the documentation translations (See the #4388 for more details). Since it utilizes the `gettext` builder, I wanted to install newer dependencies to minimize the potential for translation build errors caused by an outdated dependencies.

Besides that, I believe many have noticed that the syntax for specifying the Python version in the current [`readthedocs.yaml`](https://github.com/highlightjs/highlight.js/blob/4cd65730547d595ee63ac5e38bc92b931f4f18b8/.readthedocs.yaml#L17-L18) is no longer valid. This has been causing [build errors](https://app.readthedocs.org/projects/highlightjs/builds/33263702/) in RTD for quite some time, so I decided to correct the syntax while I was at it.

Finally, I plan to submit a new PR later to migrate the current highlight.js documentation theme from [`sphinx-rtd-theme`](https://github.com/readthedocs/sphinx_rtd_theme) to [`shibuya`](https://github.com/lepture/shibuya), which offers a built-in **"dark mode"**, and its **"three-column layout"** (global TOC, page content, local TOC) aligns perfectly with modern theme trends. Below is a preview of what it looks like after migrating to Shibuya:

Screenshot_20260706_025706
Screenshot_20260706_025656<br />



However, using the `shibuya` theme requires Sphinx version 7.x or higher. Therefore, I decided to completely upgrade the current documentation dependencies before submitting the PR for the Shibuya theme migration.

joshgoebel:
Makes sense. (staying current) I forget how our docs even work - this would only going to affect local doc builds - it won't affect (can't affect) our web builds, right?

Dunno about the theme change, but one thing at a time.

hwhsu1231:
> I forget how our docs even work - this would only going to affect local doc builds - it won't affect (can't affect) our web builds, right?

Since it installs `docs/requirements.txt` in [`.readthedocs.yaml`](https://github.com/highlightjs/highlight.js/blob/e384df94aa4a00a18457ce0e8f31b64b5482a911/.readthedocs.yaml#L20), it will affect not just the local doc build, but also the RTD build.

docs/requirements.txt
```diff
@@ -1,2 +1,3 @@
-myst-parser == 0.15.0
-sphinx-rtd-theme == 0.5.2
+sphinx == 9.1.0
+myst-parser == 5.1.0
+sphinx-rtd-theme == 3.1.0
```
joshgoebel, line 3: Are there any big changes from the theme from 0.5.2 to 3.1.0? That is a big jump

hwhsu1231: See its [changelog](https://sphinx-rtd-theme.readthedocs.io/en/stable/changelog.html) for more details.

joshgoebel: OK, We'll give it a shot.

joshgoebel (approved):

joshgoebel:
You'll need to rebase on main as it seems I don't have permission to push changes to your branch.

hwhsu1231:
cc: @joshgoebel

PR #4376, opened 2026-03-10
fix(bash): highlight command line options (--option)

KJyang-0114:
This PR adds highlighting for long command line options (--option) in bash to address issue #4288.

## Problem

Previously, command line options like `--project`, `--env`, `--deployment-package` were not highlighted consistently across lines in bash code blocks. Some would have `hljs-built_in` class while others had no class at all.

## Solution

Add an OPTION pattern that highlights long options when they appear after whitespace. This makes CLI examples more readable.

## Testing

Tested manually:
```bash
epi deployment start --project my-project --env Integration --direct-deploy true
```

Now outputs:
```
epi deployment start<span class="hljs-attribute"> --project</span> my-project<span class="hljs-attribute"> --env</span> Integration<span class="hljs-attribute"> --direct-deploy</span> <span class="hljs-literal">true</span>
```

## Test Updates

Updated 2 test expectation files to reflect the new behavior:
- `test/markup/bash/token-containing-keyword.expect.txt`
- `test/markup/shell/command-continuation.expect.txt`

All tests now pass (1570 passing).

## Fixes

Fixes #4288

PR #4367, opened 2026-03-03
fix(bash): highlight command line flags/options fixes #4288

KJyang-0114:
## Summary

This PR fixes issue #4288 where command line flags/options like `--project`, `--deployment-package`, etc. were not being consistently highlighted in bash code blocks.

## The Problem

The issue reported that in bash commands with multiple flags:
- Only some flags were highlighted as `hljs-built_in`
- Other flags had no highlighting at all
- This caused inconsistent colors between lines

## The Fix

Added a new `COMMAND_FLAGS` pattern that highlights:
- Long options: `--option-name`
- Short options: `-o`

These are highlighted with the `hljs-attr` class for better visibility.

## Testing

Before:
```
epi deployment start \
  --project my-project
```

After:
```
epi deployment start <span class="hljs-attr">--project</span> my-project
```

Closes #4288

PR #4421, opened 2026-06-27
fix(bash): Add consistent highlighting for command-line options

### LINKED ISSUE #4288: (Bash) Highlighted colors doesn't look correct

WaterXiao-git:
Add an OPTION mode to match short (-o) and long (--option=value) command-line flags, ensuring they are highlighted consistently across continuation lines instead of being partially matched by keyword or path patterns.

Closes #4288

### DIFF
--- a/src/languages/bash.js
+++ b/src/languages/bash.js
@@ -145,6 +145,14 @@ export default function(hljs) {
   // to consume paths to prevent keyword matches inside them
   const PATH_MODE = { match: /(\/[a-z._-]+)+/ };
 
+  // Match command-line options (short: -o, long: --option, --option=value)
+  // to ensure consistent highlighting across continuation lines.
+  const OPTION = {
+    className: 'attr',
+    begin: /(?:^|\s)--?\w[\w-]*(?:=?\S*)?\s*/,
+    relevance: 0
+  };
+
   // http://www.gnu.org/software/bash/manual/html_node/Shell-Builtin-Commands.html
   const SHELL_BUILT_INS = [
     "break",
@@ -397,6 +405,7 @@ export default function(hljs) {
       COMMENT,
       HERE_DOC,
       PATH_MODE,
+      OPTION,
       QUOTE_STRING,
       ESCAPED_QUOTE,
       APOS_STRING,


src/languages/bash.js
```diff
@@ -145,6 +145,14 @@ export default function(hljs) {
   // to consume paths to prevent keyword matches inside them
   const PATH_MODE = { match: /(\/[a-z._-]+)+/ };
 
+  // Match command-line options (short: -o, long: --option, --option=value)
+  // to ensure consistent highlighting across continuation lines.
+  const OPTION = {
+    className: 'attr',
+    begin: /(?:^|\s)--?\w[\w-]*(?:=?\S*)?\s*/,
```
joshgoebel, line 152: I'm not so sure about flags like

```
--mode=compact
```

Think perhaps the `=` should be operator (or nothing) and maybe `compact` should be a string?  Thoughts?

Might be nice to see this a few different ways. CC @allejo  Any thoughts?

arturict (commented):
I reproduced the current matcher against shell separators and redirections. There is a correctness issue beyond the open `=value` scoping question: `(?:=?\S*)?` is broader than an optional `=value` because `=` itself is optional.

For example, the mode consumes ` --foo; `, ` --foo|grep `, and ` --foo>out `, which would hide shell operators/redirections inside `hljs-attr`; the surrounding whitespace is scoped too. Please scope only the option token and let the existing modes parse the remainder. A multi-part match could preserve the leading boundary without assigning it the `attr` scope.

This grammar change also needs markup cases for `--mode=compact`, `--foo | grep`, `--foo>out`, quoted/variable values, and short options. Since #4376, #4367, and #4369 address the same issue, it may be worth consolidating around one tested approach.

joshgoebel:
Agree, more works and tests needed here.

arturict:
If this option highlighting change touches multiple grammars, maybe include a compact list of tested languages to reduce regression risk.

joshgoebel:
Embedding is typically just running the grammar on a subset of code, so any tests added here would also apply to anywhere this grammar is embedded in another grammar... but it does seem we need a beefier chunk of tests here and then try to come up with the minimal regex that will cover our test cases.

And false positive cases.  Keep piling on the false positive cases anyone.

src/languages/bash.js
allejo: - +1 to `compact` being a string (with or without quotes). I know in Bash, `OPTARG` is used alongside `getopts` to parse CLI options, and I'm _pretty_ sure it always comes in as a string, but I'm having a hard time finding a source that explicitly mentions the data type.
- I'm leaning towards treating `=` as an operator because in Bash, the `=` sign _is_ an operator in other syntaxes. It's technically _setting_ the value of a flag in the tool.

PR #4424, opened 2026-07-02
fix(parser): count named capture groups when rewriting backreferences

spokodev:
When several sibling rules of a mode are merged into the mode's single
multi-matcher, every numbered backreference is renumbered based on a running
count of capture groups. `_rewriteBackreferences` counts groups with
`BACKREF_RE`, whose `\(\??` branch skips a named capture group `(?<name>...)`.
So a named group in an earlier joined regex is not counted, the running total
drifts low, and every following `\1` or `\2` backreference resolves to the wrong
group.

For example a grammar with `{begin: /(?<n>@)\w+/}` before `{match: /(\w)\1/}`
highlights every character of `hello` instead of only the doubled `ll`, because
the `\1` is renumbered one short and matches the wrong group.

The fix recognizes `(?<name>` and `(?'name'` in `BACKREF_RE`, ordered before the
generic `(?` branch and guarded against the lookbehind forms `(?<=` and `(?<!`,
and counts them as captures. This matches `countMatchGroups`, which already
counts named groups through the real engine. Non-capturing, lookahead and
lookbehind groups are still not counted.

### DIFF
--- a/src/lib/regex.js
+++ b/src/lib/regex.js
@@ -101,14 +101,17 @@ export function startsWith(re, lexeme) {
   return match && match.index === 0;
 }
 
-// BACKREF_RE matches an open parenthesis or backreference. To avoid
-// an incorrect parse, it additionally matches the following:
-// - [...] elements, where the meaning of parentheses and escapes change
-// - other escape sequences, so we do not misparse escape sequences as
-//   interesting elements
-// - non-matching or lookahead parentheses, which do not capture. These
-//   follow the '(' with a '?'.
-const BACKREF_RE = /\[(?:[^\\\]]|\\.)*\]|\(\??|\\([1-9][0-9]*)|\\./;
+// BACKREF_RE matches an open parenthesis or backreference. To avoid an
+// incorrect parse, it also matches the constructs where the meaning of
+// parentheses, escapes, or capture counting changes.
+const BACKREF_RE = new RegExp(either(
+  /\[(?:[^\\\]]|\\.)*\]/, // a character class, inside which ( and \ lose their meaning
+  /\(\?<(?![=!])[^>]+>/, // a named capture group `(?<name>` (not a lookbehind `(?<=` / `(?<!`)
+  /\(\?'[^']+'/, // a named capture group `(?'name'`
+  /\(\??/, // an opening parenthesis, capturing or non-capturing / lookahead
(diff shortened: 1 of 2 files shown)


joshgoebel:
Which grammars are using named capture groups?  I think that iself may have slipped in unintentionally.

spokodev:
You are right that no bundled grammar uses them. I checked `src/languages` and there are zero named capture groups. The fix is not for a bundled grammar; it is for custom grammars registered through the public API, where `_rewriteBackreferences` miscounts a named group and renumbers a following `\1` backreference wrong. If named capture groups are out of scope for grammars, this can be closed. If custom grammars are meant to accept the full regex syntax, the counting fix keeps their backreferences correct. Your call.

src/lib/regex.js
```diff
@@ -108,7 +108,10 @@ export function startsWith(re, lexeme) {
 //   interesting elements
 // - non-matching or lookahead parentheses, which do not capture. These
 //   follow the '(' with a '?'.
-const BACKREF_RE = /\[(?:[^\\\]]|\\.)*\]|\(\??|\\([1-9][0-9]*)|\\./;
+// - named capture groups `(?<name>` and `(?'name'`, which do capture and so
+//   must be counted. The lookbehind forms `(?<=` and `(?<!` are excluded so
+//   they still fall through to the non-capturing branch below.
+const BACKREF_RE = /\[(?:[^\\\]]|\\.)*\]|\(\?<(?![=!])[^>]+>|\(\?'[^']+'|\(\??|\\([1-9][0-9]*)|\\./;
```
joshgoebel: Now that this has gotten more complex, can we use `regex.either` to make it easier to read?

joshgoebel:
>  it is for custom grammars registered through the public API,

Hmmm, I suppose that's a gap we could/should close. Ok, back to review :)

src/lib/regex.js
spokodev: Done. Rewrote BACKREF_RE with `either()`, one alternative per line with a comment. I checked the composed regex matches the old one exactly (same `match[0]`, index, and backreference group across the paren, lookbehind, named-group, backreference, and char-class cases); the parser and markup tests stay green.

joshgoebel:
Can you make linter happy please?

joshgoebel (approved):

PR #4431, opened 2026-07-06
fix(crystal): stop treating integer division // as a regex literal

### LINKED ISSUE #3730: (language name) Short description of issue...
**Describe the issue**
Using `//` in crystal will make the highlighter mark the rest of the file.
In crystal is `//` used for integer division.
Code:
```
def something()
    something // 4
end
```
image

Crystal

highlight

**Sample Code to Reproduce**
```
def something()
    something // 4
end
```
image

**Expected behavior**
The expected behavior is to not mark the rest of the code with the color green.

**Additional context**
Nope

zain-asif-dev:
Fixes #3730

## Summary of Changes

Crystal uses `//` (and `//=`) as its integer division operator, but the REGEXP detection heuristic in `src/languages/crystal.js` was borrowed from Ruby, where `//` has no special meaning and is instead treated as an empty regex literal.

This caused two related problems:

- Mid-expression, e.g. `something // 4`, the first `/` was consumed as a generic "regex starter" operator, leaving the second `/` to open an unterminated regex that swallowed the rest of the line (as reported in the issue), or even the rest of the file.
- At the start of a line/statement, e.g. `//` or `//=` on their own, the "empty regex" variant (`//[a-z]*`) matched `//` directly, again misclassifying the division operator as a regexp span. This was actually already baked into the `operators.expect.txt` test fixture as the (buggy) expected output.

This PR:

- Adds a dedicated `INTEGER_DIVISION` mode matching `//` and `//=`, given priority over the regex heuristic in the `contains` list, so it wins whenever the two would otherwise tie for the same starting position.
- Drops the ambiguous "empty regex" variant from the regexp heuristic, since Crystal doesn't actually support that syntax, and it's what let `//` seep through at the start of a line even where `INTEGER_DIVISION` wasn't a tied first alternative.
- Updates `test/markup/crystal/operators.expect.txt` to reflect the corrected (unstyled) output for `//` and `//=`.
- Adds a new `test/markup/crystal/integer-division.txt` regression test reproducing the exact snippet from the issue.

Regex literal detection after keywords and operators (`if /foo/`, `+/foo/`, etc., already covered by `test/markup/crystal/regexes.txt`) continues to work as before.

## Testing instructions

Ran the full test suite (`npx mocha`), all 1574 tests pass, plus `npx eslint --no-eslintrc -c .eslintrc.lang.js src/languages/crystal.js` with no warnings.

Manually verified before/after with the reproduction from the issue:

```crystal
def something()
    something // 4
end
```

Before: everything from the second `/` onward (`/ 4\nend`) was wrapped in `hljs-regexp`.
After: `//` is left unstyled (as an operator) and `4` is correctly highlighted as a number.

### DIFF
--- a/src/languages/crystal.js
+++ b/src/languages/crystal.js
@@ -159,6 +159,16 @@ export default function(hljs) {
         // a regex here would also make this variant swallow the rest of
         // the line (or file) whenever `//` appears at the start of a
         // statement, since there would be no closing delimiter to match.
+        //
+        // INTEGER_DIVISION (above, given priority over this REGEXP mode)
+        // already handles the common case where `//` starts matching at the
+        // same position as this mode, e.g. mid-expression like
+        // `something // 4`. But this outer REGEXP mode's `begin` can also
+        // start matching one character earlier, at a preceding newline
+        // (via its `\n` alternative), when `//`/`//=` is the first thing on
+        // a line; in that case INTEGER_DIVISION never gets a chance to
+        // compete, so this lookahead is what stops `//` from being
+        // misclassified as a regexp there too.
         begin: '/(?!\\/)',
         end: '/[a-z]*'
       }
--- a/src/languages/crystal.js
+++ b/src/languages/crystal.js
@@ -154,16 +154,13 @@ export default function(hljs) {
           hljs.BACKSLASH_ESCAPE,
           SUBST
         ],
-        variants: [
-          {
-            begin: '//[a-z]*',
-            relevance: 0
-          },
-          {
-            begin: '/(?!\\/)',
-            end: '/[a-z]*'
-          }
-        ]
+        // Unlike Ruby, Crystal has no "empty regex" (`//`) literal syntax,
+        // since `//` is already used for integer division. Treating it as
+        // a regex here would also make this variant swallow the rest of
+        // the line (or file) whenever `//` appears at the start of a
+        // statement, since there would be no closing delimiter to match.
+        begin: '/(?!\\/)',
+        end: '/[a-z]*'
       }
     ],
     relevance: 0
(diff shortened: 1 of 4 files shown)


src/languages/crystal.js
```diff
@@ -154,16 +154,13 @@ export default function(hljs) {
...
-            end: '/[a-z]*'
-          }
-        ]
+        // Unlike Ruby, Crystal has no "empty regex" (`//`) literal syntax,
+        // since `//` is already used for integer division. Treating it as
+        // a regex here would also make this variant swallow the rest of
+        // the line (or file) whenever `//` appears at the start of a
+        // statement, since there would be no closing delimiter to match.
+        begin: '/(?!\\/)',
+        end: '/[a-z]*'
```
joshgoebel, line 173: Do we still need the changes here now that we have the `//` mode in front?

zain-asif-dev: Yes, still needed — I verified this empirically by temporarily reverting just this lookahead (back to `begin: '/'`) and rebuilding (`npm run build`; the test suite runs against `build/`, not `src/` directly, so a plain `npx mocha test` without rebuilding first will silently test stale output).

With the lookahead removed, `should markup operators` fails because `//` and `//=` (each on their own line in `operators.txt`, i.e. immediately following a newline with nothing else on the line) get wrapped as `<span class="hljs-regexp">//</span>` and `<span class="hljs-regexp">//</span>=`:

```
AssertionError: expected '...===\n<span class="hljs-regexp">//</span>\n<span class="hljs-regexp">//</span>=\n&amp;+...' to be '...===\n//\n//=\n&amp;+...'
```

The reason `INTEGER_DIVISION`'s priority in `contains` doesn't fully cover this case: that ordering only wins ties when both modes' `begin` match at the *exact same* starting index (e.g. mid-expression like `something // 4`, where both `INTEGER_DIVISION` and the outer `REGEXP`'s `RE_STARTERS_RE` alternative for a lone `/` start at the same character). But `REGEXP`'s outer `begin` also has a `\n` alternative, which lets it start matching one character *earlier*, at the preceding newline itself, when `//`/`//=` is the first thing on a line. In that case the outer `REGEXP` wins outright (no tie to break by ordering), consumes the newline, and hands off to its inner `contains` starting right at the first `/` — where, without the `(?!\/)` lookahead, `begin: '/'` greedily matches that `/` and then `end: '/[a-z]*'` immediately matches the very next `/`, producing a (wrong) 2-character regexp span.

So `INTEGER_DIVISION` handles the "swallows the rest of the line/file" mid-expression danger, and this inner lookahead handles the "start of line" misclassification the issue also mentioned. Happy to add a comment above the lookahead explaining this interaction if that would help future readers — let me know and I'll push it.

zain-asif-dev: Went ahead and pushed the clarifying comment (96038a1) explaining the INTEGER_DIVISION/newline interaction, since it seemed useful regardless. Happy to adjust the wording if you'd put it differently.

joshgoebel (approved):

PR #4432, opened 2026-07-07
CI: Add node 24.x to ci and remove 21.x

paladox:
24.x is a lts release. 21.x was just a normal release, never a lts so can be dropped.

### DIFF
--- a/.github/workflows/tests.js.yml
+++ b/.github/workflows/tests.js.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         # All 'LTS' and 'Current' versions (https://nodejs.org/en/about/previous-releases)
-        node-version: [18.x, 20.x, 21.x, 22.x]
+        node-version: [18.x, 20.x, 22.x, 24.x]
         build-how: ["node", "browser", "browser -n", "cdn :common"]
 
     steps:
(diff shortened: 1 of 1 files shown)


joshgoebel:
> Both Node.js 18 and Node.js 20 have reached their official End-of-Life (EOL), meaning they are no longer supported by the [Node.js Foundation](https://nodejs.org/) and receive no further security patches.Node.js [1](https://www.herodevs.com/blog-posts/node-js-end-of-life-dates-you-should-be-aware-of)8: Reached EOL on April 30, 2025.Node.js 20: Reached EOL on April 30, 2026.

I say lets lose 18 and 20 as well...

arturict:
If this introduces CI matrix adjustments, a quick matrix summary in PR body helps triage runtime compatibility concerns.

joshgoebel:
Thanks for this — dropping 21 and adding 24 LTS looks good.

Could you also add **26.x** (current Current) so the matrix matches the comment about LTS + Current?

```yaml
node-version: [18.x, 20.x, 22.x, 24.x, 26.x]
```

We'll drop 18 later with v12.

joshgoebel:
Looks like 20 is still LTS, so no reason I can think of not to keep for now.

joshgoebel (approved):

PR #4436, opened 2026-07-12
fix(lisp): preserve highlighting after quoted multiplication

### LINKED ISSUE #3978: (lisp) stop coloring when «'(*» is found

When an expression starts with «'(\*» the coloring stops for the following lines until a «*» is found.

```lisp
(+ 6 7)   ; colored
'(* 6 7)  ; not colored
(+ 6 7)   ; not colored
(+ 6 7)   ; not colored
'(* 6 7)  ; 6 7 colored, star restarts coloring

(+ 6 7)   ; colored
'(* 6 7)  ; not colored
(+ 6 7)   ; not colored
(+ 6 7)   ; not colored
(* 6 7)   ; 6 7 colored, star restarts coloring

(* 6 7)   ; colored
'(* 6 7)  ; not colored
(+ 6 7)   ; not colored
(+ 6 7)   ; not colored
(+ *)     ; not colored, star restarts coloring

(* 6 7)   ; colored
'(* 6 7)  ; not colored
(+ 6 7)   ; not colored, * in comment restart coloring but not immediatly
(+ 6 7)   ; 6 7 colored
```

arturict:
## Summary

- replace the open-ended Lisp `VARIABLE` mode with a complete star-delimited variable match
- preserve highlighting after quoted multiplication expressions
- keep symbols such as `*lines*` visibly highlighted
- leave pure star runs such as `**` and `********` unscoped
- add markup coverage for the reported expression, the following line, a star-delimited variable, and rejected pure-star runs

## Root cause

The old mode started at every `*`. In `'(* 6 7)` the multiplication operator opened a region that did not end until a later `*`, swallowing the rest of the expression and following lines. The replacement matches only a non-empty, non-whitespace symbol body enclosed by stars. This preserves the intended `*lines*` variable highlighting without treating multiplication operators or pure star runs as variables.

Closes #3978.

## Validation

- full test suite: 1576 passing, 3 pending
- focused Lisp markup suite: 3 passing
- Lisp grammar ESLint check passed
- `git diff --check` passed

Tests run with supported Node 22.


test/markup/lisp/quoted-multiplication.expect.txt
```diff
@@ -0,0 +1,4 @@
+(<span class="hljs-name">+</span> <span class="hljs-number">6</span> <span class="hljs-number">7</span>)
+(<span class="hljs-name">defvar</span> *lines*)
```
joshgoebel: >  The mode is unnecessary because the existing Lisp identifier expression already recognizes symbols such as *lines*.

Looks like that is untrue, *lines* isn't highlighted here.

arturict: You're right. `IDENT` parses it, but doesn't give it a scope. I restored this as a single `variable` match for complete star-delimited symbols, so `*lines*` is highlighted without letting a lone multiplication `*` open a mode across lines. Added in 1cd6a4f8; the full suite still passes.

src/languages/lisp.js
```diff
@@ -35,8 +35,8 @@ export default function(hljs) {
     { relevance: 0 }
   );
   const VARIABLE = {
-    begin: '\\*',
-    end: '\\*'
+    scope: 'variable',
+    match: /\*(?:[^\s()*]+\*|\*+)/
```
joshgoebel: What exact patterns are we actively trying to match here?

This matches things like ************** which seems wrong.

arturict: Only star-delimited symbols with a non-empty, non-whitespace, non-parenthesis body, such as `*lines*`. The `|\*+` alternative was too broad; I removed it in 28ef7c13 and added markup coverage showing that `********` and `**` stay unscoped. The full suite still passes.

joshgoebel (approved):

PR #4438, opened 2026-07-14
fix(cmake): only highlight standalone numbers, not digits in identifiers

### LINKED ISSUE #4170: (cmake) Folders starting by a number highlighted incorrectly
**Describe the issue**
Folders starting by a number are highlighted incorrectly in CMake.

Exemple: 

image

Notice how the 3 at the beginning of 3rdparty is highlighted differently.

cmake

Ayu highlighting through mdbook... I don't know sorry.

**Sample Code to Reproduce**

```cmake
project(my_app)

# example of folder structure
add_subdirectory(3rdparty/libremidi)

add_executable(my_app src/main.cpp)

target_link_libraries(my_app PRIVATE libremidi)
```

**Expected behavior**
File names in CMake function calls are highlighted normally.

xianjianlf2:
## What

Fixes the CMake grammar so a digit at the start of an identifier is no longer highlighted as a number. For example, in:

```cmake
add_subdirectory(3rdparty/foo)
```

the leading `3` of `3rdparty` was being wrapped in a `hljs-number` span.

## Why

`src/languages/cmake.js` used `hljs.NUMBER_MODE`, whose `begin` (`\b\d+(\.\d+)?`) has only a **leading** word boundary, so it matched the `3` inside `3rdparty`. This matches the fix suggested by the maintainer in the issue: use a number mode with a trailing `\b` as well.

## How

Replaced `hljs.NUMBER_MODE` with an inline mode:

```js
{ className: 'number', begin: /\b\d+(\.\d+)?\b/, relevance: 0 }
```

## Testing

- Extended the `test/markup/cmake/` fixture with `add_subdirectory(3rdparty/foo)` and the expected markup where `3` is **not** wrapped in a number span.
- Confirmed the fixture fails before the change and passes after (`ONLY_LANGUAGES=cmake mocha test/markup`).
- Existing CMake version-number highlighting (`2.8.8`) is unchanged.
- Added a `CHANGES.md` entry.

Closes #4170


src/languages/cmake.js
```diff
@@ -56,7 +56,13 @@ export default function(hljs) {
       hljs.COMMENT(/#\[\[/, /]]/),
       hljs.HASH_COMMENT_MODE,
       hljs.QUOTE_STRING_MODE,
-      hljs.NUMBER_MODE
+      {
+        className: 'number',
```
joshgoebel:
```suggestion
        scope: 'number',
```

src/languages/cmake.js
```diff
@@ -56,7 +56,13 @@ export default function(hljs) {
...
+        // require a word boundary so digits that merely begin an
+        // identifier (e.g. the `3` in `3rdparty`) are not highlighted
+        begin: /\b\d+(\.\d+)?\b/,
+        relevance: 0
```
joshgoebel, line 62: Not necessary.

joshgoebel (changes requested):
Nice, looking good!

xianjianlf2:
Thanks for the review! Both addressed in 03f30a9 — switched to `scope` and dropped the comment.

joshgoebel (approved):

PR #4441, opened 2026-07-14
fix: remove obsolete Flix grammar

### LINKED ISSUE #4440: Flix overlap
During the screening of the available built-in grammars (#4439), I noticed that `Flix` exists as a built-in grammar but is referenced in supported languages as a 3rd party lib.

The author of the built-in grammar seems to be originally a `Flix` maintainer. The 3rd reference is also from `Flix` directly and seems more up-to-date.

My proposal: simply remove `flix.js` as a built-in language in `highlight.js`.

iamdhrv:
Resolves #4440

### Changes
- Remove the obsolete built-in Flix grammar.
- Remove its coupled markup and detection fixtures.
- Keep the supported-languages entry pointing to the maintained third-party package.

### Checklist
- [x] Markup tests do not apply because this removes the built-in grammar and its existing fixtures.
- [x] Updated the changelog at `CHANGES.md`.

Tests: `npm run build`, `npm test`.


Baltsat (commented):
I checked the removal path against current `main`. The built-in Flix source and coupled fixtures are still present, while `SUPPORTED_LANGUAGES.md` keeps the maintained third-party `highlightjs-flix` path. The build loader discovers core grammars from `src/languages/*.js` and third-party grammars separately, so the five-file deletion is scoped consistently.

The original 18 checks are green, but GitHub now reports this head as conflicting with `main`, so those runs do not prove the current base. I found no source-level blocker after a rebase; the remaining step is to resolve the conflict and rerun CI.

joshgoebel:
See issue, this would be a breaking change, so we can only do it when we're ready for v12.

iamdhrv:
Understood. Since removal is a breaking change and the project is not ready for v12, I will hold this PR for that release window rather than rebase it now.

PR #4442, opened 2026-07-15
fix(cpp) add word boundary to number literal regexes

### LINKED ISSUE #4231: (cpp) `1.f` is considered a floating-point literal in `a1.foo()`
**Describe the issue**
See title.

C++ (`cpp`)

`highlight`.

**Sample Code to Reproduce**
`a1.foo()`

**Expected behavior**
`a1` should be an identifier, not just `a`, followed by a `1.f` floating-point literal.

xianjianlf2:
Fixes #4231

### Changes

In C++ code like `a1.foo()`, the `NUMBER` regex matched starting mid-identifier: the `1.f` inside `a1.foo` was highlighted as a float literal (digits + `.` + suffix letter), producing broken highlighting for member access on identifiers that end in a digit.

This implements the fix suggested by @joshgoebel in the issue discussion: require a word boundary (`\b`) before the numeric literal alternatives so a match can never begin inside an identifier. Specifically, `\b` is added to:

- the decimal float variant that starts with digits (`1.5`, `1.`, `1.5e3`)
- the exponent-only variant (`1e5`)
- the hexadecimal float variant (`0x1p3`)

Notes:

- The leading-dot variants (`.5`, `0x.1p3`) are intentionally left untouched — `\b` before `.` would not match after whitespace and these forms cannot begin inside an identifier anyway.
- The integer variant already used `\b`, so this brings the float variants in line with it.
- Added markup test cases covering `a1.foo()` / `b2.size()` style member access on identifiers ending in a digit; the existing float-literal cases in the same file confirm regular literals still highlight.

### Checklist
- [x] Added markup tests (`test/markup/cpp/number-literals.txt`)


joshgoebel (approved):
Looks reasonable.

PR #4444, opened 2026-07-15
fix(cpp): highlight digit separators in defines

### LINKED ISSUE #4443: (C and C++) Defines treat digit separator in defines as string beginning
**Describe the issue**

```c
#define FOO 10'000
```

is valid code, using `'` as a digit separator. Highlight.js incorrectly treats it as a beginning of a string.

Image

I tested C and C++, both get it wrong.

No clue. I originally noticed the issue while sending code over Discord and confirmed it again with [your demo](https://highlightjs.org/demo). I did select the language manually in both cases.

**Expected behavior**

The `'` character should be treated as part of the number.

**Additional context**

Support for this separator has been around in C++ since C++14, but it's relatively new in C, added in C23.

iamdhrv:
### Changes

- run the existing C/C++ number modes before string modes inside preprocessor directives
- add markup regressions for apostrophe digit separators in `#define` values
- add the required changelog entry

This prevents literals such as `100'000` in a macro definition from treating the apostrophe as the start of a string while preserving normal preprocessor string handling.

Resolves #4443

### Validation

- `npm run build`
- `ONLY_LANGUAGES="c cpp" npm run test-markup` (`18 passed`)
- `npm test` (`1578 passed`, `3 pending`)
- ESLint on the changed C and C++ grammar files
- `git diff --check`

### Checklist

- [x] Added markup tests for both affected languages

Development note: implemented with Codex assistance and validated locally with the full Node.js test suite.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 Core Grammars:
 
 - enh(dos) add `batch` as an alias, issue #4395 [Hashim Khan][]
+- fix(c, cpp) recognize digit separators in preprocessor definitions, issue #4443 [Dhruv Maniya][]
 - fix(lisp) preserve highlighting after quoted multiplication expressions [arturict][]
 - fix(rust) recognize `\\` and `\"` char-literal escapes so highlighting doesn't leak, issue #4351 [Sarath Francis][]
 - fix(cmake) only highlight standalone numbers, not digits that begin an identifier (e.g. `3rdparty`), issue #4170 [MarkXian][]
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -119,6 +119,7 @@ export default function(hljs) {
         begin: /\\\n/,
         relevance: 0
       },
+      NUMBERS,
       hljs.inherit(STRINGS, { className: 'string' }),
       {
         className: 'string',
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -106,27 +106,54 @@ export default function(hljs) {
   ],
     relevance: 0
   };  
+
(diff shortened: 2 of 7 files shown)


joshgoebel:
We either need more parsing examples or we'll need to lock this down to just #define if that's the intended scope - not all the other pre-processor directives.

Baltsat (commented):
I checked the exact head (`854c6ed`) with the full test and lint suite. The new `NUMBERS` entry applies to every `PREPROCESSOR`, so `#if`, `#elif`, and `#line` now highlight separator literals too, while the fixtures still cover only `#define`.

There is also a continued-definition case that currently breaks in both grammars:

```cpp
#define LIMIT 10'\
000
```

This renders `10` as a number and the remainder as a string. After phase-2 line splicing, the logical token is `10'000`, so it should remain one number token.

Could you add a regression for this form and make the numeric mode survive the continuation? Please also either add examples for the intended non-`#define` directives or confine the behavior to `#define`, as requested in the earlier scope review.

iamdhrv:
Thanks @Baltsat. The current head adds the continued integer-literal fixture for C and C++, and confines numeric highlighting to #define; the #if fixture remains unhighlighted. The fix is in 5fef03a.

iamdhrv:
Rebased onto the current main and addressed the review scope: numeric literal highlighting is limited to `#define`; `#if` literals remain unhighlighted. Added continued-separator regressions for C/C++, preserved the current `#include`/`#embed` handling, and updated the affected fixtures. CI is running on `46c9aeec`.

PR #4446, opened 2026-07-18
fix(c): only match real atomic_* type names, not atomic functions

### LINKED ISSUE #3837: (c) `atomic_init`, `atomic_xyz`, etc. are all treated as types, but shouldn't be
**Describe the issue**
This rule is overly aggressive:
https://github.com/highlightjs/highlight.js/blob/5c1bc1e4ab0b4f351d0a2cf8a627886472351327/src/languages/c.js#L24-L31
As a result, atomic operations such as `atomic_store`, `atomic_load`, `atomic_init`, etc. are also highlighted as types. Instead of a general rule such as `atomic_` being a type, 

`c`

This is just based off of observations on Discord, which, to my knowledge uses highlight.js. Presumably `highlight`.

**Sample Code to Reproduce**
```c
atomic_init()
atomic_load()
atomic_store()
```

**Expected behavior**
Only actual typedef-names such as `atomic_int` should be highlighted as types. Functions should not be highlighted as types.

**Additional context**
A list of types and functions can be found here https://en.cppreference.com/w/c/atomic. To fix this, it would not be sufficient to add negative lookahead to the pattern (to exclude function calls), since it is reasonable to write code such as:
```cpp
// declare variable
auto x = std::atomic_int();
// type alias for a function returning atomic_int
using atomic_maker = atomic_int();
```
To avoid false positives, it would be better to add special cases for all the known `atomic` type aliases (there aren't that many anyway).

xianjianlf2:
## Changes

Resolves #3837.

The C grammar's `atomic_[a-z]{0,6}[kmgtptme]?` wildcard rule was matching atomic *function* names (`atomic_init`, `atomic_load`, `atomic_store`, …) as types. Per the issue discussion, this replaces the wildcard with an explicit whitelist of the real `<stdatomic.h>` type names (`atomic_bool`, `atomic_int`, `atomic_char32_t`, etc.), so functions like `atomic_init(...)` are no longer mis-highlighted as types while all genuine atomic types keep their `type` scope.

New markup test `test/markup/c/atomic-types` covers both sides: every whitelisted type highlights as `type`, and atomic function calls do not.

## Checklist

- [x] Added markup test (`test/markup/c/atomic-types.txt` / `.expect.txt`)
- [x] Updated CHANGES.md
- [x] `npm run build` + `mocha test/markup` — 546 passing


src/languages/c.js
```diff
@@ -21,11 +21,24 @@ export default function(hljs) {
...
+  // atomic_fetch_add, ...) are not mistakenly highlighted as types. See #3837.
+  const ATOMIC_TYPES = regex.concat(/\batomic_/, regex.either(
+    'bool', 'char', 'schar', 'uchar', 'short', 'ushort', 'int', 'uint',
+    'long', 'ulong', 'llong', 'ullong',
+    'char16_t', 'char32_t', 'wchar_t',
+    'int_least8_t', 'uint_least8_t', 'int_least16_t', 'uint_least16_t',
+    'int_least32_t', 'uint_least32_t', 'int_least64_t', 'uint_least64_t',
+    'int_fast8_t', 'uint_fast8_t', 'int_fast16_t', 'uint_fast16_t',
+    'int_fast32_t', 'uint_fast32_t', 'int_fast64_t', 'uint_fast64_t',
+    'intptr_t', 'uintptr_t', 'size_t', 'ptrdiff_t', 'intmax_t', 'uintmax_t'
```
joshgoebel: One item per line please (easier review/diffs in the future).  Otherwise looks great, thanks!

joshgoebel (changes requested):

src/languages/c.js
xianjianlf2: Done, split the atomic type list to one item per line.

joshgoebel (approved):

PR #4447, opened 2026-07-18
fix(csharp) support digit separators in binary literals and numeric type suffixes

### LINKED ISSUE #4258: (C#) Support digit separators
**Describe the issue**
Digit separators are not supported: if you write ``20_000_000`` for example, ``20`` will be number-colored, while ``_000_000`` will be treated as invalid.

C#

``highlight``

**Sample Code to Reproduce**
```csharp
int i = 20_000_000;
```

Screenshot of broken syntax highlighting

**Expected behavior**
the entirety of the number to be highlighted as a number, like on GitHub:
Screenshot of correct syntax highlighting

**Additional context**
None

sarathfrancis90:
Resolves #4258 — the digit-separator work in #4280 covered decimal and hex literals, but a few C# number forms are still highlighted wrong.

### Changes

The number mode had kept some C/C++ habits: `'` as the digit separator, `_` allowed only in the decimal/hex variant, and the `u/U/l/L` suffixes recognised only after a plain decimal literal. In practice that means:

```csharp
int bin = 0b1010_1010;      // stops highlighting at the first _
ulong hex = 0xFF_FFUL;      // UL falls out of the literal
ulong dec = 1_000_000UL;    // same
decimal price = 19.99m;     // m/M and d/D suffixes aren't matched at all
int _count = 0;             // the leading _ renders as a number
```

The last one is the one I noticed first — `[\d_]+` can start with an underscore, so `_count` (the usual private-field convention) gets a stray highlighted `_`, and `point._1` highlights `._1`.

I rewrote the three variants to follow the lexical grammar in the spec ([integer literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types), [real literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types)): separators sit between digits and may also follow the `0x`/`0b` prefix, integer literals take `U`/`L` in either order, and real literals take `F`/`D`/`M`.

Tested with a new markup fixture covering the forms above plus the two identifier cases; `npm run build && npm test` is green (1579 passing), as are `npm run lint` and `npm run lint-languages`. The existing `csharp/floats` fixture is unchanged.

### Checklist


test/markup/csharp/numbers.txt
```diff
@@ -0,0 +1,10 @@
+int bin = 0b1010_1010;
+int binSeparatorFirst = 0b_1010_1010;
+ulong hex = 0xFF_FFUL;
+uint hexSuffix = 0xFFu;
+ulong decSuffix = 1_000_000UL;
+decimal price = 19.99m;
+double ratio = 3.14d;
+decimal big = 1.5e10M;
+int _count = 0;
+int index = point._1;
```
joshgoebel, line 11: If this is a false positive lets add a comment making that explicit.

joshgoebel (changes requested):
One request. Looks great.

sarathfrancis90:
Yes — both of those highlighted as numbers before the fix, since the old `[\d_]+` alternative matched without requiring a leading digit, so `_count` and the `._1` member both came out as NUMBER. Added a line above them saying so.

joshgoebel (approved):

PR #4449, opened 2026-07-21
Fix C++ auto-detection for trailing return types

### LINKED ISSUE #3673: C++ : Trailing return types in function declaration result in illegal language match

`hljs.highlightAuto` does not work on the following example, while `hljs.highlightElement` does.

```js
hljs.highlightAuto("auto main() -> int {\n    auto i = 42; // test\n    return i;\n}", [ 'cpp' ]) 
```

```
// result :
{
  "language": "cpp",
  "value": "auto main() -&gt; int {\n    auto i = 42; // test\n    return i;\n}",
  "illegal": true,
  "relevance": 0,
  "_illegalBy": {
    "message": "Illegal lexeme \"-\" for mode \"function\"",
    "index": 11,
    "context": "auto main() -> int {\n    auto i = 42; // test\n    return i;\n}",
    "mode": {},
    "resultSoFar": ""
  },
 ...
}
```

This issue seems to be related to the arrow `->` (feature name : `trailing return type`), which is legal in post-modern C++.

See the documentation of function declaration, section `(2)` (C++11), here on **cppreference** : 

- https://en.cppreference.com/w/cpp/language/function

**Expected behavior**

- Consistency between `highlightAuto` and `highlightElement` behaviors
- Proper modern and post-modern C++ synthaxes match & detection

AysajanE:
Resolves #3673

### Changes

- allow the C++ function-declaration mode to accept a trailing-return arrow without accepting standalone hyphens
- add positive and negative API regression tests for the narrowed grammar behavior
- update the current changelog

### Checklist

- [x] Markup tests do not apply because this is an auto-detection regression; API regression tests cover both the reported behavior and the preserved illegal-hyphen behavior.

---
AI assistance was used. This change was reviewed by Northset, and I accept responsibility for this submission.

### Verification

[Northset proof-of-pass receipt M-1054](https://northset-oss.github.io/verification-pilot/receipts/M-1054/)  
Contributor self-run; not maintainer verification.

### DIFF
--- a/src/languages/cpp.js
+++ b/src/languages/cpp.js
@@ -478,7 +478,7 @@ export default function(hljs) {
     end: /[{;=]/,
     excludeEnd: true,
     keywords: CPP_KEYWORDS,
-    illegal: /[^\w\s\*&:<>.]/,
+    illegal: /(?!->)[^\w\s\*&:<>.]/,
     contains: [
       { // to prevent it from being confused as the function title
         begin: DECLTYPE_AUTO_RE,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 Core Grammars:
 
+- fix(cpp) recognize trailing return types in function declarations, issue #3673 [Aysajan Eziz][]
 - fix(lisp) preserve highlighting after quoted multiplication expressions [arturict][]
 - fix(rust) recognize `\\` and `\"` char-literal escapes so highlighting doesn't leak, issue #4351 [Sarath Francis][]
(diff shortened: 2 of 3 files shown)


src/languages/cpp.js
```diff
@@ -478,7 +478,7 @@ export default function(hljs) {
     end: /[{;=]/,
     excludeEnd: true,
     keywords: CPP_KEYWORDS,
-    illegal: /[^\w\s\*&:<>.]/,
+    illegal: /(?!->)[^\w\s\*&:<>.]/,
```
joshgoebel, line 481: Auto-detect is deprecated and will be removed from core in v12.  It was never very good.  So we aren't accepting changes or improvements to auto-correct related functionality like `illegal`, etc.

PR #4455, opened 2026-07-26
fix(java): support qualified generic type arguments

### LINKED ISSUE #3677: (java) Highlighting breaks on Generics with fully-qualified-name
**Describe the issue**
When the type inside of a generic (i.e. inside of `<>`) includes `.` as part of a fully qualified class name, the highlighter no longer seems to parse the rest of the line.

Java

hljs.highlightAll() with class="language-java" on the code-element
...

**Sample Code to Reproduce**
```html
<pre><code class="language-java">
interface FooBar {
  List&lt;String&gt; foo();
  List&lt;java.lang.String&gt; bar();
}
</code></pre>
```
http://jsfiddle.net/zwg2sp4x/

Only "foo" is highlighted as a function name, "bar" is not.

**Expected behavior**
both "foo" and "bar" should be highlighted as a function name.

**Additional context**
Tested with versions 11.7.0 and 10.5.0 with same behavior on each

Baltsat:
### Changes

The Java grammar accepted only one identifier inside a generic type argument, so a declaration such as `List<java.lang.String> bar()` was not recognized as a method.

This change:

- permits dot-separated qualified names inside Java generic type arguments;
- adds the exact `List<java.lang.String>` regression to the existing Java generics fixture; and
- records the fix in `CHANGES.md`.

The behavior introduced by #4346 remains covered: all 11 Java markup fixtures pass, and top-level qualified declarations are unchanged.

Fixes #3677

### Verification

- `npm run build`
- `ONLY_LANGUAGES=java npm run test-markup` — 11 passing
- Java language ESLint — passed
- `npm run lint-languages` — passed
- `npm run lint` — passed with one existing ignored-vendor warning
- `npm test` — 1,583 passing, 3 pending
- `git diff --check` — passed
- Red→green: the added fixture failed on unpatched main with 10 passing and 1 failing, then passed after the grammar change.
- Revert proof: reverting only `src/languages/java.js` reproduced the focused failure; restoring it returned all 11 Java fixtures to green.

A fresh `npm ci` was unavailable because registry DNS failed in the isolated Linux environment. Verification reused the exact unchanged lockfile dependency tree plus the official lock-resolved Rollup Linux package; `npm ls --all` reported 782 nodes and no problems.

### Checklist

- [x] Added markup tests
- [x] Updated `CHANGES.md`

### Assistance disclosure

This patch and its verification were prepared with OpenAI Codex assistance. The final diff and test outputs were reviewed before publication.

### DIFF
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -38,7 +38,7 @@ export default function(hljs) {
 
   // A simple Java type: a type name, optionally followed by type arguments and/or array brackets
   // '<@@@>' is replaced with the pattern for optional type arguments by recurRegex below.
-  const SIMPLE_TYPE_RE = JAVA_IDENT_RE + '<@@@>' + ARRAY_BRACKETS_OPTIONAL_RE;
+  const SIMPLE_TYPE_RE = JAVA_IDENT_RE + '(?:\\.' + JAVA_IDENT_RE + ')*<@@@>' + ARRAY_BRACKETS_OPTIONAL_RE;
 
   // A bounded (? extends Number) or unbounded (?) wildcard type
   const WILDCARD_TYPE_RE = '\\?(?:\\s+(?:extends|super)\\s+' + SIMPLE_TYPE_RE + ')?';
--- a/src/languages/java.js
+++ b/src/languages/java.js
@@ -36,9 +36,9 @@ export default function(hljs) {
   // Optional 1..n pairs of square brackets identifying an array type
   const ARRAY_BRACKETS_OPTIONAL_RE = '(?:(?:\\s*\\[\\s*])+)?';
 
-  // A simple Java type: a type name, optionally followed by type arguments and/or array brackets
+  // A simple Java type: a type name (optionally package-qualified), optionally followed by type arguments and/or array brackets
   // '<@@@>' is replaced with the pattern for optional type arguments by recurRegex below.
-  const SIMPLE_TYPE_RE = JAVA_IDENT_RE + '<@@@>' + ARRAY_BRACKETS_OPTIONAL_RE;
+  const SIMPLE_TYPE_RE = JAVA_IDENT_RE + '(?:\\.' + JAVA_IDENT_RE + ')*<@@@>' + ARRAY_BRACKETS_OPTIONAL_RE;
 
   // A bounded (? extends Number) or unbounded (?) wildcard type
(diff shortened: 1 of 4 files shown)


joshgoebel (commented):
Optional nits only — LGTM overall.

test/markup/java/generics.txt
```diff
@@ -12,3 +12,8 @@ public interface Map<K, V> {
   void put(K key, V value);
   V get(K key);
 }
+
+interface FooBar {
+  List<String> foo();
+  List<java.lang.String> bar();
+}
```
joshgoebel, line 21: Optional: one more case covering FQNs inside nested/wildcard type args would lock the `TYPE_ARG_RE` path a bit harder, e.g.:

```suggestion
interface FooBar {
  List<String> foo();
  List<java.lang.String> bar();
  Map<java.lang.String, java.util.List<String>> nested();
  List<? extends java.lang.Number> bounded();
}
```

(and matching lines in `generics.expect.txt` after a rebuild/`ONLY_LANGUAGES=java npm run test-markup`.)

src/languages/java.js
```diff
@@ -38,7 +38,7 @@ export default function(hljs) {
 
   // A simple Java type: a type name, optionally followed by type arguments and/or array brackets
   // '<@@@>' is replaced with the pattern for optional type arguments by recurRegex below.
-  const SIMPLE_TYPE_RE = JAVA_IDENT_RE + '<@@@>' + ARRAY_BRACKETS_OPTIONAL_RE;
+  const SIMPLE_TYPE_RE = JAVA_IDENT_RE + '(?:\\.' + JAVA_IDENT_RE + ')*<@@@>' + ARRAY_BRACKETS_OPTIONAL_RE;
```
joshgoebel: Nit: worth updating the comment so it mentions qualified names:

```suggestion
  // A simple Java type: a type name (optionally package-qualified), optionally followed by type arguments and/or array brackets
  // '<@@@>' is replaced with the pattern for optional type arguments by recurRegex below.
  const SIMPLE_TYPE_RE = JAVA_IDENT_RE + '(?:\.' + JAVA_IDENT_RE + ')*<@@@>' + ARRAY_BRACKETS_OPTIONAL_RE;
```

joshgoebel (approved):

PR #4456, opened 2026-07-28
fix(c, cpp): keep raw string delimiters from swallowing quotes

### LINKED ISSUE #3585: (C++) Incorrect highlight of raw string literals
**Incorrect highlight of raw string literals**

A raw string literal will cause any code following it to be highlighted as a string.

屏幕截图(18)

**Expected behavior**

```c++
// just to show the error
vector<string> s({R"(JKhf2n"v; 0>qwv>q " JKhf2n"l0 "v42HHf&g )","4U \"N *+ "});
int other_code;  // just to show the error
```

davidpavlovschi:
## Summary

- exclude quotes from the delimiter captured by the C/C++ raw-string rules, following the workaround proposed in #3585
- stop a false delimiter match from making the raw string consume the rest of the document
- preserve the existing behavior for unterminated raw strings
- add matching C and C++ regression fixtures for empty, named, and delimiter-like content cases

The closing rule currently allows its greedy delimiter capture to consume a quote. `END_SAME_AS_BEGIN` then rejects that longer capture, but cannot retry the correct shorter delimiter at the same position, so highlighting can remain inside the string until EOF.

## Validation

- current head: focused C/C++ markup suite, 21 passing
- current head: `npm run lint-languages` passes
- prior production head: full suite 1,566 passing, 3 pending, 1 pre-existing jsdom setup timeout (the unmodified base reports 1,565/3/1)
- differential fuzzing: 1,683 generated valid cases against a reference C++ raw-string scanner; 810 mismatches on `main`, 0 with this patch
- the existing truncated raw-string fixture still highlights to EOF

## Trade-off

C++ technically permits a quote inside the delimiter itself. As discussed in #3585, this patch deliberately stops recognizing that pathological form as a raw string; it degrades locally to ordinary string highlighting instead of corrupting everything after it.

Fixes #3585.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Core Grammars:
 - fix(csharp) support digit separators in binary literals and numeric type suffixes, and stop highlighting the leading `_` of an identifier, issue #4258 [Sarath Francis][]
 - fix(haskell) highlight `where` in GADT and closed type-family declarations, issue #3753 [Konstantin Baltsat][]
 - fix(css) support six-digit `unicode-range` values [Konstantin Baltsat][]
+- fix(c, cpp) stop a raw string's closing delimiter from swallowing quotes, which broke highlighting of everything after the literal, issue #3585 [David Pavlovschii][]
 - enh(python) add missing builtins: `aiter` and `anext` (Python 3.10), `frozendict` and `sentinel` (Python 3.15) [Hugo van Kemenade][]
 - enh(python) Support t-strings [Nicolas Le Cam][]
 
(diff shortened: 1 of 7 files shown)


Baltsat (commented):
Tested this exact head on Node 22: the build and language lint pass, focused C/C++ markup tests are 20/20, and the full suite is 1,584 passing / 3 pending. Keeping the new fixture while reverting only the production regex reproduces the leak (19 passing / 1 failing); restoring the patch returns 20/20.

One coverage gap: both `c.js` and `cpp.js` change, but the committed regression fixture exercises only C++. I reproduced the same leak/fix through the C grammar as well. Could you add the equivalent C fixture, or explain why the C grammar change does not need coverage?

I also confirmed the documented compatibility tradeoff: the C++ grammar permits `"` in a raw-string delimiter, and GCC 15 accepts `R"a"b(payload)a"b"`. On this patch that legal edge case degrades to ordinary quoted spans, while following code remains correctly highlighted. So the tradeoff is bounded, but real.

joshgoebel (approved):

Issue #3298, opened 2021-08-01
(DOS) Extra line breaks in comments


deniskovalchuk:
**Describe the issue**
Extra line breaks are added to comments if your DOS code uses Windows-style
line endings (`CRLF`). The problem is not reproduced if Unix-style line endings
are used (`LF`).

I'm not aware of the internal details of highlight.js, but I can assume a
problem with using `\s` in the `COMMENT` regular expression.
https://github.com/highlightjs/highlight.js/blob/4fb68add8dd4d8273096d24ad64f09d0aac2e58e/src/languages/dos.js#L11

**Which language seems to have the issue?**
DOS

**Are you using `highlight` or `highlightAuto`?**
`highlightElement`

**Sample Code to Reproduce**
Run the samples below and see console output.

Windows-style line endings (bad):
https://jsfiddle.net/ua3om8fb/1/

Unix-style line endings (good):
https://jsfiddle.net/8xw4vq39/3/

**Expected behavior**
![expected](https://user-images.githubusercontent.com/15797194/127786115-3cdc8d62-d441-4207-ab5b-b74aa6cc5700.png)
```html
<span class="hljs-comment">rem line1</span>
<span class="hljs-comment">rem line2</span>
```

**Actual behavior**
![actual](https://user-images.githubusercontent.com/15797194/127786124-81cc7de2-5b5e-4ac4-8af1-01aef4273b42.png)
```html
<span class="hljs-comment">rem line1</span>
<span class="hljs-comment">
rem line2</span>
```

joshgoebel:
Please confirm which web browser or browsers you are seeing this behavior with.  ￼

joshgoebel:
Nevermind, this is a problem with the engine itself (not browser/DOM).  It can be reproduced with `highlight` easily:

```
const s = "rem line1\r\nrem line2"
r = hljs.highlight(s, {language: "dos"})

// a dance just to make things visible
console.log(r.value.replace("\r","\\r").replace("\n","\\n"));
```

I'm not sure it's correct to blame the grammar though... this is quite easy to do using regex and it wouldn't surprise me if there were other examples in other grammars... I think a fix might have to be higher-level (and broader).

joshgoebel:
@highlightjs/core Ever run into this before?

[...]

PR #4457, opened 2026-08-02
fix(yaml) empty lines should not terminate block scalars

### LINKED ISSUE #4090: (YAML) Multiline strings don't support empty lines

pikammmmm:
Fixes #4090.

## The problem

A YAML block scalar stops being highlighted at the first empty line, and — worse — the remainder of the block is then re-parsed as YAML, so string content turns into keys.

Using the sample from the issue:

```yaml
foo:
  bar: |
    still: a string

    not: anymore
```

`not: anymore` is plain text inside the string, but it is emitted as `hljs-attr` (a key). On a real-world file (`chroma`'s `.github/ISSUE_TEMPLATE/bug_report.yaml`) the paragraph after the blank line is broken up into individual words, each wrapped in its own `hljs-string` span.

## Cause

The multi-line string mode has only a `begin` regex and no `end`, so the entire highlighted span is whatever `begin` matches. Continuation lines had to match:

```
\2[^\n]+
```

that is, the block indent followed by **at least one more character**. An empty line has nothing after the indent, so it can never match and the block ends there.

This is observable: a "blank" line containing *more* spaces than the block indent works today, while a truly empty line, a shorter one, or one exactly as wide as the indent all terminate the block early.

## The change

Empty lines (and lines containing only spaces) are now accepted as block content, both immediately after the header and between content lines:

```diff
-begin: '[\\|>]([1-9]?[+-])?[ ]*\\n( +)[^ ][^\\n]*\\n(\\2[^\\n]+\\n?)*'
+begin: '[\\|>]([1-9]?[+-])?[ ]*\\n(?:[ ]*\\n)*( +)[^ ][^\\n]*\\n((?:[ ]*\\n)*\\2[^\\n]+\\n?)*'
```

Two things were deliberate:

- **Blank lines are only consumed when another indented line follows.** A blank line that separates the block from the next key is left outside the span, so that key still highlights as a key. The new fixture asserts this negative case explicitly (`trailing-blank`).
- **Both added groups are non-capturing**, so the `\2` backreference still refers to the indent. The capture count is unchanged at 3.

Because both additions are optional (`*`), the new pattern is a strict superset of the old one — it can only ever extend a match, never shrink or relocate one.

## Testing

New fixture `test/markup/yaml/block_empty_lines.txt` covers the issue's own sample, a folded `>` scalar, a blank line right after the header, a spaces-only blank line, consecutive blank lines, and the trailing-blank guard described above.

- Full suite with the change: **1594 passing, 0 failing**
- Full suite with the change reverted: **1593 passing, 1 failing** — the sole failure is the new fixture, confirming it actually covers the bug
- Differential check over 400 real-world `.yaml`/`.yml` files on disk, old build vs new: the underlying text is byte-identical in every file (markup-only changes), 37 files differ, and in all 37 the string coverage grew — no file lost coverage and no file gained a spurious key

## Known limitations (pre-existing, not addressed here)

These were already broken before this change and are unchanged by it:

- **CRLF line endings** — `[ ]*\n` cannot cross the `\r`, so block scalars in CRLF files remain broken.
- **Blank lines containing a tab** still terminate the block.

Both would need separate work; I kept this patch to the empty-line case in the issue.

### DIFF
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -133,29 +133,20 @@ export default function(hljs) {
       relevance: 10
     },
     { // multi line string
-      // Blocks start with a | or > followed by a newline
-      //
-      // Indentation of subsequent lines must be the same to
-      // be considered part of the block
-      //
-      // Empty lines (and lines containing only spaces) do not end the block
-      // when another indented line follows. Trailing blanks stay outside
-      // so the next key is still a key.
-      //
-      // WHY: extra blank-line groups are non-capturing so \2 stays the
-      // indent group `( +)`. If they captured, indent would be \3 and
-      // \2 could match empty, so the block would swallow following keys.
+      // WHY: blank-line pieces stay non-capturing so \2 is the indent
+      // group `( +)`. Capturing them would make indent \3 and \2 empty,
+      // so the block would swallow following keys.
       scope: 'string',
       begin: regex.concat(
-        /[\|>]/,
-        /([1-9]?[+-])?/,
-        /[ ]*\n/,
-        regex.anyNumberOfTimes(/[ ]*\n/),
-        /( +)/,
-        /[^ ][^\n]*\n/,
-        regex.anyNumberOfTimes(
-          regex.concat(regex.anyNumberOfTimes(/[ ]*\n/), /\2[^\n]+\n?/)
-        )
+        /[\|>]/, // | or > indicator
+        /([1-9]?[+-])?/, // optional chomp / indent indicator
+        /(?:[ ]*\n)+/, // header newline plus leading blank lines
+        /( +)/, // indent of the first content line (\2)
+        /[^ ][^\n]*\n/, // first non-blank content line
+        regex.anyNumberOfTimes(regex.concat(
+          regex.anyNumberOfTimes(/[ ]*\n/), // blank lines between content
+          /\2[^\n]+\n?/ // same indent; final newline optional
+        ))
       )
     },
     { // Ruby/Rails erb
--- a/src/languages/yaml.js
+++ b/src/languages/yaml.js
@@ -136,8 +136,12 @@ export default function(hljs) {
       //
       // Indentation of subsequent lines must be the same to
       // be considered part of the block
+      //
+      // Empty lines (and lines containing only spaces) do not end the block,
+      // they are part of its content - but only when another indented line
+      // follows, so that blank lines trailing the block are left alone.
       className: 'string',
-      begin: '[\\|>]([1-9]?[+-])?[ ]*\\n( +)[^ ][^\\n]*\\n(\\2[^\\n]+\\n?)*'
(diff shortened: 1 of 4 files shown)


joshgoebel:
> CRLF line endings — [ ]*\n cannot cross the \r, so block scalars in CRLF files remain broken.

What is the impact here with regard to https://github.com/highlightjs/highlight.js/issues/3298?

Will that help?

src/languages/yaml.js
```diff
@@ -136,8 +136,12 @@ export default function(hljs) {
       //
       // Indentation of subsequent lines must be the same to
       // be considered part of the block
+      //
+      // Empty lines (and lines containing only spaces) do not end the block,
+      // they are part of its content - but only when another indented line
+      // follows, so that blank lines trailing the block are left alone.
       className: 'string',
-      begin: '[\\|>]([1-9]?[+-])?[ ]*\\n( +)[^ ][^\\n]*\\n(\\2[^\\n]+\\n?)*'
+      begin: '[\\|>]([1-9]?[+-])?[ ]*\\n(?:[ ]*\\n)*( +)[^ ][^\\n]*\\n((?:[ ]*\\n)*\\2[^\\n]+\\n?)*'
```
joshgoebel: Is there a reason you added non-capturing?  Generally (other than multi-match) our engine doesn't care about capturing or not... is this intended to be a performance improvement or were you just not aware that "capture doesn't matter"?

I wonder if we could break this into multiple lines/regex (with comments) and use our `regex.concact` to build the final regex... it's getting a little hard to follow.

pikammmmm: Required here, not perf: `\2` has to keep pointing at the indent group `( +)`. If the new groups capture, the indent becomes `\3` and `\2` refers to a group that can match zero times, so `\2[^\n]+` degrades to "any line" and the block swallows what follows it — append `baz: quux` to the test sample and it gets pulled into the string.

Happy to rebuild it as commented pieces via `regex.concat` — say the word and I'll push that.

joshgoebel: Oh right, duh. :-)

> say the word and I'll push that.

word.

joshgoebel (approved):

pikammmmm:
Verified on main_12 with CRLF input: yes, that solves it — after the normalization the grammar only ever sees LF, so this rule needs no CRLF handling of its own. One note: main_12's yaml.js still has the pre-#4457 regex, so v12 currently still shows the #4090 empty-line bug; once this rule rides over in a grammar merge, CRLF block scalars are covered too.

PR #4460, opened 2026-08-05
Update sqf.js to A3 v2.22

LouMontana:
Hi, this updates the SQF language to its current version (A3 v2.22) also adding previously-missed commands from older/other titles.


src/languages/sqf.js
```diff
@@ -96,12 +96,16 @@ export default function(hljs) {
     'exitWith',
     'for',
     'forEach',
+    'forEachMember',
+    'forEachMemberAgent',
+    'forEachMemberTeam',
+    'forEachReversed',
     'from',
     'if',
-    'local',
```
joshgoebel, line 101: No more local?

joshgoebel:
Do we have code that was used to generate or update these lists?  Looks like a LOT of changes... if so we'd like to capture that somewhere so the next maintainer could generate the lists programmatically, etc.

src/languages/sqf.js
LouMontana: no more 'local' indeed, as it is deemed a command ([local](https://community.bistudio.com/wiki/local)) rather than a keyword now ([private](https://community.bistudio.com/wiki/private) now)
thanks for checking :)

LouMontana:
> Do we have code that was used to generate or update these lists? Looks like a LOT of changes... if so we'd like to capture that somewhere so the next maintainer could generate the lists programmatically, etc.

Hi yes, there is SQF code from Leopard20 to generate the list, what I did however was using the [Scripting Commands](https://community.bistudio.com/wiki/Category:Scripting_Commands) list in order to have all commands from all games (to cover all SQF commands and not just Arma 3 ones).

src/languages/sqf.js
joshgoebel: Should it stay in a separate list perhaps that we track separately?  One thing I don't think a lot of people think about is what about all the Arma 2 code still out there being highlighted? 

Thoughts?

LouMontana: the 'local' keyword was not used a lot in Arma 2, even in official code; I think it's fine to have it as A3 (I learnt about it when A3 made it obsolete :D). the highlight will still mark it as "keyword"/"command", so I think it's fine.
I suppose it's that or we make an SQF version per game? :-|

joshgoebel: Ok, so we keep it then?

LouMontana: I would say so yes.

src/languages/sqf.js
```diff
@@ -355,9 +370,9 @@ export default function(hljs) {
     'briefingName',
     'buildingExit',
     'buildingPos',
-    'buldozer_EnableRoadDiag',
-    'buldozer_IsEnabledRoadDiag',
-    'buldozer_LoadNewRoads',
+    'buldozer_enableRoadDiag',
+    'buldozer_isEnabledRoadDiag',
+    'buldozer_loadNewRoads',
```
joshgoebel, line 375: Sorry just saw this, is there a "idiomatic" case for these even if they are insensitive?  Just wondering why they changed.

LouMontana: it was from the copy-paste; they are indeed case-insensitive
it is to keep one and same format for command naming, so people don't get auto-suggested `SetDamage` instead of `setDamage`, etc

joshgoebel: So these are the preferred idiomatic casing then, yes?

LouMontana: yes, I checked them ;-) Wiki / camelCase is the format to use

joshgoebel (approved):

PR #4461, opened 2026-08-05
fix(rust) reuse the string rules inside attributes

### LINKED ISSUE #3817: (Rust) Escaped double quotations in string at attribute are not highlighted properly

MaksZhukov:
Resolves #3817

### Changes

The `meta` mode for Rust attributes declared its own minimal string rule (`"` … `"` + `BACKSLASH_ESCAPE`), so only plain double-quoted literals were recognised inside `#[...]`. Raw strings and byte strings were mis-highlighted:

```rust
#[doc = r#"a "quoted" word"#]
```

was cut at the first inner quote, so `quoted` fell out of the string and the rest of the attribute was wrongly scoped.

Attributes hold arbitrary token trees, so any string literal that is valid elsewhere is valid there too — which answers the question raised in the issue. This extracts the existing string modes into named constants (`QUOTE_STRING`, `RAW_STRING`, `CHARACTER`) and reuses `QUOTE_STRING` / `RAW_STRING` inside `meta` instead of duplicating a weaker version. The top-level `contains` order is unchanged.

The character-literal mode is deliberately **not** added to `meta`, so lifetimes in attribute token trees (`#[foo(Bar<'a>)]`) are not treated as an unterminated char literal.

Note on the original report: the exact snippet in the issue (`#[error("\" appears in a string")]`) has been highlighted correctly since 11.10.0, but the underlying cause it pointed at — `meta` not reusing the string rules — was still present. Before/after for the remaining cases:

| input | before | after |
|---|---|---|
| `#[doc = r#"a "quoted" word"#]` | `r#` outside the string, literal split in two | whole `r#"…"#` is one string |
| `#[doc = b"bytes"]` | `b` outside the string | `b"bytes"` is one string |
| multi-line `r##"…"##` in an attribute | broken | one string |

### How to test

```
npm run build && npm test
```

Or in the demo/playground, paste the Rust snippets above.

### Checklist
- [x] Added markup tests (`test/markup/rust/strings.txt` / `.expect.txt` — raw, byte, multi-line raw strings in attributes, plus a lifetime-in-attribute regression case)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/languages/rust.js
+++ b/src/languages/rust.js
@@ -29,18 +29,6 @@ export default function(hljs) {
     hljs.C_LINE_COMMENT_MODE,
     hljs.COMMENT('/\\*', '\\*/', { contains: [ 'self' ] })
   ];
-  const META_STRING = [
-    {
-      scope: 'string',
-      begin: /"/,
-      end: /"/,
-      contains: [ hljs.BACKSLASH_ESCAPE ]
-    },
-    {
-      scope: 'string',
-      begin: /b?r(#*)"(.|\n)*?"\1(?!#)/
-    }
-  ];
   const NUMBER_SUFFIX = '([ui](8|16|32|64|128|size)|f(16|32|64|128))\?';
   const KEYWORDS = [
     "abstract",
@@ -173,6 +161,25 @@ export default function(hljs) {
     "assert_ne!",
     "debug_assert_ne!"
   ];
+  const QUOTE_STRING = hljs.inherit(hljs.QUOTE_STRING_MODE, {
+    begin: /b?"/,
+    illegal: null
+  });
+  const RAW_STRING = {
+    scope: 'string',
+    begin: /b?r(#{0,255})"(.|\n)*?"\1(?!#)/
+  };
+  const CHARACTER = {
+    scope: 'string',
+    begin: /b?'/,
+    end: /'/,
+    contains: [
+      {
+        scope: "char.escape",
+        match: /\\('|"|\\|\w|x\w{2}|u\w{4}|U\w{8})/
+      }
+    ]
+  };
   const TYPES = [
     "i8",
     "i16",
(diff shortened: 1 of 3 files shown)


src/languages/rust.js
```diff
@@ -157,6 +157,25 @@ export default function(hljs) {
     "assert_ne!",
     "debug_assert_ne!"
   ];
+  const QUOTE_STRING = hljs.inherit(hljs.QUOTE_STRING_MODE, {
```
joshgoebel, line 164: What are we still getting from QUOTE_STRING_MODE? escapes?

src/languages/rust.js
```diff
@@ -157,6 +157,25 @@ export default function(hljs) {
...
+    begin: /b?"/,
+    illegal: null
+  });
+  const RAW_STRING = {
+    scope: 'string',
+    begin: /b?r(#*)"(.|\n)*?"\1(?!#)/
```
joshgoebel: Can the run of `#` really be infinite?

joshgoebel: If so lets add a ################ sample to the specs, just cause.  If not lets add the max size (if it's reasonable) and tighten this rule.

joshgoebel (changes requested):

src/languages/rust.js
MaksZhukov: Yes — the `end: /"/` and the `BACKSLASH_ESCAPE` handling. `illegal: null` clears `QUOTE_STRING_MODE`'s `illegal: '\n'` so multiline strings still highlight. (This rule is unchanged from before — the PR just extracts it into a named constant so the attribute rule can reuse it.)

src/languages/rust.js
MaksZhukov: No — rustc caps raw string delimiters at 255 `#` symbols ("too many `#` symbols: raw strings may be delimited by up to 255 `#` symbols"). Tightened the rule to `#{0,255}` and added a many-hash sample to the specs. Verified the boundary: 255 hashes highlights as one raw string, while 256 hashes (invalid Rust) no longer matches the raw-string rule.

src/languages/rust.js
```diff
@@ -157,6 +157,26 @@ export default function(hljs) {
...
+  };
+  const CHARACTER = {
+    scope: 'string',
+    begin: /b?'/,
+    end: /'/,
+    contains: [
+      {
+        scope: "char.escape",
+        match: /\\('|"|\\|\w|x\w{2}|u\w{4}|U\w{8})/
+      }
```
joshgoebel, line 180: is CHARACTER meant to match a single character?

joshgoebel (approved):

PR #4463, opened 2026-08-05
fix(ruby): don't treat scope resolution :: as symbols

### LINKED ISSUE #4294: Ruby syntax: Wrong recognition of symbols at the place of calling a class methods
### Description of the issue
Wrong recognition of symbols at the place of calling a class methods.
Class methods may be call via:
`ClassName::method`
but it recogmized as symbol `:method`

##### Which language seems to have the issue?
**Ruby**

**I'm using `highlight`**

#### Expected behavior
`ClassName::method` must be recognized as call of method

## Patch
Just modify regexp for `symbol` from `:(?!\\s)` to **`(?<!:):(?!\\s)`**

(I changed these regexp in my local version of the script `highlight` - everything works fine!)

## Yet another wish

Please bring the syntax analysis in line with the changes in the latest version [**Ruby 3.4**](https://docs.ruby-lang.org/en/3.4/syntax_rdoc.html)

Hashim1999164:
Fixes #4294

### Problem

In Ruby, `ClassName::method` was highlighted as a class name followed by two symbols:

```
<span class="hljs-title class_">ClassName</span><span class="hljs-symbol">:</span><span class="hljs-symbol">:method</span>
```

The grammar already had a rule meant to swallow namespace qualifiers (`begin: hljs.IDENT_RE + '::'`), but it never fired. The CamelCase `CLASS_REFERENCE` rule is listed earlier and matches at the same offset, so it consumed `ClassName` on its own and left `::method` for the symbol rule, which matches any colon not followed by whitespace.

### Fix

Match the `::` operator on its own instead of `identifier::`, so it is swallowed no matter what rule handled the text before it. The issue suggested a negative lookbehind on the symbol rule, but the codebase still avoids lookbehind for browser support reasons (see the TODO notes in `swift.js` and `javascript.js`), and it would also leave the first colon highlighted as a symbol.

After:

```
<span class="hljs-title class_">ClassName</span>::method
<span class="hljs-title class_">Foo</span>::<span class="hljs-title class_">Bar</span>::baz(<span class="hljs-number">1</span>)
<span class="hljs-title class_">Math</span>::<span class="hljs-variable constant_">PI</span>
::<span class="hljs-title class_">TopLevel</span>.new
```

Regular symbols (`:symbol`, `:"quoted"`, `key:` in hashes, `obj.send(:to_s)`) are unaffected.

### Test plan

- Added `test/markup/ruby/symbols.txt` / `symbols.expect.txt` covering plain symbols, hash keys, quoted symbols, and scope resolution including nested namespaces and a leading `::`.
- `npm run build && npm test`: 1594 passing, 0 failing.
- Ran `highlightAuto` over every fixture in `test/detect` before and after the change; the detected language and relevance are identical for all 198 files, so auto detection is unaffected.

### DIFF
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -325,8 +325,10 @@ export default function(hljs) {
     CLASS_REFERENCE,
     METHOD_DEFINITION,
     {
-      // swallow namespace qualifiers before symbols
-      begin: hljs.IDENT_RE + '::' },
+      // swallow the scope resolution operator so `::` is not read as a symbol
+      begin: '::',
+      relevance: 0
+    },
     {
       className: 'symbol',
       begin: hljs.UNDERSCORE_IDENT_RE + '(!|\\?)?:',
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -326,8 +326,7 @@ export default function(hljs) {
     METHOD_DEFINITION,
     {
       // swallow the scope resolution operator so `::` is not read as a symbol
-      begin: '::',
-      relevance: 0
+      begin: '::'
     },
     {
       className: 'symbol',
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Core Grammars:
 - fix(haskell) highlight `where` in GADT and closed type-family declarations, issue #3753 [Konstantin Baltsat][]
 - fix(css) support six-digit `unicode-range` values [Konstantin Baltsat][]
 - fix(c, cpp) stop a raw string's closing delimiter from swallowing quotes, which broke highlighting of everything after the literal, issue #3585 [David Pavlovschii][]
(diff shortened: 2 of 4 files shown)


src/languages/ruby.js
```diff
@@ -325,8 +325,10 @@ export default function(hljs) {
     CLASS_REFERENCE,
     METHOD_DEFINITION,
     {
-      // swallow namespace qualifiers before symbols
-      begin: hljs.IDENT_RE + '::' },
+      // swallow the scope resolution operator so `::` is not read as a symbol
+      begin: '::',
+      relevance: 0
```
joshgoebel: I don't think we need this anymore, relevance is on it's way out.  It only ever mattered for auto-detect.

joshgoebel:
Seems simple enough, just the suggested change.

Hashim1999164:
Removed the `relevance: 0` on the `::` swallow rule as suggested.

joshgoebel (approved):

PR #4466, opened 2026-08-07
fix(c, cpp): scope angle-bracket header string to #include (issue #3505)

### LINKED ISSUE #3505: (C++) highlight incorrectly

MsfPablo:
Fixes #3505.

## Root cause

The C/C++ preprocessor grammar applied the angle-bracket quoted-string rule (`<.*?>`, intended for `#include <header>`) to **every** preprocessor directive, not just `#include`. When a `#define` body contained `>`, for example:

```cpp
#define what do { cout << ">"; } while (0)
```

the greedy `<.*?>` match consumed `<< ">`, leaving an unbalanced `"` that the string rule then extended to the end of the file, breaking highlighting for everything after the offending line.

## Fix

Per @joshgoebel's guidance in the issue ("You may be right that this needs to be scoped to `#include`..."), this scopes the angle-bracket string to a dedicated `PREPROCESSOR_INCLUDE` mode that only matches `#include` lines, and removes it from the general `PREPROCESSOR` mode. `#include <iostream>` and `#include <sys/types.h>` continue to highlight exactly as before; other directives (`#define`, `#if`, ...) no longer risk swallowing a quote.

The change is applied symmetrically to both `src/languages/c.js` and `src/languages/cpp.js`, with `PREPROCESSOR_INCLUDE` placed ahead of `PREPROCESSOR` in each `contains` array so `#include` matches the more specific mode first.

## Tests

- Adds markup fixtures `test/markup/cpp/preprocessor-define-angle-bracket` and `test/markup/c/preprocessor-define-angle-bracket` reproducing the issue (a `#define` with `>` in its body plus a normal `#include` and follow-on code).
- `npm run build`, `npm test` (1595 passing, 3 pending, 0 failing), `npm run lint`, and `npm run lint-languages` are all green.

The maintainer explicitly asked for a PR in issue #3505 ("Either of your willing to submit a PR for this?").

Developed with AI assistance and reviewed by the contributor.

### DIFF
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -118,18 +118,18 @@ export default function(hljs) {
   // which would otherwise leave an unbalanced `"` and break highlighting for
   // the rest of the file. See issue #3505.
   const PREPROCESSOR_INCLUDE = {
-    className: 'meta',
+    scope: 'meta',
     begin: /#\s*include\b/,
     end: /$/,
     keywords: { keyword: 'include' },
     contains: [
       {
+        // the `\` at the end of a line signaling continuation
         begin: /\\\n/,
-        relevance: 0
       },
-      hljs.inherit(STRINGS, { className: 'string' }),
+      STRINGS,
       {
-        className: 'string',
+        scope: 'string',
         begin: /<.*?>/
       },
       C_LINE_COMMENT_MODE,
@@ -154,6 +154,11 @@ export default function(hljs) {
       hljs.C_BLOCK_COMMENT_MODE
     ]
   };
+
+  const PREPROCESSORS = [
+    PREPROCESSOR_INCLUDE,
+    PREPROCESSOR
+  ];
 
   const TITLE_MODE = {
     className: 'title',
@@ -262,8 +267,7 @@ export default function(hljs) {
   };
 
   const EXPRESSION_CONTAINS = [
-    PREPROCESSOR_INCLUDE,
-    PREPROCESSOR,
+    ...PREPROCESSORS,
     TYPES,
     C_LINE_COMMENT_MODE,
     hljs.C_BLOCK_COMMENT_MODE,
@@ -359,8 +363,7 @@ export default function(hljs) {
       TYPES,
       C_LINE_COMMENT_MODE,
       hljs.C_BLOCK_COMMENT_MODE,
-      PREPROCESSOR_INCLUDE,
-      PREPROCESSOR
+      ...PREPROCESSORS
     ]
   };
 
@@ -377,8 +380,7 @@ export default function(hljs) {
       FUNCTION_DECLARATION,
       EXPRESSION_CONTAINS,
       [
-        PREPROCESSOR_INCLUDE,
-        PREPROCESSOR,
+        ...PREPROCESSORS,
         {
           begin: hljs.IDENT_RE + '::',
           keywords: KEYWORDS
@@ -395,7 +397,6 @@ export default function(hljs) {
       ]),
     exports: {
       preprocessor: PREPROCESSOR,
-      preprocessorInclude: PREPROCESSOR_INCLUDE,
       strings: STRINGS,
       keywords: KEYWORDS
     }
--- a/src/languages/c.js
+++ b/src/languages/c.js
@@ -111,6 +111,32 @@ export default function(hljs) {
     relevance: 0
   };  
   
+  // `#include` is the only preprocessor directive that takes an angle-bracket
+  // quoted header (`#include <header>`). Scoping that rule to `#include` keeps
+  // the greedy `<...>` match from eating a `>` that belongs to the body of
+  // another directive (e.g. `#define what do { cout << ">"; } while (0)`),
+  // which would otherwise leave an unbalanced `"` and break highlighting for
+  // the rest of the file. See issue #3505.
+  const PREPROCESSOR_INCLUDE = {
+    className: 'meta',
+    begin: /#\s*include\b/,
+    end: /$/,
+    keywords: { keyword: 'include' },
+    contains: [
+      {
+        begin: /\\\n/,
+        relevance: 0
+      },
+      hljs.inherit(STRINGS, { className: 'string' }),
+      {
+        className: 'string',
+        begin: /<.*?>/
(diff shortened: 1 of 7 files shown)


src/languages/c.js
```diff
@@ -111,6 +111,32 @@ export default function(hljs) {
...
+  // the rest of the file. See issue #3505.
+  const PREPROCESSOR_INCLUDE = {
+    className: 'meta',
+    begin: /#\s*include\b/,
+    end: /$/,
+    keywords: { keyword: 'include' },
+    contains: [
+      {
+        begin: /\\\n/,
+        relevance: 0
```
joshgoebel: Don't need this.

src/languages/c.js
```diff
@@ -111,6 +111,32 @@ export default function(hljs) {
...
+  // which would otherwise leave an unbalanced `"` and break highlighting for
+  // the rest of the file. See issue #3505.
+  const PREPROCESSOR_INCLUDE = {
+    className: 'meta',
+    begin: /#\s*include\b/,
+    end: /$/,
+    keywords: { keyword: 'include' },
+    contains: [
+      {
+        begin: /\\\n/,
```
joshgoebel, line 128:
```suggestion
        // the `\` at the end of a line signaling continuation
        begin: /\\\n/,
```

Right?

src/languages/c.js
```diff
@@ -111,6 +111,32 @@ export default function(hljs) {
...
+        relevance: 0
+      },
+      hljs.inherit(STRINGS, { className: 'string' }),
```
joshgoebel: What is this trying to do vs just including `STRINGS` directly?

src/languages/c.js
```diff
@@ -111,6 +111,32 @@ export default function(hljs) {
...
+      {
+        className: 'string',
```
joshgoebel: `className` is deprecated, please use `scope` in new work.

src/languages/c.js
```diff
@@ -353,6 +377,7 @@ export default function(hljs) {
       FUNCTION_DECLARATION,
       EXPRESSION_CONTAINS,
       [
+        PREPROCESSOR_INCLUDE,
         PREPROCESSOR,
```
joshgoebel: I feel like this trend will grow in the future (more variants) so what about moving this pattern out:

```js
PREPROCESSORS = [
   PREPROCESSOR_INCLUDE,
   PREPROCESSOR,
]
```

Then destructuring each location to pull in the whole list

```
...PREPROCESSORS
```

src/languages/c.js
```diff
@@ -370,6 +395,7 @@ export default function(hljs) {
       ]),
     exports: {
       preprocessor: PREPROCESSOR,
+      preprocessorInclude: PREPROCESSOR_INCLUDE,
```
joshgoebel: Is there a reason you added this?  I think these other things are dead code (at least in this grammar)... it looks like the only place we still use this type of thing is typescript.js and arduino.js... there is no sharing between C and CPP.

joshgoebel (changes requested):

src/languages/c.js
MsfPablo: Done — dropped the `relevance: 0` in 549c8cb.

src/languages/c.js
MsfPablo: Done exactly as suggested in 549c8cb (kept the comment, dropped `relevance: 0`). Right, the continuation child doesn't need to suppress relevance here.

src/languages/c.js
MsfPablo: Good catch — `STRINGS` already scopes to `'string'` at the top level, so the `inherit(..., { className: 'string' })` was a no-op override. Replaced with `STRINGS` directly in 549c8cb (both c.js and cpp.js).

src/languages/c.js
MsfPablo: Switched to `scope` throughout the new `PREPROCESSOR_INCLUDE` block (`scope: 'meta'` and `scope: 'string'`) in 549c8cb, in both c.js and cpp.js.

src/languages/c.js
MsfPablo: Good idea — factored out `const PREPROCESSORS = [PREPROCESSOR_INCLUDE, PREPROCESSOR]` and spread `...PREPROCESSORS` at all three call sites in both c.js and cpp.js (549c8cb).

src/languages/c.js
MsfPablo: You're right, there's no C↔CPP sharing — it was dead code. Removed the `preprocessorInclude` export from c.js in 549c8cb (cpp.js had no such export to begin with).

MsfPablo:
Pushed commit `549c8cb` addressing all of the review feedback (applied consistently to both `c.js` and `cpp.js`):

- **L128 continuation child** — dropped the unnecessary `relevance: 0` and added the suggested comment (as proposed).
- **L130 `inherit(STRINGS, …)`** — `STRINGS` already scopes to `"string"` at the top level, so the override was a no-op; now uses `STRINGS` directly.
- **L132 `className`** — switched the whole new `PREPROCESSOR_INCLUDE` block to `scope` (`meta` + `string`).
- **L381** — factored the repeated pair into `const PREPROCESSORS = [PREPROCESSOR_INCLUDE, PREPROCESSOR]` and spread `...PREPROCESSORS` at all three call sites.
- **L398** — removed the dead `preprocessorInclude` export (no C↔CPP sharing exists).

Markup output is unchanged (`scope` and `className` emit the same span); all 22 `c`/`cpp` markup fixtures still pass against the built lib, including the two `preprocessor-define-angle-bracket` fixtures this PR adds. `npm run build` and `npm run lint-languages` are clean. Happy to adjust further.

(Grammar authored with AI assistance; happy to rework any part manually.)

joshgoebel (approved):

Issue #3157, opened 2021-04-21
Future: Possibly more automated type compilation


joshgoebel:
> Getting access to the types in my main source. My source is all JS, it exports no types... all the type exports are ambient modules.

I’m proposing the inverse: define your types in your code.

> I dunno what that means? You mean declare them in highlight.js itself and then export them, import them into index.d.ts and then reexport them? I'm not sure I know how to do that if that's possible - or advisable?

Indeed: write them in JavaScript. No need to add a manual `index.d.ts`: typescript can compile javascript and generate that `index.d.ts`.

Here’s a smaller example: https://github.com/words/levenshtein-edit-distance/commit/2c3d890068b0d9405881a94f1439b31f8d08c392.
Lowlight is a bit more involved.
A bit more complex is `unist-util-select`: https://github.com/syntax-tree/unist-util-select/commit/4c1b02e34f849394bc90b42ea3bce0f52d9b2512.
And quite complex is xdm: https://github.com/wooorm/xdm.

I am not a fan of typescript. I personally don’t really love types either. But they are somewhat useful — especially for newcomers to a giant ecosystem ([unified](https://unifiedjs.com)).
I think types through jsdoc in javascript is rather nice: code is directly runnable, no building needed. But everything is still type checked by typescript (and I like `type-coverage` to ensure *everything* is completely typed).
It ensures `.d.ts` are kept up to date and work

Some more info here: https://github.com/voxpelli/types-in-js

_Originally posted by @wooorm in https://github.com/highlightjs/highlight.js/issues/3131#issuecomment-824157692_

joshgoebel:
If someone knows how this is done and would like to whip up a tiny PR, I'd love to see it.  I think what would be since is a tiny utility script that we could run periodically that would compile the types.  I also wonder if this magically will create ambient modules as we do now.

joshgoebel:
Just don't get your hopes up. This isn't a guaranteed merge.  I want to see how exactly this would work and have a chance to play with it to see if it feels right.

PR #4480, opened 2026-08-09
refactor(types): JSDoc source of truth + declaration emit

joshgoebel:
## Summary
- Move public/private type definitions into `src/lib/hljs_types.js` (JSDoc) and stop treating hand-written ambient `types/index.d.ts` as the source of truth (#3157).
- Add `npm run types:emit` (`tsconfig.emit.json`) to generate `types/generated/**`; package entry is a thin barrel over that output.
- Wire node build to ship generated decls + `types/private.d.ts` (`exports["./private"]`); drop unused `vuePlugin` from the public type surface.

## Notes
- Consumers still use `"types": "./types/index.d.ts"`; named public types are explicit exports; private aliases are `highlight.js/private` only.
- Generated files are committed for now; a CI drift check (`types:emit && git diff --exit-code`) is a natural follow-up.

## Test plan
- [x] `npm run types:emit` (exit 0)
- [x] `npm run test-parser`
- [x] Smoke `tsc` consumer: default export + `Mode` / `HLJSApi`; private types not on main entry
- [ ] Full `npm test` / node build pack types check


allejo:
I'm all for improved and automated types, but it's 2026; why are we just not using TypeScript? We already have a build step to bundle our library for distribution; what's one more step being added? We'd also get improved type safety over JSDoc that isn't necessarily enforced anywhere.

joshgoebel:
> why are we just not using TypeScript?

Oh boy another TypeScript zealot. :-). This "everything should be ported to typescript" assumption is one reason I dislike the idea... but...

Originally I thought it's easier to find people who can contribute with regular JS - no types to worry about, I only added the typing back when I used VS Code (I don't anymore) and it would help me out sometimes.  I dunno how that stands today but still feels like with TS you have to know extra "stuff" - so any TS project is potentially locking out contributors who know only JS.  And I didn't (and still don't) think we should be locking out _any_ potential contributors.

So why JSDoc?  I wanted to see if I could get SOME of the value from TS types without migrating the entire project.  It had it's moments but It was never a huge value though.  

TLDR: _Not every project needs or benefits from strict types._

joshgoebel:
And I dunno about this issue/PR really... was just playing around with the idea to see what it looked like.  Should have marked it a draft.

allejo:
> I thought it's easier to find people who can contribute with regular JS
> ...
> feels like with TS you have to know extra "stuff"

Yes, and no? The extra "stuff" I believe you're referring to in TypeScript is some of its advanced features, but I can't see us needing anything too advanced. Besides, everything you're doing in JSDoc in this PR is just TS with more fluff and in comments. And let's not forget that all valid JS is valid TS, so JS-only contributors _can_ still contribute and we can just sprinkle the types during the PR review process.

## Inline Types

**JavaScript**

```js
/** @type {Mode} */
export const C_BLOCK_COMMENT_MODE = COMMENT('/\\*', '\\*/');
```

**TypeScript**

```ts
export const C_BLOCK_COMMENT_MODE: Mode = COMMENT('/\\*', '\\*/');
```

## Importing Types

**JavaScript**

```js
/**
@typedef {import('./hljs_types.js').Mode} Mode
@typedef {import('./hljs_types.js').CompiledMode} CompiledMode
@typedef {import('./hljs_types.js').Language} Language
@typedef {import('./hljs_types.js').HLJSPlugin} HLJSPlugin
@typedef {import('./hljs_types.js').CompiledLanguage} CompiledLanguage
*/
```

**TypeScript**

```ts
import { 
  CompiledLanguage,
  CompiledMode,
  HLJSPlugin,
  Language
  Mode,
} from './hljs_types';
```

## Inline Function Types

**JavaScript**

```js
mode.contains.forEach(/** @param {CompiledMode} term */ term => mm.addRule(term.begin, { rule: term, type: "begin" }));
```

**TypeScript**

```ts
mode.contains.forEach((term: CompiledMode) => mm.addRule(term.begin, { rule: term, type: "begin" }));
```

## Callback Signatures

**JavaScript**

```js
/**
 * @callback LanguageFn
 * @param {HLJSApi} hljs
 * @returns {Language}
 */
 ```

**TypeScript**

```ts
export type LanguageFn = (hljs: HLJSApi) => Language;
```

## Unions

**JavaScript**

```js
/**
 * @typedef {Error & { mode?: Mode | Language | CompiledMode, languageName?: string, badRule?: Mode | CompiledMode }} AnnotatedError
 */
```

**TypeScript**

```ts
export type AnnotatedError = Error & { 
    mode?: Mode | Language | CompiledMode;
    languageName?: string;
    badRule?: Mode | CompiledMode;
};
```

---

> So why JSDoc? I wanted to see if I could get SOME of the value from TS types without migrating the entire project

We don't _have_ to port the entire project to TS in one go. We can use `"allowJs": true`, which you already have included in the `tsconfig.json` of this PR and [add `// @ts-check` to files we want to check with TypeScript](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html).

For me, your argument that new contributors need to learn extra "stuff" seems irrelevant, since they will need to know JSDoc syntax to contribute without breaking type safety. On the contrary, I'd argue that having clear data types throughout the codebase would _help_ new contributors know what data types functions expect/return. I feel like the only reason why using JSDoc would be easier for new contributors is that our types are just comments that can be ignored, which means they will inevitably become out of sync or outright wrong.

Whether we stick with JSDoc or move to TypeScript, we should add a GitHub Action to verify the types' sanity using `tsc`.

joshgoebel:
> Whether we stick with JSDoc or move to TypeScript, we should add a GitHub Action to verify the types' sanity using tsc.

That makes sense.  Does that work with this PR or would it work before?

Lets revisit the larger TypeScript discussion here after we get to v12... I just finally want to get that huge branch merged and released before we go and change everything. :-)

allejo:
> That makes sense. Does that work with this PR or would it work before?

I made the following changes:

- rebased this branch onto `main` (habit, sry if you wanted me to merge)
- generated files don't need to be committed; they can be generated during our release/publish pipeline
- added new GitHub Actions to ensure type safety

The only remaining change that needs to be done is to fix the type safety that `tsc` is complaining about.

PR #4488, opened 2026-08-12
refactor: replace mode Object.create chain with explicit stack

joshgoebel:
## Summary
- Replace the `Object.create(mode, { parent })` mode chain in `_highlight` with an explicit `modeStack` (push/pop).
- `endOfMode` / `endsParent` / `endsWithParent` walk stack indexes instead of `.parent`.
- Sublanguage continuations and `_top` now carry the full mode stack (`CompiledMode[]`).

## Test plan
- [x] `npx mocha test/api test/markup test/highlight` (617 passing)
- [ ] CI on the PR

### DIFF
--- a/src/highlight.js
+++ b/src/highlight.js
@@ -129,7 +129,7 @@ const HLJS = function(hljs) {
    * @property {number} relevance - the relevance score
    * @property {string} value - the highlighted HTML code
    * @property {string} code - the original raw code
-   * @property {CompiledMode} top - top of the current mode stack
+    * @property {CompiledMode[]} top - mode stack (continuation)
    * @property {boolean} illegal - indicates whether any illegal matches were found
   */
   function highlight(codeOrLanguageName, optionsOrCode, ignoreIllegals) {
@@ -179,9 +179,9 @@ const HLJS = function(hljs) {
    * @param {string} languageName - the language to use for highlighting
    * @param {string} codeToHighlight - the code to highlight
    * @param {boolean?} [ignoreIllegals] - whether to ignore illegal matches, default is to bail
-   * @param {CompiledMode?} [continuation] - current continuation mode, if any
-   * @returns {HighlightResult} - result of the highlight operation
-  */
+    * @param {CompiledMode[]?} [continuation] - mode stack to resume, if any
+    * @returns {HighlightResult} - result of the highlight operation
+   */
   function _highlight(languageName, codeToHighlight, ignoreIllegals, continuation) {
     const keywordHits = Object.create(null);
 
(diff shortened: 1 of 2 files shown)


src/highlight.js
```diff
@@ -131,7 +131,7 @@ const HLJS = function(hljs) {
    * @property {number} relevance - the relevance score
    * @property {string} value - the highlighted HTML code
    * @property {string} code - the original raw code
-   * @property {CompiledMode} top - top of the current mode stack
+    * @property {CompiledMode[]} top - mode stack (continuation)
```
allejo: what's with the extra spaces?

src/highlight.js
```diff
@@ -326,17 +326,19 @@ const HLJS = function(hljs) {
...
     }
 
     /**
-     * @param {CompiledMode } mode - the mode to potentially end
+     * @param {number} stackIndex - index in modeStack of the mode to potentially end
      * @param {RegExpMatchArray} match - the latest match
      * @param {string} matchPlusRemainder - match plus remainder of content
-     * @returns {CompiledMode | void} - the next mode, or if void continue on in current mode
+     * @returns {number | void} - stack index of mode to close through, or void to continue
      */
```
allejo: Since you're touching this, could we return `-1` to continue instead? I find it weird that we're relying on a function's implicit behavior of returning `undefined`.

allejo (changes requested):
minor nitpicks

src/highlight.js
joshgoebel: AI?
```suggestion
   * @property {CompiledMode[]} top - mode stack (continuation)
```

src/highlight.js
joshgoebel: Ugh, what is this C? ;-).  -1 is "not found" when all you have are numbers. 

Also -1 IS an array index in many languages, just from the back side of the list, so don't love it as a sentinel value when we have richer choices. 

Is it just undefined that bothers you?  Is `null` ok?  We could also have a `hljs.continue` that's a Symbol() if undefined/null aren't agreeable.

allejo: I'd be okay with an explicit `null`. What bothers me is the "return void", which in a lot of languages (TypeScript included) means the function only performs an operation and doesn't return anything; i.e., the function either returns values or doesn't. `null` is a value that means "no value" so I'm okay with returning _that_ value for clarity that something _is_ still being returned.

joshgoebel: I see now... it's a convention of "this function has no return value"... then I say we switch to `null` to be explicit.

joshgoebel:
@allejo  back to you

allejo (approved):
lgtm

PR #4493, opened 2026-08-14
fix(typescript): do not treat <string[]> assertions as HTML tags

### LINKED ISSUE #4301: (typescript) `<string[]>` breaks highlighting

arronKler:
`<string[]>` (and other angle-bracket array type assertions) were treated as JSX/XML tags, so highlighting stopped at the `[]`. `[` cannot follow an HTML tag name, so `isTrulyOpeningTag` now ignores that case the same way it already ignores nested generics (`<`) and type parameters (`,`).

Fixes #4301

### Changes
- Ignore `<Type[]>` in the shared JS/TS JSX tag check
- Add a TypeScript markup fixture for the reported case

### Checklist
- [x] Added markup tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,7 @@ Core Grammars:
 - fix(haskell) highlight `where` in GADT and closed type-family declarations, issue #3753 [Konstantin Baltsat][]
 - enh(java) improve detection of types, including generic and array types [Hannes Wallnoefer][]
 - enh(javascript) add `self` to built-in variables [Dsaquel][]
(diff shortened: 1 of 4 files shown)


joshgoebel:
Does #4478 help with this issue?

PR #4495, opened 2026-08-14
fix(php-template): a quote in a line comment should not break highlighting

### LINKED ISSUE #4152: (PHP Template) A single apostrophe ' breaks php-template highlighting

arronKler:
### Changes
php-template finds `?>` by skipping strings and block comments. A `'` inside `//` or `#` started a skip-string and swallowed the closer, so later HTML/PHP was not highlighted.

This keeps `skip` on strings (so `?>` inside a string stays in PHP). It also skips `//` and `#` line comments, stopping before `?>` so `// note ?>` still closes the block, and it does not treat `#[` as a hash comment. Escaped quotes use a skipped `\.` so `\'` does not end the skipped string.

Fixes #4152

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ New Grammars:
 
 Core Grammars:
 
+- fix(php-template) do not let a quote in a line comment or an escaped quote in a string swallow the rest of the template, issue #4152 [arronKler][]
 - fix(c) only match real `atomic_*` type names, not C11 atomic functions, issue #3837 [Mark Xian][]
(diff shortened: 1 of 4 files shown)


arronKler:
Closing: #4152 is labeled good first issue. Per the project's AI-assisted contributions policy, GFI issues should not be claimed end-to-end with AI.

PR #4496, opened 2026-08-14
fix(java): tokenize method invocations and parameter types

### LINKED ISSUE #4229: (java) Code not getting tokenized (and thus not highlighted)
**Describe the issue**

Java. It works perfectly fine for *any* other language we've tested with. Just Java seems to break for some reason, with multiple different code examples tested.

Highlighting it through `hljs.highlightElement(codeElement);` where codeElement is an HTML element of type `code`, and then rendering that HTML element to the DOM.

**Sample Code to Reproduce**
```typescript
const codeElement = DomBuilder.getInstance().build({
  type: 'code',
  classNames: [
    ...(props.language !== undefined ? [ `language-${props.language}` ] : [ 'language-plaintext' ])
  ],
  innerHTML: "Some code string here"
});
hljs.highlightElement(codeElement);
```

Image

**Expected behavior**
The code is tokenized into spans and highlighted accordingly.

**Additional context**

This issue is occurring in the open source [MynahUI](https://github.com/aws/mynah-ui) repository for AWS, specifically in the [SyntaxHighlighter](https://github.com/aws/mynah-ui/blob/main/src/components/syntax-highlighter.ts) component.

arronKler:
Fixes #4229

### Changes

Java highlighted keywords, declarations, strings, and numbers, but left ordinary call sites and many parameter types as plain text. The Fibonacci sample from the issue is a typical case: `System.out.print(...)`, `fibonacci(10)`, and `main(String[] args)` were not marked up.

- Match identifier + `(` as `title.function.invoke`, excluding statement keywords (`if`, `for`, `while`, `switch`, `catch`, `synchronized`, `try`, `when`) plus `this`/`super`.
- Match simple `Type name` / `Type[] name` inside parameter lists so `String[] args` and `int n` are scoped. Generic arguments are left alone so `? super T` / `? extends R` keep their existing highlighting.
- Add a markup fixture from the issue sample.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,7 @@ Core Grammars:
 - enh(groovy) support underscores in numeric literals [greymoth][]
 - fix(haskell) highlight `where` in GADT and closed type-family declarations, issue #3753 [Konstantin Baltsat][]
 - enh(java) improve detection of types, including generic and array types [Hannes Wallnoefer][]
+- fix(java) tokenize method invocations and parameter types, issue #4229 [Arron Zou][]
 - enh(javascript) add `self` to built-in variables [Dsaquel][]
 - enh(kotlin) add `ktm` and `ktx` aliases [DarkMatter-999][]
 - fix(leaf) fix bug in Leaf keyword highlighting [Francesco Paolo Severino][]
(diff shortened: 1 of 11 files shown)


arronKler:
Closing: #4229 is not a confirmed grammar bug (bundling vs grammar). This change was a feature-level tokenizer expansion, not a hunt-bar bugfix.

PR #4506, opened 2026-08-16
fix(python): handle nested `{}` replacement fields in f-string format specs

### LINKED ISSUE #4503: (Python) f-string nested replacement field in a format spec closes `subst` at the wrong brace
**Which language:** Python (specified explicitly, not auto-detect)
**Version:** 11.12.0
**Using:** `highlight` (not `highlightAuto`)

**Describe the issue**
A nested replacement field inside an f-string format spec (dynamic width/precision, e.g. `f"{x:{w}}"`) closes the `subst` region on the *inner* `}`. The field's own closing `}` is then emitted as plain string text, and the `subst` span ends up brace-unbalanced.

**Sample Code to Reproduce**
```python
f"{x:{w}}"
```

**Actual** — `hljs.highlight('f"{x:{w}}"', {language:'python'}).value`:
```html
<span class="hljs-string">f&quot;<span class="hljs-subst">{x:{w}</span>}&quot;</span>
```
The `subst` span is `{x:{w}` (an unmatched `{`) and the final `}` renders in string color instead of as part of the field. `f"{value:{width}.{prec}f}"` fails the same way.

**Expected behavior**
The whole `{x:{w}}` is a single replacement field, and the nested `{w}` is itself a field. This is valid Python — `{w}` sets the width dynamically (`f"{x:{w}}"` with `x=42, w=6` → `'    42'`). The closing `}` should belong to the field, not leak into the string, and the `subst` region should be brace-balanced.

**Cause**
`SUBST` in `src/languages/python.js` (`begin:/\{/, end:/\}/`) does not `contain` itself (`SUBST.contains = [STRING, NUMBER, PROMPT]`), so it ends at the first inner `}` and cannot handle nested replacement fields.

Assisted-by: Claude Opus 4 (high)

waterWang:
## Summary

Fixes #4503 — `SUBST` (begin: `{`, end: `}`) did not contain itself, so a nested replacement field inside an f-string format spec (e.g. `f"{x:{w}}"`) closed the `subst` region at the inner `}`. The outer `}` was then emitted as plain string text, leaving the `subst` span brace-unbalanced.

### `src/languages/python.js`
- Added `'self'` to `SUBST.contains` so that `SUBST` can contain nested replacement fields

## Why `'self'`

In highlight.js, `'self'` in a mode's `contains` array refers to the mode itself. This is the standard way to allow a mode to contain instances of itself, which is exactly what we need for nested `{}` replacement fields in Python f-strings.

## Test plan

```python
f"{x:{w}}"          # nested field in format spec
f"{value:{width}.{prec}f}"  # multiple nested fields
```

Before this fix:
- `f"{x:{w}}"` → `subst` span is `{x:{w}` (unmatched `{`)
- After: the whole `{x:{w}}` is a single balanced `subst`

### DIFF
--- a/src/languages/python.js
+++ b/src/languages/python.js
@@ -358,6 +358,7 @@ export default function(hljs) {
     ]
   };
   SUBST.contains = [
+    'self',
     STRING,
     NUMBER,
     PROMPT


joshgoebel:
We'll need some specs to show this works.

PR #4512, opened 2026-08-17
fix(haml) recognize implicit div tag lines starting with .class or #id

### LINKED ISSUE #3783: HAML syntax highlighting broken
**Describe the issue**
HAML syntax highlighting is broken

### [Codepen](https://codepen.io/jerasmus123/pen/YzJeBev)

This issue was previously reported (https://github.com/highlightjs/highlight.js/issues/1435) and closed, but the issue does not seem to be fixed.

Screenshot 2023-05-08 at 15 07 12

Screenshot 2023-05-08 at 15 19 47

HAML

**Sample Code to Reproduce**
```
!!! 5
%html
  %body
    %h1.foo-bar This works fine
    %span.hello.world syntax is fine here
    .broken{ broken: 'yes' }
    %anything.working{ working: 'fine' }
      .also-broken
      .i-am-broken
        .me-too
          %h1.good All good!
```

**Expected behavior (Rouge example)**
Screenshot 2023-05-08 at 15 12 38

**Additional context**
Same issue with HAML mixed with Ruby:

Screenshot 2023-05-08 at 15 17 50

pikammmmm:
Resolves #3783.

### Changes

HAML's implicit-div syntax gets no highlighting at all. Lines like

```haml
.broken{ broken: 'yes' }
#main content
```

emit zero spans, while the equivalent `%div.broken{ broken: 'yes' }` highlights fine (sample and screenshots in the issue).

The cause: the tag mode only opens on `^\s*%`, and the `selector-class` / `selector-id` rules that would handle `.class` and `#id` are nested inside it, so a line without a leading `%` never reaches them.

The fix adds a second variant of the tag mode that begins, zero-width, on a line starting with `.class` or `#id`:

```js
begin: '^\\s*(?=[.#][\\w-])',
```

and reuses the same nested rules — they're extracted into shared constants, which is most of the diff's line count. The `%` mode itself is unchanged, and I diffed the rendered output of the existing haml fixtures before and after: byte-identical.

A few deliberate choices:

- The implicit-div variant carries `relevance: 0` throughout. Lines starting with `.foo` / `#foo` are everywhere in other languages (CSS selectors, C preprocessor, assembler directives); without this, a three-rule CSS snippet ties HAML with CSS in `highlightAuto` (6 vs 6). With it, such snippets contribute 0 to HAML, while real HAML detection is unchanged (the detect sample still scores 36).
- The lookahead requires `[\w-]` after the `.`/`#`, so `#{ruby}` interpolation at the start of a line still renders as Ruby, not as a div.
- The dot/hash stays part of the `selector-class` / `selector-id` span — the Rouge-style option @j-erasmus suggested in the thread, and what the nested rules already do for `%tag.class` today. If you'd rather have them as punctuation (the other option raised there), that's an easy switch.

Out of scope: new-style hash attributes (`{ broken: 'yes' }`) get no highlighting inside the braces — but that's equally true for `%tag{ ... }` today, so it's a pre-existing, separate enhancement rather than part of this bug.

### Checklist
- [ ] Added markup tests (`test/markup/haml/implicit_div.txt`)
- [ ] I have read and followed our [AI-assisted contributions](../docs/ai-contributions.md) policy (human review, no slop)

### DIFF
--- a/src/languages/haml.js
+++ b/src/languages/haml.js
@@ -8,6 +8,65 @@ Category: template
 
 // TODO support filter tags like :javascript, support inline HTML
 export default function(hljs) {
+  const SELECTOR_ID = {
+    className: 'selector-id',
+    begin: '#[\\w-]+'
+  };
+  const SELECTOR_CLASS = {
+    className: 'selector-class',
+    begin: '\\.[\\w-]+'
+  };
+  const RUBY_HASH_ATTRS = {
+    begin: /\{\s*/,
+    end: /\s*\}/,
+    contains: [
+      {
+        begin: ':\\w+\\s*=>',
+        end: ',\\s+',
+        returnBegin: true,
+        endsWithParent: true,
+        contains: [
+          {
+            className: 'attr',
+            begin: ':\\w+'
+          },
+          hljs.APOS_STRING_MODE,
+          hljs.QUOTE_STRING_MODE,
+          {
+            begin: '\\w+',
+            relevance: 0
+          }
+        ]
+      }
+    ]
+  };
+  const HTML_ATTRS = {
+    begin: '\\(\\s*',
+    end: '\\s*\\)',
+    excludeEnd: true,
+    contains: [
+      {
+        begin: '\\w+\\s*=',
+        end: '\\s+',
+        returnBegin: true,
+        endsWithParent: true,
+        contains: [
+          {
+            className: 'attr',
+            begin: '\\w+',
+            relevance: 0
+          },
+          hljs.APOS_STRING_MODE,
+          hljs.QUOTE_STRING_MODE,
+          {
+            begin: '\\w+',
+            relevance: 0
+          }
+        ]
+      }
+    ]
+  };
+
   return {
     name: 'HAML',
     case_insensitive: true,
@@ -38,64 +97,26 @@ export default function(hljs) {
             className: 'selector-tag',
             begin: '\\w+'
           },
-          {
-            className: 'selector-id',
-            begin: '#[\\w-]+'
-          },
-          {
-            className: 'selector-class',
-            begin: '\\.[\\w-]+'
-          },
-          {
-            begin: /\{\s*/,
-            end: /\s*\}/,
-            contains: [
-              {
-                begin: ':\\w+\\s*=>',
-                end: ',\\s+',
-                returnBegin: true,
-                endsWithParent: true,
-                contains: [
-                  {
-                    className: 'attr',
-                    begin: ':\\w+'
-                  },
-                  hljs.APOS_STRING_MODE,
-                  hljs.QUOTE_STRING_MODE,
-                  {
-                    begin: '\\w+',
-                    relevance: 0
-                  }
-                ]
-              }
-            ]
-          },
-          {
-            begin: '\\(\\s*',
-            end: '\\s*\\)',
-            excludeEnd: true,
-            contains: [
-              {
-                begin: '\\w+\\s*=',
-                end: '\\s+',
-                returnBegin: true,
-                endsWithParent: true,
-                contains: [
-                  {
-                    className: 'attr',
-                    begin: '\\w+',
-                    relevance: 0
-                  },
-                  hljs.APOS_STRING_MODE,
-                  hljs.QUOTE_STRING_MODE,
-                  {
-                    begin: '\\w+',
-                    relevance: 0
-                  }
-                ]
-              }
-            ]
-          }
+          SELECTOR_ID,
+          SELECTOR_CLASS,
+          RUBY_HASH_ATTRS,
+          HTML_ATTRS
+        ]
+      },
+      {
+        className: 'tag',
+        // implicit div: the line starts directly with `.class` or `#id`;
+        // `[\w-]` keeps `#{interpolation}` out of this case
+        begin: '^\\s*(?=[.#][\\w-])',
+        // relevance 0 everywhere: lines starting with `.foo` / `#foo` are
+        // common in other languages (CSS selectors, C preprocessor,
+        // assembler directives), so they must not count towards HAML
+        relevance: 0,
+        contains: [
+          hljs.inherit(SELECTOR_ID, { relevance: 0 }),
+          hljs.inherit(SELECTOR_CLASS, { relevance: 0 }),
+          hljs.inherit(RUBY_HASH_ATTRS, { relevance: 0 }),
+          hljs.inherit(HTML_ATTRS, { relevance: 0 })
         ]
       },
       { begin: '^\\s*[=~]\\s*' },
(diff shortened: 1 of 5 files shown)


src/languages/haml.js
```diff
@@ -8,6 +8,65 @@ Category: template
 
 // TODO support filter tags like :javascript, support inline HTML
 export default function(hljs) {
+  const SELECTOR_ID = {
+    className: 'selector-id',
+    begin: '#[\\w-]+'
```
joshgoebel:
```suggestion
    begin: /#[\w-]+/
```

Please always prefer raw regex when possible.

src/languages/haml.js
```diff
@@ -8,6 +8,65 @@ Category: template
 
 // TODO support filter tags like :javascript, support inline HTML
 export default function(hljs) {
+  const SELECTOR_ID = {
+    className: 'selector-id',
```
joshgoebel: `className` is deprecated. Prefer `scope` now.

src/languages/haml.js
```diff
@@ -8,6 +8,65 @@ Category: template
...
+    begin: '#[\\w-]+'
+  };
+  const SELECTOR_CLASS = {
+    className: 'selector-class',
+    begin: '\\.[\\w-]+'
```
joshgoebel: Prefer `match` if there is no begin/end pairing.

src/languages/haml.js
```diff
@@ -8,6 +8,65 @@ Category: template
...
+            relevance: 0
+          }
+        ]
+      }
+    ]
+  };
+  const HTML_ATTRS = {
+    begin: '\\(\\s*',
+    end: '\\s*\\)',
+    excludeEnd: true,
```
joshgoebel: Prefer `end: regex.lookahead()` instead of `excludeEnd`...

src/languages/haml.js
```diff
@@ -8,6 +8,65 @@ Category: template
...
+        contains: [
+          {
+            className: 'attr',
+            begin: ':\\w+'
+          },
+          hljs.APOS_STRING_MODE,
+          hljs.QUOTE_STRING_MODE,
+          {
+            begin: '\\w+',
+            relevance: 0
```
joshgoebel: `relevance` is deprecated, you can lose all the relevance keys.

src/languages/haml.js
```diff
@@ -38,64 +97,26 @@ export default function(hljs) {
...
+        begin: '^\\s*(?=[.#][\\w-])',
+        // relevance 0 everywhere: lines starting with `.foo` / `#foo` are
+        // common in other languages (CSS selectors, C preprocessor,
+        // assembler directives), so they must not count towards HAML
+        relevance: 0,
+        contains: [
+          hljs.inherit(SELECTOR_ID, { relevance: 0 }),
+          hljs.inherit(SELECTOR_CLASS, { relevance: 0 }),
+          hljs.inherit(RUBY_HASH_ATTRS, { relevance: 0 }),
+          hljs.inherit(HTML_ATTRS, { relevance: 0 })
```
joshgoebel: Pretty sure you may not need to inherit any of these then...

joshgoebel (changes requested):
Dropped a few notes, mostly simple stuff.

pikammmmm:
Addressed all six points: raw regex, scope instead of className, match instead of begin/end, lookahead instead of excludeEnd, dropped relevance, and trimmed the now-unneeded hljs.inherit() overrides. Also rebased onto main. PTAL.

src/languages/haml.js
```diff
@@ -8,6 +8,62 @@ Category: template
...
+          {
+            match: /\w+/
+          }
+        ]
+      }
+    ]
+  };
+  const HTML_ATTRS = {
+    begin: /\(\s*/,
+    end: regex.lookahead(/\s*\)/),
```
joshgoebel: So if the \s is optional, and doesn't matter, why look for it at all and not just the `)`?

src/languages/haml.js
```diff
@@ -8,6 +8,62 @@ Category: template
...
+    scope: 'selector-id',
+    match: /#[\w-]+/
+  };
+  const SELECTOR_CLASS = {
+    scope: 'selector-class',
+    match: /\.[\w-]+/
+  };
+  const RUBY_HASH_ATTRS = {
+    begin: /\{\s*/,
+    end: /\s*\}/,
```
joshgoebel: Same question here, if spaces are optional, why do we need to match them in the end?

It CAN better in begin, where maybe we need to gobble the to present a later match from triggering, but often in begin we shouldn't do this either... look at many other grammers.

We just match `{}` without dealing with spaces at all.

src/languages/haml.js
```diff
@@ -8,6 +8,62 @@ Category: template
...
+    contains: [
+      {
+        begin: /:\w+\s*=>/,
+        end: /,\s+/,
+        returnBegin: true,
+        endsWithParent: true,
+        contains: [
```
joshgoebel, line 28: I'll try to find time to look at this in more detail... often people try to write these grammars like REAL parsers, and that results in overly complex rules.  THis isn't as bad as some but I wonder if we could just detect the individual pieces with simpler pattern matching without trying to parse key/value parings...

IE Inside of `{}` you are going to have a few things:

- any ruby value on the right side (string, number, etc...), just a list of regular matchers
- a nested hash, so you could match with magic "self" rule
- a key on the left
  - maybe a `symbol:`
  - maybe a `:symbol =>`
  - maybe a `"string" =>`
  - maybe a `:"string" =>` symbol
  - anyways a small number of possible combos...

And we'd match the key expressions first... then value expressions... we don't have to try and parse them in the right order - because we only care about valid code.

Look at JSON for examples... we don't "parse" anything (we used to)... we just detect `{},:[]` as punctuation then look for strings, etc... very simple grammar.

src/languages/haml.js
pikammmmm: Dropped — it's `begin: /\{/`, `end: /\}/` now. Same for the parens.

src/languages/haml.js
pikammmmm: Dropped the `\s*`, and reverted this to `end: /\)/`. The lookahead left the `)` outside the enclosing tag span, which changed existing output — `test/markup/haml/default.expect.txt` had been updated to match. Consuming it keeps that file untouched by this PR.

src/languages/haml.js
pikammmmm: Agreed on the approach — but these two blocks are unchanged from main, I only lifted them into consts so the implicit-div mode could reuse them. Happy to do the JSON-style rewrite, just as a separate PR so this one stays the #3783 fix.

joshgoebel: > but these two blocks are unchanged from main

Ah, that's right... I can't help but see all the context... I'm just headed to sleep, let me take another look tomorrow and see if we can move this forward.

joshgoebel (approved):

PR #4513, opened 2026-08-22
fix(css, less, scss, stylus) keep escaped characters in selector names

### LINKED ISSUE #3965: (css) Escaped special character in selector class name breaks highlighting

maximilliangrand:
Fixes #3965.

## The bug

Class and id names may contain CSS escape sequences, which Tailwind emits for its variant separators. The selector patterns stopped at the backslash, so the name was cut short and the remainder rescanned:

```css
.dark\:hover\:bg-sky-500:hover {}
```

On 11.12.0 that scopes `.dark` as the class, then the escaped `\:hover` as a pseudo class and `500` as a number. All four css-like grammars are affected.

## Root cause

Each grammar spelled its own class and id pattern from a plain character class with no notion of escapes:

- `src/languages/css.js:17` (IDENT_RE, used by selector-class) and `:40`
- `src/languages/scss.js:38` and `:43`
- `src/languages/less.js:188` and `:189`, via `INTERP_IDENT_RE`
- `src/languages/stylus.js:76` and `:82`

## The fix

Adds `ESCAPE_RE` and `SELECTOR_IDENT_RE` to `src/languages/lib/css-shared.js` and uses them for the class and id patterns in all four grammars, following your note on the issue about pushing a shared name pattern down into `css-shared.js`. Each grammar keeps its own leading-character rules, so the change is additive.

The hex branches of the escape are kept disjoint: a six digit run is only taken when a seventh follows, and every shorter run has to reach a non-hex character. Without that, `\0` plus `0` and `\00` are both valid splits of the same text and `test/regex` fails with exponential backtracking on `"\00".repeat(n)`.

## Tests

Four escape cases added to `test/markup/css/css_consistency.txt` and mirrored into less, scss and stylus with `tools/css`, so the shared fixture is what proves the four grammars now agree.

With the source change reverted those four `css_consistency` tests fail (598 passing, 4 failing); with it applied the markup suite is 602 passing and the full suite 1649 passing, 3 pending, matching main.

Assisted-by: Claude Opus 4.8 (high)

### DIFF
--- a/src/languages/lib/css-shared.js
+++ b/src/languages/lib/css-shared.js
@@ -1,3 +1,32 @@
+/**
+ * A CSS escape sequence: a backslash followed either by up to six hex digits
+ * (with an optional single trailing space) or by any one character that is not
+ * a hex digit. Newlines cannot be escaped.
+ *
+ * The hex branches are kept disjoint so the sequence cannot be split more than
+ * one way, which would let the surrounding quantifier backtrack exponentially.
+ * A run of six digits is only taken when a seventh follows, since a hex escape
+ * stops after six; every shorter run has to reach a non-hex character.
+ *
+ * https://www.w3.org/TR/css-syntax-3/#escaping
+ */
+export const ESCAPE_RE = '\\\\(?:'
+  + '[\\da-fA-F]{6}(?=[\\da-fA-F])'
+  + '|[\\da-fA-F]{1,6}(?![\\da-fA-F]) ?'
+  + '|[^\\da-fA-F\\r\\n]'
+  + ')';
+
+/**
+ * Builds a selector name pattern that also accepts CSS escape sequences, so
+ * that names such as `.dark\\:hover\\:bg-sky-500` are matched in one piece
+ * instead of breaking apart at the backslash.
+ *
+ * @param {string} first character class the name may begin with
+ * @param {string} rest character class the remaining characters may use
+ */
+export const SELECTOR_IDENT_RE = (first, rest) =>
+  '(?:' + first + '|' + ESCAPE_RE + ')(?:' + rest + '|' + ESCAPE_RE + ')*';
+
 export const MODES = (hljs) => {
   return {
     IMPORTANT: {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## Version 11.12.1
 
+Core Grammars:
+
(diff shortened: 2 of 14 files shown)


src/languages/lib/css-shared.js
```diff
@@ -1,3 +1,32 @@
...
+ * one way, which would let the surrounding quantifier backtrack exponentially.
+ * A run of six digits is only taken when a seventh follows, since a hex escape
+ * stops after six; every shorter run has to reach a non-hex character.
+ *
+ * https://www.w3.org/TR/css-syntax-3/#escaping
+ */
+export const ESCAPE_RE = '\\\\(?:'
+  + '[\\da-fA-F]{6}(?=[\\da-fA-F])'
+  + '|[\\da-fA-F]{1,6}(?![\\da-fA-F]) ?'
+  + '|[^\\da-fA-F\\r\\n]'
```
joshgoebel, line 16: Can we add a few comments on what each of these is hoping to match?

PR #4518, opened 2026-08-27
fix(ruby): don't treat :: scope resolution as symbols after single-letter constants

### LINKED ISSUE #4516: Again: Ruby syntax: Wrong recognition of symbols at the place of calling a class methods. Now in v11.12.0
I already wrote about this [issue](https://github.com/highlightjs/highlight.js/issues/4294) for version v11.11.1, the issue was closed, but the bug still remains.

**Wrong recognition of symbols Ruby at the place of calling a class methods**.
Class methods may be call via:
```ClassName::method```
but it recognized as symbol ```:method```, also now ```ClassName:``` reconized as a symbol.

Ruby

```javascript
document.querySelectorAll('pre').forEach((el) => {
  hljs.highlightElement(el);
});
```

**Sample Code to Reproduce**
```:symbol```
```symbol:```
```A::a``` - call class method

### To fix the bug:
Just replace
```':(?!\\s)'``` to ```'(?<!:):(?!\\s)'```
and
```'(!|\\?)?:'``` to ```'(!|\\?)?:(?!:)'```

**Is it really that hard to do?** The bug came from the previous version and was not fixed.

### PR DESCRIPTION
**Pull request by Xi-7yao**
Fixes #4516

## Problem
`A::a` (calling a method on a single-letter constant) was highlighted as two symbols: `A:` and `:a`. Single-letter constants are not matched by the CamelCase class rule, so the trailing-colon symbol rule (`IDENT:`) consumed the first colon and the leading-colon symbol rule (`:`) then matched the second one, splitting the scope-resolution operator across two symbol matches.

## Fix
- Trailing-colon symbol rule: added a negative lookahead `(?!:)` so `IDENT:` is not matched when followed by another colon.
- Leading-colon symbol rule: added a negative lookbehind `(?<!:)` so a `:` preceded by `:` is never treated as a symbol.

The existing `::` swallow rule still handles CamelCase constants (`Foo::Bar`, `ClassName::method`); these guards close the gap for non-CamelCase (e.g. single-letter) constants.

## Test
- Added `A::a` and `A::B::c` to `test/markup/ruby/symbols.txt` (+ matching `symbols.expect.txt`).
- `npm test`: 1650 passing, 0 failing (incl. all ruby markup fixtures).
- `npm run build`: ok.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,9 +5,14 @@ Documentation:
 - docs: switch the Sphinx docs theme from ReadTheDocs to Shibuya [Haowei Hsu][]
 - docs: revamp the docs landing page using sphinx-design [Haowei Hsu][]
 
+Core Grammars:
+
+- fix(ruby) don't treat `::` scope resolution as symbols after single-letter constants, issue #4516 [Xi-7yao][]
+
 CONTRIBUTORS
 
 [Haowei Hsu]: https://github.com/hwhsu1231
+[Xi-7yao]: https://github.com/Xi-7yao
 
 ## Version 11.12.0
 
--- a/src/languages/ruby.js
+++ b/src/languages/ruby.js
@@ -360,12 +360,12 @@ export default function(hljs) {
     },
     {
       className: 'symbol',
-      begin: hljs.UNDERSCORE_IDENT_RE + '(!|\\?)?:',
+      begin: hljs.UNDERSCORE_IDENT_RE + '(!|\\?)?:(?!:)',
       relevance: 0
     },
     {
       className: 'symbol',
-      begin: ':(?!\\s)',
+      begin: '(?<!:):(?!\\s)',
       contains: [
         STRING,
         { begin: RUBY_METHOD_RE }
--- a/test/markup/ruby/symbols.expect.txt
+++ b/test/markup/ruby/symbols.expect.txt
@@ -10,3 +10,7 @@ obj.send(<span class="hljs-symbol">:to_s</span>)
 ::<span class="hljs-title class_">TopLevel</span>.new
 A::a
 A::B::c
+
+<span class="hljs-comment"># left side of :: may be a variable, not just a CamelCase constant (#4518 regression check)</span>
+an_alias = A
+an_alias::a_method(<span class="hljs-symbol">sym:</span> <span class="hljs-number">5</span>)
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by Sega100500** [comment]
**Important!** left side of expression may be variable, not constant.
Valid Ruby code:
```ruby
module A
  def self.a_method(sym: 123)
  end
end

A::a_method(sym: 1234)
# identical:
A.a_method(sym: 1234)

an_alias = A

# Important! left side of expression may be variable, not constant:
an_alias::a_method(sym: 5)
# identical:
an_alias.a_method(sym: 5)
```

Xi-7yao:
Thanks for the valid-use-case check! Confirmed against your snippet:

- `an_alias::a_method(sym: 5)` — the `::` is no longer split into two symbol matches (commit f58179a8 still works for variable receivers), the receiver + method both render as plain text, and only `sym:` is tagged `symbol`.
- `A::a_method(sym: 1234)` — same shape.

I just pushed `870e6c00` ("test(ruby): cover variable as :: left side per #4518 review") adding a regression case to `test/markup/ruby/symbols` that locks in the variable-as-receiver behaviour, so any future tweak to the trailing-/leading-colon symbol rules can't silently regress it.

The `sym:` keyword argument still being tagged as `symbol` is a separate, pre-existing grammar limitation: regex-based highlighting can't reliably distinguish a Ruby keyword argument (`a_method(sym: 5)`) from a symbol literal (`{:sym => ...}` / `{sym: ...}`), because both share the `IDENT:` shape and the surrounding context requires a real parser. That's out of scope for #4518 (whose scope is "don't split `::` into two symbols"), but happy to file/track it as a new issue if maintainers agree it warrants a separate grammar pass — just let me know.

Sega100500:
> The `sym:` keyword argument still being tagged as `symbol` is a separate, pre-existing grammar limitation: regex-based highlighting can't reliably distinguish a Ruby keyword argument (`a_method(sym: 5)`) from a symbol literal (`{:sym => ...}` / `{sym: ...}`), because both share the `IDENT:` shape and the surrounding context requires a real parser. That's out of scope for #4518 (whose scope is "don't split `::` into two symbols"), but happy to file/track it as a new issue if maintainers agree it warrants a separate grammar pass — just let me know.

That's right! `sym:` is a keyword argument and should be interpreted as a symbol, named (keyword) arguments are practically the same as passing a hash, i.e., they are hash keys — in this case, symbol keys.

Many thanks for fixing your script!

PR #4364, opened 2026-02-25
Cleanly separates out the Elements Compiler languages from their official couterparts

dwarfland:
### Changes
* Cleanly separates out the Elements Compiler languages from their official counterparts (e.g. hydrogene vs csharp, iodine vs java, etc), adding support for `oxygene`, `hydrogene`, `iodine` `silver`, `gold` `vbmercury` as named languages
* Added new C# 14 `field` keyword fo csharp & hydrogene

PR #4519, opened 2026-08-28
enh(oxygene): update keywords and add string literal support

Iamfittz:
Oxygene has been in core since 2013 and we maintain it at RemObjects.
This brings the grammar up to date with the current language:

- refreshed keyword list
- double-quoted strings, which Oxygene supports
- raw string literals (#"""..."""), matching the closing quote count
- interpolated strings ($"..." and $'...')
- hex character literals (#$1F)

Per @joshgoebel's suggestion in #4364, this is that PR reduced to Oxygene only.


Iamfittz:
@Yang062

src/languages/oxygene.js
```diff
@@ -6,18 +6,11 @@ Website: https://www.elementscompiler.com/elements/default.aspx
 Category: build-system
 */
 
-export default function(hljs) {
+export default function (hljs) {
```
joshgoebel: Please do not change lines that aren't part of the work.

src/languages/oxygene.js
```diff
@@ -6,18 +6,11 @@ Website: https://www.elementscompiler.com/elements/default.aspx
...
...   global    goto    group    has    helper    if    implementation    implements    implies    in    index    inherited    initialization    inline    interface    into    invariants    is    iterator    join    lazy    library    lifetimestrategy    locked    locking    loop    mapped    matching    method    mod    module    namespace    nested    new    nil    not    notify    nullable    of    old    on    operator    optional    or    order    otherwise    out    overload    override    packed    parallel    param    params    partial    pascal    pinned    platform    private    procedure    property    protected    public    published    queryable    raise    raises    read    readonly    record    reference    register    reintroduce    remove    repeat    require    required    result    reverse    safecall    sealed    select    selector    self    sequence    set    shl    shr    skip    soft    static    stdcall    step    strict    strong    take    then    to    true    try    tuple    type    unconstrained    unit    unmanaged    unretained    unsafe    until    uses    using    var    virtual    volatile    weak    where    while    with    write    xor    yield  "
```
joshgoebel: Can we get one word per line (see other modern grammars for example) in the newer array form...

src/languages/oxygene.js
```diff
@@ -29,15 +22,48 @@ export default function(hljs) {
     '\\*\\)',
     { relevance: 10 }
   );
+  const RAW_STRING = {
+    className: 'string',
```
joshgoebel:
```suggestion
    scope: 'string',
```

className is deprecated.

src/languages/oxygene.js
```diff
@@ -29,15 +22,48 @@ export default function(hljs) {
...
+    begin: /#+("+)(?!")/,
+    end: /"+(?!")/,
+    'on:begin': (m, resp) => {
+      resp.data._quoteCount = m[1].length;
+    },
+    'on:end': (m, resp) => {
+      if (m[0].length !== resp.data._quoteCount) {
+        resp.ignoreMatch();
+      }
```
joshgoebel: Can we do this with a single regex and back-reference?  Callbacks are very very slow compared the built-in regex engines.  If there is nothing in contains then a single regex should be able to do the same work.

src/languages/oxygene.js
```diff
@@ -29,15 +22,48 @@ export default function(hljs) {
...
+    },
+    contains: [],
+    relevance: 1
```
joshgoebel: We can lose all reference, also deprecated.

src/languages/oxygene.js
```diff
@@ -29,15 +22,48 @@ export default function(hljs) {
...
+  };
+  const INTERPOLATED_DOUBLE_STRING = {
+    className: 'string',
+    begin: /\$"/,
+    end: '"',
+    contains: [{ begin: '""' }]
```
joshgoebel: Use `match` when there is no `end`.

src/languages/oxygene.js
```diff
@@ -29,15 +22,48 @@ export default function(hljs) {
...
+    className: 'string',
+    begin: /\$'/,
+    end: '\'',
+    contains: [{ begin: '\'\'' }]
+  };
+  const DOUBLE_STRING = {
+    className: 'string',
+    begin: '"',
+    end: '"',
+    contains: [{ begin: '""' }]
```
joshgoebel: Whenever possible (tons of places in this PR) use literal regex, not strings...
```suggestion
    contains: [{ begin: /""/ }]
```

joshgoebel (changes requested):

Iamfittz:
All 7 points are addressed in the new commit.

src/languages/oxygene.js
```diff
@@ -23,47 +202,36 @@ export default function (hljs) {
...
-    },
-    'on:end': (m, resp) => {
-      if (m[0].length !== resp.data._quoteCount) {
-        resp.ignoreMatch();
-      }
-    },
-    contains: [],
-    relevance: 1
+    scope: 'string',
+    match: /#+("+)(?!")[\s\S]*?[^"]\1(?!")/
```
joshgoebel, line 206: What is the diff between `.*` and `\s\S*`?

joshgoebel (changes requested):
Much closer, just the one question.

src/languages/oxygene.js
Iamfittz: Oxygene raw strings can span lines, which `.*` wouldn't cover:

    var r := #"""
      line one
      line two
      """;

Same shape as the triple-quoted string in erlang.js, which uses `[\s\S]*?`
for the same reason.

joshgoebel (approved):

joshgoebel:
I can't touch your branch so you'll need to rebase.

Iamfittz:
Rebased.

Iamfittz:
Need to run tests again to be able to merge

PR #4526, opened 2026-08-31
fix(x86asm, mipsasm, llvm) don't swallow the preceding blank line into label spans

### LINKED ISSUE #4525: (x86asm, mipsasm, llvm) Label and directive matches swallow the preceding blank line
**Describe the issue**

In the assembly grammars, the label and directive rules anchor with `^\s*`. Because
`\s` matches `\n`, a match that starts at the beginning of a *preceding blank or
whitespace-only line* can run through the newline(s) and into the label on the next
line. Since JS regexes prefer the leftmost match, that earlier start wins over the
correct one, and the emitted `<span>` begins one line too early:

```html
<span class="hljs-symbol">
.exit:</span>
```

The blank line is now inside the symbol span, so it picks up the label's styling
(visible in any theme that gives `.hljs-symbol` a background, and it also means the
highlighted region simply doesn't correspond to the token). It also breaks anything
that splits the highlighted HTML per line.

This only triggers when the label or directive is preceded by a blank (or
whitespace-only) line, which is why it has gone unnoticed. That happens to be the
normal shape of compiler-generated assembly, where functions are separated by blank
lines.

It's already baked into one of your own fixtures: `test/markup/mipsasm/default.expect.txt`
currently expects

```html
<span class="hljs-symbol">
AckermannFunc:</span>
```

Affected rules:

- `src/languages/x86asm.js:127` — `'^\\s*[A-Za-z._?][A-Za-z0-9_$#@~.?]*(:|\\s+label)'`
- `src/languages/x86asm.js:129` — `'^\\s*%%[A-Za-z0-9_$#@~.?]*:'`
- `src/languages/x86asm.js:147` — `/^\s*\.[\w_-]+/`
- `src/languages/mipsasm.js:90,92` — `'^\\s*[a-z_\\.\\$][a-z0-9_\\.\\$]+:'`, `'^\\s*[0-9]+:'`
- `src/languages/llvm.js:38` — `/^\s*[a-z]+:/`

The trailing `\s+label` in the x86asm rule has the same defect independently: it can
join a bare identifier to a `label` keyword on the *following* line.

`x86asm`, `mipsasm` and `llvm` (explicit language, not auto-detection).

`highlight`.

**Sample Code to Reproduce**

```js
hljs.highlight("main:\n    mov eax, 1\n\n.exit:\n    ret\n", { language: "x86asm" }).value
```

Actual:

```html
<span class="hljs-symbol">main:</span>
    <span class="hljs-keyword">mov</span> <span class="hljs-built_in">eax</span>, <span class="hljs-number">1</span>
<span class="hljs-symbol">
.exit:</span>
    <span class="hljs-keyword">ret</span>
```

Same for directives (`\n\n    .section .text` puts the newline inside `hljs-meta`), for
`mipsasm` and `llvm` labels, and for the x86asm `label` keyword form across two lines.

**Expected behavior**

The blank line stays outside the span:

```html
<span class="hljs-symbol">main:</span>
    <span class="hljs-keyword">mov</span> <span class="hljs-built_in">eax</span>, <span class="hljs-number">1</span>

<span class="hljs-symbol">.exit:</span>
    <span class="hljs-keyword">ret</span>
```

**Additional context**

The fix is to use `[ \t]` rather than `\s` for the leading indentation (and for the
x86asm `label` separator): indentation before a label or directive is by definition
same-line. I have a patch with markup tests for all three grammars; PR to follow.
Verified against `main` (fc3f0639): with the fix the only existing expectation that
changes is the `mipsasm/default` one quoted above, and highlighting relevance is
unchanged for every case I tested, so auto-detection is unaffected.

Two related things I noticed while in here but deliberately left alone, as they're
separate defects with less obvious fixes:

- `src/languages/mipsasm.js:92` — the numbered-local-label rule `^\s*[0-9]+:` is
  effectively dead: the `number` mode appears earlier in `contains`, so at a true
  line start it always wins and `1:` highlights as a number followed by a bare colon.
  Before this fix, the symbol rule only ever won when it could start on an earlier
  blank line.
- `src/languages/mipsasm.js:57` — the keyword mode's `end: '\\s'` consumes the
  terminating character, so a mnemonic at the end of a line emits
  `<span class="hljs-keyword">nop\n</span>`. Same class of problem, but tightening it
  would change existing expectations for the trailing space in `addi `.

<sub>Investigated and drafted by Claude Code (Claude Opus 5) working with me; posted
from my account with my permission. I have reviewed the analysis and the patch and
can speak to them in review. `Assisted-by: Claude Opus 5 (high)`</sub>

mattgodbolt:
Resolves #4525

Both halves of this were already fixed in `armasm` in a34abcb5 / #2410, in passing
while fixing #2408. This applies the same two fixes to the sibling grammars that
were missed.

### Labels and directives starting a line early

The label and directive rules anchor with `^\s*`. `\s` matches `\n`, so a match
starting at the beginning of a preceding blank (or whitespace-only) line runs through
the newline and into the label on the next line. JS regexes prefer the leftmost match,
so that earlier start wins and the span begins one line early:

```html
<span class="hljs-symbol">
.exit:</span>
```

Indentation in front of a label or directive is same-line by definition, so this uses
`[ \t]` instead of `\s`:

- `x86asm` — global/local labels, macro-local labels, and the `.directive` meta rule
- `mipsasm` — GNU-syntax labels and numbered local labels
- `llvm` — block labels

This is character-for-character the change made to `armasm`'s label rule in a34abcb5
("symbols should not include prior newline", "also accept tabs as white space").
`armasm` and `mipsasm` were added with the same rule in 2015; `armasm` got the fix in
2020 and the others didn't. `avrasm` never had the bug, as its label rule doesn't
allow leading whitespace at all.

The trailing `\s+label` separator in the x86asm rule had the same defect
independently (it could join a bare identifier to a `label` keyword on the *following*
line), so that becomes `[ \t]+label`. The valid single-line `foo label near` form is
unaffected, including when tab-separated.

### Mnemonic spans swallowing their terminator

`mipsasm`'s mnemonic mode used `end: '\s'`, which consumes the character that ends
the match, so a mnemonic ending a line emitted `<span class="hljs-keyword">nop\n</span>`
and one mid-line kept its trailing space. a34abcb5's first bullet ("instructions
should not include the space after") fixed this in `armasm` by replacing the `end`
with a lookahead on `begin`; this does the same.

One deliberate deviation: `(?=\s|$)` rather than `armasm`'s `(?=\s)`, so a mnemonic
at the very end of the input, with no trailing whitespace, still highlights. `armasm`
currently drops it in that case — happy to send a follow-up for that if you'd like.

### Also

A separate commit switches the touched modes from `className` to `scope`, per
`AGENTS.md`.

### Verification

The blank-line bug was baked into the existing `mipsasm/default` expectation, which
this corrects:

```diff
-<span class="hljs-symbol">
-AckermannFunc:</span>
+
+<span class="hljs-symbol">AckermannFunc:</span>
```

I diffed before/after highlighting over a corpus of asm/LLVM snippets (labels at
column 0, space- and tab-indented labels, directives, macro-local labels, NASM `label`
keyword form, mnemonics mid-line, ending a line and at EOF, CRLF input, consecutive
blank lines, whitespace-only lines, labels inside strings and comments, plus the
existing fixtures). Relevance is unchanged throughout, so auto-detection is
unaffected.

Every markup difference is whitespace moving out of a span, with one exception: a
mipsasm numbered local label preceded by a blank line (`2:`) changes from `symbol` to
`number` followed by a bare colon. That's the dead-rule issue noted below — the symbol
rule only ever won there by starting on the earlier blank line, and at a true line
start the `number` mode has always won.

With the patch reverted the new markup tests fail; with it the suite is green.

### Noted but deliberately not fixed

One more pre-existing `mipsasm` defect, left alone to keep this reviewable (described
in #4525): the numbered-local-label rule `^\s*[0-9]+:` is effectively dead, because
the `number` mode appears earlier in `contains` and wins at a true line start.

### Checklist

---

<sub>Investigated, patched and drafted by Claude Code (Claude Opus 5) working with me,
and posted from my account with my permission. I have reviewed the analysis, the diff
and the test expectations and can speak to them in review.
`Assisted-by: Claude Opus 5 (high)`</sub>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## Version 11.12.1
 
+Core Grammars:
+
+- fix(x86asm, mipsasm, llvm) stop label and directive matches swallowing the preceding blank line, issue #4525 [Matt Godbolt][]
+
 Documentation:
 
 - docs: switch the Sphinx docs theme from ReadTheDocs to Shibuya [Haowei Hsu][]
@@ -8,6 +12,7 @@ Documentation:
 CONTRIBUTORS
 
 [Haowei Hsu]: https://github.com/hwhsu1231
+[Matt Godbolt]: https://github.com/mattgodbolt
 
 ## Version 11.12.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 Core Grammars:
 
 - fix(x86asm, mipsasm, llvm) stop label and directive matches swallowing the preceding blank line, issue #4525 [Matt Godbolt][]
+- fix(mipsasm) stop mnemonic spans swallowing the whitespace that terminates them, issue #4525 [Matt Godbolt][]
 
 Documentation:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,5 @@
 ## Version 11.12.1
 
-Core Grammars:
-
-- fix(x86asm, mipsasm, llvm) stop label and directive matches swallowing the preceding blank line, issue #4525 [Matt Godbolt][]
-- fix(mipsasm) stop mnemonic spans swallowing the whitespace that terminates them, issue #4525 [Matt Godbolt][]
-
 Documentation:
 
 - docs: switch the Sphinx docs theme from ReadTheDocs to Shibuya [Haowei Hsu][]
@@ -13,7 +8,6 @@ Documentation:
 CONTRIBUTORS
 
 [Haowei Hsu]: https://github.com/hwhsu1231
-[Matt Godbolt]: https://github.com/mattgodbolt
 
 ## Version 11.12.0
 
--- a/src/languages/llvm.js
+++ b/src/languages/llvm.js
@@ -34,7 +34,7 @@ export default function(hljs) {
     relevance: 0
   };
   const LABEL = {
-    className: 'symbol',
+    scope: 'symbol',
     variants: [ { begin: /^[ \t]*[a-z]+:/ }, // labels
     ],
     relevance: 0
--- a/src/languages/llvm.js
+++ b/src/languages/llvm.js
@@ -35,7 +35,7 @@ export default function(hljs) {
   };
   const LABEL = {
     className: 'symbol',
-    variants: [ { begin: /^\s*[a-z]+:/ }, // labels
+    variants: [ { begin: /^[ \t]*[a-z]+:/ }, // labels
     ],
     relevance: 0
   };
--- a/src/languages/mipsasm.js
+++ b/src/languages/mipsasm.js
@@ -84,7 +84,7 @@ export default function(hljs) {
         relevance: 0
       },
       {
-        className: 'symbol',
+        scope: 'symbol',
         variants: [
           { // GNU MIPS syntax
             begin: '^[ \\t]*[a-z_\\.\\$][a-z0-9_\\.\\$]+:' },
--- a/src/languages/mipsasm.js
+++ b/src/languages/mipsasm.js
@@ -87,9 +87,9 @@ export default function(hljs) {
         className: 'symbol',
         variants: [
           { // GNU MIPS syntax
-            begin: '^\\s*[a-z_\\.\\$][a-z0-9_\\.\\$]+:' },
+            begin: '^[ \\t]*[a-z_\\.\\$][a-z0-9_\\.\\$]+:' },
           { // numbered local labels
-            begin: '^\\s*[0-9]+:' },
+            begin: '^[ \\t]*[0-9]+:' },
           { // number local label reference (backwards, forwards)
             begin: '[0-9]+[bf]' }
         ],
--- a/src/languages/mipsasm.js
+++ b/src/languages/mipsasm.js
@@ -53,8 +53,10 @@ export default function(hljs) {
             + 'break|cache|d?eret|[de]i|ehb|mfc0|mtc0|pause|prefx?|rdhwr|'
             + 'rdpgpr|sdbbp|ssnop|synci?|syscall|teqi?|tgei?u?|tlb(p|r|w[ir])|'
             + 'tlti?u?|tnei?|wait|wrpgpr'
-        + ')',
-        end: '\\s'
+        + ')'
+        // lookahead, so the span doesn't swallow the whitespace (or newline)
+        // that terminates the mnemonic
+        + '(?=\\s|$)',
       },
       // lines ending with ; or # aren't really comments, probably auto-detect fail
       hljs.COMMENT('[;#](?!\\s*$)', '$'),
--- a/src/languages/x86asm.js
+++ b/src/languages/x86asm.js
@@ -124,9 +124,9 @@ export default function(hljs) {
         className: 'symbol',
         variants: [
           // Global label and local label
-          { begin: '^\\s*[A-Za-z._?][A-Za-z0-9_$#@~.?]*(:|\\s+label)' },
+          { begin: '^[ \\t]*[A-Za-z._?][A-Za-z0-9_$#@~.?]*(:|[ \\t]+label)' },
           // Macro-local label
-          { begin: '^\\s*%%[A-Za-z0-9_$#@~.?]*:' }
+          { begin: '^[ \\t]*%%[A-Za-z0-9_$#@~.?]*:' }
         ],
         relevance: 0
       },
@@ -144,7 +144,7 @@ export default function(hljs) {
       },
       {
         className: 'meta',
-        begin: /^\s*\.[\w_-]+/
+        begin: /^[ \t]*\.[\w_-]+/
       }
     ]
   };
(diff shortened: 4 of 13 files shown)


mattgodbolt:
Thanks — `className` → `scope` applied for the modes this PR touches.

On dropping `relevance: 0`, I'd push back, because I don't think it's the no-op the
"touch it, modernize it" rule assumes. Mode `relevance` is still live in the engine
(`src/highlight.js`, `relevance += top.relevance` as modes are popped), and the mode
compiler defaults an unset mode to 1. So removing `relevance: 0` from these modes
isn't dropping a redundant field, it's changing 0 to 1 — and since these are label
rules, that adds relevance for *every label in the file*, scaling with label count.

Markup is byte-identical either way, so it's purely a detection change. The x86asm
label rule matches more or less any `identifier:` at the start of a line, so making it
relevance-bearing would make x86asm materially louder in `highlightAuto` against
anything else with leading `label:` lines. `test-detect` still passes, but that corpus
seems small enough that I don't think it's strong evidence either way.

Happy to do it if you want it — it just seemed like a detection change that deserves
to be a deliberate, separately-reviewed decision rather than a drive-by in a
whitespace fix. Your call; say the word and I'll push it.

Two small corrections while I'm here:

- The suppressed comment on the x86asm `meta` mode mentions `relevance`; that mode
  only had `className`, no `relevance` field.
- I've corrected an overstatement in my own PR description: "no token gains or loses
  highlighting" wasn't quite right, as a mipsasm numbered local label after a blank
  line does change scope. See the revised Verification section.

mattgodbolt:
Turns out there's direct precedent for this in the repo, which I should have led with:
`armasm` had exactly the same bug and you fixed it in a34abcb5 / #2410, while passing
through for #2408. From that commit message:

> * fix(armasm) instructions should not include the space after
> * **fix(armasm) symbols should not include prior newline**
> * also accept tabs as white space

and the diff:

```diff
-            {begin: '^\\s*[a-z_\\.\\$][a-z0-9_\\.\\$]+:'}, //GNU ARM syntax
+            {begin: '^[ \\t]*[a-z_\\.\\$][a-z0-9_\\.\\$]+:'}, //GNU ARM syntax
```

That is character-for-character the change this PR makes to `mipsasm`. The two
grammars were near-identical siblings — `mipsasm` was added in 6ed30890 with the same
rule — and `armasm` got the fix in 2020 while `mipsasm`, `x86asm` and `llvm` were
missed. `avrasm` never had the bug because its label rule doesn't allow leading
whitespace at all.

So this isn't a new judgment call about how labels should be anchored; it's applying
a decision already made here six years ago to the three grammars that didn't get it.
The same commit also introduced `hljs.COMMENT('^[ \\t]*(?=#)', ...)`, so `[ \t]` for
"indentation on this line" is the established idiom in `armasm`.

That commit also contains the fix for the other `mipsasm` defect I mentioned above.
Its first bullet, "instructions should not include the space after", replaced the
mnemonic mode's `end: '\\s'` with a `(?=\\s)` lookahead on `begin`. `mipsasm` still
has the `end: '\\s'` form, which is why a mnemonic ending a line emits
`<span class="hljs-keyword">nop\n</span>`.

I've tried porting that across as well, and it's clean: mnemonic spans stop including
the trailing whitespace, the only expectations that change are the `mipsasm` ones
(`addi ` becomes `addi`, etc.), and nothing else moves. I've left it out to keep this
PR to one concern, but it's a two-line change if you'd like it here or as a follow-up.

mattgodbolt:
Pushed the `end: '\s'` fix too, so this now carries both halves of a34abcb5 across to
the grammars that missed them.

`mipsasm`'s mnemonic mode now ends with a `(?=\s|$)` lookahead on `begin` instead of
`end: '\s'`. Mnemonic spans no longer include the whitespace that terminates them, so
`<span class="hljs-keyword">nop\n</span>` becomes `<span class="hljs-keyword">nop</span>`
and mid-line mnemonics drop their trailing space. Only the `mipsasm` expectations
change, and relevance is unchanged. There's a new
`mipsasm/mnemonic-trailing-whitespace` fixture covering mid-line, end-of-line and
tab-separated cases.

One deliberate deviation from the precedent: `(?=\s|$)` rather than `(?=\s)`. With the
pure lookahead, a mnemonic at the very end of the input with no trailing whitespace
stops matching entirely and loses its highlighting. `armasm` has that gap today
(`hljs.highlight('main:\n    nop', {language: 'armasm'})` leaves `nop` unhighlighted).
Adding `|$` avoids it here. I've left `armasm` alone since it's outside this PR, but
it's a one-character follow-up if you want it.

I've updated the PR description to lead with the a34abcb5 precedent rather than
bury it.

joshgoebel (changes requested):
Remove the changelog entries please, I'm managing those via AI now at release time based on commit history, so not something we need with each PR.  After that I can merge.

Thanks!

mattgodbolt:
Thanks @joshgoebel - much appreciated! On it now (human Matt here)

mattgodbolt:
Done — `CHANGES.md` is untouched now, back to byte-identical with `main`. Dropped both entries and the CONTRIBUTORS link line that went with them.

Removed in a separate commit rather than by rewriting the branch, so your existing review context still lines up. Suite is green.

PR #4545, opened 2026-09-20
fix(css, less, scss, stylus) do not highlight digits in preserve-3d

### LINKED ISSUE #4531: (CSS) preserve-3d gets highlighted incorrectly
**Describe the issue**
The `transform-style: preserve-3d` keyword gets highlighted incorrectly, turning the number 3 into a different color, presumably due to the `-3` part.

css

highlight

**Sample Code to Reproduce**
```css
div {
  transform-style: preserve-3d;
}
```

which hljs highlights into:

Image

**Expected behavior**
Same as GitHub's highlighting - the `preserve-3d` part should be the same color.

**Additional context**
This doesn't repro for values such as `f00-b4r`, where the number is not directly following a hyphen.

dyk1454683243-sudo:
Fixes #4531

### Changes
`transform-style: preserve-3d` highlighted the `3` as `hljs-number` because shared `CSS_NUMBER_MODE` is built on `NUMBER_RE` (`\b\d+…`), and `\b` matches between `-` and a digit.

`on:begin` now ignores a number whose digit starts after an identifier-continuing hyphen (`preserve-3d`, `*-2d`, …). Ordinary lengths (`3px`, `0.5rem`, `100%`) and negative lengths (`-3px`) still highlight as numbers. No per-keyword `*-3d` list.

Lookbehind would express the same rule, but language grammars are linted as ES2015, so the callback is the 11.x-safe version.

Stylus does not apply `UNICODE_RANGE` on property values, so `U+0025-00FF` used to paint `00` after the hyphen as a number. That expect is updated for the same boundary.

Markup coverage is a dedicated `preserve-3d` CSS case plus the same snippet on the shared css/less/scss/stylus consistency fixtures.

### Checklist
- [x] Added markup tests

Assisted-by: Grok 4.6 (high)


fguzman-stack (commented):
Ran this locally: with the branch checked out, `npm test` reports **1651 passing / 0 failing** across the whole markup + detect + api + special + builds suites, so the fix looks safe against the full grammar set (not just the css-family suites). Nice.

Two concrete points from the review:

1. **The `on:begin` vs. lookbehind tradeoff is correctly justified.** I double-checked `.eslintrc.lang.js`: it sets `ecmaVersion: 2015` and its comment explicitly says the file exists to make sure "no ES2018 or newer syntax has crept in" - so a `(?<=[\w-])-` lookbehind would actually be rejected by `npm run lint-languages`. Some reviewers may not know this; thanks for spelling it out in the description.

2. **Suggestion: add one adversarial case that exercises the "keep" branch of the `on:begin` guard.** The existing markup fixture covers `preserve-3d` (ignoreMatch path) and `margin: -3px` (keep path, since the char before the `-` is a space). It does **not** lock in the common real-world negative-length case where the hyphen is preceded by an opening bracket - e.g. `transform: translate(-100px, 0);` or `grid-area: 1 / -2 / 3 / 4;`. I ran both against your branch and they highlight correctly, but a future refactor that changes the `match.input[i - 2]` check to a stricter `/[\s]/` (or drops the `-` from the char class) would silently regress them. Adding one line to `css_consistency.txt` would make that a compile-time failure instead of a runtime one. Not blocking, just a small hardening suggestion.

Otherwise clean, focused, and well-scoped. The unicode-range change in `css_consistency` expectations (`U+0025-00FF` no longer paints `00` as a number) is a behavior change but arguably a bug fix in its own right, and CHANGES.md captures it.

Assisted-by: opencode (Claude Sonnet) - verified locally, not blindly reviewed from the web UI.

fguzman-stack (commented):
Ran this locally: with the branch checked out, `npm test` reports **1651 passing / 0 failing** across the whole markup + detect + api + special + builds suites, so the fix looks safe against the full grammar set (not just the css-family suites). Nice.

Two concrete points from the review:

1. **The `on:begin` vs. lookbehind tradeoff is correctly justified.** I double-checked `.eslintrc.lang.js`: it sets `ecmaVersion: 2015` and its comment explicitly says the file exists to make sure no ES2018 or newer syntax has crept in. A `(?<=[\w-])-` lookbehind would actually be rejected by `npm run lint-languages`. Some reviewers may not know this; thanks for spelling it out.

2. **Suggestion: add one adversarial case that exercises the 'keep' branch of the `on:begin` guard.** The existing markup fixture covers `preserve-3d` (ignoreMatch path) and `margin: -3px` (keep path, hyphen preceded by space). It does not lock in the common real-world negative-length case where the hyphen is preceded by an opening bracket -- e.g. `transform: translate(-100px, 0);` or `grid-area: 1 / -2 / 3 / 4;`. I ran both against your branch and they highlight correctly, but a future refactor that changes the `match.input[i - 2]` check to a stricter `/[\s]/` (or drops `-` from the char class) would silently regress them. Adding one line to `css_consistency.txt` would make that a compile-time failure. Not blocking, just a small hardening suggestion.

Otherwise clean, focused, and well-scoped. The unicode-range change in `css_consistency` expectations (`U+0025-00FF` no longer paints `00` as a number) is a behavior change but arguably a bug fix in its own right.

*Assisted-by: opencode (I ran every verification command above on my local checkout; not a rubber stamp).*

fguzman-stack:
Apologies for the duplicated review - my `gh pr review` command fired twice by mistake. Please read the later of the two (they are the same content; the substantive point is the optional `translate(-100px, 0)` / `grid-area: 1 / -2` hardening suggestion). Everything else: verified locally with 1651 passing, LGTM from a reviewer's standpoint.

